Issue #116, opened 2015-03-12
Add Traversable.minBy(Comparator), maxBy(Comparator)


danieldietrich:

danieldietrich:
Type bounds of Comparator not correct.

PR #148, opened 2015-04-14
Fixes javaslang/javaslang#116

danieldietrich:


src/main/java/javaslang/collection/Traversable.java
```diff
@@ -682,6 +699,22 @@ default T min() {
...
 
+    /**
+     * <p>Calculates the minimum of this elements.</p>
+     *
+     * @param comparator A non-null element comparator
+     * @return the minimum of this elements.
+     * @throws java.lang.NullPointerException if comparator is null
+     * @throws java.lang.UnsupportedOperationException if no elements are present
+     */
+    default T minBy(Comparator<T> comparator) {
```
szarnekow: You may want to use Comparator<? super T> here and for maxBy(..), too

danieldietrich: Yes, thanks! Will fix that.

PR #243, opened 2015-05-16
Minor bugfixes + spelling and grammar tweaks to documentation

Sounie:
The main functional changes involve a null check and exception behaviour when an index exceeds the bounds within an iterator.


src/main/java/javaslang/collection/Stream.java
```diff
@@ -148,6 +148,9 @@ public boolean hasNext() {
 
             @Override
             public T next() {
+                if (i >= elements.length) {
+                    throw new NoSuchElementException();
+                }
                 return elements[i++];
```
danieldietrich: In general you are of course right. Here, this check is unnecessary because the Iterator is internally used only. The `Stream.ofAll(Iterator)` uses a for-loop to iterate the elements. Please remove the check to save cycles.

src/main/java/javaslang/collection/Tree.java
```diff
@@ -247,7 +247,7 @@ String toString(Tree<T> tree, int depth) {
                         final String children = tree.getChildren()
                                 .map(child -> toString(child, depth + 1))
                                 .join();
-                        return String.format("\n%s%s%s", indent, value, children);
+                        return String.format("%n%s%s%s", indent, value, children);
```
danieldietrich: This has to be a `\n`, not a `%n` to create a new line.

danieldietrich:
Thank you very much for your contribution! Please see my comments in 'Files changed' and amend the pull request.

src/main/java/javaslang/collection/Tree.java
martin-g: `String#format()` is rather slow operation.
Since there is no special formatting I'd recommend to use normal String concatenation or using StringBuilder. It is faster and still readable.

danieldietrich: Yes, right. Here we can use `"\n" + indent + value + children`. Could you do that, please?

In general I prefer readability, because the String representations are mostly used for Debug purpose or probably in the future Java 9 REPL :-)

When it comes to high performance, we will use StringBuilder. See also http://stackoverflow.com/a/13815477/1110815

martin-g: If it is for debugging then it is OK. No need to change it.

danieldietrich: here both solutions are ok! you can decide.

martin-g:
```java
 Returns a list string representation of this tree.
     *
     * @return A new string
     */
    default String toLispString() {
```
The javadoc says lis**t**. The name of the method is lis**p**. 
Looking at the code it does look like LISP (produces many braces :-) ).
Is there a typo in the javadoc?

Sounie:
I have reviewed Daniel's comments and updated my code changes accordingly.

danieldietrich:
LGTM, thanks!

I can't merge your changes automatically. Can you merge origin/master please? Here is how:

```shell
git remote -v
git remote add upstream https://github.com/javaslang/javaslang.git
git remote -v
git fetch upstream
git checkout master
git merge upstream/master
```

Then resolve conflicts, if any, and `git commit -a -m "merged origin/master"` and `push` as usual.

Note: Alternatively I can merge your changes manually but then this Pull Request gets lost.

Sounie:
Merged successfully :)

danieldietrich:
Great! Many thanks!!

PR #245, opened 2015-05-16
Minor performance improvement - use String concatenation instead of S…

martin-g:
…tring#format()

The improvement is used only when the concatenation representation is still readable. All usages of String#format() used in exception's message, or with formatting or with multiple placeholders are preserved.


danieldietrich:
LGTM

PR #255, opened 2015-05-19
Removed use try-with-resources on System.out and System.err

jorander:
Seems strange to use try-with-resources (that closes the resource after the end of the block) on System.out and System.err. Those streams should never be closed by application code.

This seems to work fine in most execution environments but causes problem when executing unit tests inside IntelliJ IDEA.


jorander:
This is the change needed to solve the issue with running tests in IntelliJ as were discussed with Daniel over mail.

danieldietrich:
Nice finding, thank you! Also for investigating it in depth!

Issue #126, opened 2015-04-04
Implement the first 99 Project Euler problems using javaslang collections


danieldietrich:
See [collection.euler.ProjectEulerTest](https://github.com/rocketscience-projects/javaslang/blob/master/src/test/java/javaslang/collection/euler/ProjectEulerTest.java)

danieldietrich:
The single test for all problems is growing. I will split the Euler solutions into separate classes.

danieldietrich:
See also [github.com/shadaj/euler](https://github.com/shadaj/euler) and [youtube.com/shadajProgramming](https://www.youtube.com/channel/UCzsv1Xi4X7D-oif5E4RT0JQ).

danieldietrich:
This will be part of every release as ongoing issue (until finished).

danieldietrich:
This is ongoing and done via PRs. We don't need this issue any more.

PR #260, opened 2015-05-21
Issue #126: Addition to Euler Problem 1 and implementation of Problem 2

jorander:
Fleshed out the solution of Problem 1 a bit and added a solution for Problem 2 including pattern matching and memoization.

Suggested future steps in development of these examples:
1. Move whole package to reflect the fact that the tests show examples of using more of Javaslang that just collections.
2. Make memoization part of the api, e.g. as a static method on Function* and CheckedFunction* interfaces.


danieldietrich:
@jorander Wow! Thanks.

You may move the package, e.g. to `javaslang.euler`. This would be great.

The memoization idea is very cool! I will add it to the (Checked)Function Generator:

```java
static <T1, R> Function1<T1, R> memoize(Function1<T1, R> f) {
    final Map<T1, R> cache = new ConcurrentHashMap<>();
    return input -> cache.computeIfAbsent(input, f::apply);
}

static <T1, T2, R> Function2<T1, T2, R> memoize(Function2<T1, T2, R> f) {
    final Map<Tuple2<T1, T2>, R> cache = new ConcurrentHashMap<>();
    return (t1, t2) -> cache.computeIfAbsent(Tuple.of(t1, t2), f.tupled()::apply);
}

// ...
```

danieldietrich:
@jorander memoize is now part of (Checked)Functions!

danieldietrich:
Functions now have a new memoized() method and updated the Euler test. For better integration now Function1 extends java.util.function.Function.

jbgi:
ConcurrentHashMap is, by default, optimized for large number of writer (default concurrency level = 16). This is rarely a good default. I would suggest using the alternate constructor that allow to specify the concurrency level (and use something like 1 or 2). This should provide better performances for the common single-threaded use-case.

danieldietrich:
Thanks for the hint, jb!

danieldietrich:
@jbgi I double-checked the source of ConcurrentHashMap. The concurrency level is used for calculating the initial capacity only. The constant `DEFAULT_CONCURRENCY_LEVEL` is

> Unused but defined for compatibility with previous versions of this class.

Using the default constructor should be fine.

jbgi:
oups my bad, I still had the old ConcurrentHashMap implementation in mind.

danieldietrich:
I've learned about the old - didn't know that.

Issue #227, opened 2015-05-14
Howto process further with Match.orElse()


danieldietrich:
On the one hand it is common to use `otherwise`. On the other hand, `orElse()` is unified throughout Java and Javaslang.

We are not able to let `Match` fully implement `Valences.Univalent`. However, `Match` should implement:

```java
T orElse(T other);
T orElseGet(Supplier<? extends T> other);
<X extends Throwable> T orElseThrow(Supplier<X> exceptionSupplier) throws X;
```

danieldietrich:
After thinking again, I also tend to use `otherwise()`. There is nothing to `get()` from a `Match` before applying it to an `Object`.

jbgi:
You could also design match to use
```java
Number i = Integer.valueOf(1);

BigDecimal asBigDecimal =
Match.of(i)
     .when((Integer i) -> new BigDecimal(i))
     .when((BigDecimal bd) -> bd)
     .getOrlElse(BigDecimal.ZERO);

Option<BigDecimal> asBigDecimalOption =
Match.of(i)
     .when((Integer i) -> new BigDecimal(i))
     .when((BigDecimal bd) -> bd)
     .get();
```
also only Function<? extends Number, R> should be allowed as when argument, IMO.
and the first when argument determine the type of R. (that would require a smarter Match builder)

danieldietrich:
Wow, this would be exactly the solution in order to make the `when`-part of `Match` implement [Univalent](https://github.com/javaslang/javaslang/blob/master/src/main/java/javaslang/control/Valences.java) (difference: we call `toOption()` instead of `get()` then). Awesome!

The existing `Match` syntax allows us to match arbitrary objects (like Scala). Additionally it makes `Match` reusable (see example below) and also composable, because it is a `Function`.

Do you think that `Match` will be too complicated, if there are two ways to use it: (1) as value, (2) as function ? Simplicity is very important. Ideally there is only one way to achieve a goal. In case (1) I would use `orElse()`, in case (2) I would use `otherwise()` for the default case.

```java
class Test {

    static final Match<BigDecimal> TO_BIG_DECIMAL = Match
        .when((Integer i) -> new BigDecimal(i))
        .when((BigDecimal bd) -> bd)
        .otherwise(BigDecimal.ZERO);

    List<BigDecimal> mapToBigDecimal(List<Number> numbers) {
        return numbers.map(TO_BIG_DECIMAL);
    }
}
```

danieldietrich:
[...]

PR #269, opened 2015-05-24
126 euler problems

jorander:
Continued work with The Euler problems. Working implementation for Problem #7.

Found that readability can be enhanced through implementation of a new convenience method on Stream.

Added a bit more JavaDoc on Match.Case to document a lessons learned during the work with Euler #7.


src/main/java/javaslang/collection/Stream.java
```diff
@@ -97,6 +98,20 @@
...
     /**
+     * Generates an (theoretically) infinitely long Stream using a function to calculate the next value
+     * based on the previous.
+     *
+     * @param seed The first value in the Stream
+     * @param supplier A function to calculate the next value based on the previous
+     * @param <T>      value type
+     * @return A new Stream
+     */
+    static <T> Stream<T> gen(T seed, Function1<T, T> supplier) {
```
danieldietrich: We should use `java.util.function.Function` instead of `javaslang.Function1` here. `Function1` is a `Function`, so we widen the type and increase the interoperability with Java a little bit.

src/test/java/javaslang/collection/euler/ProjectEulerTest.java
```diff
@@ -85,9 +84,37 @@ private static long fibonacci(int order) {
      * <p>See also <a href="https://projecteuler.net/problem=7">projecteuler.net problem 7</a>.</p>
      */
     @Test
-    @Ignore
-    public void shouldSolveProblem7() {
-        final int actual = Primes.asStream().drop(10_000).head();
-        assertThat(actual).isEqualTo(0); // TODO, see Primes#sieve(Stream)
+    public void test7() {
```
danieldietrich: We should choose similar names for the test methods. I like the `shouldXxx` notion because is a readable spec and `testXxx` is redundant with `@Test`.

danieldietrich:
Nice! Great to sea how you changed the generator template. I was a little bit in 'fead' that the generator might be confusing to committers.
I'm also impressed about the prime number solution! My solution produced a stack overflow. I have to update the twitter background of `@_Javaslang` with your solution :-)

src/test/java/javaslang/collection/euler/ProjectEulerTest.java
```diff
@@ -85,9 +84,37 @@ private static long fibonacci(int order) {
...
+        return Match
+                .when(1, (i) -> 2)
+                .orElse(() -> knownPrimes.drop(index - 1).head())
+                .apply(index);
+    }
+
+    private static final Stream<Integer> knownPrimes = Stream.gen(2, (p) -> nextPrime(p));
+
+    private static int nextPrime(int previousPrime) {
+        final Integer nextPrime = Match
```
danieldietrich: In general, I would directly return the match result. No need to introduce a variable, it has the same name as the method.

src/test/java/javaslang/collection/euler/ProjectEulerTest.java
```diff
@@ -85,9 +84,37 @@ private static long fibonacci(int order) {
...
+        assertThat(primeNo(5)).isEqualTo(11);
+        assertThat(primeNo(6)).isEqualTo(13);
+        assertThat(primeNo(10_001)).isEqualTo(104_743);
+    }
+
+    private static int primeNo(int index) {
+        return Match
+                .when(1, (i) -> 2)
+                .orElse(() -> knownPrimes.drop(index - 1).head())
+                .apply(index);
```
danieldietrich: Without having tried it, I think it should work to directly `return knownPrimes.drop(index - 1).head();`. If `index == 1` we are dropping `0` and then Stream.gen seed of `2` is returned.

src/test/java/javaslang/collection/euler/ProjectEulerTest.java
danieldietrich: `Match` is good for recognizing objects of a specific type. I would suggest to use the good old `if` instead, also because Match uses deserialization of lambdas which costs cycles.

```java
if (previousPrime == 2) {
    return 3;
} else {
    return Stream.gen(previousPrime + 2, v -> v + 2)
            .filter(i -> !isEvenlyDiversableByKnownPrimes(previousPrime, i))
            .take(1).head();
}
```

or even

```java
return (previousPrime == 2) ? 3 : Stream.gen(previousPrime + 2, v -> v + 2)
            .filter(i -> !isEvenlyDiversableByKnownPrimes(previousPrime, i))
            .take(1).head();
```

This should be much faster.

src/test/java/javaslang/collection/euler/ProjectEulerTest.java
```diff
@@ -85,9 +84,37 @@ private static long fibonacci(int order) {
...
+                .when(2, (i) -> 3)
+                .orElse(() -> Stream.gen(previousPrime + 2, (v) -> v + 2)
+                        .filter((i) -> !isEvenlyDiversableByKnownPrimes(previousPrime, i))
+                        .take(1).head())
+                .apply(previousPrime);
+        return nextPrime;
+    }
+
+    private static boolean isEvenlyDiversableByKnownPrimes(int previousPrime, int val) {
+        return knownPrimes.takeWhile((p) -> p < previousPrime).append(previousPrime)
```
danieldietrich: Instead of appending the previousPrime I would takeWhile with `p <= previousPrime` instead of `p < previousPrime`. This should give the same result.

danieldietrich: I omit the braces `(p)` and write `p` when there is not type in a single lambda argument. Braces are a little redundant.

danieldietrich:
Phew - many comments :-) I'm thrilled to test the solution but I'm waiting with the merge. I don't want to make changes myself (see comment), it is your solution!

generator/Generator.scala
```diff
@@ -279,10 +279,25 @@ def generateMainClasses(): Unit = {
...
+                /**
+                 * <p>Provides a default value which is returned if no case matches.</p>
+                 * <p>Note that this method takes the default by value which means that the input is
+                 * <em>eagerly evaluated</em> even if the orElse clause of the expression is not executed.
+                 * Unless you already have a default value calculated or as a literal it might be better
+                 * to use the {@link Match.Case#orElse(Supplier)} alternative to gain lazy evaluation.</p>
+                 *
+                 * @param defaultValue The default value.
+                 * @return a Match-expression
+                 */
```
danieldietrich: I also fought with javadoc / doclint. and discovered that we can use for paragraphs also a single `<p>` as a kind of separator. This javadoc is totally ok!! I just wanted to mention it because it surprised me that a single `<p>` is allowed. Feels a little bit like invalid html. However, I use it in newer javadocs, because it is shorter.

These are the same:

```java
Some text.
<p>
More text.
```

```java
<p>Some text.</p>
<p>More text.</p>
```

Additionally code blocks and lists do not need no paragraph separator.

```java
Some text.

<pre>
<code>
// code
</code>
</pre>

More text.

<ul>
<li>item</li>
</ul>
```

src/main/java/javaslang/collection/Stream.java
jorander: Of course, I'll make the change.

src/test/java/javaslang/collection/euler/ProjectEulerTest.java
jorander: Yes, my mistake.

src/test/java/javaslang/collection/euler/ProjectEulerTest.java
jorander: In this case, I can't see any difference in execution speed on the overall problem. Cycles are mostly used in other areas. In addition I like the readability and symmetry of the Match- clause.

With the extra variable, I totally agree. It was needed in an earlier solution and just got hanging around. I'll remove it.

danieldietrich: This is great to hear. The whole Match API was an experiment. It is good to see that it makes sense.

danieldietrich: Btw (offtopic) - the Match API will change slightly with #227. The current `orElse` will be named `otherwise`. And there will be an _additional_, more monadic API that returns a value instead of a function which has to be applied.

```java
Match.of(value)
   .when(...)
   .when(...)
   .otherwise(...);
```

jorander: I like that. The only thing that has disturbed my sense of readability was that the value to match on came last. This looks much better.

src/test/java/javaslang/collection/euler/ProjectEulerTest.java
jorander: Yep, you're right. It worked fine. And it make sense.

src/test/java/javaslang/collection/euler/ProjectEulerTest.java
jorander: No, using `p <= previousPrime` doesn't work since `Stream.takeWhile(Predicate)` runs until it finds an element that doesn't satisfy the Predicate and since we here are inside the functions that builds up the tail of the `knownPrimes` stream `nextPrime(...)` will be called again and we will get a StackOverflow.

danieldietrich:
Awesome, I love it!

PR #285, opened 2015-05-30
126 euler no 3

jorander:
Here is an implementation of Euler no 3 using the stream of primes from problem no 7.

Added an additional method (```build(..)```) on Stream to gain better readability over using ```new String.Cons(...)```. I'm not sure "build" is a good method name, but for now I haven't been able to come up with something better. "gen" is not possible to use since there will be a name clash with an existing method due to type erasure.

### DIFF
--- a/src/main/java/javaslang/collection/Stream.java
+++ b/src/main/java/javaslang/collection/Stream.java
@@ -111,6 +111,21 @@ public interface Stream<T> extends Seq<T> {
     }
 
     /**
+     * Builds an (theoretically) infinitely long Stream using a function to calculate the tail of the stream
+     * based on the head.
+     *
+     * @param head         The first value in the Stream
+     * @param tailFunction A function to calculate the tail values based on the head. To end the stream,
+     *                     return {@link Stream#nil}.
+     * @param <T>          value type
+     * @return A new Stream
+     */
+    static <T> Stream<T> build(T head, Function<T, Stream<T>> tailFunction) {
+        Objects.requireNonNull(tailFunction, "tailFunction is null");
+        return new Stream.Cons<>(head, () -> tailFunction.apply(head));
+    }
+
+    /**
      * Returns the single instance of Nil. Convenience method for {@code Nil.instance()}.
      * <p>
      * Note: this method intentionally returns type {@code Stream} and not {@code Nil}. This comes handy when folding.
--- a/src/main/java/javaslang/collection/Stream.java
+++ b/src/main/java/javaslang/collection/Stream.java
@@ -111,18 +111,17 @@ public interface Stream<T> extends Seq<T> {
     }
 
     /**
-     * Builds an (theoretically) infinitely long Stream using a function to calculate the tail of the stream
+     * Generates an (theoretically) infinitely long Stream using a function to calculate the tail of the stream
      * based on the head.
      *
      * @param head         The first value in the Stream
-     * @param tailFunction A function to calculate the tail values based on the head. To end the stream,
-     *                     return {@link Stream#nil}.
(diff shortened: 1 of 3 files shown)


danieldietrich:
Thank you, Jörgen! It is great to see the Euler tests evolving!
I will provide comments in a minute...

src/main/java/javaslang/collection/Stream.java
```diff
@@ -111,6 +111,21 @@
...
+     *
+     * @param head         The first value in the Stream
+     * @param tailFunction A function to calculate the tail values based on the head. To end the stream,
+     *                     return {@link Stream#nil}.
+     * @param <T>          value type
+     * @return A new Stream
+     */
+    static <T> Stream<T> build(T head, Function<T, Stream<T>> tailFunction) {
+        Objects.requireNonNull(tailFunction, "tailFunction is null");
+        return new Stream.Cons<>(head, () -> tailFunction.apply(head));
```
danieldietrich: We apply the head that is already given, so we can apply it before the call of build. I will give an example in a comment below...

Then we can rename `build` to `gen` and use `Supplier<Stream<T>>` instead of the `Function`.

src/test/java/javaslang/collection/euler/ProjectEulerTest.java
```diff
@@ -78,6 +79,40 @@ private static long fibonacci(int order) {
...
+    }
+
+    private static Stream<Tuple2<Long, Long>> primeFactorsAndResultingValTail(Tuple2<Long, Long> previousFactorAndResultingVal) {
+        return Match
+                .when(1L, p -> Stream.<Tuple2<Long, Long>>nil())
+                .otherwise(() -> {
+                    final long nextPrimeFactor = knownPrimes.filter(p -> previousFactorAndResultingVal._2 % p == 0).take(1).head();
+                    return Stream.build(
+                            new Tuple2<>(nextPrimeFactor, previousFactorAndResultingVal._2 / nextPrimeFactor),
+                            ProjectEulerTest::primeFactorsAndResultingValTail);
```
danieldietrich:
```java
// maybe tuple is not a good name...
final Tuple2<...> tuple = new Tuple2<>(nextPrimeFactor, previousFactorAndResultingVal._2 / nextPrimeFactor);
return Stream.build(tuple, () -> primeFactorsAndResultingValTail(tuple));
```

src/main/java/javaslang/collection/Stream.java
danieldietrich: `Stream.cons(T head, Supplier<Stream<T>> tail)` would be also a candidate instead of `gen` but by using `gen` we keep the amount of factory method names small. Should be ok!

jorander:
I agree with your comments. Great you saw the possibility, I didn't.

src/main/java/javaslang/collection/Stream.java
```diff
@@ -111,6 +111,20 @@
...
+     * based on the head.
+     *
+     * @param head         The first value in the Stream
+     * @param tailSupplier A function to calculate the tail values. To end the stream, return {@link Stream#nil}.
+     * @param <T>          value type
+     * @return A new Stream
+     */
+    static <T> Stream<T> gen(T head, Supplier<Stream<T>> tailSupplier) {
+        Objects.requireNonNull(tailSupplier, "tailSupplier is null");
+        return new Stream.Cons<>(head, () -> tailSupplier.get());
```
danieldietrich: Just one minor thing - this can be simplified to

```java
return new Stream.Cons<>(head, tailSupplier);
```

danieldietrich:
Many thanks!!

Issue #289, opened 2015-06-01
Replace 'woops' message in negative tests.


ignasi35:
Some negative tests require the API to fail. In those cases there's a log trace stating the error but that's actually expected. Current output is:

```
Running javaslang.test.PropertyCheck8Test
test: Errored after 0 passed tests with message: Applying predicate: woops
test: OK, passed 1000 tests.
test: Errored after 0 passed tests with message: Arbitrary 1 of size 100: woops
```

It could be something like:

```
Running javaslang.test.PropertyCheck8Test
test: Errored after 0 passed tests with message: Applying predicate: yay! (this is a negative test)
test: OK, passed 1000 tests.
```

PR #290, opened 2015-06-01
Solves #289

ignasi35:


danieldietrich:
Nice, thanks!

generator/Generator.scala
```diff
@@ -1246,7 +1247,7 @@ def generateTestClasses(): Unit = {
 
             @$test
             public void shouldRecognizeArbitraryError() {
-                final Arbitrary<?> arbitrary = n -> { throw new RuntimeException("woops"); };
+                final Arbitrary<?> arbitrary = n -> { throw new RuntimeException("$woops"); };
```
martin-g: Shouldn't `"$woops"` actually be `s"$woops"`.
The same happens few times below too.

danieldietrich: yep, will fix it. thx!

PR #292, opened 2015-06-03

jorander:
…llections

Solution for problem #4.

Not entirely sure there is not a neater solution to producing the products using some built-in API. I couldn't find any so I went with flatMapping one stream over the other.

### DIFF
--- a/src/test/java/javaslang/collection/euler/ProjectEulerTest.java
+++ b/src/test/java/javaslang/collection/euler/ProjectEulerTest.java
@@ -113,6 +113,37 @@ public class ProjectEulerTest {
     }
 
     /**
+     * <strong>Problem 4: Largest palindrome product</strong>
+     * A palindromic number reads the same both ways. The largest palindrome made
+     * from the product of two 2-digit numbers is 9009 = 91 × 99.
+     * <p>
+     * Find the largest palindrome made from the product of two 3-digit numbers.
+     * <p>
+     * See also <a href="https://projecteuler.net/problem=4">projecteuler.net problem 4</a>.
+     */
+    @Test
+    public void shouldSolveProblem4() {
+        assertThat(largetsPalindromeOfProductsFromFactorsInRange(10, 99)).isEqualTo(9009);
+        assertThat(largetsPalindromeOfProductsFromFactorsInRange(100, 999)).isEqualTo(906609);
+    }
+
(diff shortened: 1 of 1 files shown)


src/test/java/javaslang/collection/euler/ProjectEulerTest.java
```diff
@@ -113,6 +113,37 @@ private static long largestPrimeFactorOf(long val) {
...
+     */
+    @Test
+    public void shouldSolveProblem4() {
+        assertThat(largetsPalindromeOfProductsFromFactorsInRange(10, 99)).isEqualTo(9009);
+        assertThat(largetsPalindromeOfProductsFromFactorsInRange(100, 999)).isEqualTo(906609);
+    }
+
+    private static int largetsPalindromeOfProductsFromFactorsInRange(final int min, final int max) {
+        return Stream.rangeClosed(min, max)
+                .flatMap((i) -> Stream.rangeClosed(i, max).map((j) -> i * j))
```
danieldietrich: Lovely! I did not know StringBuilder.reverse() yet.
Could you replace (i) and (j) with i and j?

src/test/java/javaslang/collection/euler/ProjectEulerTest.java
```diff
@@ -113,6 +113,37 @@ private static long largestPrimeFactorOf(long val) {
...
+     * <p>
+     * See also <a href="https://projecteuler.net/problem=4">projecteuler.net problem 4</a>.
+     */
+    @Test
+    public void shouldSolveProblem4() {
+        assertThat(largetsPalindromeOfProductsFromFactorsInRange(10, 99)).isEqualTo(9009);
+        assertThat(largetsPalindromeOfProductsFromFactorsInRange(100, 999)).isEqualTo(906609);
+    }
+
+    private static int largetsPalindromeOfProductsFromFactorsInRange(final int min, final int max) {
```
danieldietrich: largets vs. largest

src/test/java/javaslang/collection/euler/ProjectEulerTest.java
danieldietrich: I've thought about the `flatMap`. It is really nice - I can't think of another solution.

danieldietrich:
Yay! One step further towards 99 :-)

PR #299, opened 2015-06-16
Adding additional andThen method to allow chaining of previous Try re…

martincooper:
Adding in additional "andThen" method to the Try class allowing a nicer chaining of results from previous Try to the next expression.

### DIFF
--- a/src/main/java/javaslang/control/Try.java
+++ b/src/main/java/javaslang/control/Try.java
@@ -5,6 +5,7 @@
  */
 package javaslang.control;
 
+import javaslang.CheckedFunction1;
 import javaslang.collection.TraversableOnce;
 
 import java.util.Collections;
@@ -232,6 +233,35 @@ public interface Try<T> extends TraversableOnce<T> {
      */
     default Try<Void> andThen(CheckedRunnable runnable) {
         return flatMap(ignored -> Try.run(runnable));
+    }
+
+    /**
+     * Runs the given checked function if this is a {@code Success},
+     * passing the result of the current expression to it.
+     * If this expression is a {@code Failure} then it'll return a new
+     * {@code Failure} of type R with the original exception.
+     *
+     * The main use case is chaining checked functions using method references:
+     *
+     * <pre>
+     * <code>
+     * Try.of(() -&gt; 100)
+     *    .andThen(x -&gt; x + 100)
+     *    .andThen(x -&gt; x * 20);
+     *
+     * </code>
+     * </pre>
+     *
+     * @param f A checked function taking a single argument.
+     * @return a new {@code Try}
+     */
+    default <R> Try<R> andThen(CheckedFunction1<T, R> f) {
+
+        if (this.isFailure()) {
+            return new Failure<>(this.failed().get());
+        }
+
+        return Try.of(() -> f.apply(this.get()));
     }
 
     @Override
--- a/src/main/java/javaslang/control/Try.java
+++ b/src/main/java/javaslang/control/Try.java
@@ -255,14 +255,7 @@ public interface Try<T> extends TraversableOnce<T> {
      * @param f A checked function taking a single argument.
      * @return a new {@code Try}
      */
-    default <R> Try<R> andThen(CheckedFunction1<T, R> f) {
-
-        if (this.isFailure()) {
-            return new Failure<>(this.failed().get());
-        }
-
-        return Try.of(() -> f.apply(this.get()));
-    }
+    <R> Try<R> andThen(CheckedFunction1<T, R> f);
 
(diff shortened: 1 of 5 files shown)


martincooper:
The change I proposed seem functionally fine, but fail on the "TypeConsistencyTest". I'm not too sure what is incorrect, if it's something i've missed?

danieldietrich:
Hi Martin,

that is a great addition because it solves an issue in an elegant way I wrapped my head around again and again. Let me explain: Contrary to most other languages Java has checked exceptions. We can't use method signatures with checked and unchecked exceptions side-by-side because of ambiguities. On the one hand the Try monad deals with exceptions. Because of this we want methods take checked functions as arguments, so that we are able to deal with code that throws. On the other hand, all other monads are using the unchecked functions in their method signatures. It is important to have a common behavior because we (will) need a common interface for these monadic types.

The method you added was there a few weeks ago but it had another name: `map`. Before `map` took a `CheckedFunction`. I had to change the signature, now it takes a `Function`. The reason is the common behavior of monadic types I mentioned above. The current solution is to `flatMap` throwing code. Your method is syntactic sugar for that. I will comment your change - we can replace the `if` statement with the shorter `flatMap` variant.

> The change I proposed seem functionally fine, but fail on the "TypeConsistencyTest". I'm not too sure what is incorrect, if it's something i've missed?

There are many cases where it makes sense to use covariant (= more specific) return types in sub classes. The "TypeConsistencyTest" checks that. Example:

```java
// defined in Try - we know nothing specific about the return type
Try<T> recover(Function<Throwable, ? extends T> f);

// defined in Success - we know that calling success.recover(...) will definitely return a Success
Success<T> recover(Function<Throwable, ? extends T> f) { return this; }

// defined in Failure - we know nothing specific about the return type
Try<T> recover(Function<Throwable, ? extends T> f) {
    return Try.of(() -> f.apply(cause.getCause()));
}
```

But there are also cases, like this, where the TypeConsistencyTest falsely reports an error. Then we just add an exception to the test. Please add the following lines to the `TypeConsistencyTest` class. I copy-and-pasted most of these from the unit test output (format: `<class> '//' <method>`). Hopefully they are correct, you can test it with a `mvn clean test`.

```java
// control.Failure
...
"javaslang.control.Failure//public default javaslang.control.Try javaslang.control.Try.andThen(javaslang.CheckedFunction1)",

// control.Success
...
"javaslang.control.Success//public default javaslang.control.Try javaslang.control.Try.andThen(javaslang.CheckedFunction1)",
```

Many thanks!

src/main/java/javaslang/control/Try.java
```diff
@@ -234,6 +235,35 @@
...
+     * @param f A checked function taking a single argument.
+     * @return a new {@code Try}
+     */
+    default <R> Try<R> andThen(CheckedFunction1<T, R> f) {
+
+        if (this.isFailure()) {
+            return new Failure<>(this.failed().get());
+        }
+
+        return Try.of(() -> f.apply(this.get()));
```
danieldietrich: Please replace the method with:

```java
    default <R> Try<R> andThen(CheckedFunction1<T, R> f) {
        return flatMap(value -> Try.of(() -> f.apply(value)));
    }
```

Scala folks would say _flatMap that shit_ :-)

danieldietrich:
Sorry for the inconvenience, had to think again. The best way to do it (from the API view point) is the following:

```java
Try.java
--------
<R> Try<R> andThen(CheckedFunction1<T, R> f);

Success.java
------------
@Override
public <R> Try<R> andThen(CheckedFunction1<T, R> f) {
    return flatMap(value -> Try.of(() -> f.apply(value)));
}

Failure.java
------------
@Override
public <R> Failure<R> andThen(CheckedFunction1<T, R> f) {
    return (Failure<R>) this;
}

TypeConsistencyTest.java
------------------------
// control.Success
...
"javaslang.control.Success//public default javaslang.control.Try javaslang.control.Try.andThen(javaslang.CheckedFunction1)",
```

Please note that `TypeConsistencyTest` now only contains an exception-rule for the Success case.

I really hope you do not run away regarding this mess I made :-)

martincooper:
Hi Daniel. Thanks for the detailed response :-) I'll go through it in detail when I finish work tonight and update accordingly.

danieldietrich:
great!

martincooper:
Hi Daniel

I've just been reading through your explanation and it all makes perfect sense :-) I've spent a few months working with Scala, but i'm still trying learning when and where I should *flatMap shit* :-) 

I've updated the code and all tests pass.

Thanks again.

danieldietrich:
Thx!

PR #304, opened 2015-06-23
Fixes unexpected list behavior

### LINKED ISSUE #298: Unexpected List.map behavior for independent values
@ggalmazor Gave an example, where a List seems to be in the wrong order after mapping values:

https://gist.github.com/ggalmazor/c122cc15ae971dbd08b8

Also other List operations should be checked regarding the use of `foldLeft` and `foldRight`.

This fix will change the behavior of List operations for some special cases.

danieldietrich:
Fixes #298

### DIFF
--- a/src/main/java/javaslang/collection/List.java
+++ b/src/main/java/javaslang/collection/List.java
@@ -263,7 +263,17 @@ public interface List<T> extends Seq<T>, Stack<T> {
     @Override
     default <U> List<U> flatMap(Function<? super T, ? extends Iterable<U>> mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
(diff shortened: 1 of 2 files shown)


ggalmazor:
I see that this fix is not available yet in 2.0.0-beta version... Is this right?

danieldietrich:
It should be all fine.

Here is 2.0.0-beta: https://github.com/javaslang/javaslang/blob/javaslang-2.0.0-beta/src/main/java/javaslang/collection/List.java#L594

and RC1: https://github.com/javaslang/javaslang/blob/master/src/main/java/javaslang/collection/List.java#L593

Do we have a regression? Do you have an example?

Thx!

ggalmazor:
Sorry! My local maven cache was giving me a hard time and I was seeing 1.2.2 as 2.0.0...

danieldietrich:
Thx for the info - then I expect it to work (it should because of the unit tests).

Issue #220, opened 2015-05-04
[Epic] Extend the collection library


danieldietrich:
This includes:

* #9 Tree transformation X
* #113 Tree extends Traversable ✓

Already finished:

* #102 Traversable.countOccurrences X
* #110 Traversable.groupBy ✓
* #170 Immutable HashMap ✓
* #214 Immutable TreeSet ✓
* #256 Re-organize collection.Tree ✓
* #271 More common TraversableOnce methods ✓
* #278 Immutable Stack ✓
* #279 Immutable Queue ✓
* #295 Immutable Array ✓
* #428 Add an immutable TreeMap to the javaslang.collection package ✓
* #429 Add an immutable HashSet to the javaslang.collection package ✓

Useful links:

* http://en.wikipedia.org/wiki/List_of_data_structures
* http://rosettacode.org/wiki/Rosetta_Code
* http://xlinux.nist.gov/dads/

danieldietrich:
I've solved functionaljava/functionaljava#149 already for Javaslang but it will be interesting to track it anyway...

danieldietrich:
An overview of the next version of Javaslang, especially the collection API (work in progress):

![javaslang-classes](https://cloud.githubusercontent.com/assets/743833/8150475/f0ca51d6-12eb-11e5-8f68-6d5730657740.png)

danieldietrich:
`RedBlackTree` and `HashArrayMappedTrie` are backing data structures for the high-level collections `Map`/`SortedMap`, `Set`/`SortedSet` etc. I think it does not make sense to implement `Traversable`.

Here I want to share some thoughts on how to provide an appropriate abstraction for such data structures.

## Idea

1) The interfaces should be kept slim. We may add functionality to special interfaces like `RedBlackTree` in form of static methods. This functionality is called by high-level collections to implement the `Traversable` interface. I've seen this in Scala's [TreeSet](https://github.com/scala/scala/blob/v2.11.7/src/library/scala/collection/immutable/TreeSet.scala) which is used in conjunction with a [RedBlackTree companion object](https://github.com/scala/scala/blob/v2.11.7/src/library/scala/collection/immutable/RedBlackTree.scala).

2) Parent references will not be held.

[...]

PR #306, opened 2015-06-24
Improves List.partition(Predicate)

### LINKED ISSUE #305: Improve List.partition(Predicate)
Currently `List.partition()` uses two passes (modulo reverse) to split the underlying List using `filter(perdicate)` and `filter(predicate.negate())`.

functionaljava for example partitions a List using one pass but creating intermediate objects (product/tuple):

```java
public P2<List<A>, List<A>> partition(F<A, Boolean> f) {
    P2<List<A>, List<A>> p2 = foldLeft(acc -> a ->
        f.f(a) ? P.p(acc._1().cons(a), acc._2()) : P.p(acc._1(), acc._2().cons(a)),
        P.p(nil(), nil())
    );
    return P.p(p2._1().reverse(), p2._2().reverse());
}
```

Scala uses mutable TraversableLike builders to build the result in one pass without using intermediate objects:

```scala
def partition(p: A => Boolean): (Repr, Repr) = {
    val l, r = newBuilder
    for (x <- this) (if (p(x)) l else r) += x
    (l.result, r.result)
}
```

The goal here is to use one pass and create no intermediate objects.

danieldietrich:
Fixes #305

### DIFF
--- a/src/main/java/javaslang/collection/List.java
+++ b/src/main/java/javaslang/collection/List.java
@@ -426,7 +426,15 @@ public interface List<T> extends Seq<T>, Stack<T> {
     @Override
     default Tuple2<List<T>, List<T>> partition(Predicate<? super T> predicate) {
         Objects.requireNonNull(predicate, "predicate is null");
-        return Tuple.of(filter(predicate), filter(predicate.negate()));
+        List<T> left = nil(), right = nil();
+        for (T t : this) {
+            if (predicate.test(t)) {
+                left = left.prepend(t);
+            } else {
+                right = right.prepend(t);
+            }
+        }
+        return Tuple.of(left.reverse(), right.reverse());
     }
 
     @Override


src/main/java/javaslang/collection/List.java
```diff
@@ -426,7 +426,15 @@ default int lastIndexOf(T element) {
...
-        return Tuple.of(filter(predicate), filter(predicate.negate()));
+        List<T> left = nil(), right = nil();
+        for (T t : this) {
+            if (predicate.test(t)) {
+                left = left.prepend(t);
+            } else {
+                right = right.prepend(t);
+            }
+        }
+        return Tuple.of(left.reverse(), right.reverse());
```
szarnekow: I'm pretty sure it doesn't harm, but don't we have a second pass here with the reverse() operation?

danieldietrich: Yes, you're right :-) It is better than the solution before, which used 2 x filter(), which uses foldLeft(), which also calls reverse(), i.e. 4 passes in sum. There is no improvement when using ArrayList for building lists because List.ofAll(Iterable) has to be called to convert the ArrayList to a javaslang.collection.List, which internally also needs to call reverse(). I think there is no better solution for a single-linked list.

danieldietrich: Mmhhh, nope, that's not right what I wrote. An ArrayList could be iterated backward to create a javaslang List, i.e. there is no need to call reverse(). That would be 1 pass in sum for partition(). However, I need to provide a better implementation for List.ofAll().

Thanks Sebastian, I will do that!

danieldietrich: Note: List.ofAll(Interable) could be improved in a future version of Javaslang in the same way as for ArrayList by reverse-iterating all IndexedSeq instances in O(1). See https://github.com/javaslang/javaslang/issues/220#issuecomment-111876895

danieldietrich: @szarnekow of course the ArrayLists have also to be passes. That means we need 3 passes instead of 1. I updated the comments above accordingly.

I see a slight improvement using ArrayList compared to the first solution using javaslang's linked List and reverse() because no intermediate Cons objects are created while interating the original List.

danieldietrich:
This pull request was finally merged with #308

src/main/java/javaslang/collection/List.java
danieldietrich: In fact we now have 2 passes overall because left.size() + right.size() = n.

PR #309, opened 2015-06-24
Improvement of List creation given java.util.Vector

danieldietrich:
Addition to #305

### DIFF
--- a/src/main/java/javaslang/collection/List.java
+++ b/src/main/java/javaslang/collection/List.java
@@ -133,12 +133,12 @@ public interface List<T> extends Seq<T>, Stack<T> {
             @SuppressWarnings("unchecked")
             final List<T> list = (List<T>) elements;
             return list;
-        } else if (elements instanceof ArrayList) {
+        } else if (elements instanceof ArrayList || elements instanceof Vector) {
             @SuppressWarnings("unchecked")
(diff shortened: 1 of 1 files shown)


src/main/java/javaslang/collection/List.java
```diff
@@ -133,12 +133,12 @@
             @SuppressWarnings("unchecked")
             final List<T> list = (List<T>) elements;
             return list;
-        } else if (elements instanceof ArrayList) {
+        } else if (elements instanceof ArrayList || elements instanceof Vector) {
             @SuppressWarnings("unchecked")
-            final ArrayList<T> arrayList = (ArrayList<T>) elements;
+            final java.util.List<T> indexedList = (java.util.List<T>) elements;
```
szarnekow: you could always use List.listIterator with previous() and hasPrevious() to traverse all kinds of j.u.List backwards. No need for special cases for ArrayList and Vector.

danieldietrich: @szarnekow That helps! Didn't know that.

Do you want to create a pull request or should I...?

szarnekow: see https://github.com/javaslang/javaslang/pull/310

PR #310, opened 2015-06-25
Minor optimization for List.allOf(NavigableSet) and List.allOf(j.u.List)

szarnekow:


danieldietrich:
Great!!

src/main/java/javaslang/collection/List.java
```diff
@@ -133,13 +136,19 @@
             @SuppressWarnings("unchecked")
             final List<T> list = (List<T>) elements;
             return list;
-        } else if (elements instanceof ArrayList || elements instanceof Vector) {
-            @SuppressWarnings("unchecked")
-            final java.util.List<T> indexedList = (java.util.List<T>) elements;
+        } else if (elements instanceof java.util.List<?>) {
```
danieldietrich: Hi Sebastian,
just a question - is there an advantage providing a `<?>` when using `instanceof`? I haven't seen it before. Also javac does not complain when omitting it (using `-Xlint:all -Werror`).

szarnekow: No, there is no advantage as far as I can tell. It's probably just me: when I see a generic type reference without any type parametrization I feel inclined to add that.

danieldietrich: Reflexes are a good thing - I also do that in different contexts. In general Java's type system is funny (sort of). IMO for example the assignment `List<T> list2 = list` should be valid, where `list` is of type `List<? extends T>`, but it does not compile. I do that in some locations of Javaslang and suppress the 'unchecked' warnings...

szarnekow: Yes, it would be great if the compiler could check that List<T> is effectively an immutable list. But since it's just an interface, it's impossible to prove that no instance of List<? extends T> could ever violate the statically known type of a List<T>

PR #312, opened 2015-06-27
Additional Try functionality. Adding Consumers & value chaining.

martincooper:
I've made some slight additions to the Try class, allowing a bit more flexibility.

Firstly, updating "andThen" to handle Consumers, allowing methods that return Void to also be able to accept the previous value EG...

```java
Try.of(() -> 100).thenRun(val -> System.out.println(val))
```

Leading on from that, I thought it would be extremely useful to perform a Try.Run (to perform some type of side effect) without losing your previous value. Currently...

```java
Try.of(() -> 100)
  .thenRun(val -> System.out.println(val)) // Do something with val which returns void.
  .thenRun(val -> val + 100) // ** Doesn't work. The previous "theRun" has lost val.
```
So, I modified the "thenRun" to (on success) return the previous value (Try[T] instead of Try[Void] ). This allows us to collect the value and use it with both Consumers and Suppliers, eg...

```java
Try<Integer> actual = Try.of(() -> new ArrayList<Integer>()) // Returns a new array list.
                .thenRun(arr -> arr.add(10)) // Calls a methods which returns a void, passing the array down.
                .thenRun(arr -> arr.add(30)) // Same...
                .thenRun(arr -> arr.add(20)) // Same
                .thenRun(arr -> System.out.println(arr.size()) //Same
                .then(arr -> arr.get(1)); // Still have access to the array here after all the "thenRun"'s
```
I renamed "andThen" to "then", adding a "thenRun" as it seemed a bit shorter than "andThenRun"

I hope this functionality seems like a useful change. 

Thanks

### DIFF
--- a/src/main/java/javaslang/control/Failure.java
+++ b/src/main/java/javaslang/control/Failure.java
@@ -150,14 +150,8 @@ public final class Failure<T> implements Try<T>, Serializable {
 
     @SuppressWarnings("unchecked")
     @Override
-    public <R> Failure<R> then(CheckedFunction1<T, R> f) {
+    public <R> Failure<R> mapTry(CheckedFunction1<T, R> f) {
         return (Failure<R>) this;
-    }
-
-    @SuppressWarnings("unchecked")
-    @Override
-    public Failure<T> thenRun(CheckedConsumer<T> f) {
-        return this;
     }
 
     @Override
--- a/src/main/java/javaslang/control/Failure.java
+++ b/src/main/java/javaslang/control/Failure.java
@@ -156,8 +156,8 @@ public final class Failure<T> implements Try<T>, Serializable {
 
     @SuppressWarnings("unchecked")
     @Override
-    public Failure<Void> thenRun(CheckedConsumer<T> f) {
-        return (Failure<Void>) this;
+    public Failure<T> thenRun(CheckedConsumer<T> f) {
+        return this;
     }
 
     @Override
--- a/src/main/java/javaslang/control/Failure.java
+++ b/src/main/java/javaslang/control/Failure.java
@@ -150,8 +150,14 @@ public final class Failure<T> implements Try<T>, Serializable {
 
     @SuppressWarnings("unchecked")
     @Override
-    public <R> Failure<R> mapTry(CheckedFunction1<T, R> f) {
+    public <R> Failure<R> mapTry(CheckedFunction1<? super T, ? extends R> f) {
         return (Failure<R>) this;
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public Failure<T> andThen(CheckedConsumer<? super T> f) {
+        return this;
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


danieldietrich:
Hi Martin,

nice feature! For simple logging/System.out there is a `peek(Consumer)` method but the checked case is very useful.

You previously added the nice `andThen(CheckedFunction1)`. As we see (and the compiler complains) `andThen(Consumer)` is potentially ambiguous with that. I agree that `andThenRun` is too long. The origin of the method `andThen` is from the class `java.util.function.Consumer`. I want to preserve this name to be consistent with the standard Java naming.

Let's solve the ambiguity by renaming `andThen(CheckedFunction1)` to `mapTry(CheckedFunction1)`. This is what it really is. Then, it would be natural to also have a `flatMapTry(CheckedFunction1)` (same signature as flatMap but using a CheckedFunction1).

The other `andThen` methods remain `andThen(CheckedRunnable)` and `andThen(CheckedConsumer)`.

Please see also my notes on the code (in a sec).

Thanks

Daniel

src/main/java/javaslang/control/Failure.java
```diff
@@ -150,10 +150,16 @@ public void orElseRun(Consumer<? super Throwable> action) {
 
     @SuppressWarnings("unchecked")
     @Override
-    public <R> Failure<R> andThen(CheckedFunction1<T, R> f) {
+    public <R> Failure<R> then(CheckedFunction1<T, R> f) {
```
danieldietrich: will be mapTry

src/main/java/javaslang/control/Failure.java
```diff
@@ -150,10 +150,16 @@ public void orElseRun(Consumer<? super Throwable> action) {
...
         return (Failure<R>) this;
     }
 
+    @SuppressWarnings("unchecked")
+    @Override
+    public Failure<T> thenRun(CheckedConsumer<T> f) {
```
danieldietrich: Will be `andThen(CheckedConsumer<? super T> f)`. Please note the generic type bounds.

src/main/java/javaslang/control/Success.java
```diff
@@ -153,11 +153,17 @@ public void orElseRun(Consumer<? super Throwable> action) {
...
     @Override
-    public <R> Try<R> andThen(CheckedFunction1<T, R> f) {
+    public <R> Try<R> then(CheckedFunction1<T, R> f) {
         return flatMap(value -> Try.of(() -> f.apply(value)));
     }
 
     @Override
+    public Try<T> thenRun(CheckedConsumer<T> f) {
+        Try<Void> result = flatMap(value -> Try.run(() -> f.accept(value)));
+        return result.isSuccess() ? this : new Failure<T>(result.failed().get());
```
danieldietrich: This implementation is a bit shorter - we do not need a variable:

```java
public Try<T> andThen(CheckedConsumer<? super T> consumer) {
    return Try.run(() -> consumer.accept(value)).flatMap(ignored -> this);
}
```

martincooper:
Hi Daniel

I've modified the code according to your comments. I hope i've got all the method names as you requested. 

I wasn't entirely sure what the implementation of ```flatMapTry(CheckedFunction1)``` would be, so I've left that out for now.

Thanks!

Martin

danieldietrich:
I like it to see Try emerging. Thanks for your contribution!

Issue #302, opened 2015-06-22
Match to allow Scala style matching on a range of values.


martincooper:
It would be nice to extend Match functionality to allow a range of possible values to match on. An example being the following case in Scala...

```scala
def toBool(choice: String): Boolean = choice match {
    case "Y" | "YES" | "T" | "TRUE" | "1" => true
    case "N" | "NO" | "F" | "FALSE" | "0" => false
    case _ => false
  }
```
Currently this would require multiple 'caze' statements and is a bit more verbose. EG:
```java
Match
	.caze("Y", o -> true)
	.caze("YES", o -> true)
	.caze("T", o -> true)
	.caze("TRUE", o -> true)
	.caze("1", o -> true)
    .caze("N", o -> false) ... etc
```
Maybe allowing...
```java
Match.caze("Y", "YES", "T", o -> true)
```
Or if this is not possible, something like...
```java
Match.cazeIn(new String[] { "Y", "YES", "T", "TRUE" }, o -> true)
```

danieldietrich:
See PR #315

danieldietrich:
Now we are able to write:

```java
Match
    .when(1).then(i -> i)   // before: when(1, i -> i)
    .when(2).then(() -> 1)  // before: when(2, ignored -> 1)
    .whenIn(3, 4, 5).then(i -> i)
    .whenIn(6, 7, 8).then(() -> 1)
    .whenNull().then(n -> 1)
    .whenNull().then(() -> 1)
    .apply(1000);
```

In the meantime the Match API emerged. Instead of using it as function (as shown above), we now may use it as Monad by capturing the value to be matched at the beginning with 'of'.

The new when().then() / whenIn().then() syntax has also to be added to the Match.of() branch of the API. Then we will also able to write:

```java
Match.of(1000)
    .when(1).then(i -> i)   // before: when(1, i -> i)
    .when(2).then(() -> 1)  // before: when(2, ignored -> 1)
    .whenIn(3, 4, 5).then(i -> i)
    .whenIn(6, 7, 8).then(() -> 1)
    .whenNull().then(n -> 1)
    .whenNull().then(() -> 1)
    .get(); // .orElse(...) .orThrow(...) etc.
```

danieldietrich:
**Note:**

Because of ambiguities we cannot use `when(null).then(...)`. Therefore `whenNull().then(...)` was introduced.

[...]

PR #315, opened 2015-06-29
Adding matching on a range of values.

martincooper:
Hi Daniel,

I've implemented the matching which I raised an issue for the other day. I think this looks like a reasonable way to do it. 

```java
        final String result = Match
                .whenIn(new Integer[] { 1, 3, 5, 7 }, x -> "true")
                .whenIn(new Integer[] { 2, 4, 6, 8 }, x -> "false")
                .when(10, x -> "other")
                .apply(6);

        assertThat(result).isEqualTo("false");
```
The array creation inside the "whenIn" is a bit more verbose than I would have liked. It's a pity I can't use ```T... prototypes``` as the argument for the method as it's not the last argument.

Maybe something along the lines of this would be nice...

        final String result = Match
                .whenIn(1, 3, 5, 7).then(x -> "true")
                .whenIn(2, 4, 6, 8).then(x -> "false")
                .when(10).then(x -> "other")
                .apply(6);

Although this would most likely require a significant re-work.

Thanks

Martin

### DIFF
--- a/src-gen/main/java/javaslang/control/Match.java
+++ b/src-gen/main/java/javaslang/control/Match.java
@@ -11,7 +11,6 @@ package javaslang.control;
 
 import java.io.Serializable;
 import java.lang.invoke.MethodType;
-import java.util.Arrays;
 import java.util.Objects;
 import java.util.function.Function;
 import java.util.function.Predicate;
@@ -19,7 +18,6 @@ import java.util.function.Supplier;
 import javaslang.Function1;
 import javaslang.Lazy;
 import javaslang.collection.List;
-import javaslang.collection.Stream;
 
 /**
  * {@code Match} is a better switch for Java. Some characteristics of {@code Match} are:
--- a/src-gen/main/java/javaslang/control/Match.java
+++ b/src-gen/main/java/javaslang/control/Match.java
@@ -11,6 +11,7 @@ package javaslang.control;
 
 import java.io.Serializable;
 import java.lang.invoke.MethodType;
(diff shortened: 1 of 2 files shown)


danieldietrich:
Hi Martin,

this looks good! I like the `whenIn(...).then(...)` and `when(...).then(...)` syntax.

It seems you changed the generated `Match` class. Changes will be overwritten when re-generating sources, e.g. by calling `mvn clean generate-sources`. I'm currently also changing the `Match` generator (see #227) and about to merge it into master this evening (there will be a new `Match.of(...)` syntax that allows strong typed match expressions and monadic behavior like `map`, `flatMap`, `filter`).

If you don't mind, I will rework the `Match` generator according to your suggestion, because I'm already on it (see [Generator.scala](https://github.com/danieldietrich/javaslang/blob/17319a1123767d864562b190637a1d79f3d6ee2b/generator/Generator.scala#L39-L885)).

What do you think?

\- Daniel

martincooper:
Hi Daniel

I didn't realise the Match class was generated :astonished:

Yep, it's cool for you to make these changes as you're already working on it. I look forward to testing it out after. :smiley:

Thanks!

danieldietrich:
Thanks for making Match awesome! I will push a first version until tomorrow...

danieldietrich:
> I will push a first version until tomorrow...

Need sleep :) Will go on tomorrow...

danieldietrich:
The new syntax is now in place, details provided in #302.

Issue #318, opened 2015-06-29
Get lambda reflection right


danieldietrich:
A Java lambda expression does capture all effectively final variables. Internally these are _hidden_ lambda arguments.

&lambda;.getType() returns a `MethodType` reflecting the parameter count, the return type and the parameter types of the lambda expression / the SAM.

In order to get lambda reflection right, we need to remove the internally added _hidden_ variables.

danieldietrich:
Note: `Match` is currently the only 'user' of lambda reflection

Issue #277, opened 2015-05-26
scala-maven-plugin: "character to be escaped is missing" on win7/x64


danieldietrich:
@chagmed reported an UTF-8 issue with building Javaslang 1.2.3 on a Win7-x64 box. The scala-maven-plugin 3.2.0 reported: "character to be escaped is missing". More specifically: "IO error while decoding javaslang\target\.scalaScriptGen\Generator_1.scala with UTF-8"

danieldietrich:
Are there any win7/x64 users here which are able to compile the Javaslang Github repo on their machine via maven? Need some feedback!

danieldietrich:
I could reproduce this using VirtualBox with Win7/x64, Java 1.8.0_45, git-bash and maven 3.3.3.

Setting the scalac file encoding in the pom.xml did not help:

```shell
<args>
    <arg>-encoding</arg>
    <arg>UTF8</arg>arg>
</args>
```

This needs to be further investigated.

ignasi35:
Hi @danieldietrich ,  

I did a quick native2ascii and apart from comments there's a ' lambda-sign' that got escaped. Could that be it?

```
(excerpt)
712c712
<           public interface $className$fullGenerics extends λ<R>$additionalExtends {
---
>           public interface $className$fullGenerics extends \u03bb<R>$additionalExtends {
```

Sorry I can't test it on Win machine at the moment.

Cheers,




PS: full diff:

```

511[generator](master)$ /Library/Java/JavaVirtualMachines/jdk1.8.0_40.jdk/Contents/Home/bin/native2ascii Generator.scala > out

512[generator](master)$ diff Generator.scala out
642c642
<               // ¬(p => q) ≡ ¬(¬p ∨ q) ≡ p ∧ ¬q
---
>               // \u00ac(p => q) \u2261 \u00ac(\u00acp \u2228 q) \u2261 p \u2227 \u00acq
712c712
<           public interface $className$fullGenerics extends λ<R>$additionalExtends {
---
>           public interface $className$fullGenerics extends \u03bb<R>$additionalExtends {
1197c1197
<                 // (∀a,b ∈ ℝ+ ∃c ∈ ℝ+ : a²+b²=c²) ≡ (∀a,b ∈ ℝ+ : √(a²+b²) ∈ ℝ+)
---
[...]

Issue #87, opened 2015-01-22
Match is (probably) a Monad


danieldietrich:
Make match an interface with `caze` factory methods - it's like an 'extended builder pattern'.

```java
@FunctionalInterface
interface Match<R> extends Function1<Object, R>, Monad<R, Match<?>> /*, Univalent */ {

    static Case<R> caze(...) {
        return new Case<>(...);
    }

    // matches o
    @Override
    R apply(Object o);

    @Override
    <U> Match<U> map(Function<? super R, ? extends U> f);

    @Override
    default <U, MATCH extends HigherKinded<U, Match<?>>> Match<U> flatMap(Function<? super R, MATCH> f) {
        return o -> f.apply(Match.this.apply(o));
    }

    Match<R> filter(Predicate<? super R> predicate);

    static class Case<R> extends Match<R> {
        final Stream<...> cases;
    
        Case(Stream<...> cases) {
            this.cases = cases;
        }

        Case<R> caze(...) {
            return new Case<>(cases.append(caze));
        }

        OrElse<R> orElse(...) {
            return new OrElse<>(cases.append(orElse));
        }

        @Override
        public <U> Match<U> map(Function<? super R, ? extends U> f) {
            return new Case<>(/*something like this*/ o -> f.apply(apply(o)));
        }
    }

    static class OrElse extends Match<R>

        final Stream<...> cases;

        OrElse(Stream<...> cases) {
            this.cases = cases;
        }

        // no caze(), orElse() - just apply() possible

        @Override
        public <U> Match<U> map(Function<? super R, ? extends U> f) {
            return new OrElse<>(/*something like this*/ o -> f.apply(apply(o)));
        }
    }

} // end of Match
```

danieldietrich:
With #227 Match will implement Monad

danieldietrich:
Follow-up: #317

PR #319, opened 2015-06-29
Adds Match.of() with monadic properties

danieldietrich:
Fixes

- #277 Howto process further with Match.orElse()
- #87 Match is (probably) a Monad
- #318 Get lambda reflection right


danieldietrich:
In order to eventually reach the next milestone, I now concentrate on adding features. Full coverage will be restored before the release.

PR #336, opened 2015-07-18

jorander:
…llections

Solution for problem #5.


danieldietrich:
Thank you Jörgen, very nice!
Sorry for the delayed answer, I just returned from a journey.
Have a great week,
Daniel

PR #339, opened 2015-07-22

jorander:
…llections

Solution for problem #6.


danieldietrich:
thumb-ub

PR #366, opened 2015-07-27
Stream.appendSelf(Function<Stream,Stream> mapper)

ruslansennov:
I believe we're a little closer to scala now :)


danieldietrich:
Wow, really nice. Never heard of this, yet.
I ask myself if a similar/slightly different functionality could be used to calculate primes with the sieve of eratosthenes?

ruslansennov:
I'm not sure, but I think the realization of this algorithm in a functional style requires a lot of memory.
But with addSelf() method searching for prime numbers may become easier
```scala
val primes: Stream[Int] = 2 #:: Stream.from(3, 2).filter(i => 
    primes.takeWhile(j => j * j <= i).forall(k => i % k > 0)
)
```
We can make refactoring PrimeNumbers.java because I believe it's completely non-obvious

danieldietrich:
Sound good. Would be great if you can do it!

ruslansennov:
#370 done

Issue #388, opened 2015-07-30
Make travis-ci build faster


danieldietrich:
With evolving Euler tests, the travis-ci build takes longer and longer...

**Task:** Analyse possibilities to make the build faster.

danieldietrich:
Fixed (~ 35% faster)

PR #389, opened 2015-07-30
Maven: skip unit tests where possible

danieldietrich:
Towards #388 Make travis-ci build faster


danieldietrich:
It is not ~ 35% faster.

PR #397, opened 2015-07-30
Stream.containsSlice(Iterable)

ruslansennov:


danieldietrich:
Our collection lib starts to get a great toolbox!

danieldietrich:
@ruslansennov I [added](https://github.com/javaslang/javaslang/pull/399) Traversable.unit() because I need it for a Monad abstraction. Here it comes handy to hide impl details from the public API.

Btw. I like your unit tests - all of them, not only these! The are very mature. Good tests help so much when refactoring code.

ruslansennov:
I thought about something similar to Traversable.unit(), thank you

PR #400, opened 2015-07-30
#126 Solution for problem 11.

ruslansennov:


danieldietrich:
Awesome, also the Javadoc!

PR #401, opened 2015-07-30
Renamed Traversable.unit() to .cons()

danieldietrich:
because `unit()` and `join()` are names from the Haskell world. We use `cons()´ and `flatMap()` instead.

### DIFF
--- a/src/main/java/javaslang/collection/List.java
+++ b/src/main/java/javaslang/collection/List.java
@@ -1202,7 +1202,7 @@ public interface List<T> extends Seq<T>, Stack<T> {
     }
 
     @Override
-    default List<T> unit(Iterable<? extends T> iterable) {
+    default List<T> cons(Iterable<? extends T> iterable) {
         return List.ofAll(iterable);
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/javaslang/collection/List.java
```diff
@@ -1202,7 +1202,7 @@ default T peek() {
     }
 
     @Override
-    default List<T> unit(Iterable<? extends T> iterable) {
+    default List<T> cons(Iterable<? extends T> iterable) {
```
mperry: @danieldietrich I'm not sure cons is a good term for this method.  Cons usually takes an explicit head and tail.

danieldietrich: @mperry Thank you for the hint! I will revert the change.

danieldietrich: @mperry What do you think of having a `Monad` / `FilterMonadic` interface with methods (in conjunction with the `Kind`for semi-higher order kinds`):

* `flatMapM`
* `flattenM`

Have these methods a different meaning, too, in other languages?

I can't use the original names `flatMap` and `flatten` because they have a different, relaxed signature (with Iterable instead of Monad/FilterMonadic for  Traversables).

Thanks in advance!

PR #404, opened 2015-07-31
Seq.indexOf(element,start); Seq.lastIndexOf(element,end)

ruslansennov:
+ unit tests
+ missing unit tests

### DIFF
--- a/src/main/java/javaslang/collection/List.java
+++ b/src/main/java/javaslang/collection/List.java
@@ -841,8 +841,8 @@ public interface List<T> extends Seq<T>, Stack<T> {
     @Override
     default int lastIndexOf(T element, int end) {
         int result = -1, index = 0;
-        for (List<T> list = this; !list.isEmpty(); list = list.tail(), index++) {
-            if (index <= end && Objects.equals(list.head(), element)) {
+        for (List<T> list = this; index <= end && !list.isEmpty(); list = list.tail(), index++) {
+            if (Objects.equals(list.head(), element)) {
                 result = index;
             }
         }
--- a/src/main/java/javaslang/collection/List.java
+++ b/src/main/java/javaslang/collection/List.java
@@ -776,10 +776,10 @@ public interface List<T> extends Seq<T>, Stack<T> {
     }
 
     @Override
-    default int indexOf(T element) {
+    default int indexOf(T element, int from) {
         int index = 0;
         for (List<T> list = this; !list.isEmpty(); list = list.tail(), index++) {
-            if (Objects.equals(list.head(), element)) {
+            if (index >= from && Objects.equals(list.head(), element)) {
                 return index;
             }
         }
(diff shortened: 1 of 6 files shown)


src/main/java/javaslang/collection/Stream.java
```diff
@@ -927,10 +927,10 @@ default int indexOf(T element) {
     Stream<T> intersperse(T element);
 
     @Override
-    default int lastIndexOf(T element) {
+    default int lastIndexOf(T element, int end) {
         int result = -1, index = 0;
         for (Stream<T> stream = this; !stream.isEmpty(); stream = stream.tail(), index++) {
```
danieldietrich: Maybe the for condition `!stream.isEmpty()` can be extended to `index <= end && !stream.isEmpty()`? ...

src/main/java/javaslang/collection/Stream.java
```diff
@@ -927,10 +927,10 @@ default int indexOf(T element) {
...
-            if (Objects.equals(stream.head(), element)) {
+            if (index <= end && Objects.equals(stream.head(), element)) {
```
danieldietrich: ... then we can remove the check here and the loop ends, if `index > end`.

Would this be the same semantic?

src/main/java/javaslang/collection/Queue.java
```diff
@@ -645,13 +645,13 @@ public T head() {
...
-    public int indexOf(T element) {
-        final int frontIndex = front.indexOf(element);
+    public int indexOf(T element, int from) {
+        final int frontIndex = front.indexOf(element, from);
         if (frontIndex != -1) {
             return frontIndex;
         } else {
             // we need to reverse because we search the first occurrence
-            final int rearIndex = rear.reverse().indexOf(element);
+            final int rearIndex = rear.reverse().indexOf(element, from - front.length());
```
danieldietrich: Did we check the case that `front.length() > from`. Just asking for the case ... :) I didn't dig deep into it...

src/main/java/javaslang/collection/List.java
```diff
@@ -839,10 +839,10 @@ default int indexOf(T element) {
     }
 
     @Override
-    default int lastIndexOf(T element) {
+    default int lastIndexOf(T element, int end) {
         int result = -1, index = 0;
         for (List<T> list = this; !list.isEmpty(); list = list.tail(), index++) {
-            if (Objects.equals(list.head(), element)) {
+            if (index <= end && Objects.equals(list.head(), element)) {
```
danieldietrich: See my comment at Stream.java below (for-loop)

src/main/java/javaslang/collection/Stream.java
ruslansennov: Of course you're right, shame on me :)

src/main/java/javaslang/collection/Queue.java
ruslansennov: I believe it's all right here... we just continue searching in rear part of queue with start < 0

danieldietrich: ok, thx!

src/main/java/javaslang/collection/Stream.java
danieldietrich: Don't shame on you - we team up to get the best result. There is nearly always room for improvement.

src/main/java/javaslang/collection/List.java
ruslansennov: fixed in same commit

danieldietrich:
Great work!

PR #405, opened 2015-07-31
Traversable.[last]indexOfSlice(Iterable[,int])

ruslansennov:


src/main/java/javaslang/collection/Traversable.java
```diff
@@ -670,6 +722,95 @@ default T last() {
...
+
+    /**
+     * Finds last index where this sequence contains a given sequence as a slice.
+     * <p>
+     * Note: will not terminate for infinite-sized collections.
+     * @param that the sequence to test
+     * @return the last index such that the elements of this sequence starting a this index match the elements
+     * of sequence that, or -1 of no such subsequence exists.
+     */
+    default int lastIndexOfSlice(Iterable<T> that) {
```
danieldietrich: It makes sense to widen the type bounds here. Instead of `Iterable<T>` I would use `Iterable<? extends T>`.

src/main/java/javaslang/collection/Traversable.java
```diff
@@ -670,6 +722,95 @@ default T last() {
...
+    }
+
+    /**
+     * Finds last index before or at a given end index where this sequence contains a given sequence as a slice.
+     * @param that the sequence to test
+     * @param end the end index
+     * @return the last index &lt;= end such that the elements of this sequence starting at this index match
+     * the elements of sequence that, or -1 of no such subsequence exists.
+     */
+    default int lastIndexOfSlice(Iterable<T> that, int end) {
```
danieldietrich: Also here `Iterable<? extends T>`

src/main/java/javaslang/collection/Traversable.java
```diff
@@ -670,6 +722,95 @@ default T last() {
...
+        class Util {
+            int lastIndexOfSlice(Traversable<T> t, Traversable<T> slice, int end) {
```
danieldietrich: but here _not_ - within Util we know that we have exactly type `T` (also the slice has the right type returned by `unit()`).

src/main/java/javaslang/collection/Stream.java
```diff
@@ -1541,11 +1541,6 @@ private Nil() {
         }
 
         @Override
-        public boolean containsSlice(Iterable<? extends T> that) {
```
danieldietrich: For user-friendly error messages I check for null when args are other than simple values.

Here I would insert `Objects.requireNonNull(that, "that is null");`

and also declare a javadoc param `@throws NullPointerException if that is null`.

src/main/java/javaslang/collection/Traversable.java
```diff
@@ -553,6 +538,73 @@ default T fold(T zero, BiFunction<? super T, ? super T, ? extends T> op) {
...
+    /**
+     * Finds first index where this sequence contains a given sequence as a slice.
+     * <p>
+     * Note: may not terminate for infinite-sized collections.
+     * @param that the sequence to test
+     * @return the first index such that the elements of this sequence starting at this index match
+     * the elements of sequence that, or -1 of no such subsequence exists.
+     */
+    default int indexOfSlice(Iterable<? extends T> that) {
+        return indexOfSlice(that, 0);
```
danieldietrich: Also here I would insert `Objects.requireNonNull(that, "that is null");`

and also declare a javadoc param `@throws NullPointerException if that is null`.

src/main/java/javaslang/collection/Traversable.java
```diff
@@ -553,6 +538,73 @@ default T fold(T zero, BiFunction<? super T, ? super T, ? extends T> op) {
...
+    /**
+     * Finds first index after or at a start index where this sequence contains a given sequence as a slice.
+     * <p>
+     * Note: may not terminate for infinite-sized collections.
+     * @param that the sequence to test
+     * @param from the start index
+     * @return the first index >= from such that the elements of this sequence starting at this index match
+     * the elements of sequence that, or -1 of no such subsequence exists.
+     */
+    default int indexOfSlice(Iterable<? extends T> that, int from) {
```
danieldietrich: Also here I would insert `Objects.requireNonNull(that, "that is null");`

and also declare a javadoc param `@throws NullPointerException if that is null`.

src/main/java/javaslang/collection/Traversable.java
danieldietrich: Also here I would insert `Objects.requireNonNull(that, "that is null");`

and also declare a javadoc param `@throws NullPointerException if that is null`.

src/main/java/javaslang/collection/Traversable.java
danieldietrich: Also here I would insert `Objects.requireNonNull(that, "that is null");`

and also declare a javadoc param `@throws NullPointerException if that is null`.

danieldietrich:
Many thanks, great addition! It is good to have a _fresh_ mind looking over the / and adding new functionality.

src/main/java/javaslang/collection/Stream.java
danieldietrich: Oh - just recognized that this line was removed :)

ruslansennov:
it's typical for me to forget <? extends T>

danieldietrich:
Me too!

src/main/java/javaslang/collection/Traversable.java
```diff
@@ -553,6 +540,77 @@ default T fold(T zero, BiFunction<? super T, ? super T, ? extends T> op) {
...
+     * Finds first index after or at a start index where this sequence contains a given sequence as a slice.
+     * <p>
+     * Note: may not terminate for infinite-sized collections.
+     * @param that the sequence to test
+     * @param from the start index
+     * @return the first index >= from such that the elements of this sequence starting at this index match
+     * the elements of sequence that, or -1 of no such subsequence exists.
+     * @throws NullPointerException if {@code that} is null.
+     */
+    default int indexOfSlice(Iterable<? extends T> that, int from) {
```
szarnekow: This could even be Iterable<?> rather than Iterable<?>. Consider this scenario:
```java
List<CharSequences> that = Arrays.asList("a", "b");
Traversable<String> traversable = ..
traversable.indexOf(that, 0)
```
there could be a valid index for the given list `that`. On the other hand, it'll open the door for the same kind of bug-prone signatures as `java.util.List<T>.containsAll(Collection<?> c);`.

danieldietrich: @szarnekow Good Morning! Did you mean

> `Iterable<?>` rather than `Iterable<? extends T>`

or

> `Iterable<? super T>` rather than `Iterable<? extends T>`

We need the type-safety in general I think. It would be possible to first 'downgrade' the traversable, e.g.

```java
traversable                     // is a Traversable<String>
        .as(CharSequence.class) // returns a Traversable<CharSequence>
        .indexOfSlice(that, 0)
```

Unfortunately it is not possible in Java to define a generic lower bound `<U super T>` (which I do not understand but which will have to do with the compiler algorithm to that finds correct bounds):

```java
<U super T> Traversable<U> as(Class<U> type) {
    return (Traversable<U>) this;
}
```

That means that `Iterable<?>` would be the only solution to allow lower bounds (`Iterable<? super T> that`) and upper bounds (`Iterable<? extends T> that`).

My impression also is, that `Iterable<?>` will be error-prone.

szarnekow: Hi @danieldietrich ,
I meant `Iterable<?>` which would indeed be a workaround for the missing `<U super T>` in Java. I think in general it's find to have type constraint on `indexOf`, `containsAll` and the like, even though it may render certain alien usage patterns cumbersome. It's just something that should be considered briefly when an API is designed to make sure we don't miss some usage scenarios.
Best, Sebastian

danieldietrich: @szarnekow yes, haven't thought of these cases... Good to have your X-Files ([Xtext](https://eclipse.org/Xtext) &amp; [Xtend](https://eclipse.org/xtend)) know-how here! :-)

PR #407, opened 2015-08-01
Refactored Function*.getType(). Got interface/class design right.

danieldietrich:
We traded the hiding of Lambda serialization for code repetition (see Function*.Type). It is ok for now, the repetitive code is generated automatically and defined only once within `Generator.scala`.

However, I added a TODO to find a better solution.

### DIFF
--- a/src/main/java/javaslang/λ.java
+++ b/src/main/java/javaslang/λ.java
@@ -117,7 +117,9 @@ public interface λ<R> extends Serializable {
      *
      * @param <R> Result type of the function
      * @since 2.0.0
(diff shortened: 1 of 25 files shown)


src/main/java/javaslang/λ.java
```diff
@@ -173,7 +105,103 @@ public String toString() {
          *
          * @return the parameter types
          */
-        Class<?>[] parameterArray();
+        Class<?>[] parameterTypes();
```
danieldietrich: The array is intentionally not cloned/duplicated, i.e. it may be modified. However, each new call of `getType()` will create a new parameterTypes array.

Mutation is generally discouraged (by me). Don't do that - or at most carefully.

Issue #288, opened 2015-05-31
Add substitute for removed Monad interface


danieldietrich:
The `Monad` interface was mainly used to standardize `map` and `flatMap` across types. This did not work well because we _relaxed_ the flatMap signature for Traversables (`flatMap(Function<? super T, ? extends Iterable<T>>`), which in turn did not work well with other monadic types like `Try`, `Option`, etc.

Changing the `flatMap` signature of Traversables made them semi-monadic, because the original monad laws are not applicable any more but structural properties keep roughly the same.

Scala uses the trait [FilterMonadic](http://www.scala-lang.org/api/2.11.5/index.html#scala.collection.generic.FilterMonadic) to express properties which are needed for _for-comprehensions_. It would be nice, if Javaslang would have also one super-type for (semi-)monadic types.

The `flatMap` signature of such a type would have to be compatible with the existing `flatMap` definitions. Or the signature is extended by an instance builder, which is in Scala currently implicit and needed to convert an `Iterable` to an instance of the current type, e.g. `Option` (which may only contain one element). If this problem is solved, we may also have a common super interface `FilterMonadic`.

danieldietrich:
Check if `peek()` fits in here, too.

danieldietrich:
Additionally to `FilterMonadic` the `javaslang.control` types and the upcoming `javaslang.concurrent.Future` type need a common interface. I recently deleted `Valences.Univalent` and `Valences.Bivalent` which were meant to be common interfaces for these types. But the names are unfamiliar.

We need a simple name that reflects that we can get a value or else provide an alternative or throw an exception. Some of these values are of one type (univalent) or of one of two possible types/states (bivalent).

Possible names for such an interface are

* `Value`
* `Val`
* `Getable`
* `GetOrElse`

I prefer `Value` because it is the basis of all functional programs.

Values are either present or absent (reflected by `isEmpty` and maybe `isDefined`). Functions are transformations of values.

Examples:

```java
Value<T> val1 = Option.of(...);

Value<T> val2 = Try.of(...);

Value<T> val3 = Match.of(...).when(...);

T t = value.get()
```

danieldietrich:
What do you think of having a `Monad` / `FilterMonadic` interface with methods (in conjunction with the `Kind` for semi-higher order kinds):

* `flatMapM`
* `flattenM`

Have these methods a different meaning, too, in other languages?

[...]

Issue #170, opened 2015-04-19
Add an immutable HashMap to the javaslang.collection package


danieldietrich:
Interesting read: http://javarevisited.blogspot.de/2011/02/how-hashmap-works-in-java.html

danieldietrich:
Maps work best backed by a Trie, see "Purely Functional Data Structures", p. 163 ff.

danieldietrich:
Good read on functional/immutable Map: http://blog.ezyang.com/2010/03/the-case-of-the-hash-array-mapped-trie/

danieldietrich:
Ideal Hash Trees (Phil Bagwell/Clojure): http://lampwww.epfl.ch/papers/idealhashtrees.pdf

(See also wikipedia [Trie](http://en.wikipedia.org/wiki/Trie) and [Hash array mapped trie](http://en.wikipedia.org/wiki/Hash_array_mapped_trie))

danieldietrich:
Fast mergable integer maps: http://citeseer.ist.psu.edu/viewdoc/download?doi=10.1.1.37.5452&rep=rep1&type=pdf

danieldietrich:
@ignasi35 Here I provided some links for the Map implementation. I think, Map is one of the most complex collections to implement.

Without having read on Maps yet, a naive impl would be to use a balanced binary search tree with (informally) `Entry(key.hashCode, List(values))`, where `equals` and `compareTo` of an `Entry` just compare the `key.hashCode` and not the values. Given that, a tree node could be found just by the `hashCode` and the value could be obtained from the associated List. We need a List of values at each node because of hash collisions.

But my understanding is that `HashMap.get()` performs in O(1). The above impl would implement in `O(log n)`...

Better implementations use the so called `Trie` data structure with 'marked' edges (and values at the leafs only?). This has to be further investigated... How does Java implement (mutable) HashMap? How does Scala implement HashMap?

ignasi35:
Some more tips on the subject. http://stackoverflow.com/a/9054660

ruslansennov:
@danieldietrich  please review this HAMT implementation ruslansennov/javaslang@b9fcfb185175b02be7ddc99e8641309f178f6942
What other methods need to be implemented?

danieldietrich:
@ruslansennov Wow, Ruslan, I'm really impressed. This is \*exactly\* what we need. This is the best implementation I can imagine - head to head with Clojure and Scala.

I will add some line notes in a few minutes.

[...]

PR #409, opened 2015-08-02
Towards FilterMonadic and Map

danieldietrich:
Towards #288 
Towards #170


danieldietrich:
@ruslansennov I've made some slight modifications - I hope this does not cause merge-conflicts at your fork. I'm now jumping on the 'collection' train. First I setup the FilterMonadic stuff, then Set interface, then SortedSet based on Red-Back-Tree...

HashSet should be relatively easy because it is backed on a HashMap, right?

ruslansennov:
@danieldietrich I believe any <code>Set</code> is a scecial case of <code>Map</code> :)

danieldietrich:
@ruslansennov Technically yes, and Tree is a List of Lists :)

PR #410, opened 2015-08-03
Move some methods from Traversable to Seq

ruslansennov:
Because <code>Traversable</code> will be extended by <code>Map</code> in near future


danieldietrich:
Great, that makes sense!

PR #418, opened 2015-08-04
extract HashSet & HashMap common code

ruslansennov:


src/main/java/javaslang/collection/HashMap.java
```diff
@@ -6,434 +6,74 @@
...
-import java.io.Serializable;
-import java.util.Objects;
+import java.util.Iterator;
 
 /**
  * A {@code HashMap} implementation based on a
  * <a href="https://en.wikipedia.org/wiki/Hash_array_mapped_trie">Hash array mapped trie (HAMT)</a>.
  */
-public interface HashMap<K, V> extends Map<K, V> {
+class HashMap<K, V> implements Map<K, V> {
```
danieldietrich: _Just some thoughts for the future:_

Maybe there will be a future refactoring and we will make collection classes interfaces. This enables us to use lazy values:

```java
// real lazy value: evaluated and memoized on demand!
HashMap<?, ?> map = Lazy.asVal(() -> HashMap.empty(), HashMap.class);
```

**But for now lets use classes!**

src/main/java/javaslang/collection/HashMap.java
```diff
@@ -6,434 +6,74 @@
...
 
     static <K, V> HashMap<K, V> empty() {
-        return EmptyNode.instance();
+        return new HashMap<>(HashTree.empty());
```
danieldietrich: Maybe it makes sense to make the empty HashMap a singleton, too. It will be frequently used.

src/main/java/javaslang/collection/HashTree.java
```diff
@@ -0,0 +1,404 @@
...
+import javaslang.control.Option;
+
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.Iterator;
+
+/**
+ * <a href="https://en.wikipedia.org/wiki/Hash_array_mapped_trie">Hash array mapped trie (HAMT)</a>.
+ */
+abstract class HashTree<K, V> implements Iterable<Tuple2<K, V>> {
```
danieldietrich: Maybe we should name it `HashArrayMappedTrie` (as [specialization of `HashTree`](https://en.wikipedia.org/wiki/Hash_tree_(persistent_data_structure))). This is your original name. I will most probably also do this with `RedBlackTree`, which is a `BinarySearchTree`.

src/main/java/javaslang/collection/HashMap.java
ruslansennov: Yes, I watch closely your commits

src/main/java/javaslang/collection/Map.java
```diff
@@ -6,16 +6,78 @@
...
+            return (HashMap<K, V>) entries;
+        } else {
+            HashMap<K, V> map = HashMap.empty();
+            for (Entry<? extends K, ? extends V> entry : entries) {
+                map = map.put(entry.key, entry.value);
+            }
+            return map;
+        }
+    }
+
```
danieldietrich: Please let's have static factory methods in _terminal collection interfaces/classes_ (like List, Stream, HashMap, ...) which are like classes. In the image these are the leafs.

classes-overview

I'm not sure, if it helps to have in other interfaces static factory methods. My experience is, that it feels a little bit magic, when users are not familiar with the collection lib and don't know which instance they get. It is better to be explicit:

```java
Map<?, ?> map = HashMap.empty().put("Hi", "Ruslan");
```

src/main/java/javaslang/collection/TraversableOnce.java
```diff
@@ -109,16 +109,16 @@ default void forEach(Consumer<? super T> action) {
...
-     * Converts this TraversableOnce to a HashMap.
+     * Converts this TraversableOnce to a Map.
      *
      * @param f   A function that maps an element to a Map.Entry
      * @param <K> The key type of a Map Entry
      * @param <V> The value type of a Map Entry
      * @return a new HashMap containing the elements mapped to entries
      */
-    default <K, V> HashMap<K, V> toHashMap(Function<? super T, ? extends Map.Entry<? extends K, ? extends V>> f) {
+    default <K, V> Map<K, V> toMap(Function<? super T, ? extends Map.Entry<? extends K, ? extends V>> f) {
```
danieldietrich: Here I thought of `toHashMap` and later add `toTreeMap`. But I start to see, what you may have in mind. Alternatively we could provide a `toMap` and `toSortedMap`, which is great, too. On the other hand is always good to be as specific as possible. What do you think?

danieldietrich:
Nice refactoring!

src/main/java/javaslang/collection/TraversableOnce.java
ruslansennov: It is not so principal, but I prefer <code>toMap</code> and <code>toSortedMap</code>

src/main/java/javaslang/collection/HashMap.java
danieldietrich: Thanks for being uncomplaining regarding all my commits. I bet you had some merge conflicts the last days... It is under heavy development :-)

src/main/java/javaslang/collection/TraversableOnce.java
danieldietrich: Then let's do it with toMap and toSortedMap!

danieldietrich:
You are fast, I even haven't brewed my first coffee :)

PR #420, opened 2015-08-05
HashSet implementation

ruslansennov:


src/main/java/javaslang/collection/HashArrayMappedTrie.java
```diff
@@ -17,7 +17,7 @@
 /**
  * <a href="https://en.wikipedia.org/wiki/Hash_array_mapped_trie">Hash array mapped trie (HAMT)</a>.
  */
-abstract class HashArrayMappedTrie<K, V> implements Iterable<Tuple2<K, V>> {
+abstract class HashArrayMappedTrie<K, V> implements Iterable<Tuple2<K, V>>, Serializable {
```
danieldietrich: Hi, I think the build fails because the serialVersionUID = 1L is missing...

danieldietrich:
Wow, thanks! That was fast :)

src/main/java/javaslang/collection/HashArrayMappedTrie.java
ruslansennov: Yes, I see, fixed.
But there is another error message when I do <code>mvn test</code>:
```text
Failed tests:   shouldHaveAConsistentTypeSystem(javaslang.TypeConsistencyTest): Unoverriden methods found. Check if this is ok and copy & paste the necessary lines into the TypeConsistencyTest.WHITELIST:
```

src/main/java/javaslang/collection/HashSet.java
```diff
@@ -0,0 +1,515 @@
...
+            return tree.isEmpty() ? empty() : new HashSet<>(tree);
+        }
+    }
+
+    private final static Object O = new Object();
+
+    private final HashArrayMappedTrie<T, Object> tree;
+    private final Lazy<List<T>> list;
+    private final Lazy<Integer> hash;
+
```
danieldietrich: With `list` we trade space against performance here. This is nice regarding performance. I wonder how Scala does this for Map/Set... I'm just curious.

src/main/java/javaslang/collection/HashArrayMappedTrie.java
danieldietrich: Ah - there is a unit test that checks, if all return types are as special as they can be.

Example:

Option.map returns Option but we know that None.map returns a None.

If we do not overwrite Option.map within None that way, it is put on the WHITELIST in the unit test.

You can copy the String 1:1 from the console and add a WHITELIST entry.

src/main/java/javaslang/collection/HashSet.java
danieldietrich: But this prototype is fine. It is exactly what we wanted!

src/main/java/javaslang/collection/HashArrayMappedTrie.java
ruslansennov: I prefer to change the signature :)

PR #427, opened 2015-08-06
Minor changes to collections

danieldietrich:
@ruslansennov What do you think of removing the `static Object O = new Object()` from `HashSet`. Do the ~~Trie~~ semantics remain unchanged?


ruslansennov:
it seems to be OK

danieldietrich:
thx!

PR #430, opened 2015-08-06
minor improvements (HAMT, HashSet)

ruslansennov:


danieldietrich:
This is most probably the reason why Scala has an Iterator with map/flatMap capabilities - I stumbled today over this topic ([Views vs. Iterators vs. Streams](http://docs.scala-lang.org/tutorials/FAQ/stream-view-iterator.html)).

I will implement a Scala-like `javaslang.collection.Iterator` for such purpose. But until we have it, we will do it like this.

Issue #433, opened 2015-08-06
Make complex hashCodes lazy


danieldietrich:
**This is more a question...**

All immutable non-empty collections could implement `hashCode()` as follows, because they may store a huge amount of data:

```java
private final transient Lazy<Integer> hashCode = Lazy.of(() -> Traversable.hash(this));

@Override
public int hashCode() {
    return hashCode.get();
}
```

An immutable empty collection should return `1` in favor of performance.

```java
@Override
public int hashCode() {
    return 1; // faster than Traversable.hash(this), but exposes implementation details
}
```

where `Traversable.hash()` looks like this:

```java
interface Traversable<T> /*...*/ {

    static int hash(Iterable<?> objects) {
        int hashCode = 1;
        for (Object o : objects) {
            hashCode = 31 * hashCode + Objects.hashCode(o);
        }
        return hashCode;
    }

    // ...
}
```

**We would trade performance against memory.**

danieldietrich:
Nope. Directly returning `1` for empty collections is a bad idea. Safety first. Empty collections should return the following. Then we are able to swap the hash-routine internally.

```java
@Override
public int hashCode() {
    return Traversable.hash(this);
}
```

danieldietrich:
Double checked all collections - this is done.

Issue #113, opened 2015-03-01
Make interface Tree extend Traversable


danieldietrich:
map function for Tree: http://en.wikibooks.org/wiki/Haskell/The_Functor_class

danieldietrich:
## Traversing a Tree

A `Tree` has an order, pre-order by default. The order may be changed, however, then a new Tree instance is created - for that node only. Different nodes of a tree may have a different order associated with. If a node is traversed, the order of the current node is taken into account when traversing all nodes. There are different methods to traverse a tree:

* `traverse()` - uses the actual order of the node
* `traverse(Tree.Order)` - uses a specific order and ignores the actual order of the current node
* `preOrder()` - depth first
* `postOrder()` - post order
* `inOrder()` - in order
* `levelOrder()` - breadth first

## Implementing Traversable and Monad

When implementing `Traversable` most methods will return a `Seq`, according to the underlying order of the tree. Internally the tree is traversed to perform these operations.

Interesting is, how to flatten or flatMap a tree - this is not clear. A binary search tree for example, having only values at the leafs, could be flatMapped as follows:

```java
interface Tree<A> {
    default Tree<B> flatMap(Function<A, Tree<B>> f) {
        return Match.ofType(Tree.class)
                .caze((Nil nil) -> nil)
                .caze((Leaf leaf) -> f.apply(leaf.value))
                .caze((Node node) -> new Node<>(node.left.flatMap(f), node.right.flatMap(f)))
                .apply(this);
    }
}
```

A good point to start is to take a look at the behavior of scalaz tree:

* Example of flatten: http://stackoverflow.com/a/20926958/1110815
* Howto start Scalaz REPL: http://stackoverflow.com/a/16528434/1110815

danieldietrich:
Please consider to let `traverse()` return `Seq<Tree<T>>` instead of `Seq<T>`

danieldietrich:
A scala worksheet is probably not the ideal way to test it, but it works:

```scala
import scalaz._, Scalaz._

object App extends TreeFunctions {

  val t = node(List(List(1)), Stream(node(List(List(2)), Stream(leaf(List(List(3, 4), List(5))))), leaf(List(List(4)))))

  def test1 =
    t.flatMap(l => leaf(l.flatten)).drawTree

  def test2 =
    t.flatMap(l => node(l.flatten, Stream(leaf(l.flatten), leaf(l.flatten)))).drawTree

}

App.test1

App.test2
```

**Output:**

```scala
[...]

PR #439, opened 2015-08-07
Unleashed javaslang.collection.Iterator. Refactored Collections.

### LINKED ISSUE #256: Re-organize collection.Tree
The current `Tree` type hierarchy is a little bit bulky.

As discussed [recently](https://github.com/javaslang/javaslang/issues/252), interfaces and implementations go hand-in-hand. Let me describe here, how we take the collection interfaces to the next level:

```java
            Traversable
                 |
               Tree
              /    \
     BinaryTree    ...
           /  \
RedBlackTree  ...
```

All trees in Javaslang are designed to have a value at each node. Also the empty tree exists for all tree interfaces.

```java
Tree<T> := Nil | Branch(T value, Seq<Tree> children)

BinaryTree<T> := Nil | Branch(BinaryTree<T> left, T value, BinaryTree<T> right)

// enum Color { RED, BLACK } handled internally
RedBlackTree<T> := Nil | Branch(RedBlackTree<T> left, T value, RedBlackTree<T> right)
```

`Tree` (_formerly: RoseTree_) is the most general tree data structure. It has

* a root
* (inner) nodes
* leaves
* an arbitrary amount of children

`BinaryTree` is an unbalanced binary search tree. Each node has at most two children: left and right. They are the simplest form of binary search trees. They perform poorly on ordered data and might take up to O(n) time.

`RedBlackTree` is a very efficient balanced binary search tree, maybe the fastest solution (with some tweaks) which exists so far. A color is internally used to keep the tree _aproximately_ balanced. Then no operation takes more than O(log n) time.

### LINKED ISSUE #291: Remove internal abstract classes from public API
E.g. the abstract collections which implement equals, hashCode and toString

### LINKED ISSUE #431: Implement Scala like Iterator with lazy map/flatMap capabilities
See [What is the difference between view, stream and iterator?](http://docs.scala-lang.org/tutorials/FAQ/stream-view-iterator.html).

There are use-cases within the collection library, e.g. HAMT and HashSet, where currently `Stream.ofAll(...).iterator()` is used to obtain a lazy Iterator.

Having a special `Iterator` is more lightweight compared to `Stream.iterator()`.

danieldietrich:
* Fixes #431 Implement Scala like Iterator with lazy map/flatMap capabilities
* Fixes #291 Remove internal abstract classes from public API
* Fixes #256 Re-organize collection.Tree
* Towards #433 Make complex hashCodes lazy  // but Deserialization of transient Lazy does not work!
* Towards #113 Make interface Tree extend Traversable
  * // it will at least extend Iterable and FilterMona, maybe TraversableOnce but not Traversable


danieldietrich:
**@ruslansennov I'm on vacation with my family for one week, starting tomorrow. Notebook has to stay home o_O... I will try to go online at most once a day. We will see.**

danieldietrich:
@ruslansennov I refactored the collections a little bit. Please have a look (HashSet, HashMap). There is now `javaslang.collection.Iterator`...

ruslansennov:
Have a nice trip! It's a good idea to leave the laptop at home

danieldietrich:
Yes...

danieldietrich:
I made HAMT and Interface. That leaks the inner classes to the public.

**Pro:** this enables pattern matching by subtype, 'introspection' and Proxying of the HAMT.

**Con:** the public API shows more classes

I'm not sure. We can switch back, when in doubt. The RedBlackTree is also designed as Interface with inner static classes. I like interfaces in general. They are more flexible, and it seems a good design practice to me.

ruslansennov:
> Con: the public API shows more classes

This was the reason I chose <code>abstract class</code> instead of <code>interface</code>
But there is not problem, I believe it's just a matter of taste

danieldietrich:
Ok. Abstract classes were always some intermediate types to me, not intended to be directly used other than extending them for own types. We will go with interfaces for now. I hope that Java 9 gives us some extra syntax regarding interfaces to hide private methods and maybe private inner classes. We will see.

PR #441, opened 2015-08-07
Added Map.Entry.map() and flatMap(). Added Iterator.take(n) and drop(n).

danieldietrich:


src/main/java/javaslang/collection/Iterator.java
```diff
@@ -155,6 +167,75 @@ public T next() {
...
+        } else if(!hasNext()) {
+            return Iterator.empty();
+        } else {
+            final Iterator<T> that = this;
+            return new Iterator<T>() {
+
+                int count = n;
+
+                @Override
+                public boolean hasNext() {
```
ruslansennov: this method must be cached

PR #445, opened 2015-08-08
fixes infinite loop in iterator implementation

ruslansennov:


danieldietrich:
Thx! Iterator is hacked from the brain to the keyboard. All untested code...

PR #451, opened 2015-08-09
Iterator extends TraversableOnce

ruslansennov:


src/test/java/javaslang/collection/AbstractTraversableOnceTest.java
```diff
@@ -0,0 +1,147 @@
...
+import org.assertj.core.api.*;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.stream.Collector;
+
+/**
+ * Tests all methods defined in {@link Traversable}.
+ */
+public abstract class AbstractTraversableOnceTest {
```
danieldietrich: This is a great idea to extract it! We have some many classes beside collections, which implement TraversableOnce already!

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -166,6 +167,19 @@ public T next() {
         };
     }
 
+    @Override
+    Iterator<T> peek(Consumer<? super T> action);
```
danieldietrich: Because Interable is lazy, i.e. evaluates values only, when needed, when can take an approch similar to Stream.peek(Consumer):

```java
@Override
default Iterator<T> peek(Consumer<? super T> action) {
    Objects.requireNonNull(action, "action is null");
    if (!hasNext()) {
        return Iterator.empty();
    } else {
        final Iterator<T> that = this;
        return new Iterator<T>() {
            @Override
             public boolean hasNext() {
                 return that.hasNext();
             }

            @Override
             public T next() {
                 if (!hasNext()) {
                 }
                 final T next = that.next();
                 consumer.accept(next);
                 return next;
             }
}
```

Please remove the abstract class `Impl`, we do not need it.

I hope you dont't mind to revert a part of your work. I really appreciate your help!

Greets,

Daniel

PS: I only have an iPad here. The code above is not tetsted, it may not work.

ruslansennov: you're right

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -166,6 +167,16 @@ public T next() {
         };
     }
 
+    @Override
+    default Iterator<T> iterator() {
+        return this;
```
ruslansennov: @danieldietrich  should we create new instance in this method?

danieldietrich: I think not, because Iterators are mutable/not reusable. The new Iterator would wrap the other and be the identity operation. It is semantically the same as `this`.

danieldietrich: The same with ofAll(Iterator) and ofAll(Iterable) they are effectively the identity and will be removed. I created an issue yesterday. Then the method `iterator()` will be removed completely!

You can do this (remove ofAll(Iterable), ofAll(Iterator) and iterator()).

When we need elsewhere an Iterable and we have already an Iterator we just do. `() -> myIterator`

ruslansennov: But <code>Iterator<T> iterator()</code> requires by <code>TraversableOnes</code> which we extends

danieldietrich: Oh, I've overseen it. Then iterator() stays with us and the ofAll() methods pass away. (Note: we still need ofAl(T...))

danieldietrich: I meant, if we have an Iterable, we just call myIterable.iterator() to obtain an iterator...

danieldietrich: Oh - CAUTION - we still need ofAll(*java.util.Iterator*) AND ALSO ofAll(Iterable) because they **lift** java.util.Iterator to javaslang.collection.Iterator. (Wish I had my Notebook here, mobile phone is not good for programming...)

ruslansennov: :) I believe there is all OK with this PR now

danieldietrich: :) thx ☕️☕️

Issue #454, opened 2015-08-10
Add Seq.remove(int)


danieldietrich:
To remove an element at a specific index.

```java
remove(index) = splitAt(index).transform(t -> t._2.isEmpty() ? t._1 : t._1.appendAll(t._2.tail()))
```

Given this, remove `HashArrayMappedTrie.IndexedNode.removeIndx(List, int)`.

Also provide a better implementation of `List.splitAt()` which currently costs O(2n) = O(n). This can be reduced by constant factor 2, by removing `take` and `drop` calls. Instead the elements should be iterated until the index is reached. Instead of calling drop, we just use the tail then. We have to take care of changing the interface. take and drop do not throw. splitAt should throw(?), on an illegal index. We have to take a look at Scala here, how non-existing indices are handled.

ruslansennov:
Scala accept illegal index in <code>splitAt()</code>

danieldietrich:
ok, splitAt should not throw

danieldietrich:
Thanks @ruslansennov, awesome!

PR #456, opened 2015-08-11
#454 Seq.removeIndx(); Seq.splitAt() improvements

ruslansennov:


src/main/java/javaslang/collection/List.java
```diff
@@ -1316,6 +1329,18 @@ public T head() {
...
+        public Tuple2<List<T>, List<T>> splitAt(int n) {
+            List<T> head = Nil.instance();
+            List<T> tail = this;
+            while (n > 0 && !tail.isEmpty()) {
+                head = head.prepend(tail.head());
+                tail = tail.tail();
+                n--;
+            }
+            return Tuple.of(head.reverse(), tail);
+        }
```
danieldietrich: great, that's the waynto do it!

src/main/java/javaslang/collection/Stream.java
```diff
@@ -1310,6 +1311,16 @@ public T head() {
...
 
         @Override
+        public Tuple2<Stream<T>, Stream<T>> splitAt(int n) {
+            if(n <= 0) {
+                return Tuple.of(empty(), this);
+            } else {
+                Tuple2<Stream<T>, Stream<T>> t = tail().splitAt(n - 1);
+                return Tuple.of(new Cons<>(head, () -> t._1), t._2);
+            }
+        }
```
danieldietrich: For Stream the Tuple.of(take(n), drop(n)) solution is the best, because take() is lazy and no intermediate Tuple objects are required.

src/main/java/javaslang/collection/Seq.java
```diff
@@ -540,6 +540,14 @@ private boolean checkPrefix(Traversable<T> t, Traversable<T> prefix) {
     @Override
     Seq<T> remove(T element);
 
+    /**
+     * Removes the element at the specified position in this sequence. Shifts any subsequent elements to the left
+     * (subtracts one from their indices).
+     * @param indx position of element to remove
+     * @return a sequence containing all elements of this without the element at the specified position.
+     */
+    Seq<T> removeIndx(int indx);
```
danieldietrich: Let's name it `remove(int)` to align it with `set(int, T)`. Will this cause ambiguities when we have a `Seq<Integer>`? Then we should name it `removeAt()` to align it with `splitAt()`.

danieldietrich: We should throw on illegal indices, like `insert` and `set` do:

```java
@throws IndexOutOfBoundsException if this is empty, index &lt; 0 or index &gt;= length()
```

src/main/java/javaslang/collection/List.java
```diff
@@ -1316,6 +1329,18 @@ public T head() {
         }
 
         @Override
+        public Tuple2<List<T>, List<T>> splitAt(int n) {
+            List<T> head = Nil.instance();
```
danieldietrich: I would name it `init`

src/main/java/javaslang/collection/Stream.java
ruslansennov: yes, you're right

src/main/java/javaslang/collection/Seq.java
ruslansennov: OK, but with this feature implementation will be a little more complicated

danieldietrich:
Super-cool, thx!

Issue #455, opened 2015-08-11
Further improvements of HAMT


danieldietrich:
@ruslansennov To get the most out of List used in HAMT we should do the following:

* use prepend() and prependAll() instead of append() and appendAll(). Note: I hope the order of entries does not matter in a HAMT!? However, the iterator() will then return the entries in a different order. Does this matter for HashMaps/HashSets?
* use findFirst() instead of filter().headOption()
* in mergeLeaves() please replace List.empty() with this.entries() and delete the appendAll(this.entries) below

Additionally I have created two issues to improve [removeIndx()](https://github.com/javaslang/javaslang/issues/454) and [iterator()](https://github.com/javaslang/javaslang/issues/452).

danieldietrich:
What if we put the entries in a (Sorted)Set instead? E.g. containing Tuple2 with compared by ._1/key?

ruslansennov:
As I know nobody promises that the return set of keys should be in certain order. We can do it as our feature, but I'm not sure. All another your proposals are very good.

danieldietrich:
Using List, we have O(n), using Set it is O(log_2(n)) for operations on entries.

I have to take a look on other HAMT impls and need also to understand, why a HAMT backed by a Set (RedBlackTree) performs better than using just a RedBlackTree (containing Entries compared by key). How do Scala and Closure store the entries in a HAMT?

Interesting!

danieldietrich:
Good reading: http://lampwww.epfl.ch/papers/idealhashtrees.pdf

ruslansennov:
just for clarify: <code>HashMap.keySet()</code> should be return <code>Set</code> and only <code>Set</code>. But sorted set or hash set, that is the question

danieldietrich:
Our API will declare return type `Set`, that makes it flexible for other impls.

Our implementation may use SortedSet for performance reasons, if it makes sense.

danieldietrich:
(minor thing) We could make the number of slots (currently 32) a constant SLOTS, just for clarity.

danieldietrich:
@ruslansennov Is there still something to do here? https://github.com/javaslang/javaslang/blob/master/src/main/java/javaslang/collection/HashArrayMappedTrie.java#L208-L216

ruslansennov:
Of course no! After last <code>List</code> improvements It can be changed to 
<code>
{ List.removeAt() } else { List.set() }
</code>
I'll fix it later

ruslansennov:
[...]

PR #457, opened 2015-08-11
Further improvements of HAMT #455

ruslansennov:


danieldietrich:
Top! This brought the collections forward - not only HAMT.

Issue #452, opened 2015-08-09
Add Iterator.ConcatIterator


danieldietrich:
Currently we use for Tree structures (like HAMT) the method Iterator.flatMap to concatenate child iterators. To prevent stack overflows it might be better to introduce a mutable [ConcatIterator](https://github.com/scala/scala/blob/v2.11.7/src/library/scala/collection/Iterator.scala#L165-L189) which allows to add child Iterators.

ruslansennov:
I called builders <code>Iterator.ofIterators()</code> because of conflicts with other statics

danieldietrich:
Closed with #461

Issue #352, opened 2015-07-26
Fix result bound of function argument of higher order functions


danieldietrich:
`java.util.stream.Stream` defines

```java
<R> Stream<R> flatMap(Function<? super T, ? extends Stream<? extends R>> mapper);
```

Javaslang defines throughout all classes:

```java
<R> Xyz<R> method(Function<? super T, ? extends Xyz<R>> f);
```

Should `<Xyz<R>>` be replaced with `<Xyz<? extends R>>` in Javaslang?

@szarnekow Do you have a hint for me? Does it really make a difference? The Java's Stream.flatMap method can't directly return the mapper result IMO, it has to cast `Stream<? extends R>` to `Stream<R>`. The two types are formally not _compatible_ in Java. However, a cast would be semantically correct (but is unchecked).

It is not clear to me if we need arguments `Function<? super T, ? extends Xyz<? extends R>>` or `Function<? super T, ? extends Xyz<R>>` in general.

I've created a simple test-case which shows, that it _seems_ to make no difference from the user perspective. But maybe it makes a difference when calling more methods after flatMap which also have generic method parameters induced through the current context!?

```java
import java.util.function.Function;

public class Test {

    public static void main(String[] args) {

        // DOES COMPILE, AS EXPECTED.
        Arr<LifeForm> lifeFormArr1 = new Arr<>(1).flatMap1(i -> new Arr<>(new Animal()));

        // DOES COMPILE FINE, TOO. NOT 100% CLEAR TO ME...
        Arr<LifeForm> lifeFormArr2 = new Arr<>(1).flatMap2(i -> new Arr<>(new Plant()));

        // update: this does work
        Arr<LifeForm> lifeFormArr3 = new Arr<>(new Plant());

        // update: this does *not* work
        Arr<LifeForm> lifeFormArr4 = new Arr<Plant>(new Plant());

        // update: of course this does work again
        Arr<? extends LifeForm> lifeFormArr5 = new Arr<Plant>(new Plant());
    }

    static class Arr<T> {

        final T t;

        Arr(T t) {
            this.t = t;
        }

        @SuppressWarnings("unchecked")
        <R> Arr<R> flatMap1(Function<? super T, ? extends Arr<? extends R>> mapper) {
            return (Arr<R>) mapper.apply(t);
        }

        <R> Arr<R> flatMap2(Function<? super T, ? extends Arr<R>> mapper) {
            return mapper.apply(t);
        }
    }

    interface LifeForm {
    }

[...]

PR #461, opened 2015-08-12
ConcatIterator #452

ruslansennov:


src/main/java/javaslang/collection/Iterator.java
```diff
@@ -128,6 +128,33 @@ public T next() {
...
+     * @return A new {@code javaslang.collection.Iterator}
+     */
+    @SafeVarargs
+    static <T> Iterator<T> ofIterators(Iterator<T>... iterators) {
+        Objects.requireNonNull(iterators, "iterators is null");
+        List<Iterator<T>> queue = List.empty();
+        for (Iterator<T> iterator : iterators) {
+            queue = queue.prepend(iterator);
+        }
+        return queue.isEmpty() ? Iterator.empty() : new ConcatIterator<>(queue.reverse());
```
danieldietrich: This could also work here:

```java
return (iterators.length == 0) ? Iterator.empty() : new ConcatIterator<>(List.of(iterators));
```

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -128,6 +128,33 @@ public T next() {
     }
 
     /**
+     * Creates an Iterator which traverses along all given iterators.
+     * @param iterators The list of iterators
+     * @param <T>       Component type.
+     * @return A new {@code javaslang.collection.Iterator}
+     */
+    @SafeVarargs
+    static <T> Iterator<T> ofIterators(Iterator<T>... iterators) {
```
danieldietrich: Does the body compile with arg type `Iterator<? extends T>`?

src/main/java/javaslang/collection/Iterator.java
ruslansennov: Unfortunately this produces compilation warning 
```
Varargs method could cause heap pollution from non-reifiable varargs parameter iterators
```
and build fails. I'll change loop direction like in <code>List.of()</code>

danieldietrich:
Hi Ruslan,

I fiddled around some time and think we have to go lazy here. This will return an Iterator for a Map in O(1), regardless of the size. My intention is that we modify ConcatIterator like this:

```java
class ConcatIterator<T> implements Iterator<T> {

    private final Iterable<? extends Iterator<? extends T>> iterators;
    private Iterator<? extends T> current = Iterator.empty(); // important for lazy processing

    private ConcatIterator(Iterable<? extends Iterator<? extends T>> iterators) {
        this. iterators = iterators;
    }

    // hasNext and next as you implemented them
```

We use Iterator.next() and hasNext() instead of Traversable.head() and tail() here.

Our Iterator then has these static factory methods

```java
    static <T> ConcatIterator<T> ofIterables(Iterable<? extends Iterable<? extends T>> iterables) {
        final Iterable<? extends Iterator<? extends T>> iterators = Stream.of(iterables).map(Iterable::iterator);
        return new ConcatIterator<>(iterators);
    }

    @SuppressWarnings({"unchecked", "varargs"})
    @SafeVarargs
    static <T> ConcatIterator<T> ofIterables(Iterable<? extends T>... iterables) {
        final Iterable<? extends Iterator<? extends T>> iterators =  Stream.of(iterables).map(Iterable::iterator());
        return new ConcatIterator<>(iterators);
    }

    static <T> ConcatIterator<T> ofIterators(Iterable<? extends Iterator<? extends T>> iterables) {
        final Iterable<? extends Iterator<? extends T>> iterators = Stream.of(iterables).map(Iterable::iterator);
        return new ConcatIterator<>(iterators);
    }

    @SuppressWarnings({"unchecked", "varargs"})
    @SafeVarargs
    static <T> ConcatIterator<T> ofIterators(Iterator<? extends T>... iterators) {
        final Iterable<? extends Iterator<? extends T>> iterators =  Stream.of(iterators);
        return new ConcatIterator<>(iterators);
    }
```

And finally the iterator() method of our two HAMT nodes returns:

```java
return Iterable.ofIterables(subNodes);`
```

What do you think?

danieldietrich:
Update: I Fixed many bugs in my previous comment.

ruslansennov:
beautiful, I'll try to implement it in the evening

danieldietrich:
great, I'm looking forward to it

danieldietrich:
Added GC frindly checks to factory methods. Repetition is ok here.

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -128,6 +128,34 @@ public T next() {
...
+        return queue.isEmpty() ? Iterator.empty() : new ConcatIterator<>(queue);
+    }
+
+    /**
+     * Creates an Iterator which traverses along all given iterators.
+     * @param iterators The list of iterators
+     * @param <T>       Component type.
+     * @return A new {@code javaslang.collection.Iterator}
+     */
+    static <T> Iterator<T> ofIterators(Traversable<Iterator<T>> iterators) {
```
szarnekow: Should be ok to use `Traversable<? extends Iterator<T>>` here

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -429,4 +457,32 @@ public T next() {
         }
     }
 
+    class ConcatIterator<T> implements Iterator<T> {
+
+        private Traversable<Iterator<T>> queue;
+        private Iterator<T> current;
+
+        private ConcatIterator(Traversable<Iterator<T>> queue) {
```
szarnekow: Also here

src/main/java/javaslang/collection/Iterator.java
ruslansennov: thank you, but I'll use <code>Iterator&lt;Iterator&lt;T&gt;&gt;</code>
see [this](#issuecomment-130607691) comment

ruslansennov:
@danieldietrich @szarnekow pls review

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -128,6 +128,82 @@ public T next() {
...
+        return iterables.length == 0 ? Iterator.empty() : new ConcatIterator<>(Stream.of(iterables).map(Iterator::ofAll).iterator());
+    }
+
+    /**
+     * Creates an Iterator which traverses along all given iterators.
+     * @param iterators The iterator over iterators
+     * @param <T>       Component type.
+     * @return A new {@code javaslang.collection.Iterator}
+     */
+    static <T> Iterator<T> ofIterators(Iterator<Iterator<T>> iterators) {
```
szarnekow: The input parameter can be slightly more general, e.g. `Iterator<? extends Iterator<T>>` if not even Iterator<? extends Iterator<? extends T>>` which would allow to concat two iterators `Iterator<Integer>` and `ConcatIterator<Double>` to an iterator of Number.

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -128,6 +128,82 @@ public T next() {
...
+        return iterators.isEmpty() ? Iterator.empty() : new ConcatIterator<>(Stream.ofAll(iterators).iterator());
+    }
+
+    /**
+     * Creates an Iterator which traverses along all given iterables.
+     * @param iterables The iterator over iterables
+     * @param <T>       Component type.
+     * @return A new {@code javaslang.collection.Iterator}
+     */
+    static <T> Iterator<T> ofIterables(Iterator<? extends Iterable<T>> iterables) {
```
szarnekow: should also be `Iterable<? extends T>`

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -128,6 +128,82 @@ public T next() {
...
+        return iterables.isEmpty() ? Iterator.empty() : new ConcatIterator<>(Stream.ofAll(iterables).map(Iterator::ofAll).iterator());
+    }
+
+    /**
+     * Creates an Iterator which traverses along all given iterators.
+     * @param iterators The iterable of iterators
+     * @param <T>       Component type.
+     * @return A new {@code javaslang.collection.Iterator}
+     */
+    static <T> Iterator<T> ofIterators(Iterable<Iterator<T>> iterators) {
```
szarnekow: Dito

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -128,6 +128,82 @@ public T next() {
...
+        return new ConcatIterator<>(Stream.ofAll(iterators).iterator());
+    }
+
+    /**
+     * Creates an Iterator which traverses along all given iterables.
+     * @param iterables The iterable of iterables
+     * @param <T>       Component type.
+     * @return A new {@code javaslang.collection.Iterator}
+     */
+    static <T> Iterator<T> ofIterables(Iterable<? extends Iterable<T>> iterables) {
```
szarnekow: Dito

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -429,4 +505,36 @@ public T next() {
...
+        private final Iterator<? extends Iterator<? extends T>> iterators;
+        private Iterator<? extends T> current;
+
+        private ConcatIterator(Iterator<? extends Iterator<? extends T>> iterators) {
+            this.current = Iterator.empty();
+            this.iterators = iterators;
+        }
+
+        private static boolean maybeEmpty(Iterable<?> iterable) {
+            return (iterable instanceof TraversableOnce && ((TraversableOnce<?>) iterable).isEmpty())
```
szarnekow: Should be save to use iterable.iterator().hasNext().

src/main/java/javaslang/collection/Iterator.java
danieldietrich: @ruslansennov yes, I would also suggest the double-extends variant. See https://github.com/javaslang/javaslang/issues/352#issuecomment-125876580

danieldietrich: The other stuff looks great!

src/main/java/javaslang/collection/Iterator.java
danieldietrich: Yes, let's delete this method and substitue the call with Sebastian's suggestion

ruslansennov:
@szarnekow @danieldietrich thank you

danieldietrich:
Yes! Awesome.

PR #466, opened 2015-08-14
List.length() improvement

ruslansennov:
list always know its length


danieldietrich:
Nice addition. The additional bytes for an instance or ok. Object serialization/deserialization still works. Great!

PR #467, opened 2015-08-15
Euler 55, 99 solved #126

ruslansennov:


danieldietrich:
Looks great!

(I'm offline for the rest of the day because we drive back home)

Issue #460, opened 2015-08-12
Keep memory clean on collection operations


danieldietrich:
If a Traversable / Seq / List / ... operation results in the identity, we should remove the original instance and let gc clean up intermediate objects.

Example: (1, 2, 3).remove(4) currently produces a second (1, 2, 3)

Reducing the number of instances scales regarding memory consumption and when that equals can return immediately because `==` matches. This implies better `Map` performance in practice.

danieldietrich:
Fixed by #472

PR #472, opened 2015-08-16
help GC #460

ruslansennov:


src/main/java/javaslang/collection/HashSet.java
```diff
@@ -350,7 +360,7 @@ public boolean isEmpty() {
 
     @Override
     public HashSet<T> remove(T element) {
-        return new HashSet<>(tree.remove(element));
+        return contains(element) ? new HashSet<>(tree.remove(element)) : this;
```
danieldietrich: what about the tree? does it return the same instance when removing an element which is not contained? if so, we may compare tree == removed, where removed = tree.remove(element)

src/main/java/javaslang/collection/HashSet.java
```diff
@@ -380,7 +390,7 @@ public boolean isEmpty() {
         for (T element : elements) {
             trie = trie.remove(element);
         }
-        return new HashSet<>(trie);
+        return trie.size() == list.get().length() ? this : new HashSet<>(trie);
```
danieldietrich: also here. if true.remove(element) returns the same instance, we may test with == instead of size and length

src/main/java/javaslang/collection/HashSet.java
ruslansennov: great

src/main/java/javaslang/collection/List.java
```diff
@@ -679,6 +679,12 @@ public Short next() {
 
     @Override
     default List<T> dropRight(int n) {
+        if (n <= 0) {
+            return this;
+        }
+        if (n >= length()) {
+            return empty();
+        }
```
danieldietrich: Very cool. We are now collecting the fruits after your length implementation!

src/main/java/javaslang/collection/List.java
```diff
@@ -695,7 +701,18 @@ public Short next() {
     @Override
     default List<T> filter(Predicate<? super T> predicate) {
         Objects.requireNonNull(predicate, "predicate is null");
-        return isEmpty() ? this : foldLeft(List.<T> empty(), (xs, x) -> predicate.test(x) ? xs.prepend(x) : xs).reverse();
+        List<T> list = this;
+        List<T> result = List.empty();
+        boolean testFailed = false;
```
danieldietrich: i would name it `isFiltered` to suggest that nothing failed

src/main/java/javaslang/collection/List.java
```diff
@@ -695,7 +701,18 @@ public Short next() {
...
+        while (!list.isEmpty()) {
+            if (predicate.test(list.head())) {
+                result = result.prepend(list.head());
+            } else {
+                testFailed = true;
+            }
+            list = list.tail();
+        }
+        return testFailed ? result.reverse() : this;
```
danieldietrich: Mmhh, we need to iterate all elements. Me may save the last reverse() even shorter without variables:

```java
final List<T> filtered = foldLeft(List.<T> empty(), (xs, x) -> predicate.test(x) ? xs.prepend(x) : xs);
return (this.length() == filtered.length()) ? this : filtered.reverse();
```

src/main/java/javaslang/collection/Stream.java
```diff
@@ -793,7 +793,11 @@ public Short next() {
 
     @Override
     default Stream<T> dropRight(int n) {
-        return reverse().drop(n).reverse();
+        if(n <= 0) {
+            return this;
```
danieldietrich:
```java
if (n >= length) {
    return empty();
}
```

danieldietrich: Oh no! It's a Stream :-)

src/main/java/javaslang/collection/List.java
ruslansennov: ok

src/main/java/javaslang/collection/Stream.java
```diff
@@ -984,7 +991,11 @@ default int lastIndexOf(T element, int end) {
 
     @Override
     default Stream<T> removeLast(Predicate<T> predicate) {
-        return reverse().removeFirst(predicate).reverse();
+        if(findFirst(predicate).isDefined()) {
+            return reverse().removeFirst(predicate).reverse();
+        } else {
+            return this;
+        }
```
danieldietrich: I would do a `reverse().removeFirst(predicate)` instead of `findFirst` in order to not process elements 2 times, e.g.

```java
final Stream<T> removed = reverse().removeFirst(predicate);
return removed.length() == length() ? this : removed.reverse();
```

src/test/java/javaslang/collection/AbstractTraversableTest.java
```diff
@@ -1522,7 +1562,12 @@ public void shouldTakeWhileAllOnFalseCondition() {
 
     @Test
     public void shouldTakeWhileAllOnTrueCondition() {
-        assertThat(of(1, 2, 3).takeWhile(x -> true)).isEqualTo(of(1, 2, 3));
+        final Traversable<Integer> t = of(1, 2, 3);
+        if(isThisLazyCollection()) {
+            assertThat(t.takeWhile(x -> true)).isEqualTo(t);
+        } else {
+            assertThat(t.takeWhile(x -> true)).isSameAs(t);
+        }
```
danieldietrich: Cool!

src/main/java/javaslang/collection/Stream.java
```diff
@@ -1155,7 +1166,11 @@ default int lastIndexOf(T element, int end) {
 
     @Override
     default Stream<T> takeRight(int n) {
-        return reverse().take(n).reverse();
+        if(length() <= n) {
```
danieldietrich: we should also check for `n <= 0` here

src/main/java/javaslang/collection/Stream.java
danieldietrich: Just saw that length is Lazy. When we do it elsewhere, e.g. in Stream.takeRight we should also check `n >= length()` here.

ruslansennov:
done, thank you

danieldietrich:
Super-awesome! This improves not only memory-performance but also equals-performance.

PR #473, opened 2015-08-17
Value & Traversable interfaces

danieldietrich:
* First shot: Refactored Value vs TraversableOnce vs Traversable.
* Towards Set an Map interfaces.
* Towards Traversable Tree. TODO: traverse order has to be saved in tree nodes.
* Disabled TypeConsistencyCheck temporarily.


danieldietrich:
I hope this will not cause too much trouble (merge-conflicts).

danieldietrich:
@ruslansennov My vacation is now over. That means I can only spend time on Javaslang now after may daylight job for the next weeks. :-/

Issue #464, opened 2015-08-14
Add an immutable Vector (implementds IndexedSeq)


danieldietrich:

ruslansennov:
What algorithm you prefer to implement here? If something like <code>java.util.ArrayList</code> then we can use it in <code>HashArrayMappedTrie</code>. If no, then I'd like to remove <code>HashArrayMappedTrie.ArrayNode</code> because of high cost of <code>List.set(index, element)</code> operation

danieldietrich:
An internal `HAMT<Integer, T>` makes sense for `Vector<T>`.

Please tell me a little bit more about the details why we can remove ArrayNode and what changes regarding our performance characteristics (time/memory) if we do so.

ruslansennov:
<code>HashArrayMappedTrie.ArrayNode</code> is useful only if cost of <code>set()</code> operation is <code>O(1)</code>. We have <code>O(n)</code>, so we can use <code>HashArrayMappedTrie.IndexedNode</code> only

ruslansennov:
... with some modifications, of course :)

danieldietrich:
Why is the cost of set O(n)? I know Tries in general but haven't read the paper about HAMT in depth. The Integer-keys should be distributed equally. I would expect O(1) then. Or has the complete Trie to be rewritten on set()? If so, why is the performance then better for other key-types. If the performance is not better for other key-types, why the ArrayNode is then used at all?

danieldietrich:
Nevertheless, do it with the best performance possible. You know the characteristics best. Although I want to understand it in depth.

ruslansennov:
No, I mean that cost of <code>List.set(index, element)</code> is <code>O(n)</code>. If <code>HashArrayMappedTrie.ArrayNode</code> backed by <code>List</code>, then there is no general difference between <code>HashArrayMappedTrie.ArrayNode</code> and <code>HashArrayMappedTrie.IndexedNode</code>, and they may be combined into <code>NoLeafNode</code> without loosing CPU or memory

danieldietrich:
From what I understand is, that we generally traded in our HAMT List against Array. In other words, this affects also our Map impls.

Then I would keep our ArrayNode and substitute List with Array when we implemented it (see #295). This should boost it from O(n) to O(1), right?

ruslansennov:
Oh, I have not seen #295 before. Of course in this case we should keep ArrayNode

danieldietrich:
Fine, let's do it this way.

[...]

PR #480, opened 2015-08-17
javaslang.collection.Vector #464

ruslansennov:


src/main/java/javaslang/collection/Vector.java
```diff
@@ -0,0 +1,1481 @@
...
+                vector = new Impl<>(temp);
+            }
+
+            private Object readResolve() {
+                return vector;
+            }
+        }
+    }
+
+    class Nil<T> implements Vector<T>, Serializable {
```
danieldietrich: And _boom_, Ruslan comes with Vector around the corner :-)) I love this HAMT!!

I've learned that Nil is somewhat 'reserved' for the empty (lazy) linked list. Lets name it `Empty`.

src/main/java/javaslang/collection/Vector.java
```diff
@@ -0,0 +1,1481 @@
...
+            return Traversable.hash(this);
+        }
+
+        @Override
+        public String toString() {
+            return "Vector()";
+        }
+
+        @Override
+        public HashArrayMappedTrie<Integer, T> toHashArrayMappedTrie() {
```
danieldietrich: This is an implementation detail. A vector may be backed by another data structure. Let's make it not public API and remove this method.

src/main/java/javaslang/collection/Vector.java
```diff
@@ -0,0 +1,1481 @@
...
+                return vector;
+            }
+        }
+    }
+
+    class Nil<T> implements Vector<T>, Serializable {
+        private static final long serialVersionUID = 1L;
+        private static final Vector<?> INSTANCE = new Nil<>();
+
+        private final HashArrayMappedTrie<Integer, T> trie = HashArrayMappedTrie.empty();
```
danieldietrich: if we really need the trie instance in another place, let's make it package private but for now leave it private if possible.

danieldietrich: Or better: remove it if possible. An empty Vector should be identified by `isEmpty()`.

src/main/java/javaslang/collection/Vector.java
danieldietrich: Let's make it public.

src/main/java/javaslang/collection/Vector.java
ruslansennov: I need <code>Vector.toHashArrayMappedTrie()</code> method to be able to implement 99% of code in interface body. This hack is not so beautiful but very effective

danieldietrich: I see. I would prefer to implement all methods for the empty case in the `Empty` instance. But that's a huge amount of extra work. For now let's keep it this way. It's ok.

ruslansennov: No, I'll do it :)

src/main/java/javaslang/collection/Vector.java
ruslansennov: It's public :)

src/main/java/javaslang/collection/Vector.java
danieldietrich: That's the very thing! We need the abstract interface though. Bonus: We can adjust the return types to `Impl` and `Empty` where possible.

danieldietrich: In exchange I will not commit & push until you are finished :-) So if there's s.th. to merge I'll do it!

src/main/java/javaslang/collection/Vector.java
```diff
@@ -0,0 +1,1481 @@
...
+import java.io.*;
+import java.util.*;
+import java.util.function.*;
+import java.util.stream.Collector;
+
+/**
+ * TODO javadoc
+ * @param <T>
+ */
+public interface Vector<T> extends IndexedSeq<T> {
```
danieldietrich: Beside the trie method all other things look awesome to me!

ruslansennov: What do you think about <code>Vector.ofTrie()</code>?

danieldietrich: Please remove all public API related to Trie. We do not really need it. The new `Value` I committed yesterday has all conversion methods between 'regular' collection one could need. I see HAMT, RedBlackTree, ... as backing collections, not as core collections. They do not extends Traversable et al. If in doubt, we have implemented the Iterable interface, which is sufficient to create any Traversable collection via `ofAll`.

ruslansennov: Well

danieldietrich: Hold fire! HashMap, HashSet and Queue are currently classes. I thought about making them interfaces. But we may also stay with them as classes and make Vector a class instead. Then you could keep your implementation and remove the inner classes.

danieldietrich: It needs to be `final`.

danieldietrich: And the trie stuff can be private...

ruslansennov:
done

danieldietrich:
Thx!

Issue #465, opened 2015-08-14
Add an immutable String implements IndexedSeq<Character>


danieldietrich:
java.lang.String is already immutable but javaslang.collection.String will be feature-rich.

I expect, that this feature will be a magnet (similar to Try) because Strings are probably the most popular native language feature (of Java) `> Hello world!`.

Example:

```java
// = {
//   'J' -> 1,
//   'a' -> 3,
//   'v' -> 1,
//   's' -> 1,
//   'l' -> 1,
//   'n' -> 1,
//   'g' -> 1
// }
final Map<Character, Integer> occurrences = String.of("Javaslang").countOccurrences();
```

patrox:
hi @danieldietrich, so the future `javaslang.collection.String` will be very similiar to: 
`scala.collection.immutable.WrappedString` ?

danieldietrich:
Hi @patrox, yes - exactly. It will be an IndexedSeq of Character that wraps a java.lang.String

danieldietrich:
Then we should name it WrappedString

ruslansennov:
I'll name it <code>WrappedString</code> right now

patrox:
@ruslansennov are you working currently on the implementation? Because to be honest I like this project a lot and if it's possible I would like to help you somehow - and I was wondering if I can work on this one? :)

danieldietrich:
@patrox :+1: you're welcome!

ruslansennov:
@patrox It seems all ready, I'll finish soon

danieldietrich:
@patrox to start getting your hands dirty you could implement 1-2 euler tests. I like the Scala project of @shadaj as reference: https://github.com/shadaj/euler. The existing tests are located in src/test/java/javaslang/collection/euler

patrox:
@ruslansennov so I was a bit late ... but it will be quite interesting to compare the implementations :)
@danieldietrich I'll have a look at these tests

danieldietrich:
Great! There will be more todo - much more :-)

ruslansennov:
@danieldietrich There is problem to make test for String. I see no other way except make it independent from <code>AbstractSeqTest</code>

danieldietrich:
Alright!

danieldietrich:
@ruslansennov @patrox Guys, I urgently need some high quality sleep - I will come back tomorrow...

danieldietrich:
@ruslansennov @patrox 

[...]

PR #485, opened 2015-08-18
WrappedString #465

ruslansennov:


ruslansennov:
#465

src/main/java/javaslang/collection/Vector.java
```diff
@@ -1342,9 +1342,9 @@ private Object readResolve() {
     public boolean equals(Object o) {
         if (o == this) {
             return true;
-        } else if (o instanceof Vector) {
+        } else if (o instanceof IndexedSeq) {
```
danieldietrich: Hi Ruslan, what was your intention here? Changing it to super-interface IndexedSeq should work. This is a general philosophy: when are objects equal.

To keep it consistent with the other classes I would keep `Vector` here.

ruslansennov: This allows comparison between <code>String</code> and <code>Vector</code>

danieldietrich: Ok, I see. Have to think about it. Will come back later...

ruslansennov: Maybe you're right and I should rewrite <code>StringTest.assertThat()</code> method if we don't need in this functionality outside of tests

src/main/java/javaslang/WrappedString.java
```diff
@@ -0,0 +1,2198 @@
...
+ */
+public final class WrappedString implements CharSequence, IndexedSeq<Character>, Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final WrappedString EMPTY = new WrappedString("");
+    
+    private final java.lang.String back;
+    
+    private WrappedString(java.lang.String javaString) {
```
patrox: is it really needed to use a fully qualified class name ?

ruslansennov: This has remained since those times when the class was also named <code>String</code> :)

patrox: yes, that exactly what I thought - I had the same problem - but I was far less imaginative and called my `WrappedString` just String2 ... :)

src/main/java/javaslang/WrappedString.java
```diff
@@ -0,0 +1,2198 @@
...
+        return new WrappedString(new java.lang.String(chrs));
+    }
+
+    /**
+     * Creates a String of {@code java.lang.String}.
+     *
+     * @param javaString {@code java.lang.String} instance.
+     * @return A new {@code javaslang.String}
+     */
+    public static WrappedString of(java.lang.String javaString) {
```
danieldietrich: I would make this `of(CharSequence)`

danieldietrich: also check for null (Objects.requireNonNull)

src/main/java/javaslang/WrappedString.java
```diff
@@ -0,0 +1,2198 @@
...
+     * Creates a String of the given elements.
+     *
+     * The resulting string has the same iteration order as the given iterable of elements
+     * if the iteration order of the elements is stable.
+     *
+     * @param elements An Iterable of elements.
+     * @return A string containing the given elements in the same order.
+     * @throws NullPointerException if {@code elements} is null
+     */
+    public static WrappedString ofAll(Iterable<? extends Character> elements) {
```
danieldietrich: dito null-check

src/main/java/javaslang/WrappedString.java
```diff
@@ -0,0 +1,2198 @@
...
+        return of(sb.toString());
+    }
+
+    @Override
+    public WrappedString clear() {
+        return EMPTY;
+    }
+
+    @Override
+    public Vector<Tuple2<Character, Character>> crossProduct() {
```
danieldietrich: I wanted to think about the usage of Vector. This is what I've come to:

Scala uses for `WrappedString.combinations(int)` as return type `Iterator<WrappedString>`. I also prefer not to use `Vector` _directly_. Instead I would use `Seq` or `Iterator`. `Seq` is not a good alternative because of the wildcards in return types, e.g. `Tuple2<? extends Seq, ...>`. This leads to `Iterator` in the case of this method (which could be obtained by calling `result.iterator()`).

In general there is another thing which comes to my mind. Instead of wrapping a sequence of chars, i.e. `java.lang.String`, we could wrap a sequence of arbitrary object, for example stored in `Vector<?> back`. This would `flatMap` allow to return a `WrappedString`. When a concrete String representation is needed, `back.mkString()` could be called. Example: `replaceAll(regEx, Str) = back.mkString().replaceAll(regEx, Str)`.

This is just an initial thought. Before going further I've to think more about it (later today).

@ruslansennov @patrox What are your thoughts?

src/main/java/javaslang/collection/Vector.java
danieldietrich: Yes, the new (yet unknown) solution will require that

src/main/java/javaslang/WrappedString.java
ruslansennov: > In general there is another thing which comes to my mind. Instead of wrapping a sequence of chars, i.e. java.lang.String, we could wrap a sequence of arbitrary object, for example stored in Vector<?> back. This would flatMap allow to return a WrappedString. When a concrete String representation is needed, back.mkString() could be called. Example: replaceAll(regEx, Str) = back.mkString().replaceAll(regEx, Str).

We can store both <code>CharSequence</code> and <code>Vector</code> in <code>Lazy</code> wrappers.

ruslansennov: Sorry, <b>one of</b> <code>CharSequence</code> or <code>Vector</code>

danieldietrich: yes, we may do that but before starting implementation please wait until later today...

ruslansennov: First of all I should to implement the remaining tests

patrox: Let me share some of my thoughts regarding Vector as a backing type:

In my impl of `WrappedString` I was building a `Stream` out of the backing `String` contents and then I was delegating methods to a corresponding `Stream` method, so if we would to store a `Vector<Character>` as a backing type, then we could delegate `WrappedString` method calls to already implemented `Vector` methods.

But I'm not certain how this would affect the performance, but this can (and should) be measured and compared. Does it make any sense?

ruslansennov: @patrox I completely agree with you about performance. <code>Vector</code> is slower than <code>CharSequence</code> in any case

ruslansennov: And another one: what about the long list of methods inherited from <code>java.lang.String</code>?

patrox: yes, that complicates using `Vector` a bit, as in this case it's far easier to just delegate method calls to the backing `java.lang.String` ... But I think that the most functionality is coming from `IndexSeq` - nevertheless, as this feature (`WrappedString`) will probably be used quite a lot, then I would say that performance should be a priority, correct?

danieldietrich: I've made some manual tests on the Scala REPL. We should keep it as is, with `back` of type String. Also the return type `Vector` is good is almost all places. I will provide comments in a minute where we need changes...

src/main/java/javaslang/WrappedString.java
```diff
@@ -0,0 +1,2200 @@
...
+    }
+
+    @Override
+    public WrappedString findAll(Predicate<? super Character> predicate) {
+        Objects.requireNonNull(predicate, "predicate is null");
+        return filter(predicate);
+    }
+
+    @Override
+    public  <U> Vector<U> flatMap(Function<? super Character, ? extends Iterable<? extends U>> mapper) {
```
danieldietrich: With `flatMap` we have a TODO in general but not here, with this pull-request.

Example:

```scala
scala> "123".flatMap(Some(_))
res0: String = 123

scala> "123".flatMap(c => Some(c + 0))
res1: scala.collection.immutable.IndexedSeq[Int] = Vector(49, 50, 51)

scala> val words = Set("Patryk", "Ruslan", "Daniel")
words: scala.collection.immutable.Set[String] = Set(Patryk, Ruslan, Daniel)

scala> words flatMap (word => word.toSet)
res2: scala.collection.immutable.Set[Char] = Set(e, s, n, y, t, u, a, i, l, P, r, R, k, D)

scala> words flatten (word => word.toSet)
res3: scala.collection.immutable.Set[Char] = Set(e, s, n, y, t, u, a, i, l, P, r, R, k, D)
```

Our flatMap works totally different. We need to check what is possible with Java, what makes sense and what is the expected behavior of flatMap. Maybe we need a Monad.unit(val) and Monoid.combine(m1, m2)...

src/main/java/javaslang/WrappedString.java
```diff
@@ -0,0 +1,2200 @@
...
+            if(i > 0) {
+                sb.append(element);
+            }
+            sb.append(get(i));
+        }
+        return sb.length() == 0 ? EMPTY : of(sb.toString());
+    }
+
+    @Override
+    public <U> Vector<U> map(Function<? super Character, ? extends U> mapper) {
```
danieldietrich: This looks really great! Compared to Scala:

```scala
scala> "abc".map(_ + "z")
res0: scala.collection.immutable.IndexedSeq[String] = Vector(az, bz, cz)

scala> "abc".map(_ + 0)
res1: scala.collection.immutable.IndexedSeq[Int] = Vector(97, 98, 99)
```

danieldietrich:
This is really huge! The more I read, the more I'm impressed!

src/main/java/javaslang/WrappedString.java
ruslansennov: It would be nice to create issue with link to this note

ruslansennov:
4,663 lines in 2 classes! :)

danieldietrich:
Yes, my computer has problems to display the web page :-)

src/main/java/javaslang/WrappedString.java
patrox: haha, that's funny - it reminds me of operator overloading in C++ ;)

danieldietrich: yes :-))

Issue #484, opened 2015-08-18
Get Traversable and Seq right (towards clean Set and Map interface)


danieldietrich:
Missing Traversable methods (from [Scala's Traversable](http://www.scala-lang.org/api/2.12.0-M2/index.html#scala.collection.immutable.Traversable)):

```scala
// more general form of fold and reduce 
aggregate[B](z: ⇒ B)(seqop: (B, A) ⇒ B, combop: (B, B) ⇒ B): B

// not sure about this. we have no PartialFunction yet.
// we count use a Function<A, Value<B>> and we test for value.isEmpty / value.isDefined
collect[B](pf: PartialFunction[A, B]): Traversable[B]
collectFirst[B](pf: PartialFunction[A, B]): Option[B]

// counts elements which satisfy the given predicate
count(p: (A) ⇒ Boolean): Int

// dual of dropWhile/takeWhile, same as dropWhile(!p)/takeWhile(!p)
dropUntil(p: (A) ⇒ Boolean): Traversable[A]
takeUntil(p: (A) ⇒ Boolean): Traversable[A]

// same as filter(!p)
filterNot(p: (A) ⇒ Boolean): Traversable[A]

// these basically check for isEmpty and wrap the result of the corresponding reduce methods
// in a Some, if not empty
reduceLeftOption[B >: A](op: (B, A) ⇒ B): Option[B]
reduceOption[A1 >: A](op: (A1, A1) ⇒ A1): Option[A1]
reduceRightOption[B >: A](op: (A, B) ⇒ B): Option[B]

// ...
scan[B >: A, That](z: B)(op: (B, B) ⇒ B)(implicit cbf: CanBuildFrom[Traversable[A], B, That]): That
scanLeft[B, That](z: B)(op: (B, A) ⇒ B)(implicit bf: CanBuildFrom[Traversable[A], B, That]): That
scanRight[B, That](z: B)(op: (A, B) ⇒ B)(implicit bf: CanBuildFrom[Traversable[A], B, That]): That

// returns the simple class name by default or the interface name (e.g. List/Stream).
// we need to change the toString methods to use stringPrefix instead of hard-coded strings.
stringPrefix: String

// creates uniform nested collections, i.e. List(Set(...), Set(...)).transform() = List(List(...),List(...)).
// may be realized analog to flatten() with a Match
transpose[B](implicit asTraversable: (A) ⇒ GenTraversableOnce[B]): Traversable[Traversable[B]]
```

[...]

PR #486, opened 2015-08-19
Add solution for Euler problem no. 13

patrox:
Also extracted helper methods `file` & `readLines` from `Euler99Test` to `Utils`,
so they can be reused in other tests (which involved reading from files).


danieldietrich:
Nice, many thanks!

---
Other topic:

@ruslansennov @patrox Next is to get Traversable and Seq right. If you both are in search for more, this is the right place to start: #484 

This may have side-effects to existing collections.

Because we are now three, we have to sync (fork/join) us in order not to work on the same things.

PR #499, opened 2015-08-22
Add tests for CharSeq.{map,mapChars}

patrox:
These two simple tests validate if the types returned by `CharSeq.map` & `CharSeq.mapChars`,
are those which we expect.

### DIFF
--- a/src/test/java/javaslang/collection/CharSeqTest.java
+++ b/src/test/java/javaslang/collection/CharSeqTest.java
@@ -420,7 +420,7 @@ public class CharSeqTest {
 
     @Test
     public void shouldFoldLeftNil() {
-        assertThat(this.<String> empty().foldLeft("", (xs, x) -> xs + x)).isEqualTo("");
+        assertThat(this.<String>empty().foldLeft("", (xs, x) -> xs + x)).isEqualTo("");
     }
 
     @Test(expected = NullPointerException.class)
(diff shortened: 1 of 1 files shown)


danieldietrich:
Great, many thanks!

It is confusing that Vector.equals checks for instanceof IndexedSeq. Could you chabge the check to Vector there?

patrox:
it's not that simple, as this will narrow the types with which `Vector` can be compared to only `Vector` itself. So it won't be possible to compare it with for example `CharSeq` - which as a result will brake some tests.

I've found out that there might be some incosistencies in them, as in this test:

```java
@Test
public void shouldMapNil() {
    assertThat(this.<Integer> empty().map(i -> i + 1)).isEqualTo(empty());
}
```

a result of `CharSeq.map` is compared against `CharSeq`, which is incorrect as `CharSeq.map` returns `Vector`. I'll try to review the failing tests and fix them.

danieldietrich:
yes, the behavior should be consistent. Vector should not be equal to CharSeq in any case.

patrox:
I'm almost done - maybe it would be best to do it in another PR, to keep it atomic ?

patrox:
I was lucky - 500th PR :) https://github.com/javaslang/javaslang/pull/500

danieldietrich:
:))

patrox:
I rebase the code in this PR to current master, so it can be merged without any conflicts.

danieldietrich:
roger that!

PR #500, opened 2015-08-22
Vector can be equal to only other Vector

patrox:
* Previously Vector could be compared against other instances,
 of the `IndexedSeq` base  class.
* Tests have to be updated, as they were using this functionality.


danieldietrich:
this is great! many thx

Issue #504, opened 2015-08-23
Set.equals fails to compare two identical Sets


patrox:
This tests failes:

```java
@Test public void shouldMapNonEmpty() {
  final Set<Integer> expected = HashSet.<Integer>of(1, 2);
  final Set<Integer> actual =
    HashMap.<Integer, String>of(
      Map.Entry.of(1, "1"),
      Map.Entry.of(2, "2"))
    .map(entry -> entry.key);
  assertThat(actual).isEqualTo(expected);
}
```

with rather interesting message:

```java
java.lang.AssertionError: 
Expecting:
 <"HashSet(1, 2) (HashSet@8001)">
to be equal to:
 <"HashSet(1, 2) (HashSet@8001)">
but was not.
```

The minimal test to reproduce it is:

```java
@Test
public void shouldBeEqual() {
    assertTrue(HashSet.of(1).equals(HashSet.of(1)));
}
```

This seems to be caused by using `? (any type)` wildcard when casting `HashSet.equals` param to a HashSet class:

```
  final HashSet<?> that = (HashSet<?>) o;
```

The solution for this is issue is to change the type wildcard to `T` which is the type of elements for `this` `HashSet` instance, but this could result in `ClassCastException` when a `HashSet` of different type will be passed to `equals`.

danieldietrich:
The underlying HashArrayMappedTrie may have stored the elements in an arbitrary order, so that the iterators do not return the elements in a specific order.

This lead to this implementation:

@SuppressWarnings("unchecked")
@Override
public boolean equals(Object o) {
    if (o == this) {
        return true;
    } else if (o instanceof HashSet) {
        final HashSet<?> that = (HashSet<?>) o;
        return this.length() == that.length() && ((HashSet<Object>) this).containsAll(that);
    } else {
        return false;
    }
}
We do not cast to <T> in order to prevent ClassCastExceptions.

danieldietrich:
Fixed with #505

Issue #488, opened 2015-08-19
Rethink flatMap et al.


danieldietrich:
The following is taken from a [pull request comment](https://github.com/javaslang/javaslang/pull/485#discussion_r37454043):

With `flatMap` we have a TODO in general but not here, with this pull-request.

Example:

```scala
scala> "123".flatMap(Some(_))
res0: String = 123

scala> "123".flatMap(c => Some(c + 0))
res1: scala.collection.immutable.IndexedSeq[Int] = Vector(49, 50, 51)

scala> val words = Set("Patryk", "Ruslan", "Daniel")
words: scala.collection.immutable.Set[String] = Set(Patryk, Ruslan, Daniel)

scala> words flatMap (word => word.toSet)
res2: scala.collection.immutable.Set[Char] = Set(e, s, n, y, t, u, a, i, l, P, r, R, k, D)

scala> words flatten (word => word.toSet)
res3: scala.collection.immutable.Set[Char] = Set(e, s, n, y, t, u, a, i, l, P, r, R, k, D)
```

Our flatMap works totally different. We need to check what is possible with Java, what makes sense and what is the expected behavior of flatMap. Maybe we need a Monad.unit(val) and Monoid.combine(m1, m2)...

patrox:
I can pick this up - and will come back tomorrow with some ideas, ok ?

ruslansennov:
I'll take a break :)

danieldietrich:
@patrox ok, this will be fun!

danieldietrich:
@patrox if any questions (e.g. regarding existing code) - I'm here!

danieldietrich:
@ruslansennov 

> I'll take a break :)

you've more than deserved it :)

patrox:
@ruslansennov you deserved some :beers: :)

ruslansennov:
@patrox it's really a good idea

patrox:
@danieldietrich, @ruslansennov, regarding `flatMap` and/or `map`, here are my findings:

I found that when we run any transformation function on a plain `String` in **Scala**, we will go through the following sequence:
1.) First the `String` is converted (implicitly) to `StringOps` instance (http://www.scala-lang.org/api/current/index.html#scala.collection.immutable.StringOps)
2.) Then the requested function (method) is called
3.) Lastly the result is converted back to `String`
 
[...]

PR #505, opened 2015-08-23
Implement HashMap.map, fix HashSet.equals

### LINKED ISSUE #503: NPE thrown while combining Stream.groupBy with Map.map
While working on one of Euler problems I've found a nasty bug,
for now I've isolated it in a test and I'm working on a fix for it,
please see below for the details.

The following (not yet existing in master) test is failing:

```java
@Test
public void shouldGroupByAndMapProperly() {
   assertThat(Stream.of(1)
      .groupBy(Function.identity())
      .map(entry -> entry.key)
      .toList()).isEqualTo(List.of(1));
}
```
```
java.lang.NullPointerException
	at javaslang.collection.StreamTest.shouldGroupByAndMapProperly(StreamTest.java:294)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:78)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:212)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:68)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:140)
```
whereas, the corresponding scala code produces the expected results, without any problems:

```scala
Stream(1)
  .groupBy(e => e)
  .map(entry => entry._1)
  .toList
```

```
List[Int] = List(1)
```

patrox:
As a "side effect" of running tests for `HashMap.map`,
a problem with `HashSet.equals` was identified and fixed.

Fixes #503 and #504.

### DIFF
--- a/src/main/java/javaslang/collection/HashMap.java
+++ b/src/main/java/javaslang/collection/HashMap.java
@@ -258,10 +258,10 @@ public final class HashMap<K, V> implements Map<K, V>, Serializable {
     public <U, W> Map<U, W> map(BiFunction<? super K, ? super V, ? extends Entry<? extends U, ? extends W>> mapper) {
         return null;
     }
-
     @Override
     public <U> Set<U> map(Function<? super Entry<K, V>, ? extends U> mapper) {
-        return null;
+        Objects.requireNonNull(mapper, "mapper is null");
+        return foldLeft(HashSet.empty(), (acc, entry) -> acc.add(mapper.apply(entry)));
     }
 
     @Override
--- a/src/main/java/javaslang/collection/HashSet.java
+++ b/src/main/java/javaslang/collection/HashSet.java
@@ -480,14 +480,14 @@ public final class HashSet<T> implements Set<T>, Serializable {
         return hash.get();
     }
 
-    @Override
     @SuppressWarnings("unchecked")
+    @Override
     public boolean equals(Object o) {
         if (o == this) {
             return true;
         } else if (o instanceof HashSet) {
-            final HashSet<T> that = (HashSet<T>) o;
-            return this.iterator().equals(that.iterator());
+            final HashSet<?> that = (HashSet<?>) o;
+            return this.length() == that.length() && ((HashSet<Object>) this).containsAll(that);
         } else {
             return false;
         }
--- a/src/main/java/javaslang/collection/HashSet.java
+++ b/src/main/java/javaslang/collection/HashSet.java
@@ -481,11 +481,12 @@ public final class HashSet<T> implements Set<T>, Serializable {
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public boolean equals(Object o) {
         if (o == this) {
             return true;
         } else if (o instanceof HashSet) {
-            final HashSet<?> that = (HashSet<?>) o;
+            final HashSet<T> that = (HashSet<T>) o;
             return this.iterator().equals(that.iterator());
         } else {
             return false;
--- a/src/test/java/javaslang/collection/HashMapTest.java
+++ b/src/test/java/javaslang/collection/HashMapTest.java
@@ -1,0 +1,36 @@
+/*     / \____  _    ______   _____ / \____   ____  _____
+ *    /  \__  \/ \  / \__  \ /  __//  \__  \ /    \/ __  \   Javaslang
+ *  _/  // _\  \  \/  / _\  \\_  \/  // _\  \  /\  \__/  /   Copyright 2014-2015 Daniel Dietrich
+ * /___/ \_____/\____/\_____/____/\___\_____/_/  \_/____/    Licensed under the Apache License, Version 2.0
+ */
+package javaslang.collection;
+
+import org.junit.Test;
+
+import static org.assertj.core.api.StrictAssertions.assertThat;
+
+public class HashMapTest {
+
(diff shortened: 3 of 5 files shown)


src/main/java/javaslang/collection/HashMap.java
```diff
@@ -258,10 +258,10 @@ public int length() {
     public <U, W> Map<U, W> map(BiFunction<? super K, ? super V, ? extends Entry<? extends U, ? extends W>> mapper) {
         return null;
     }
-
     @Override
     public <U> Set<U> map(Function<? super Entry<K, V>, ? extends U> mapper) {
-        return null;
+        Objects.requireNonNull(mapper, "mapper is null");
+        return foldLeft(HashSet.empty(), (acc, entry) -> acc.add(mapper.apply(entry)));
```
danieldietrich: this is nice!

patrox: I took the functional route to do it. I think that mr. Odersky would be satisfied ;)

src/main/java/javaslang/collection/HashSet.java
```diff
@@ -481,11 +481,12 @@ public int hashCode() {
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public boolean equals(Object o) {
         if (o == this) {
             return true;
         } else if (o instanceof HashSet) {
-            final HashSet<?> that = (HashSet<?>) o;
+            final HashSet<T> that = (HashSet<T>) o;
```
danieldietrich: I tested a little bit and recognized that we need another equals implementation.

The underlying HashArrayMappedTrie may have stored the elements in an arbitrary order, so that the iterators do not return the elements in a specific order.

This lead to this implementation:

```java
@Override
    if (o == this) {
        return true;
    } else if (o instanceof HashSet) {
        final HashSet<?> that = (HashSet<?>) o;
        return this.length() == that.length() && ((HashSet<Object>) this).containsAll(that);
    } else {
        return false;
    }
}
```

We do not cast to `<T>` in order to prevent ClassCastExceptions.

Thanks for for your finding!!

src/main/java/javaslang/collection/HashMap.java
danieldietrich: :+1:

patrox:
OK, so I will close this one, as `HashMap.map` will be implemented in a separate PR.
Bottom line is that implementing Euler project problems (using javaslang) is a very good way to test it very thoroughly - it's a bit like an integration test - as we're mixing various types (Streams, Maps, Sets, CharSeqs) with various methods.

danieldietrich:
Yes, the Euler tests are great for testing Javaslang.

Please implement the equals method in this PR I suggested above. I will pull it then. HashMap needs no separate PR.

Thx!

danieldietrich:
I will take care now of flatMap and map in general as described here: https://github.com/javaslang/javaslang/issues/488#issuecomment-133760856

I can pull this into master today...

patrox:
Done, coverage increased by 0.2% :)
But I'm still wondering how the order of elements in `HashArrayMappedTrie` could affect the `HashSet`s from my 'minimal' example - containing 1 element ?

danieldietrich:
Many thanks :)

For your minimal example all should be ok. Currently it is not clear why the Iterator.equals failed - we will further investigate. The order of similar elements may change in general (having more than one element) because the Map (or HAMT) has a key set with no guarantee of order.

Issue #501, opened 2015-08-22
Rename MonadOps, IterableOps and ConversionOps


danieldietrich:
These are bad names for Java classes because in Scala XxxOps are implicit classes which define extension methods for specific types.

Here, these interfaces are used to group related methods of Value. Additionally these interfaces are not intended to be public API.

We may name them `MonadLike`, `IterableLike` and `ConvertibleLike`. It is crucial to understand that these types are not meant to be general purpose. They are related to Value only.

PR #507, opened 2015-08-24
map and flatMap

### LINKED ISSUE #488: Rethink flatMap et al.

danieldietrich:
Fixes #488


danieldietrich:
Also renamed `(Hash)Map.merged()` to `merge()` and added `SortedMap`.

danieldietrich:
Also fixes #501

danieldietrich:
@patrox This will probably cause merge conflicts in HashMap ...
/cc @ruslansennov

patrox:
@danieldietrich not a problem, altough I used them a bit differently:
```java
// TODO: please implement me
throw new UnsupportedOperationException("Not implemented yet!");
```

Because such comments are usually picked up by the IDE (in my case Idea), it's very easy to quicky find them all :)

danieldietrich:
yes, that's nice. I will do that now

PR #509, opened 2015-08-24
Add test for Euler 14

patrox:
Example of how to use memoization and currying.

### DIFF
--- a/src/test/java/javaslang/collection/euler/Euler14Test.java
+++ b/src/test/java/javaslang/collection/euler/Euler14Test.java
@@ -1,0 +1,57 @@
+/*     / \____  _    ______   _____ / \____   ____  _____
+ *    /  \__  \/ \  / \__  \ /  __//  \__  \ /    \/ __  \   Javaslang
+ *  _/  // _\  \  \/  / _\  \\_  \/  // _\  \  /\  \__/  /   Copyright 2014-2015 Daniel Dietrich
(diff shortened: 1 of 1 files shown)


danieldietrich:
How cool is that!? I love this collection API :-))

ruslansennov:
:+1:

danieldietrich:
A collegue of mine said we can start from `500_000L` instead of `1L`. I will test that later to reduce the time (currently 20sec on travis-ci).

patrox:
yes, I'm aware that it takes some time to complete, for me it was around 6-7 seconds, but as for a test it was a bit slow - altough it's not a *unit test* per se. The most frustrating thing is that the corresponding Scala code completes below 1 sec ... I was wondering if we could make the Euler problems tests optional and run the remaining tests - these which actually test the core API.

danieldietrich:
it's ok - it doesn't really matter. I like these high-level unit tests to test real-world application of our code. Scala has tail-recursion to execute tail-ended recursion in constant stack...

patrox:
yes, similarily like in Clojure, as these languages support TCO (tail call optimization), which basically transforms tailrec calls in loops :)
BTW - I really miss Scala's @tailrec annotation :)

danieldietrich:
Yes, @tailrec for Java would be really cool. Technically now it is possible said Brian Goetz but unfortunately it is not targetet to be implemented, yet...

PR #512, opened 2015-08-25
HashSet special cases

ruslansennov:


danieldietrich:
Great you refactor existing code to make it simpler!!

PR #513, opened 2015-08-25
Infinite loop in Function* fixed

ruslansennov:
@danieldietrich please check this test on current branch
```java
private static Function1<Integer, Integer> recurrent1 = (i1) -> i1 <= 0 ? i1 : Function1Test.recurrent2.apply(i1 - 1) + 1;
private static Function1<Integer, Integer> recurrent2 = Function1Test.recurrent1.memoized();

@Test
public void shouldCalculatedRecursively() {
    assertThat(recurrent1.apply(11)).isEqualTo(11);
    assertThat(recurrent1.apply(22)).isEqualTo(22);
}
```
I believe you'll see infinite loop because of internal architecture of <code>ConcurrentHashMap</code>. We should use <code>java.util.HashMap</code> inside <code>synchronized</code> block...

### DIFF
--- a/generator/Generator.scala
+++ b/generator/Generator.scala
@@ -505,12 +505,26 @@ def generateMainClasses(): Unit = {
                           return ($className$fullGenerics & Memoized) Lazy.of($mappingFunction)::get;
                         """ else if (i == 1) xs"""
                           final Lazy<R> forNull = Lazy.of($forNull);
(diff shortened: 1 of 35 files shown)


danieldietrich:
You're right - I checked it on current branch and it results in an infinite loop. Good finding!

PR #515, opened 2015-08-25
sliding(int,int) bug fixed

ruslansennov:
O_o

### DIFF
--- a/src/main/java/javaslang/collection/List.java
+++ b/src/main/java/javaslang/collection/List.java
@@ -1143,9 +1143,8 @@ public interface List<T> extends LinearSeq<T>, Stack<T> {
         List<List<T>> result = Nil.instance();
         List<T> list = this;
         while (!list.isEmpty()) {
-            final Tuple2<List<T>, List<T>> split = list.splitAt(size);
-            result = result.prepend(split._1);
-            list = split._2.isEmpty() ? Nil.instance() : list.drop(step);
+            result = result.prepend(list.take(size));
+            list = list.drop(step);
         }
         return result.reverse();
     }
--- a/src/main/java/javaslang/collection/Stream.java
+++ b/src/main/java/javaslang/collection/Stream.java
@@ -1082,8 +1082,7 @@ public interface Stream<T> extends LinearSeq<T> {
         if (isEmpty()) {
             return Nil.instance();
         } else {
-            final Tuple2<Stream<T>, Stream<T>> split = splitAt(size);
-            return new Cons<>(() -> split._1, () -> split._2.isEmpty() ? empty() : drop(step).sliding(size, step));
+            return new Cons<>(() -> take(size), () -> drop(step).sliding(size, step));
         }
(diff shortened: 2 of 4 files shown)


danieldietrich:
Mmhh, I first had an intuitive solution, similar to yours, but changed it to the current because Scala REPL says:

```scala
scala> List(1, 2, 3, 4, 5).sliding(2).toList
res0: List[List[Int]] = List(List(1, 2), List(2, 3), List(3, 4), List(4, 5))

scala> List(1, 2, 3, 4).sliding(5, 3).toList
res1: List[List[Int]] = List(List(1, 2, 3, 4))
```

I don't think we have a bug here.

ruslansennov:
O_o

danieldietrich:
8-)

ruslansennov:
But
```scala
scala> List(1, 2, 3, 4, 5).sliding(2, 4).toList
res0: List[List[Int]] = List(List(1, 2), List(5))
```
This should be investigated...

BTW, we HAVE a bug!
This code is completely wrong.
```java
        while (!list.isEmpty()) {
            final Tuple2<List<T>, List<T>> split = list.splitAt(size);
            result = result.prepend(split._1);
            list = split._2.isEmpty() ? Nil.instance() : list.drop(step);
        }
```
Should be something like this
```java
        while (list.length() >= size) {
            result = result.prepend(list.take(size));
            list = list.drop(step);
        }
```
or like this
```java
        while (list.length() >= size) {
            result = result.prepend(list.take(size));
            list = list.length() <= step ? empty() : list.drop(step);
        }
```

danieldietrich:
Javaslang your first example right:

```java
List.of(1, 2, 3, 4, 5).sliding(2, 4); // = List(List(1, 2), List(5))
```

To prove the code wrong, we need a counter-example, i.e. an example where Scala's result is not equal to Javaslang's result.

```java
while (!list.isEmpty()) {

    // = Tuple2(take(size), drop(size))
    // splitAt is great here, because take and drop is calculated in one turn (e.g. for List)
    final Tuple2<List<T>, List<T>> split = list.splitAt(size);

    // essentially the same as result.prepend(list.take(size))
    result = result.prepend(split._1);

    // we need the drop(size) part (= split._2) for the case step < size.
    // because drop(size) = empty => we are finished. but drop(step) may not be empty.
    list = split._2.isEmpty() ? Nil.instance() : list.drop(step);

}
```

My tests led to similar results like Scala.

I know you have a good sense for code. What do you think is wrong exactly?

ruslansennov:
It was an obsession, sorry :)

danieldietrich:
obsession is good - keep it up and running :)

danieldietrich:
Half the way there. Moved many static factory methods to Iterator. Collections may use these by delegating their static factory methods via `ofAll(() -> Iterator.xxx(...))`. TODOs added in `Iterator`. Currently the stuff is in my fork... Will be pulled soon into master.

PR #521, opened 2015-08-26
Iterator.dropRight() fixed

ruslansennov:
@danieldietrich  I believe your implementation is wrong and is same as drop(n)


danieldietrich:
You are totally right.

(I push as much in as I can. Error will pop up at the latest when we have 99.99% coverage again...)

ruslansennov:
btw, List.dropRight() now is faster: only one reverse operation

danieldietrich:
Nice!
Thanks for all the effort. Our collections will be very useful for the Java world.

PR #522, opened 2015-08-27
Eulers 18, 67 solved

ruslansennov:


danieldietrich:
Thx!
I've stared a while on `naive` and `smart` to see the difference :-)
It is only `memoized`, right?

patrox:
BTW, I'm working on Euler 17 :)

ruslansennov:
> It is only <code>memoized</code>, right?

exactly :smile: 
and with <code>ConcurrentHashMap</code> in <code>Memoized</code> implementation this trick doesn't work :cry:

danieldietrich:
@patrox Ok!

danieldietrich:
@ruslansennov that's really a weakness of ConcurrentHashMap. I wonder, if the creator(s) are aware of it...

PR #529, opened 2015-08-29
Seq.padTo() #484

ruslansennov:
...and startsWith tests

### DIFF
--- a/src/main/java/javaslang/collection/CharSeq.java
+++ b/src/main/java/javaslang/collection/CharSeq.java
@@ -388,8 +388,11 @@ public final class CharSeq implements CharSequence, IndexedSeq<Character>, Seria
 
     @Override
     public CharSeq padTo(int length, Character element) {
+        if(length <= back.length()) {
+            return this;
+        }
         final StringBuilder sb = new StringBuilder(back);
-        for (int i = 0; i < length; i++) {
+        for (int i = 0; i < length - back.length(); i++) {
             sb.append(element);
         }
         return new CharSeq(sb.toString());
--- a/src/main/java/javaslang/collection/CharSeq.java
+++ b/src/main/java/javaslang/collection/CharSeq.java
@@ -386,6 +386,15 @@ public final class CharSeq implements CharSequence, IndexedSeq<Character>, Seria
         return result;
     }
 
+    @Override
+    public CharSeq padTo(int length, Character element) {
+        final StringBuilder sb = new StringBuilder(back);
+        for (int i = 0; i < length; i++) {
+            sb.append(element);
+        }
+        return new CharSeq(sb.toString());
+    }
+
     public CharSeq mapChars(CharUnaryOperator mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
         if (isEmpty()) {
--- a/src/main/java/javaslang/collection/CharSeq.java
+++ b/src/main/java/javaslang/collection/CharSeq.java
@@ -392,7 +392,8 @@ public final class CharSeq implements CharSequence, IndexedSeq<Character>, Seria
             return this;
         }
         final StringBuilder sb = new StringBuilder(back);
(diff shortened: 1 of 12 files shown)


src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -386,6 +386,15 @@ public CharSeq intersperse(Character element) {
         return result;
     }
 
+    @Override
+    public CharSeq padTo(int length, Character element) {
+        final StringBuilder sb = new StringBuilder(back);
+        for (int i = 0; i < length; i++) {
```
danieldietrich: or is it `i < length - back.length()`?

`"12345".padTo(10, 'a')` should be `12345aaaaa`

src/main/java/javaslang/collection/List.java
```diff
@@ -856,6 +856,15 @@ default int lastIndexOf(T element, int end) {
     }
 
     @Override
+    default List<T> padTo(int length, T element) {
+        if(length == 0) {
+            return this;
+        } else {
+            return appendAll(Stream.gen(() -> element).take(length));
```
danieldietrich: here also - it should be take(length - this.length()), am I right?

src/main/java/javaslang/collection/Stream.java
```diff
@@ -931,6 +931,15 @@ default int lastIndexOf(T element, int end) {
     }
 
     @Override
+    default Stream<T> padTo(int length, T element) {
+        if(length == 0) {
+            return this;
+        } else {
+            return appendAll(Stream.gen(() -> element).take(length));
```
danieldietrich: dito

src/main/java/javaslang/collection/Vector.java
```diff
@@ -881,6 +881,15 @@ public int length() {
     }
 
     @Override
+    public Vector<T> padTo(int length, T element) {
+        if(length == 0) {
+            return this;
+        } else {
+            return appendAll(Stream.gen(() -> element).take(length));
```
danieldietrich: dito

src/main/java/javaslang/collection/CharSeq.java
ruslansennov: wow, I forgot to add <code>@Test</code> annotations to all pad tests :)

src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -386,6 +386,18 @@ public CharSeq intersperse(Character element) {
         return result;
     }
 
+    @Override
+    public CharSeq padTo(int length, Character element) {
+        if(length <= back.length()) {
+            return this;
+        }
+        final StringBuilder sb = new StringBuilder(back);
+        for (int i = 0; i < length - back.length(); i++) {
```
danieldietrich: just a little thing: let's introduce a `final int bound = length - back.length();` that saves us calculating the bound each time checking the loop condition.

src/main/java/javaslang/collection/Vector.java
```diff
@@ -881,6 +881,15 @@ public int length() {
     }
 
     @Override
+    public Vector<T> padTo(int length, T element) {
+        if(length <= length()) {
+            return this;
+        } else {
+            return appendAll(Stream.gen(() -> element).take(length - length()));
```
danieldietrich: Now that we have `Iterator` (mostly) up and running, we should use `Iterator` instead of `Stream`. It is more lightweight in the sense of number of objects instantiated. I will create a separate issue for that to add `Iterator.gen(...)` and `Iterator.from(...)`.

ruslansennov: it's a good idea

danieldietrich:
thx!

danieldietrich:
@ruslansennov @patrox btw - there is a new blog... http://blog.javaslang.com

PR #532, opened 2015-08-30
HAMT backed by Array

ruslansennov:


ruslansennov:
There is no significant difference in travis (-10%), but runs 2 times faster on my computer.

danieldietrich:
Great improvement!!

That shows that we can't rely on travis-ci for measuring performance. It is a cloud, cpu is utilized virtually - we don't know what other boxes are running on the same hardware and eating cpu in this virtual env.

ruslansennov:
I don't know... I tested it with data size = 100k instead of 5k on travis...

src/main/java/javaslang/collection/HashArrayMappedTrie.java
```diff
@@ -328,21 +328,21 @@ private IndexedNode(int bitmap, List<AbstractNode<K, V>> subNodes) {
             int bit = mask;
             int count = 0;
             List<AbstractNode<K, V>> sub = subNodes;
-            List<AbstractNode<K, V>> arr = List.empty();
+            final Object[] arr = new Object[BUCKET_SIZE];
```
danieldietrich: This line brought me to the question: why we aren't using `T[]` instead of `Object[]`?

ruslansennov: <code>javaslang.collection.Array</code> backed by <code>Object[]</code> but not by <code>T[]</code> because of generic types. It is not possible in Java create arrays like <code>Tuple2&lt;Integer,Integer&gt;[]</code>

danieldietrich: Also in Array. Wouldn't it be better to make back of type `T[]` and additionally provide factory methods to create typed, empty arrays of a specific size? E.g. `static <T> T[] create(int size)` or maybe `static <T> T[] create(int size, Class<T> type)`. Then the package internal `wrap` method can be possibly removed.

```java
// T[] instead?
private final Object[] back;

private Array(Object[] back) {
    this.back = back;
}
```

ruslansennov: > Wouldn't it be better to make back of type T[] and additionally provide factory methods to create typed, empty arrays of a specific size?

No, it's forbidden and will fails in runtime

ruslansennov: But with <code>Class&lt;T&gt; type</code> as parameter it is possible...

ruslansennov: I don't like to pass a Class&lt;T&gt; as a parameter because it will change the Array's methods signatures as compared with all the other

danieldietrich: Yes, we shouldn't change the signatures.

danieldietrich:
Thx!

PR #535, opened 2015-08-30
Revert "made Queue.hashCode transient"

danieldietrich:
Reverts javaslang/javaslang#534

actually me can't make Queue.hashCode `transient`. Wonder why this works for List.Cons!?


ruslansennov:
mmm... List.Cons has a SerializationProxy, but Queue is not

Issue #477, opened 2015-08-17
Fix the test coverage in direction of 100% (until release)


danieldietrich:
> Untested code is broken code.

ruslansennov:
@danieldietrich please reopen this issue. Now we have &lt;90% tests coverage, this is not good

danieldietrich:
@ruslansennov You're right - I closed it accidentally

arnonmoscona:
Hello!

 

New message, please read <http://lauraclauvi.ro/conduct.php?wb3g8>

 

arnon@moscona.com

danieldietrich:
See #936

danieldietrich:
Missing conversion tests:

* Map/HashMap/LinkedHashMap/TreeMap#toJavaMap()
* Set/HashSet/LinkedHashSet/TreeSet#toJavaSet()

Missing narrow tests:

* Future#narrow(Future)
* Lazy#narrow(Lazy)
* Option#narrow(Option)
* Promise#narrow(Promise)
* Try#narrow(Try)
* Value#narrow(Value)

danieldietrich:
we are at 96% - I will close this issue, tests are an ongoing task

PR #539, opened 2015-08-31
Value tests added #477

ruslansennov:


src/main/java/javaslang/collection/HashMap.java
```diff
@@ -440,9 +440,15 @@ public int hashCode() {
     public boolean equals(Object o) {
         if (o == this) {
             return true;
-        } else if (o instanceof HashSet) {
-            final HashSet<?> that = (HashSet<?>) o;
-            return this.iterator().equals(that.iterator());
```
ruslansennov: @danieldietrich this calls Object.equals() but not Iterator.equals() because of well known problem with Map.Entry generics

src/main/java/javaslang/collection/HashMap.java
```diff
@@ -440,9 +440,15 @@ public int hashCode() {
...
+        } else if (o instanceof HashMap) {
+            final Iterator<?> thisIt = this.iterator();
```
danieldietrich: Let's change `Iterator.equals(Iterator<? extends T> that)` to `Iterator.equals(Iterator<?> that)` and then

```
final HashMap<?, ?> that = (HashMap<?, ?>) o;
return this.iterator().equals(that.iterator())`?
```

this reduces duplicate code and may be also used in other collections.

src/main/java/javaslang/collection/HashMap.java
danieldietrich: ok. what if we relax `Iterator.equals(...)` as I described below?

ruslansennov: yes, it works! :)

ruslansennov:
+20% to Value tests coverage :)

danieldietrich:
Very cool!!! (c:

PR #541, opened 2015-09-01
Seq.patch() #484

ruslansennov:


ruslansennov:
Something wrong in this PR...
I should implement all override methods in all classes

ruslansennov:
Now it looks better

danieldietrich:
Nice!

PR #543, opened 2015-09-02
some TODOs in Iterator fixed #487

### LINKED ISSUE #487: Make javaslang.collection.Iterator a Traversable
To do this, we (re-)add TraversableOnce. All methods of Traversable are move to TraversableOnce and Traversable extends TraversableOnce.

Then Iterator extends TraversableOnce.

Additionally we add two methods to Traversable~~Once~~:

* `boolean hasDefiniteSize()` // ~~Iterator,~~ Stream ~~, Gen and Arbitrary~~ return(s) false, all others return true
* `boolean isTraversableAgain()` // ~~Iterator returns false,~~ all (others) return true

Note: TraversableOnce has undefined size and is not traversable again by definition.

ruslansennov:


src/main/java/javaslang/collection/HashSet.java
```diff
@@ -186,7 +186,7 @@ public boolean contains(T element) {
 
     @Override
     public Option<T> findLast(Predicate<? super T> predicate) {
-        throw new UnsupportedOperationException("TODO");
+        return findFirst(predicate);
```
danieldietrich: great!!

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1334,8 +1452,35 @@ public T next() {
...
+            return empty();
+        } else {
+            final Iterator<T> that = this;
+            return new Iterator<T>() {
+                private Queue<T> queue = Queue.empty();
+
+                @Override
+                public boolean hasNext() {
+                    while (that.hasNext()) {
+                        queue = queue.append(that.next());
```
danieldietrich: I would call `queue.enqueue(that.next())` (which is the same)

danieldietrich:
Thx for taking care of HashSet & Iterator!

Today we celebrated the birthday of my 5 yo daugther - 12 kids(!) where there. Now I'm relaxing by implementing the RedBlackTree :)

ruslansennov:
> 12 kids(!) where there

local hell

danieldietrich:
yep

PR #546, opened 2015-09-03
Iterator 80 LVL

ruslansennov:


src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1679,6 +1667,22 @@ public T next() {
 
     abstract class AbstractIterator<T> implements Iterator<T> {
 
+        public boolean equals(Object o) {
```
danieldietrich: What is the use case for overriding Iterator.equals? If we compare it, it is gone.

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1679,6 +1667,22 @@ public T next() {
...
+            if(o instanceof java.lang.Iterable) {
```
danieldietrich: Throughout Javaslang equals is implemented as follows:

1. o instance of this.getClass(), i.e. o != null && this.getClass() is assignable from o.getClass()
2. fields are equal

To compare values only, one could do s.th. similar to Stream.ofAll(iter1).equals(Stream.ofAll(iter2)).

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1679,6 +1667,22 @@ public T next() {
...
+                java.lang.Iterable<?> iterable = (java.lang.Iterable<?>) o;
+                Objects.requireNonNull(iterable, "iterable is null");
```
danieldietrich: unnecessary check. o instanceof ... returns false if o is null.

src/main/java/javaslang/collection/RedBlackTree.java
```diff
@@ -297,7 +297,7 @@ public int blackHeight() {
 
         @Override
         public boolean contains(T value) {
-            final int result = empty.comparator.compare(value, value);
+            final int result = empty.comparator.compare(value, this.value);
```
danieldietrich: whoops :)

src/main/java/javaslang/collection/Iterator.java
ruslansennov: This is important fix. It was surprise for me, but Object.equals() overrides and hides Iterator.equals() method. We should implement it in class but not in interface.

danieldietrich: yes, you are right, interfaces in Java 8 cannot have default methods which override methods of Object. I intentionally left it unoverridden. However, let's do it your way. But we need to check `o instanceof Iterator` IMO. If someone needs to compare an iterable, one could call `iterator.equals(iterable.iterator())`.

src/main/java/javaslang/collection/RedBlackTree.java
ruslansennov: :sunglasses:

src/main/java/javaslang/collection/Iterator.java
danieldietrich: on the other hand

```java
Iterator<?> that = List.of(0, 1, 2, 3).iterator();
that = that.next();
List.of(1, 2, 3).iterator().equals(that); // = true
```

ruslansennov: OK

ruslansennov: agrrrhhh!!! :) now I changed checking to Iterator. You can fix it later as you wish :)

danieldietrich: haha, ok :)

danieldietrich: Iterator equality is a **very** interesting question. I asked Scala. It says:

```
scala> List(1, 2, 3) == List(1, 2, 3)
res0: Boolean = true

scala> List(1, 2, 3) == Stream(1, 2, 3)
res1: Boolean = true

scala> Stream(1, 2, 3) == List(1, 2, 3)
res2: Boolean = true

scala> Iterator(1, 2, 3) == List(1, 2, 3)
res3: Boolean = false

scala> Iterator(1, 2, 3) == Iterator(1, 2, 3)
res4: Boolean = false

scala> Iterator(1, 2, 3)
res5: Iterator[Int] = non-empty iterator

scala> res5 == res5
res6: Boolean = true
```

danieldietrich: Mmhh, Scala's equals `==` is odd:

```scala
scala> Some(1) == List(1)
<console>:8: warning: comparing values of types Some[Int] and List[Int] using `==` will always yield false
              Some(1) == List(1)
                      ^
res10: Boolean = false

scala> Map(1 -> 'a', 2 -> 'b') == List((1 -> 'a'), (2 -> 'b'))
res11: Boolean = false

scala> (1 -> 'a')
res12: (Int, Char) = (1,a)

scala> Map(1 -> 'a')
res13: scala.collection.immutable.Map[Int,Char] = Map(1 -> a)

scala> Map(1 -> 'a').toList
res14: List[(Int, Char)] = List((1,a))

scala> Map(1 -> 'a').toList == List((1 -> 'a'))
res15: Boolean = true

scala> Map(1 -> 'a') == List((1 -> 'a'))
res16: Boolean = false
```

Javaslang's Option/Some/None is Iterable. Miming Scala's behavior we could compare Iterables in general (e.g. a default equals method in `Iterable` interface). Then Some(1).equals(List(1)) and None.equals(List.empty()) etc.
Or we may provide a method `eq` for that beside `equals`.

Issue #547, opened 2015-09-03
Rewrite equals of all Iterable classes


danieldietrich:
There is a pretty good reason to let `iterable1.equals(iterable2) == true`, if iterable1 containts the same elements as iterable2: **inheritance**.

```java
Seq<T> seq1 = List.of(1, 2, 3);
Seq<T> seq2 = Stream.of(1, 2, 3);
seq1.equals(seq2); // = true, as expected!
```
Exception: Iterator - because it is short-lived and therefore should not be evaluated

/cc @ruslansennov - thx for talking about it!

danieldietrich:
This can be used for all _sequential_ classes:

```java
boolean equals(Iterable<? extends T> iterable1, Iterable<? extends T> iterable2) {
    Objects.requireNonNull(iterable1, "iterable1 is null");
    Objects.requireNonNull(iterable2, "iterable2 is null");
    final Iterator<T> iterator1 = iterable1.iterator();
    final Iterator<T> iterator2 = iterable2.iterator();
    while (iterator1.hasNext() && iterator2.hasNext()) {
        if (!Objects.equals(iterator1.next(), iterator2.next())) {
            return false;
        }
    }
    return iterator1.hasNext() == iterator2.hasNext();
}
```

For classes with no specific order (e.g. Set, Map, ...) we need to take another approach.

For Values with one element (e.g. Option, Try, Lazy, ...) we should also take the sequential approach (see above):

```java
new Some<>(1).equals(List.of(1)); // = true

new Some<>(1).equals(List.of(1, 2, 3)); // = false

None.instance().equals(List.empty()); // = true
```

danieldietrich:
Maybe it makes sense to add an additional, exact equality match (by comparing the type via instanceof), e.g.

```java
new Some<>(1).equals(List.of(1)); // = false
new Some<>(1).equals(new Some<>(1)); // = true
```

ruslansennov:
Our goal is to make <code>equals()</code> behavior exactly as in Scala, right?

danieldietrich:
> When you call equals directly, you are skipping all that softening logic and instead treated to java's theory that two boxed values of different types are always unequal. 

\- Paul Phillips

We should keep our equals methods as they are. This is how Java devs would expect them to work. However, I will change Iterator.equals most probably to work like `==`. This will be documented in the interface.

[...]

PR #549, opened 2015-09-05
Seq.corresponds()

ruslansennov:


danieldietrich:
Hi Ruslan,

this is basically the same as `seq.zip(that).forAll(predicate)`. We should not broaden our interfaces too much. It's better to have a _thin_ but _deep_ API:

* _thin_ in the sense of number of basic building blocks, opposed to the number of special methods
* _deep_ in the sense of power of composition et al., and a nice type hierarchy

Or did you have s.th. in mind I've overseen?

danieldietrich:
(I'm now offline for a few hours - have to repair a broken computer of a friend o_O)

danieldietrich:
online again :)

btw - `iterator()` can be used to circumvent two iterations in the case of List for example:

```java
list.iterator().zip(that).forAll(predicate) // calculated in one iteration
```

ruslansennov:
Your solution doesn't work if sequences have different lengths... am I right?

danieldietrich:
Of course you're right. after taking my first coffee I see clear now :)

Let's rename it to `equals` and change the signature like this:

```java
default <U> boolean equals(java.lang.Iterable<? extends U> that, Predicate<Tuple2<? super T, ? super U>> predicate)
```

I'm not 100% sure, if we really need `Iterable<? extends U>` instead of `Iterable<U>`. Using extends isn't incorrect but it may be unnecessary. If it is really unnecessary, this would imply that we need to change many method signatures in Javaslang, where we also don't need the `extends` part...

danieldietrich:
Let's keep `? extends U`.

Considering U = Animal, the signature would match best an Iterable of Lion, Elefant, Snake, ...

danieldietrich:
(aargh - my RedBlackTree.union method has an error. need to deep dive into the rabbit hole now...)

ruslansennov:
I'm out of my computer now, I'll rename it tomorrow

danieldietrich:
Thx! Have a nice day!

danieldietrich:
Oooh - looked at the Scala API - didn't know there's the same method. Let's keep the name `corresponds` then. We only need to adjust the bounds then:

```java
default <U> boolean corresponds(java.lang.Iterable<? extends U> that, Predicate<Tuple2<? super T, ? super U>> predicate)
```

danieldietrich:
(you see, I'm very 'flappy' when it comes to clear decisions - that's me :-) )

src/main/java/javaslang/collection/Seq.java
```diff
@@ -198,6 +198,25 @@ default boolean containsSlice(java.lang.Iterable<? extends T> that) {
...
 
     /**
+     * Tests whether every element of this sequence relates to the corresponding element of another sequence by satisfying a test predicate.
+     *
+     * @param that the other sequence
+     * @param predicate the test predicate, which relates elements from both sequences
+     * @return {@code true} if both sequences have the same length and {@code predicate(x, y)}
+     * is {@code true} for all corresponding elements {@code x} of this sequence and {@code y} of {@code that}, otherwise {@code false}.
+     */
+    default <U> boolean corresponds(java.lang.Iterable<U> that, Predicate<Tuple2<T, U>> predicate) {
```
danieldietrich: As already mentioned I would change the bounds:

```java
default <U> boolean corresponds(java.lang.Iterable<? extends U> that, Predicate<Tuple2<? super T, ? super U>> predicate)
```

Also I would move corresponds to `javaslang.Iterable`.

Use case: The following RedBlackTrees (and therefore SortedSets) are equal/corresponding:

```java
      -1       -1
      / \      / \
    -2   1   -2   0
        /          \
       0            1
```

because both Iterators return elements in the following order (tree in-order): `-2, -1, 0, 1`.

danieldietrich: Hey Ruslan,

I urgently need corresponds to fix my local branch. Therefore I will merge the PR now!! I hope this is ok for you.

(I removed Iterator.equals as I explained in #547 and not shouldConvertToMap tests fail. The corresponds will heal that.)

danieldietrich: changing the signature as suggested above (? extends, ? super) then the following test does not compile any more:

```java
// error: operator '-' not applicable to Object
assertThat(of(1, 2, 3).corresponds(of(3, 4, 5), t -> t._1 == t._2 - 2)).isTrue();
```

Therefore I will leave the signature as you committed it.

danieldietrich: We can use a `BiPredicate` instead of `Predicate<Tuple2>`. That will save intermediate object creations. I will adapt that.

ruslansennov:
Very simple commit but produced lot of comments :)

danieldietrich:
Yes :) I like to think of implications. All is connected somehow...

Have to fix some IteratorTests now before I can create a PR of the new stuff...

PR #559, opened 2015-09-06
performance table fixed #479

### LINKED ISSUE #479: Add a table of collection performance characteristics to collection/package-info.java
Like this: http://docs.scala-lang.org/overviews/collections/performance-characteristics.html

Could be a HTML table:

```java
/**
 * <table cellpadding="0" cellspacing="0" border="1">
 *     <thead><tr><th>&nbsp;</th><th>contains</th><th>...</th></tr></thead>
 *     <tbody>
 *         <tr><td>Array</td><td>O(1)</td><td>...</td></tr>
 *         <tr>...</tr>
 *     </tbody>
 * </table>
 */
```

ruslansennov:
Vector's cells changed:
tail() -&gt; linear
prepend() -&gt; linear
insert() -&gt; linear


danieldietrich:
mmhh, tail() must be possible in eff. const. time... what if we do tail() like so?

```java
final Iterator<T> iterator = iterator();
if (!iterator.hasNext()) {
    throw new UnsupportedOperationException("tail of empty Vector");
}
final T key = iterator.next();
return new Vector<>(tree.remove(key));
```

(same for HashSet et. al)

ruslansennov:
No, we should decrease all indexes in tree

danieldietrich:
I ask myself how Scala does the trick of eff. const. time? As far as I known they also use HAMT, at least for HashSet. http://docs.scala-lang.org/overviews/collections/performance-characteristics.html
Also the [LAMP paper](http://lampwww.epfl.ch/papers/idealhashtrees.pdf) say that HAMT operations are O(1), e.g. remove(key).
Do you have an idea?

ruslansennov:
wow! I haven't any idea, it should be investigated

danieldietrich:
ok - I will create an issue for that. thx!

PR #562, opened 2015-09-06
Implemented some HashMap operations. Added TODOs

danieldietrich:
@ruslansennov Could you review the changes, please? E.g. the dropXxx methods. Is it how you would do it?

If @patrox is ok with it, you may implement more methods...

PR #565, opened 2015-09-06
HashMap values cache

ruslansennov:
minor changes to #562


danieldietrich:
I'm not sure about this. I prefer O(n) time and O(1) memory over O(1) time and O(n) memory consumption.

Additionally we win nothing when using a List because it has also O(n) time on contains().

Let's keep iterator() here, please.

ruslansennov:
Ok, just close this

danieldietrich:
roger that

PR #574, opened 2015-09-07
Performance improvements of HashMap.put

danieldietrich:


src/main/java/javaslang/collection/HashArrayMappedTrie.java
```diff
@@ -273,12 +281,10 @@ public int size() {
 
         private final int bitmap;
         private final List<AbstractNode<K, V>> subNodes;
-        private final int size;
 
         private IndexedNode(int bitmap, List<AbstractNode<K, V>> subNodes) {
             this.bitmap = bitmap;
             this.subNodes = subNodes;
-            this.size = subNodes.map(HashArrayMappedTrie::size).sum().intValue();
```
ruslansennov: bravo!

ruslansennov:
```
Java HashMap bench took 0.121 sec.
Java TreeSet bench took 0.24 sec.
Javaslang HashMap bench took 2.811 sec.
Javaslang TreeSet bench took 0.829 sec.
```
10x faster on my computer (was about 20 sec)

danieldietrich:
After having slept about it I'm sure we can boost internal operations from O(n) to O(log n) or even eff. O(1). I will check that.

Issue #578, opened 2015-09-08
Remove optimizations that rely on HashArrayMappedTrie.size()


danieldietrich:
We use size() to determine if the newly created collections has the same size as the current collection. To save mem, we return the current instance if sizes are equal.

Because HAMT.size() is now O(n) (at least once, if lazy evaluated), then cannot use this optimization any more.

Candidates to check are HashSet, HashMap and Vector. Maybe also other collections...

danieldietrich:
We will also move back from storing the size/length in a Stream. The use-case of a stream is lazy evaluation, like Iterator. Both have indefinite size. Therefore it is not common to ask a Stream for its size... If it is really, really needed, the user may memoize the size him/herself.

ruslansennov:
I don't understand this, sorry, can you explain? <code>HashArrayMappedTrie</code>'s length() can be O(1), like <code>List</code>, <code>Vector</code> and any other not lazy collection

danieldietrich:
It can be - at the cost of computing the size at construction time, e.g. via `subNodes.map(AbstractNode::size).sum().intValue()`. But my performance measures showed, that this costs too much time.

An alternative is, that we add a lazy field `Lazy<Integer> size = ...`. But when we want to calculate the size of a big collection, we first have to calculcate (once) all these lazy values in O(n). Ok, we do it only once, but at the cost of extra garbage/memory.

Having *not* a Lazy length is not really a disadvantage. If we need the size in an app more than once, we can introduce a local var instead of having an instance variable, e.g.

```java
final int length = myMap.length(); // calculated once
```

Let's have a look at other languages.

danieldietrich:
Or did you mean Stream? I also look at the memory footprint. And at the use-case for a specific collection type. Stream's strength is not size(). It's size is indefinite and maybe infinite. Vector is the right collection for that.

danieldietrich:
This is copied from Scala's source code (just for teaching purpose, we will not include that in Javaslang!):

```scala
override def length: Int = {
    var len = 0
    var left = this
    while (!left.isEmpty) {
      len += 1
      left = left.tail
    }
    len
  }
```

Source: https://github.com/scala/scala/blob/v2.11.5/src/library/scala/collection/immutable/Stream.scala#L307-L315

ruslansennov:
> It can be - at the cost of computing the size at construction time, e.g. via subNodes.map(AbstractNode::size).sum().intValue(). But my performance measures showed, that this costs too much time.

[...]

PR #581, opened 2015-09-08
HashMap test

ruslansennov:
Not all TODOs are removed, to be continued


danieldietrich:
Huge, thx!

Please prepare yourself mentally that with #578 the `n >= length()` et al. checks will disappear because length() is now O(n). This is the right direction. I'm sure, other libs act the same to keep the performance of the _core_ operations at the limit of the possible.

PR #582, opened 2015-09-08
HAMT.size() improved #578

ruslansennov:
something like this


danieldietrich:
Thank you, Ruslan, LGTM!

ruslansennov:
what happened with travis?

danieldietrich:
Mmhh, sometimes the cloud has no more capacity. I think it will start in a few minutes...

danieldietrich:
Seems to be stuck. I will merge now. Then a new build is started.

danieldietrich:
Travis is still not operational. Does not matter, if local builds are working via `mvn clean test`.

Issue #568, opened 2015-09-07
Implement Vector as bit-mapped vector trie


danieldietrich:
...to obtain eff. const. time complexity. Maybe we will not target this for 2.0.0 but taking only a look at it...

More on that here:

* [Scala's immutable Vector](http://docs.scala-lang.org/overviews/collections/concrete-immutable-collection-classes.html#vectors)
* [Implementation in Rust](https://bitbucket.org/astrieanna/bitmapped-vector-trie)
* [How does Scala's Vector work?](http://stackoverflow.com/questions/20612729/how-does-scalas-vector-work)
* [Extreme Cleverness: Functional Data Structures in Scala [video]](http://www.infoq.com/presentations/Functional-Data-Structures-in-Scala)
* [Strangeloop 2011 'Functional Data' Slides by Daniel Spiewak](https://github.com/strangeloop/2011-slides/blob/master/Spiewak-FunctionalData.pdf)
* [Understanding Clojure's Persistent Vectors, Part 1 ](https://news.ycombinator.com/item?id=6445628)

Understanding Clojure's Persistent Vectors

* [pt. 1](http://hypirion.com/musings/understanding-persistent-vector-pt-1)
* [pt. 2](http://hypirion.com/musings/understanding-persistent-vector-pt-2)
* [pt. 3](http://hypirion.com/musings/understanding-persistent-vector-pt-3)

danieldietrich:
As examined in #571 we need to replace occurrences of

* `List.insert(int, Object)`
* `Array.update(int, Object)`

within HashArrayMappedTrie (HAMT).

We accomplish this by using Vector as default Seq implementation throughout our HAMT.

paplorinc:
FYI: https://github.com/andrewoma/dexx/blob/master/collection/src/main/java/com/github/andrewoma/dexx/collection/Vector.java

danieldietrich:
Thx!

paplorinc:
@danieldietrich, please assign this to me.

cc? @mvh77

paplorinc:
@danieldietrich I'm currently testing the correctness of the `slice` method for the new `Vector`, but since it has very many states, I would only be able to sleep at night by being able to validate the majority of the corner cases.
Obviously that's not possible - it would probably take months to run the test that tries out every combination.

```java
@Test
public void shouldSlice() {
    for (byte depth = 1; depth <= 4; depth++) {
[...]

Issue #560, opened 2015-09-06
Boost our HAMT time complexity for some operations from O(n) to effectively O(1)


danieldietrich:
As stated in #559 our following HAMT operations are currently linear instead of effectively constant:

* tail()
* prepend()
* insert()

Scala also does use the HAMT, at least for HashSet, with eff. const. time for all operations. See http://docs.scala-lang.org/overviews/collections/performance-characteristics.html. Also the [LAMP paper](http://lampwww.epfl.ch/papers/idealhashtrees.pdf) says that HAMT operations are O(1), e.g. remove(key).

We need to investigate further if and how to get eff. const. time for the operations mentioned above.

ruslansennov:
@danieldietrich no! All HAMT operations costs O(1). But some Vector operations cost O(n)

danieldietrich:
@ruslansennov Ok, that's nice!

But then this whole operation should be O(1) and no further index update is needed, right?

```java
Vector<T> tail() {
    final Iterator<T> iterator = iterator(); // O(1)
    if (!iterator.hasNext()) { // O(1)
        throw new UnsupportedOperationException("tail of empty Vector");
    }
    final T key = iterator.next(); // O(1)
    return new Vector<>(tree.remove(key)); // O(1) and tree indices are aligned new, right?
}
```

danieldietrich:
And yes, now I understand why prepend() and insert() are O(n). Because indices need an update. Yes, yes, yes...

And also yes: when removing an element, subsequent indices have to be updated.

This issue should be obsolete. But still the question how Scala does the trick :-)

ruslansennov:
> return new Vector<>(tree.remove(key)); // O(1) and tree indices are aligned new, right?

No :)
Let's imagine Vector(a, b, c). It's corresponds to Trie(0 -&gt; a, 1 -&gt; b, 2 -&gt; c)
After removing first element we will have  Trie(1 -&gt; b, 2 -&gt; c)
But it is not correct trie object, because first element has index=1 but not 1

ruslansennov:
> But still the question how Scala does the trick :-)

Yes!

danieldietrich:
This does not to be changed/fixed before release 2.0.0. We can go live and boost the performance in a follow-up release...

ruslansennov:
I believe all is OK for HAMT and this issue can be closed

danieldietrich:
:+1: Done.

danieldietrich:
[...]

PR #585, opened 2015-09-09
Vector performance improved #560 #568

ruslansennov:
First of all,  table of Scala performance characteristics says nothing about <code>Vector.insert()</code>. Maybe it is not exists? :)

Now about <code>Vector.prepend()</code> and <code>Vector.tail()</code>
Let's define additional parameter <code>indexShift</code>. It shows difference between index in vector and corresponding index in HAMT. With it both operations can be implemented in O(1) (because all HAMT operations is O(1), see code).

Please review it carefully. This trick has come out of my head directly and it need to be verified.

### DIFF
--- a/src/main/java/javaslang/collection/Vector.java
+++ b/src/main/java/javaslang/collection/Vector.java
@@ -675,18 +675,17 @@ public final class Vector<T> implements IndexedSeq<T>, Serializable {
     @Override
     public Iterator<T> iterator() {
         return new Iterator<T>() {
-            private int index = 0;
+            private int index = indexShift;
+            private final int size = trie.size() + indexShift;
 
             @Override
             public boolean hasNext() {
-                return index < trie.size();
+                return index < size;
             }
 
             @Override
             public T next() {
-                final int trieIndex = index + indexShift;
-                index++;
-                return trie.get(trieIndex).get();
+                return trie.get(index++).get();
             }
         };
     }
@@ -782,8 +781,8 @@ public final class Vector<T> implements IndexedSeq<T>, Serializable {
         List<T> list = List.ofAll(elements);
         final int newIndexShift = indexShift - list.length();
         HashArrayMappedTrie<Integer, T> newTrie = trie;
-        for (int i = 0; !list.isEmpty(); i++) {
-            newTrie = newTrie.put(newIndexShift + i, list.head());
+        for (int i = newIndexShift; !list.isEmpty(); i++) {
+            newTrie = newTrie.put(i, list.head());
             list = list.tail();
         }
         return new Vector<>(newIndexShift, newTrie);
--- a/src/main/java/javaslang/collection/Vector.java
+++ b/src/main/java/javaslang/collection/Vector.java
@@ -29,10 +29,16 @@ public final class Vector<T> implements IndexedSeq<T>, Serializable {
     private static final Vector<?> EMPTY = new Vector<>(HashArrayMappedTrie.empty());
 
     private final HashArrayMappedTrie<Integer, T> trie;
+    private final int indexShift;
     private final transient Lazy<Integer> hashCode = Lazy.of(() -> Traversable.hash(this));
 
     private Vector(HashArrayMappedTrie<Integer, T> trie) {
+        this(0, trie);
+    }
+
+    private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
         this.trie = trie;
+        this.indexShift = indexShift;
     }
 
     /**
@@ -394,16 +400,16 @@ public final class Vector<T> implements IndexedSeq<T>, Serializable {
 
     @Override
     public Vector<T> append(T element) {
-        return new Vector<>(trie.put(trie.size(), element));
+        return new Vector<>(indexShift, trie.put(length() + indexShift, element));
     }
 
     @Override
     public Vector<T> appendAll(java.lang.Iterable<? extends T> elements) {
         HashArrayMappedTrie<Integer, T> result = trie;
         for (T element : elements) {
-            result = result.put(result.size(), element);
+            result = result.put(result.size() + indexShift, element);
         }
-        return new Vector<>(result);
+        return new Vector<>(indexShift, result);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


danieldietrich:
ok - I will deep-dive into it now......

src/main/java/javaslang/collection/Vector.java
```diff
@@ -678,7 +684,9 @@ public boolean hasNext() {
 
             @Override
             public T next() {
-                return trie.get(index++).get();
+                final int trieIndex = index + indexShift;
+                index++;
+                return trie.get(trieIndex).get();
```
danieldietrich: first impression here is, that we can do:

```java
return new Iterator<T>() {
    private int index = indexShift;
    private final size = trie.size() + indexShift;
    @Override public boolean hasNext() {
        return index < size;
    }
    @Override public T next() {
        return trie.get(index++).get();
    }
}
```

src/main/java/javaslang/collection/Vector.java
```diff
@@ -765,12 +773,20 @@ public int length() {
 
     @Override
     public Vector<T> prepend(T element) {
-        return insert(0, element);
+        final int newIndexShift = indexShift - 1;
```
danieldietrich: Can this be < 0? is the indexShift always >= 0? I have to read further through the code...

src/main/java/javaslang/collection/Vector.java
```diff
@@ -765,12 +773,20 @@ public int length() {
...
+        return new Vector<>(newIndexShift, trie.put(newIndexShift, element));
     }
 
     @Override
     public Vector<T> prependAll(java.lang.Iterable<? extends T> elements) {
-        return insertAll(0, elements);
+        List<T> list = List.ofAll(elements);
```
danieldietrich: use elements.iterator() (of type java.util.Iterator) instead of List.ofAll(). this will save cycles and memory

src/main/java/javaslang/collection/Vector.java
ruslansennov: can be

danieldietrich: ok

src/main/java/javaslang/collection/Vector.java
ruslansennov: you're right, looks better

src/main/java/javaslang/collection/Vector.java
```diff
@@ -1025,11 +1041,12 @@ public int length() {
...
             throw new UnsupportedOperationException("tail of empty vector");
         }
-        HashArrayMappedTrie<Integer, T> trie = HashArrayMappedTrie.empty();
-        for (int i = 1; i < length(); i++) {
-            trie = trie.put(i - 1, get(i));
+        if(length() == 1) {
+            return empty();
+        } else {
+            final int newIndexShift = indexShift + 1;
+            return new Vector<>(newIndexShift, trie.remove(indexShift));
```
danieldietrich: wow, I start to understand. your idea is really cunning

src/main/java/javaslang/collection/Vector.java
```diff
@@ -765,12 +773,20 @@ public int length() {
...
+        return new Vector<>(newIndexShift, trie.put(newIndexShift, element));
     }
 
     @Override
     public Vector<T> prependAll(java.lang.Iterable<? extends T> elements) {
-        return insertAll(0, elements);
+        List<T> list = List.ofAll(elements);
+        final int newIndexShift = indexShift - list.length();
+        HashArrayMappedTrie<Integer, T> newTrie = trie;
+        for (int i = 0; !list.isEmpty(); i++) {
```
danieldietrich: start with i = newIndexShift ...

src/main/java/javaslang/collection/Vector.java
```diff
@@ -765,12 +773,20 @@ public int length() {
...
+            newTrie = newTrie.put(newIndexShift + i, list.head());
```
danieldietrich: ... and put(i, list.head()) instead ...

src/main/java/javaslang/collection/Vector.java
ruslansennov: no, I need to know length of elements

danieldietrich: mmhh, we could convert the elements to an IndexedSeq. If it is already an indexed seq (checked via instanceof) it is just a cast in O(1), otherwise we could put the elements in an Object[] and wrap it in an Array.

Later in the for-loop we do a indexedSeq.get(i) instead of list.head().

Don't know if that is really better...

ruslansennov: > Don't know if that is really better...

I don't know too :) I believe it's unnecessary complication

danieldietrich: This logic can be encapsulated in a static factory method of IndexedSeq:

```java
interface IndexedSeq<T> extends Seq<T> {
    static<T> IndexedSeq<T> ofAll(java.lang.Iterable<? extends T> elements) {
        if (elements instanceof IndexedSeq) {
            return (IndexSeq<T>) elements;
        } else {
            int i = 0;
            final Object[] array = new Object[???]; // oohh - we need to know the size...
            for (T t : elelemts) {
                array[i++] = t;
            }
            return new Array<>(array); // constructor needs to be package private
        }
    }
}
```

danieldietrich: or just

```java
interface IndexedSeq<T> extends Seq<T> {
    static<T> IndexedSeq<T> ofAll(java.lang.Iterable<? extends T> elements) {
        if (elements instanceof IndexedSeq) {
            return (IndexSeq<T>) elements;
        } else {
            return Vector.ofAll(elements); // Vector is our default IndexSeq!!!
        }
    }
}
```

ruslansennov: <code>Vector.ofAll(elements)</code> and then <code>for(...) { get(i) }</code> is very expensive :(

danieldietrich: then we stay as is with List

danieldietrich:
The L_really_GTM! It is a great Idea to internally shift the index for front operations!

I will now merge it...

ruslansennov:
:metal:

PR #603, opened 2015-09-11
Iterator.zipAll() fixed; duplicated code removed

ruslansennov:


danieldietrich:
oh, wow! Iterator pays off!

PR #631, opened 2015-09-18
Improve Coverage for javaslang.control

jasoet-gdp:
- Add Coverage for Success class
- Add Coverage for Failure.NonFatal class
- Cannot access Fatal.equals(), Fatal.toString(), Fatal.hashCode()

### DIFF
--- a/src/test/java/javaslang/control/OptionTest.java
+++ b/src/test/java/javaslang/control/OptionTest.java
@@ -8,12 +8,7 @@ package javaslang.control;
 import javaslang.Serializables;
 import org.junit.Test;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.NoSuchElementException;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -133,15 +128,15 @@ public class OptionTest {
 
     @Test
     public void shouldConsumePresentValueOnIsPresentWhenValueIsPresent() {
-        final int[] actual = new int[]{-1};
+        final int[] actual = new int[] { -1 };
         Option.of(1).forEach(i -> actual[0] = i);
         assertThat(actual[0]).isEqualTo(1);
     }
 
     @Test
     public void shouldNotConsumeAnythingOnIsPresentWhenValueIsNotPresent() {
-        final int[] actual = new int[]{-1};
-        Option.<Integer>none().forEach(i -> actual[0] = i);
+        final int[] actual = new int[] { -1 };
+        Option.<Integer> none().forEach(i -> actual[0] = i);
         assertThat(actual[0]).isEqualTo(-1);
     }
 
@@ -159,7 +154,7 @@ public class OptionTest {
 
     @Test
     public void shouldReturnNoneOnFilterWhenValueIsNotPresentAndPredicateNotMatches() {
-        assertThat(Option.<Integer>none().filter(i -> i == 1)).isEqualTo(Option.none());
+        assertThat(Option.<Integer> none().filter(i -> i == 1)).isEqualTo(Option.none());
     }
 
     // -- flatten()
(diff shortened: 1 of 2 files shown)


jasoet-gdp:
@danieldietrich 

What the purpose of `Try.filterTry(CheckedPredicate)` ? 

```java
    default Try<T> filterTry(CheckedPredicate<? super T> predicate) {
        if (isFailure()) {
            return this;
        } else {
            return Try.of(() -> predicate.test(get())).flatMap(b -> filter(ignored -> b));
        }
    }
```

src/test/java/javaslang/control/OptionTest.java
```diff
@@ -131,15 +133,15 @@ public void shouldBePresentOnIsEmptyWhenValueIsPresent() {
 
     @Test
     public void shouldConsumePresentValueOnIsPresentWhenValueIsPresent() {
-        final int[] actual = new int[] { -1 };
+        final int[] actual = new int[]{-1};
```
danieldietrich: The first thing I see is that you probably use IntelliJ with the default settings because the format changed slightly. Don't mind - it is ok! I provided the Javaslang settings in `./ide/idea-settings.jar`. If you use them it is good, if not it does not really matter...

danieldietrich: (The benefit of using the same format is, that only _real_ changes are shown in the diff)

src/test/java/javaslang/control/OptionTest.java
```diff
@@ -203,7 +205,19 @@ public void shouldFlatMapSome() {
...
+    @Test
+    public void shouldFlatMapNonEmptyIterable() {
+        final java.lang.Iterable<Integer> iterable = Arrays.asList(2, 3, 4);
+        assertThat(Option.of(1).<Integer>flatMap(i -> iterable)).isEqualTo(Option.of(2));
+    }
+
+    @Test
+    public void shouldFlatMapEmptyIterable() {
+        final java.lang.Iterable<Integer> iterable = Collections.emptyList();
+        assertThat(Option.of(1).<Integer>flatMap(i -> iterable)).isEqualTo(Option.<Integer>none());
```
danieldietrich: great!

danieldietrich:
Many methods have currently zero usage because of missing unit-tests. We provide an API but don't use it ourself so far. It's ok!

The methods `map`, `flatMap` and `filter` are inherited from `Value` - the mother of most of Javaslang's types. (Side note: I decided to take the name 'value' because it reflects the core meaning of functional programming: thinking in (immutable) values and transformations of values.) Because our `Try` deals with exceptions by capturing failure state, it is great to have convenience methods if something goes wrong while (flat-)mapping or filtering results. The methods `mapTry`, `filterTry` and `flatMapTry` take checked functional interfaces, i.e. functional interfaces that throw checked exceptions.

Unfortunately Java does not allow us to change the parameter type of `map`, ... to checked functions, so the best we can come up with is adding new methods `mapTry`, ...

danieldietrich:
Here is an example of how to access Failure.Fatal:

```java
// import static ...Assertions.fail ...

try {
    Try.of(() -> {
        throw new UnknownError();
    });
    fail("Exception expected");
} catch(Failure.Fatal x) {
    assertThat(x.toString()).isEqualTo("Fatal(java.lang.UnknownError)");
    // TODO: equals & hashCode
}
```

danieldietrich:
Nice work! Coverage is one of the last gaps towards a release 2.0.0. I like to go into the code again and rethink API from another perspective - like we've done it with `mapTry`, ...

I can merge if you want - or wait until you commit the Fatal.toString equals hashCode tests. Just give me a signal! :-)

jasoet-gdp:
@danieldietrich now i use javaslang idea setting, now it nicer :D
I'll work on `Fatal.toString` and `Fatal.hashCode`, then you can merge it.

src/test/java/javaslang/control/OptionTest.java
```diff
@@ -8,10 +8,7 @@
 import javaslang.Serializables;
 import org.junit.Test;
 
-import java.util.Iterator;
-import java.util.NoSuchElementException;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
```
jasoet-gdp: I usually set idea to never use wildcard import such as `import java.util.*;` Now i use setting from javaslang standard.

jasoet-gdp:
@danieldietrich Ready to go!

danieldietrich:
Awesome!

PR #633, opened 2015-09-19
FutureTest fixes

DillonJettCallis:
Used Java's ComparableFuture to help test our Futures, with the idea being that ComparableFuture will always work. 

Added more tests to get 100% code coverage of Future and Promise. 

Found a case where a callback was slipping through the synchronization and not being called so I wrote a new hopefully fool-proof way of dealing with the problem.


danieldietrich:
Many thanks!!

Issue #642, opened 2015-09-25
Provide Map.Entry key() and value() methods to enable method references


danieldietrich:
This allows us to write

```java
// BEFORE
entrySet().map(entry -> entry.key)

// AFTER
entrySet().map(Entry::key)
```

Issue #570, opened 2015-09-07
appendAll vs addAll vs union vs concat


danieldietrich:
We already have `Seq.appendAll(Iterable)`, `

For some data structures it is possible to provide a fast concat method that concats a collection of the same type, i.e. `Iterator.concat(Iterator)` using `ConcatIterator`.

**(Updated)**

* ~~we remove `appendAll(Iterable)` from the `Seq` interface~~
* ~~`Set` should have `union(Iterable)` which merges a Set/TreeSet fast~~ union is only for Sets, not for Iterable
* ~~`Seq` should have `concat(Seq)` which concats an `Iterable.iterator()` fast~~
* `Iterator` should have `concat(Iterable)` which concats an `Iterable.iterator()` fast
* ~~`Map` should have `putAll(Iterable<Entry<? extends K, ? extends V>>)` which subsequently inserts all entries~~ `merge()` will do the job
* `Tree` should have `insert(Tree)` and `insert(int, Tree)` (?)

danieldietrich:
> Set should have union(Iterable) which merges a Set/TreeSet fast

Yes, but - `union(Set<T> that)` and `addAll(java.lang.Iterable<T> elements)`

danieldietrich:
> Tree should have insert(Tree) and insert(int, Tree) (?)

Will move this to #113

Issue #428, opened 2015-08-06
Add an immutable TreeMap to the javaslang.collection package


danieldietrich:
```java
class TreeMap<K, V> implements SortedMap<K, V>, Iterable<Map.Entry<K, V>> {

    final RedBlackTree<Entry<K, V>> entries;

    // or Comparable<? super K>? See RedBlackTree
    static <K extends Comparable<K>, V> TreeMap<K, V> empty() { ... }

    static <K, V> TreeMap<K, V> empty(Comparator<? super K> comparator) { ... }

    Iterator<Entry<K, V>> iterator() {
        return entries.iterator();
    }
}
```

danieldietrich:
We need to take care / test different scenarios regarding serializability of SortableSets/SortableMaps.

Especially the complete chain of objects has to be serializable, e.g. `TreeMap` has a `RedBlackTree` has a `Comparator`.

The Comparator may be

* a method reference, e.g. `K::compareTo` for a generic type `K extends Comparator<? super K>`
* or a lambda expression, e.g. `(k1, k2) -> k1.compareTo(k2)`

We cast lambda expressions to `(Comparator<K> & Serializable)`. Is this also possible with method references? If not, the complete chain (see above) is not serializable. Solution: do not use method references, use lambda expressions instead.

**Example of method reference usage:**

```java
public static <K extends Comparable<? super K>, V> TreeMap<K, V> empty() {
    return new TreeMap<>(RedBlackTree.empty(entryComparator(K::compareTo)));
}
```

danieldietrich:
```java
public interface RedBlackTree<T> extends java.lang.Iterable<T> {

    // variant 1
    static <T extends Comparable<? super T>> Empty<T> empty() {
        final Comparator<? super T> comparator = (Comparator<? super T> & Serializable) (t1, t2) -> t1.compareTo(t2);
        return new Empty<>(comparator);
    }

    // variant 2 - I think this will result in a ClassCastException
    static <T extends Comparable<? super T>> Empty<T> empty() {
        final Comparator<? super T> comparator = (Comparator<? super T> & Serializable) T::compareTo;
        return new Empty<>(comparator);
    }

}
```

danieldietrich:
Get TreeSet.length() and TreeMap.length() right. Do Map and Set interfaces need size() method?

danieldietrich:
As mentioned above, the following Comparator method reference is not serializable:

```java
import java.io.Serializable;
import java.util.Comparator;

[...]

PR #643, opened 2015-09-25
Towards TreeMap. General Map house keeping

danieldietrich:


danieldietrich:
Fixes #642 
Towards #570, #428

PR #649, opened 2015-09-27
Seq.range*() tests added; duplicate code removed

ruslansennov:


src/main/java/javaslang/collection/TreeSet.java
```diff
@@ -92,13 +92,14 @@
 
     public static <T extends Comparable<? super T>> TreeSet<T> ofAll(java.lang.Iterable<? extends T> values) {
         Objects.requireNonNull(values, "values is null");
-        return new TreeSet<>(RedBlackTree.ofAll(values));
+        return values.iterator().hasNext() ? new TreeSet<>(RedBlackTree.ofAll(values)) : empty();
```
danieldietrich: There is one caveat: javaslang.collection.Iterator is Iterable. With Iterator it does not work. Let's revert this line.

Does this happen elsewhere, too? I will double-check the code...

src/main/java/javaslang/collection/TreeSet.java
```diff
@@ -92,13 +92,14 @@
...
     }
 
+    @SuppressWarnings("unchecked")
     public static <T> TreeSet<T> ofAll(Comparator<? super T> comparator, java.lang.Iterable<? extends T> values) {
         Objects.requireNonNull(comparator, "comparator is null");
         Objects.requireNonNull(values, "values is null");
-        return new TreeSet<>(RedBlackTree.ofAll(comparator, values));
+        return values.iterator().hasNext() ? new TreeSet<>(RedBlackTree.ofAll(comparator, values)) : (TreeSet<T>) empty();
```
danieldietrich: Dito

src/main/java/javaslang/collection/TreeSet.java
ruslansennov: mmm... I don't understand you :( We have <code>java.lang.Iterable</code> here as parameter.

danieldietrich:
```java
/*
java.lang.Iterable
       ^
       |
javaslang.Iterable
       ^
       |
javaslang.Value
       ^
       |
javaslang.collection.TraversableOnce
       ^
       |
javaslang.collection.Iterator
*/

TreeSet.ofAll(Iterator.of(1, 2, 3)); // BOOM :)
```

ruslansennov:
```java
    @Test
    public void vvv() {
        TreeSet<Integer> t = TreeSet.ofAll(Iterator.of(1, 2, 3));
        System.out.println(t);  //  TreeSet(B:2 B:1 B:3)
    }
```
It seems all is OK...

danieldietrich: You are right. hasNext() has also to be called by next() as of the API contract in Java docs. That means that it can be called twice or more without changing the public visible state. Thx for clarification.

PR #653, opened 2015-09-28
groupBy tests added

ruslansennov:


danieldietrich:
Thanks, Ruslan! I will also ~~try~~ do my best to bring 2.0.0 'through the door' now... was busy with doing other thinks like reviewing chapters of a book, which is time-consuming. I target the 13.10.15 to make the jars available on maven central and publish the new (yet not existing) landing page at javaslang.com.

ruslansennov:
I believe 13.10.2015 may be very optimistic :) We should implement TreeMap and tons of tests. And I will busy this (and maybe next) week too...

danieldietrich:
Ok, we will take our time. Then it will be after the 13th. Thx!!

PR #685, opened 2015-10-08
RBTree 'class cast exception' fixed; tests added

ruslansennov:


danieldietrich:
Oh, just saw it :) Thx, Ruslan!

I target a release candidate this weekend - we will see...

PR #694, opened 2015-10-10
tests again

ruslansennov:


danieldietrich:
Thx! I will sync take a look at Future and Promise today and will then also put more focus on tests. Tuesday I target a first Release Candidate...

PR #711, opened 2015-10-15
Tests

ruslansennov:


danieldietrich:
Hi, I was two days offline (in Munich)...

It is good that we got rid of `result()` because it was used only in Tree. However, the new IterableAssert and ObjectAssert introduce a little background \*magic\*. But it works well and is ok.

Thx!

PR #714, opened 2015-10-19
Map.flatten() tests fixed

ruslansennov:


src/main/java/javaslang/control/Option.java
```diff
@@ -72,6 +72,11 @@ default boolean isDefined() {
 
     T get();
 
+    @Override
+    default Option<T> getOption() {
+        return this;
+    }
+
```
danieldietrich: good!!

src/test/java/javaslang/collection/IntMap.java
```diff
@@ -121,9 +121,7 @@ private Object readResolve() {
 
     @Override
     public IntMap<Object> flatten() {
-        final Seq<Object> seq = original.values().flatten();
-        final Map<Integer, Object> map = seq.zipWithIndex().toMap(t -> Tuple.of(t._2, t._1));
-        return IntMap.of(map);
+        throw new UnsupportedOperationException();
```
danieldietrich: really? don't we need it?

ruslansennov: Yep. In flatten() tests I use emptyMap() constructor which provides Map directly, instead of empty() which provides IntMap wrapper

danieldietrich: ok!

Issue #692, opened 2015-10-10
A Tuple is a Value


danieldietrich:
Functional programming is all about _values_ and transformation of values using _functions_.

Compared to other types Tuples have still a special role in Javaslang. To reflect my statement above in a consistent way, Tuple _has to be_ a [Value](https://github.com/javaslang/javaslang/blob/master/src/main/java/javaslang/Value.java) in Javaslang.

This is the current type hierarchy:

![javaslang-2 0 0](https://cloud.githubusercontent.com/assets/743833/10410566/0f472f6a-6f46-11e5-906d-5f480ea702c0.png)

Up to now I emphasized the viewpoint that _Tuples are Tuples_, not more and not less. Especially no other types should be derived from tuples, i.e. they should not be extended.

But what is the value of a Tuple? In my opinion, the value of a Tuple is the Tuple itself. This is different to all other `Value` types of Javaslang, which are always _wrappers_ of values in some way (e.g. Option, List, ...).

**Now the big question: Is there any benefit to see a Tuple as a Value of itself, i.e. a Value of a Tuple?**

One benefit is, that Tuple would inherit [all operations on Values](https://github.com/javaslang/javaslang/blob/master/src/main/java/javaslang/Value.java).

Is a Tuple not lightweight enough any more if it inherits from Value?

```java

Value<Tuple1<Integer>> val = Tuple.of(1);

class Tuple1<T> implements Tuple, Value<T> {
    ...
}
```

ggalmazor:
When I am working with Tuples I often want to handle their contents the same way I would with a List: map a Tuple3 into a Tuple2, fold all objects into one...

I see Tuples as a sequence of objects that are heterogeneously typed. According to the current type hierarchy, imo they should extend Value.

I guess that they couldn't extend Traversable because a Tuple doesn't store objects with the same type, but in an imaginary Java where this is possible, I'd see Tuple at the same level as Seq.

danieldietrich:
Thx, it is good to hear your view on that, I felt the same.

danieldietrich:
There will be some changes...

1. Value.filter will return Option instead of throwing NoSuchElementException for some types, e.g. for Lazy, Either, ... and of cause Tuple. In general we should not rely on throwing exceptions. Instead we should provide an algebraic value in order to _not break the control flow_!
[...]

PR #717, opened 2015-10-25
Tuple to seq

ruslansennov:
I believe it will be useful (and it already useful in jackson module)

### DIFF
--- a/generator/Generator.scala
+++ b/generator/Generator.scala
@@ -772,11 +772,6 @@ def generateMainClasses(): Unit = {
             }
 
             @Override
-            public $Seq<?> toSeq() {
-                return $List.empty()${(i to (1, -1)).gen(j => xs".prepend(_$j)")};
-            }
-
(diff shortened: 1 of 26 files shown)


danieldietrich:
Hi Ruslan,

we should find a general way to do it. I re-opened #692 and will work on it.

Please keep this PR open until we see in which direction #692 evolves.

Greets

Daniel

danieldietrich:
I fiddled around for hours just to learn (again) that there is no good way to let extend Tuple the Iterable interface. Also inheriting other methods is overkill.

I committed an impl of the Tuple.toSeq method (which I regret now because of the merge conflicts). Could you please commit your tests for that method? I think the simplest way is to copy your tuples test of the Generator.scala to a fresh copy of master...

Thanks for your patience with me!

ruslansennov:
done

danieldietrich:
thx!!

Issue #449, opened 2015-08-09
Further improvement of Future and Promise


danieldietrich:
Thanks to @DillonJettCallis we have a first version of Future and Promise.

Lets take a look at the Scala classes, add missing features and round the corners. E.g. Future extends Awaitable

* http://docs.scala-lang.org/overviews/core/futures.html
* http://docs.scala-lang.org/sips/completed/futures-promises.html
* http://danielwestheide.com/blog/2013/01/16/the-neophytes-guide-to-scala-part-9-promises-and-futures-in-practice.html
* http://stackoverflow.com/questions/13381134/what-are-the-use-cases-of-scala-concurrent-promise
* Future trait: http://www.scala-lang.org/api/current/index.html#scala.concurrent.Future
* Future impl: https://github.com/scala/scala/blob/v2.11.7/src/library/scala/concurrent/impl/Future.scala
* Promise trait: http://www.scala-lang.org/api/current/index.html#scala.concurrent.Promise
* Promise impl: https://github.com/scala/scala/blob/v2.11.7/src/library/scala/concurrent/impl/Promise.scala

Additionally we need to take a look, how Future and Promise integrate with existing Javaslang interfaces. Candidates to be extended by Future are

* Value
* TraversableOnce
* FilterMonadic

Future can be understood as future Value. Promise can be understood as writable-once wrapper(?) around a future.

danieldietrich:
Future:

* add Awaitable super interface
* There should be only one constructor. Maybe we can remove the dependency to Java's CompletableFuture completely and provide a static factory method which wraps one in a Future.
* multiple callbacks vs. mutable state
* remove complete() method. This is a member of Promise only
* it should implement FilterMonadic and additionally implement mapTry, flatMapTry, like Try does. Bug: mapTry and flatMapTry need to take a Try.CheckedFunction
* the recover* et al. methods should be aligned to Try.*. Maybe add an interface `Recoverable` extended by Future and Try?

danieldietrich:
We should also take a Scala-like `Awaitable` interface into account.

danieldietrich:
We should provide cancellation of futures (which is currently not part of Scala's Future but supported by Java's Future).

**Semantics:**

* cancellation is guarded by the same lock as value (= result), actions (= callbacks) and job (= computation) but has no separate instance variable
* if Future is constructed but does not run() (e.g. when used within Promise) then cancellation does nothing
* if Future.isComplete() then cancellation does nothing
[...]

PR #730, opened 2015-10-30
Updated Futute javadoc.

danieldietrich:


danieldietrich:
Towards #449 

The impl is so clean and nice now - I'm really satisfied.

The monadic operations are just decoration now, I already have them in the pipeline...

PR #753, opened 2015-11-03
Tuple.ofAll() fabric

ruslansennov:


danieldietrich:
Hi Ruslan,

what is the use-case for this method?

If it is needed in `javaslang-jackson` I would consider to put it there as _internal_ helper method. The main reason for that is, that it is currently unsafe. The generic return type `<T extends Tuple>` allows us to do incorrect things:

```java
Tuple0 tuple = Tuple.ofAll(List.of(1, 2, 3));
```

Where do you use the method?

Greets,

Daniel

ruslansennov:
Yes <code>javaslang-jackson</code> already has such helper, but I did not take into account the unsafe behavior when it used in global scope. Seems it cannot be both safe <strong>and</strong> useful.
OK, you're right, this method is unnecessary.

danieldietrich:
Yes, let's remove it. I currently do the same with `Value.flatten()`.

:+1:

Issue #764, opened 2015-11-05
Move from coveralls.io to codecov.io


danieldietrich:
We had too many infrastructure problems in the past with coveralls.io. Again the source is not linked. Now we move on...

danieldietrich:
done

PR #765, opened 2015-11-05
Moving from coveralls.io to codecov.io. Towards #764

danieldietrich:


ruslansennov:
94% -> 88% (!) :smile:

danieldietrich:
:-D oh noes!

ruslansennov:
Perfect is the enemy of good

danieldietrich:
yes - we don't need to reach 100%. the tool is better...

PR #773, opened 2015-11-06
min & max by(Function) improved

ruslansennov:
This improvement is not obvious, but in case <code>Euler14Test</code> running time on my computer decreases by 20%. In case of delegating these methods to the comparator, function was called 2*n-1 times, where n is the length of sequence. It is bad if function is not primitive (even in case memoized (!) like in <code>Euler14Test</code>)

### DIFF
--- a/src/main/java/javaslang/collection/TraversableOnce.java
+++ b/src/main/java/javaslang/collection/TraversableOnce.java
@@ -592,10 +592,16 @@ public interface TraversableOnce<T> extends Value<T> {
         if (isEmpty()) {
             return None.instance();
         } else {
-            return new Some<>(foldLeft((Tuple2<U, T>) null, (u, t) -> {
-                final U tu = f.apply(t);
-                return u != null && u._1.compareTo(tu) >= 0 ? u : Tuple.of(tu, t);
-            })._2);
+            T tm = null;
+            U um = null;
+            for (T t : Iterator.ofAll(this)) {
+                U u = f.apply(t);
+                if (um == null || u.compareTo(um) >= 0) {
+                    um = u;
+                    tm = t;
+                }
+            }
+            return new Some<>(tm);
         }
     }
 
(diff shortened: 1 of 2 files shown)


ruslansennov:
hmm... +20% on travis :smile:

ruslansennov:
In new implementation Tuple creates n times, it is also bad...

ruslansennov:
OK, travis shows different results, and can't be arbiter. Anyway I believe <code>shouldCallMinFunctionOncePerElement()</code> is not bad idea :smile:

src/main/java/javaslang/collection/TraversableOnce.java
```diff
@@ -591,7 +592,16 @@ default T last() {
...
+            T tm = null;
+            U um = null;
+            for (T t : Iterator.ofAll(this)) {
+                U u = f.apply(t);
+                if (um == null || u.compareTo(um) >= 0) {
+                    um = u;
+                    tm = t;
+                }
+            }
+            return new Some<>(tm);
```
danieldietrich: Then let's also save the null-check. And to be conform with the previous impl and maxBy(Comparator), we need to use `>` instead of `>=` (see internals of reduce and foldLeft... the zero is the previous value - but here we use the current value)

```java
final Iterator<T> iter = Iterator.ofAll(this);
T tm = iter.next();
U um = f.apply(tm);
while (iter.hasNext()) {
    final T t = iter.next();
    final U u = f.apply(t);
    if (u.compareTo(um) > 0) {
        um = u;
        tm = t;
    }
}
return new Some<>(tm);
```

src/main/java/javaslang/collection/TraversableOnce.java
```diff
@@ -639,7 +649,16 @@ default T last() {
...
+            T tm = null;
+            U um = null;
+            for (T t : Iterator.ofAll(this)) {
+                U u = f.apply(t);
+                if (um == null || u.compareTo(um) <= 0) {
+                    um = u;
+                    tm = t;
+                }
+            }
+            return new Some<>(tm);
```
danieldietrich: Same here...

```java
final Iterator<T> iter = Iterator.ofAll(this);
T tm = iter.next();
U um = f.apply(tm);
while (iter.hasNext()) {
    final T t = iter.next();
    final U u = f.apply(t);
    if (u.compareTo(um) < 0) {
        um = u;
        tm = t;
    }
}
return new Some<>(tm);
```

danieldietrich:
I like this addition. Previously there were too many function applications - as you say
Thx!

danieldietrich:
Please note that I made a slight change - now using `iterator()` instead of `Iterator.ofAll(this)`

src/main/java/javaslang/collection/TraversableOnce.java
ruslansennov: :+1:

PR #775, opened 2015-11-06
Abstracted over folds

### LINKED ISSUE #735: Adding folds that use Monoid
Hello folks.
I think it would be useful to have folds in `Traversable<T>` that accept a `Monoid<T>`

A simple use case would be:
```java
public static Monoid<String> concat = new Monoid<String>() {
	public String combine(String a1, String a2) { return a1 + a2; }
	public String zero() { return ""; }
};
...
Stream.of("1", "2", "3").foldLeft(concat);
```
which would produce `"123"`.

danieldietrich:
Fixes #735

### DIFF
--- a/src/main/java/javaslang/Value.java
+++ b/src/main/java/javaslang/Value.java
@@ -5,7 +5,9 @@
  */
 package javaslang;
 
+import javaslang.algebra.Foldable;
 import javaslang.algebra.Monad;
+import javaslang.algebra.Monoid;
 import javaslang.collection.*;
 import javaslang.control.*;
 
@@ -14,10 +16,7 @@ import java.io.PrintWriter;
 import java.util.ArrayList;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.function.Consumer;
-import java.util.function.Function;
-import java.util.function.Predicate;
-import java.util.function.Supplier;
+import java.util.function.*;
 import java.util.stream.StreamSupport;
 
 /**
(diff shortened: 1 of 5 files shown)


jbgi:
foldRight ```combine``` should be lazy in it second arguments otherwise you cannot implements short-circuiting logic with it.
=>
```foldRight(U zero, BiFunction<? super T, ? extends Supplier<? super U>, ? extends U> combine)```
(with the usual ugliness due to supporting subtyping ;-)

**Comment by danieldietrich** [comment]
Mmh, I haven't seen this signature yet, e.g. in Scala [TraversableOnce](http://www.scala-lang.org/api/current/index.html#scala.collection.TraversableOnce). Haskell is lazy by default, I think therefore it will just work out-of-the-box there.

I've found this example for using shortcut-circuiting logic, implementing `take` with `foldr` in Haskell:

```haskell
take :: Int -> [a] -> [a]
take n xs = foldr step [] (tagFrom 1 xs)
    where step (a, i) rest
               | i > n     = []
               | otherwise = a:rest

tagFrom :: Enum i => i -> [a] -> [(a, i)]
tagFrom i xs = zip xs [i..]
```

Source: http://stackoverflow.com/questions/15879940/implementing-take-using-foldr

jbgi:
there is Foldable#foldRight from scalaz:
https://github.com/scalaz/scalaz/blob/87058f198d0690632a9eaee8c531c1a64d4841af/core/src/main/scala/scalaz/Foldable.scala#L23

PR #780, opened 2015-11-07
Renamed of(T...) to ofAll(T...)

### LINKED ISSUE #721: Collection creation behavior changed from 1.2.2 to 2.0.0
@ggalmazor wrote:

```
Currently, IntelliJ can't compile this: public List<String> toList(String... someStrings) {
List<String> theList = List.of(someStrings);
return theList;
}
But it works if I change the first line on the method to: List<String> theList = List.<String>of(someStrings); 

My intention is to handle the vararg someStrings as a list inside the method, but expose a nice API for consumers of this methods in which they can invoke it like: toList("a", "b", "c", ...)

This is curious because in 1.2.2 it worked flawlessly
Maybe we could have a List.ofAll(T[] tArray)
I mean, List.ofAll(T[] tArray) => List<T>
```

danieldietrich:
Fixes #721

### DIFF
--- a/src-gen/main/java/javaslang/Tuple1.java
+++ b/src-gen/main/java/javaslang/Tuple1.java
@@ -104,7 +104,7 @@ public final class Tuple1<T1> implements Tuple, Comparable<Tuple1<T1>>, Serializ
 
     @Override
     public Seq<?> toSeq() {
-        return List.of(_1);
+        return List.ofAll(_1);
     }
 
     // -- Object
--- a/src-gen/test/java/javaslang/Tuple1Test.java
+++ b/src-gen/test/java/javaslang/Tuple1Test.java
@@ -40,7 +40,7 @@ public class Tuple1Test {
     @Test
     public void shouldConvertToSeq() {
         final Seq<?> actual = createIntTuple(1).toSeq();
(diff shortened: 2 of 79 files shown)


src-gen/main/java/javaslang/Tuple1.java
```diff
@@ -104,7 +104,7 @@ public T1 _1() {
 
     @Override
     public Seq<?> toSeq() {
-        return List.of(_1);
+        return List.ofAll(_1);
```
danieldietrich: See [this comment](https://github.com/javaslang/javaslang/issues/721#issuecomment-154803154) regarding readability.

src-gen/test/java/javaslang/Tuple1Test.java
```diff
@@ -40,7 +40,7 @@ public void shouldReturnElements() {
     @Test
     public void shouldConvertToSeq() {
         final Seq<?> actual = createIntTuple(1).toSeq();
-        assertThat(actual).isEqualTo(List.of(1));
+        assertThat(actual).isEqualTo(List.ofAll(1));
```
danieldietrich: see Tuple1

Issue #779, opened 2015-11-07
foldRight combine should be lazy


danieldietrich:
For a description see [this issue comment](https://github.com/javaslang/javaslang/pull/775#issuecomment-154659034) of @jbgi

We have to ensure that the semantics of current `TraversableOnce.foldRight` usages do not change if we make foldRight lazy.

danieldietrich:
Follow this article: http://voidmainargs.blogspot.de/2011/08/folding-stream-with-scala.html

Also see this: http://gaabbe.win/issue/1975741/foldright-on-infinite-lazy-structure

danieldietrich:
A lazy foldRight is good for infinite data structures like Iterator and Stream because an eager implementation evaluates all elements.

A general Javaslang foldRight (of TraversableOnce) could look like this:

```java
<U> U foldRight(U zero, BiFunction<? super T, ? super U, ? extends U> combine) {
    return isEmpty() ? zero : combine(head(), tail().foldRight(zero, combine));
}
```

However, the following example ends up with a StackOverflowError when eager evaluated, although it should terminate from a logical viewpoint:

```java
// StackOverflowError
boolean b = Stream.repeat(true).foldRight(false, (a, b) -> a || b);
```

In order to get the example working we make the second argument of the combine function lazy:

```java
<U> U foldRight(U zero, BiFunction<? super T, ? extends Supplier<? super U>, ? extends U> combine) {
    return isEmpty() ? zero : combine(head(), () -> tail().foldRight(zero, combine));
}
```

This leads to:

```java
// = true
boolean b = Stream.repeat(true).foldRight(false, (a, b) -> a || b::get);
```

**@jbgi: There is only one problem - _the recursion_ could now cause a stack overflow error.**

Any suggestions? Sadly we have no tail recursion. Are trampolines a workaround here? How would a constant stack solution look like. Would it be still efficient regarding creation of intermediate objects? foldRight already has a recursive call and an extra supplier for the lazy evaluation. Creating more objects each step would stress the GC.

jbgi:
yeah, trampolining is slow. It would be nonetheless interesting to benchmark trampoline based folds against different GC/GC tuning. functional-java has trampoline based ```foldRight```s for Stream and List, as well as constant stack impl base on ```reverse```+```foldLeft``` but that does not work for infinite data structures.

danieldietrich:
Yes, I also use reverse.foldLeft for Seq types.

[...]

PR #783, opened 2015-11-07
Repeat an element infinitely often

danieldietrich:


danieldietrich:
Needed for foldRight tests, see #779

PR #791, opened 2015-11-09
duplicate code again

ruslansennov:


danieldietrich:
I love duplicate code - when it disappears :-) Great!

PR #792, opened 2015-11-09
TreeSet.addAll(It) fixed

ruslansennov:


src/main/java/javaslang/collection/TreeSet.java
```diff
@@ -409,18 +409,14 @@
     @Override
     public TreeSet<T> addAll(java.lang.Iterable<? extends T> elements) {
         Objects.requireNonNull(elements, "elements is null");
-        if (isEmpty() && elements instanceof TreeSet) {
-            return (TreeSet<T>) elements;
```
ruslansennov: it was wrong in case of <code>this.comparator != elements.comparator</code>

danieldietrich:
You are right! Many thanks

danieldietrich:
Btw, for your planning, I will push RC1 to maven central on 15. Nov.

I need to do s.th. different than coding in between, s.th. healthy :-)

ruslansennov:
gogogo

danieldietrich:
yes, will do.

PR #794, opened 2015-11-11
Added Future.andThen()

danieldietrich:
Towards #449 

TODO: isEmpty(), get(), get(timeout, unit) - these are currently throwing


danieldietrich:
Note: Also fixed a corner case in the internal FutureImpl.run() method

PR #799, opened 2015-11-12
Try.getCause() returns original exception

### LINKED ISSUE #798: Let Try.getCause() return Throwable instead of NonFatal
Returning a NonFatal has historical reasons because there was a time the cause could also be a Fatal.

Because the NonFatal wraps the original exception, we currently have to call `myTry.getCause().getCause()` to get the exception. This is distracting.

We want to simplify the API by returning the original exception instead of a NonFatal.

danieldietrich:
Fixes #798


danieldietrich:
I'm aware that this might change the behavior of applications when migrating from Javaslang 1.2.2 to 2.0.0. Beside the introduction of the new collections, making the API better at the cost of backward compatibility issues is the reason for having a new major version.

PR #813, opened 2015-11-15
Moved applicability checks to λ, added more unit tests

### LINKED ISSUE #811: Functions: add check for partial applicability
We should allow to check partial applicability because Function-_n_ has _n_ `apply` methods, of arity 1..n.

We change the methods `isApplicableTo()` and `isApplicableToTypes()` in the way that they take a varargs parameter. This will not break existing API.

Additionally these methods can also be added to our &lambda; interface.

danieldietrich:
Fixes #811


danieldietrich:
Towards #477

PR #817, opened 2015-11-15
added unzip3 as in Scala collections + removed a few unnecessary SuppressWarnings

mvh77:


src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -881,6 +894,24 @@ public CharSeq takeWhile(Predicate<? super Character> predicate) {
...
+		Vector<T3> zs = Vector.empty();
+        for (int i = 0; i < length(); i++) {
+            final Tuple3<? extends T1, ? extends T2, ? extends T3> t = unzipper.apply(back.charAt(i));
+            xs = xs.append(t._1);
+            ys = ys.append(t._2);
+			zs = zs.append(t._3);
+        }
+        return Tuple.of(xs.isEmpty() ? Vector.<T1> empty() : xs, 
+						ys.isEmpty() ? Vector.<T2> empty() : ys,
+						zs.isEmpty() ? Vector.<T3> empty() : zs);
```
danieldietrich: Hey @mvh77, that's a really nice addition!

I think we don't need the isEmpty() checks here. For example when xs.isEmpty() then xs is already Vector.empty() and we can just return xs. What do you think, have I overseen s.th.?

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -69,7 +70,6 @@ public Object next() {
      * @param <T>       Component type.
      * @return A new {@code javaslang.collection.Iterator}
      */
-    @SuppressWarnings("varargs")
```
danieldietrich: It is good to remove unnecessary stuff. After the pull I will test it again in Eclipse, IntelliJ etc. I had cases where one IDE says it is ok, the other complains... We will see :-)

src/main/java/javaslang/collection/Vector.java
```diff
@@ -1180,6 +1191,23 @@ public int length() {
...
+		HashArrayMappedTrie<Integer, T3> zs = HashArrayMappedTrie.empty();
+        for (T element : this) {
+            final Tuple3<? extends T1, ? extends T2, ? extends T3> t = unzipper.apply(element);
+            xs = xs.put(xs.size(), t._1);
+            ys = ys.put(ys.size(), t._2);
+			zs = zs.put(zs.size(), t._3);
+        }
+        return Tuple.of(xs.isEmpty() ? empty() : new Vector<>(xs), 
+						ys.isEmpty() ? empty() : new Vector<>(ys),
+						zs.isEmpty() ? empty() : new Vector<>(zs));
```
danieldietrich: Please check here also, if the extra isEmpty() handling is really needed. My feeling says it is not.

src/main/java/javaslang/collection/CharSeq.java
mvh77: Hi. I pretty much copied the unzip method, and that one actually had xs.length() == 0, but anyway, you're right, now that I look at it with my brains on the isEmpty makes no sense. :)

src/main/java/javaslang/collection/Vector.java
mvh77: Yeah it isn't, it comes from the unzip method. That's what you get copying other people's code. :)

src/main/java/javaslang/collection/CharSeq.java
danieldietrich: Oh you are right :)

You carried out the changes carefully, very nice! Please do the following:

* remove the extra isEmpty() checks in CharSeq and Vector
* please also change CharSeq.unzip and Vector.unzip and remove `== 0` checks

Then I will pull it and immediately release 2.0.0-RC1

Great work!

danieldietrich:
Ah yes, we were bitten by the `-Xlint:all -Werror` compiler flags - the CI build currently does not compile because of the removed @SuppressWarnings.

It does not matter - I will pull your change and apply the changes we discussed in order to release RC1.

It's ok!

Issue #612, opened 2015-09-12
Add LinkedHashMap & LinkedHashSet


danieldietrich:
These will be backed by List + HashMap/HashSet.

See also https://github.com/abasko/linkedhashmap/tree/master/src/Data

It is interesting to see that Scala (and Clojure?) currently do not seem to have immutable LinkedHash* implementations. It is a very common feature in real-world applications. IMO we should providede them, even if some ops make take O(n). This may be further optimized by internally using Vector instead of List.

danieldietrich:
Map implementations:

* HashMap implements Map (no order)
* LinkedHashMap extends HashMap (insertion order)
* TreeMap implements SortedMap (ascending key order, natural order by default, alternatively by user-defined comparator)

danieldietrich:
One limitation of Java's type system is that method return types are only _partially_ covariant. This comes into play when providing more hash-based Set or Map implementations.

Let's take a look at the Set interface.

```java
interface Set<T> {

    Set<T> take(int n);

    Option<? extends Set<T>> initOption();
}
```

Then this is a simple example to make the return type of a method more specific in a sub-class (public modifiers omitted):

```java
class HashSet<T> implements Set<T> {
    @Override
    HashSet<T> take(int n) { ... }
}
```

But if the return type has more complex generics, this is not possible any more:

```java
class HashSet<T> implements Set<T> {
    @Override
    Option<HashSet<T>> initOption() { ... }
}

class LinkedHashSet<T> extends HashSet<T> {
    @Override
    // DOES NOT COMPILE
    Option<LinkedHashSet<T>> initOption() { ... }
}
```

There are two solution:

**1) Open the return type for extension**

```java
/*
     Set
      ^
      |
   HashSet
      ^
      |
LinkedHashSet
*/
class HashSet<T> implements Set<T> {
    @Override
    Option<? extends HashSet<T>> initOption() { ... }
}

class LinkedHashSet<T> extends HashSet<T> {
    @Override
    Option<? extends LinkedHashSet<T>> initOption() { ... }
}
```

**2) Flatten the hierarchy**

```java
/*
       Set
        ^
       / \
HashSet   LinkedHashSet
*/
final class HashSet<T> implements Set<T> {
    @Override
    Option<HashSet<T>> initOption() { ... }
}

final class LinkedHashSet<T> extends Set<T> {
    @Override
    Option<LinkedHashSet<T>> initOption() { ... }
}
```

[...]

PR #823, opened 2015-11-18
Linked hash map

ruslansennov:


ruslansennov:
#612

src/main/java/javaslang/collection/HashSet.java
```diff
@@ -800,7 +800,7 @@ public boolean equals(Object o) {
 
     @Override
     public String toString() {
-        return mkString(", ", "HashSet(", ")");
+        return mkString("HashSet(", ", ", ")");
```
danieldietrich: Oh, thanks for catching - I thought I've catched all after changing `mkString(infix, prefix, suffix)` to `mkString(prefix, infix, suffix)`.

danieldietrich:
Wow, THAT. WAS. FAST. Thx!

danieldietrich:
Fixes #612

src/main/java/javaslang/collection/HashSet.java
ruslansennov: that PR was very dangerous :smile:

PR #825, opened 2015-11-19
Complexity

ruslansennov:
Here is actual table.
* complexity of <code>HashSet.min()</code> and <code>HashMap.min()</code> is <strong>linear</strong>
* complexity of all <code>Linked*</code> operations except <code>contains/Key</code> is <strong>linear</strong>

<code>LinkedHashSet.put()</code> and <code>LinkedHashMap.put()</code> can be <code>const</code> if we change backing sequence from <code>List</code> to <code>Queue</code>


danieldietrich:
of cause - right!

PR #826, opened 2015-11-19
Conditional option create

v1ctor:
Hi there. 
I propose this enhancement because sometimes it's really convenient.
Hope that it can be useful for another people.

### DIFF
--- a/src/main/java/javaslang/control/Option.java
+++ b/src/main/java/javaslang/control/Option.java
@@ -58,11 +58,9 @@ public interface Option<T> extends Value<T> {
      *
      * @return return {@code Some} of supplier's value if condition is true, or {@code None} in other case
      */
-    static <T> Option<T> when(boolean condition, Supplier<T> supplier) {
-        if (!condition) {
-            return none();
-        }
-        return of(supplier.get());
+    static <T> Option<T> when(boolean condition, Supplier<? extends T> supplier) {
+        Objects.requireNonNull(supplier, "supplier is null");
+        return condition ? of(supplier.get()) : none();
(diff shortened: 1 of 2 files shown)


danieldietrich:
Hi @v1ctor, thank you for your proposal!
Yes, this might come handy. I will comment the code in a minute...

src/main/java/javaslang/control/Option.java
```diff
@@ -53,6 +53,19 @@
     }
 
     /**
+     *
+     * Creates {@code Some} of suppliers value if condition is true, or {@code None} in other case
+     *
+     * @return return {@code Some} of supplier's value if condition is true, or {@code None} in other case
+     */
+    static <T> Option<T> when(boolean condition, Supplier<T> supplier) {
```
danieldietrich: Let's use `Supplier<? extends T>`. There are situations where it is necessary.

Also we do null checks in order to provide user-friendly messages, e.g. `Objects.requireNonNull(supplier, "supplier is null");`

src/main/java/javaslang/control/Option.java
```diff
@@ -53,6 +53,19 @@
...
+        if (!condition) {
+            return none();
+        }
+        return of(supplier.get());
```
danieldietrich: I would to a `condition ? of(supplier.get()) : none();` here but that is a matter of taste. Using your guard method is ok!

v1ctor:
Thank you. I have fixed it in accordance to your comments.

danieldietrich:
Very nice, many thanks!

PR #828, opened 2015-11-20
Queue.dropRight() fixed; Linked* collections improved

ruslansennov:


src/main/java/javaslang/collection/Queue.java
```diff
@@ -627,7 +627,7 @@ public T peek() {
         if (n <= 0) {
             return this;
         }
-        return new Queue<>(front.dropRight(n), rear.drop(n - front.length()));
```
ruslansennov: this fails only if <code>Queue</code> created by <code>Queue.empty().append(T).append(T)...</code>

danieldietrich:
Great, thx!

PR #829, opened 2015-11-20
add optional wrapping and tests

v1ctor:
Hi!
There is another useful static method to wrap Java Optional. 
I haven't found the same.

### DIFF
--- a/src/main/java/javaslang/control/Option.java
+++ b/src/main/java/javaslang/control/Option.java
@@ -6,6 +6,7 @@
 package javaslang.control;
 
 import java.util.Objects;
+import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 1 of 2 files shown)


src/main/java/javaslang/control/Option.java
```diff
@@ -64,6 +65,20 @@
...
 
     /**
+     *
+     * Wraps a Java Optional to a new Option
+     *
+     * @param optional a given optional to wrap in {@code Option}
+     * @param <T> type of the value
+     * @return {@code Some(optional.get())} if value is Java {@code Optional} is present, {@code None} otherwise
+     */
+    static <T> Option<T> wrap(Optional<T> optional) {
```
danieldietrich: Great, thx! I've two remarks:

* Let's take `Optional<? extends T>`, it is useful in some cases
* I would not call it _wrap_, because the `Optional` instance is not wrapped but converted somehow. `toOption` is already defined in `Value` and `asOption` would be confusing because we also have `toOption`. It is syntactic sugar/a shortcut for `Option.of(optional.orElse(null))`. Using `of(Optional)` would raise ambiguities regarding `of(T)`. Let's call it `ofOptional(Optional)`.

src/main/java/javaslang/control/Option.java
```diff
@@ -64,6 +65,20 @@
...
+        Objects.requireNonNull(optional, "optional is null");
+        return optional.isPresent() ? of(optional.get()) : none();
```
danieldietrich: It could also be `return of(optional.orElse(null));` - your choice :-)

v1ctor:
Thanks! I have fixed it in accordance to your notes.

danieldietrich:
LGTM, thx!

PR #834, opened 2015-11-21
Segment length

mvh77:
segment length as in Scala + IndexedSeq.indexWhere optimization


danieldietrich:
Nice, thx! Looks very good!

Issue #824, opened 2015-11-18
Stream.append produces StackOverFlow


danieldietrich:
@netzwerg discovered a StackOverflowError: https://github.com/netzwerg/paleo/commit/e962a43fa435aa54211ad17b8c093f5cc0ec0991

danieldietrich:
I imported paelo into IntelliJ and started the `ProfilingTest.stackOverflow()` test method. Streaming just Strings and using concatenation as fold operation works fine (without StackOverflow):

```java
@Test
public void stackOverflow() {
    StringColumn.Builder columnBuilder = StringColumn.builder(ColumnIds.stringCol("test"));
    Function2<StringColumn.Builder, String, StringColumn.Builder> append = (b, s) -> b.add(s);
//    StringColumn column = Stream.range(0, COUNT).map(String::valueOf).foldLeft(columnBuilder, append).build();
    Stream.range(0, COUNT).peek(System.out::println).map(String::valueOf).foldLeft("", String::concat);
}
```

Now I investigate which portion of the `append` Function leads to the exception in `foldLeft`...

danieldietrich:
Yes, it is now clear to me. The column builder also uses internally a Stream to accumulate strings. This is done by calling Stream.append. Because Stream is lazy, all append operations are 'rewinded' on the stack when finally calling `build()`. This leads to the StackOverflowException.

I'm currently not sure, if we can do the final `append` for all elements in a lazy manner on a constant stack. I will check this...

danieldietrich:
This one is semantically equivalent and also leads to a StackOverflowError:

```java
Stream.range(0, 1_000_000).map(String::valueOf).foldLeft(Stream.empty(), Stream::append).mkString();
```

The equivalent Scala code looks like this:

```scala
Stream.from(0).take(1000000).map(_.toString).foldLeft(Stream(): Stream[String])(_ :+ _).mkString
```

The Scala-code surprisingly (for me) does compute the String on a constant stack! I've investigated the Scala implementation of `Stream.append`. It looks similar to our implementation. However, Scala as special language support for lazy values, denoted as `=> expression`. Java does not have this and needs to call a `Supplier` function.

I will further investigate, if we can provide an implementation that evaluates the Stream contents on constant Stack...

danieldietrich:
Note: While investigating how to omit the deferred recursion, I sumbled upon two places where the number of objects instantiated when creating/iterating streams can be improved.

[...]

PR #835, opened 2015-11-21
tweaking generated bytecode...

danieldietrich:
…in order to prevent lambda meta factory from capturing `this`


danieldietrich:
towards #824

PR #837, opened 2015-11-22
Scans

mvh77:
Implemented scan, scanLeft, scanRight. More tests to come.


src/main/java/javaslang/collection/ScanImpl.java
```diff
@@ -0,0 +1,49 @@
...
+            scanned = scanned.prepend(acc);
+        }
+        List<B> builder = List.empty();
+        for(B elem: scanned) {
+            builder = builder.prepend(elem);
+        }
+        return builder.reverse();
+    }
+    
+    private static <A> Traversable<A> reverse(Traversable<A> t) {
```
danieldietrich: I think this method can be omitted, it should be the same as calling `t.reverse()` directly because of the existing `List.reverse()` impl.

mvh77: no, scanRight takes a Traversable. Traversables don't have a reverse, it has to be reversed in case it isn't a Seq.

src/main/java/javaslang/collection/ScanImpl.java
```diff
@@ -0,0 +1,49 @@
+package javaslang.collection;
+
+import java.util.Objects;
+import java.util.function.BiFunction;
+
+class ScanImpl {
```
danieldietrich: We should move this functionality to `javaslang.collection.Iterator` and remove the `ScanImpl` class. Whe then remove the `thiz` arg and just use `this` instead. The other classes call `iterator().scan...` instead of `ScanImpl.scan...`.

src/main/java/javaslang/collection/ScanImpl.java
danieldietrich: yes, you're right

src/main/java/javaslang/collection/ScanImpl.java
mvh77: possible, but we still need to have that reverse (actually should be reversed if we copy Scala) method. in an interface it will be visible.

mvh77: it should actually go to Traversable then, like in Scala.

src/main/java/javaslang/collection/ScanImpl.java
```diff
@@ -0,0 +1,49 @@
...
+
+    static <A, B, X extends Traversable<A>> List<B> scanLeft(B zero, BiFunction<? super B, ? super A, ? extends B> operation, X thiz){
+        Objects.requireNonNull(zero, "zero is null");
```
danieldietrich: what happens if we permit `zero` to be `null`- should work!?

src/main/java/javaslang/collection/ScanImpl.java
danieldietrich: alright!

mvh77: I like the idea of putting it in Traversable, cleaner, but we need a Utils class for that static reverse method. What do you think?

danieldietrich: > possible, but we still need to have that reverse (actually should be reversed if we copy Scala) method. in an interface it will be visible.

In this case I created auxiliary classes, in this case

```
public interface Traversable<T> { ... }

interface TraversableModule<T> {
   
    class ScanUtil {
        static ... reverse(...) { ... }
    }
}
```

danieldietrich: The compiler flag `-Xlint:all -Werror` prevents compiling if auxiliary classes are accessed from outside of the class file.

src/main/java/javaslang/collection/ScanImpl.java
danieldietrich: an alternative would be to create `Seq` factory methods which default to `List` for example. Then we would have

```java
Seq.ofAll(traversable).reverse()
```

instead of

```java
helper.reverse(traversable);
```

src/main/java/javaslang/collection/ScanImpl.java
mvh77: ah, it doesn't work, you'd have to implemet it on every interface.

src/main/java/javaslang/collection/ScanImpl.java
mvh77: well, I guess it's the functions responsibility to handle it, so why not.

src/main/java/javaslang/collection/Seq.java
```diff
@@ -97,6 +97,14 @@
  */
 public interface Seq<T> extends Traversable<T>, IntFunction<T> {
 
+    static <T> Seq<T> ofAll(Traversable<T> traversable) {
```
danieldietrich: Awesome! Overall it now looks really great.

One more thing: let's relax `ofAll(Traversable)` and do the following:

```java
static <T> Seq<T> ofAll(java.lang.Iterable<? extends T> elements) { ... }
```

Thx!

danieldietrich:
Awesome!

PR #841, opened 2015-11-22
test cases + bug fixes

mvh77:
should be 100% coverage on scan methods

### DIFF
--- a/src/main/java/javaslang/collection/CharSeq.java
+++ b/src/main/java/javaslang/collection/CharSeq.java
@@ -742,7 +742,7 @@ public final class CharSeq implements CharSequence, IndexedSeq<Character>, Seria
     @Override
     public Vector<Character> scan(Character zero,
             BiFunction<? super Character, ? super Character, ? extends Character> operation) {
-        return scan(zero, operation);
+        return scanLeft(zero, operation);
     }
     
     @Override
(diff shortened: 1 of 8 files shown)


danieldietrich:
Great!

I've dived into the scan methods today. IMO we should get the hands on them again an do the following:

* The _main_ interfaces remain the same, e.g. Traversable, Seq/IndexedSeq/LinearSeq, Map/SortedMap, Set/SortedSet.
* The implementing classes (List, Stream, Queue, HashSet, ...) should provide individual implementations in order not to convert the collections. Here some code-duplication will take place in favor of better performance.
* Iterator should be lazy for all non-aggregating methods. We need to return an Iterator of prefixes. This can be used for the Map impls as many of the Map methods use Iterator.

This is some sort of reverse-refactoring because we unfold the existing code but I think it is worth it.

danieldietrich:
We may also abstract over it and provide an empty result, e.g. `HashSet.empty()`, an aggregate function, e.g. `result::add`, and a finished, e.g. `identity` or `result.reverse()´ when the aggregator is `prepend`.

It is like the java.util.stream.Stream collector. Then we have no code duplication. The only difference will be Iterator. Here we will have individual scan impls I think.

mvh77:
Maybe it's not the first thing to optimize. Does anybody even use those methods? :smile:  I didn't see the Scala library use any optimized versions. Maybe for now one optimization would be to change the first for loop in scanRight to use the reverseIterator which is optimized for indexedSeqs
```java
        Iterator<T> it = Seq.ofAll(this).reverseIterator();
        while(it.hasNext()){
            acc = operation.apply(it.next(), acc);
            scanned = scanned.prepend(acc);
        }
```

danieldietrich:
Yes maybe you're right. There are still plenty of things to do until 2.0.0. I tend to move some things from 2.0.1 to 2.0.0 (sequence and transform) to get a more complete lib with little effort. We target start of december for 2.0.0 release but IMO it is more important to get it right and stable than shipping a few days earlier.

PR #855, opened 2015-11-24
Overwrite forEach method in Map to use BiConsumer

v1ctor:
Hi!

I have written an forEach method for map which is using BiConsumer.
It's convenient to write like this:
```java
map.forEach((k, v) -> { doSomething(k, v); });
```
instead of:
```
map.forEach(t -> { doSomething(t._1, t._2); });
```

### DIFF
--- a/src/main/java/javaslang/collection/Map.java
+++ b/src/main/java/javaslang/collection/Map.java
@@ -5,15 +5,20 @@
  */
 package javaslang.collection;
 
-import javaslang.Tuple2;
-import javaslang.Tuple3;
-import javaslang.control.Option;
-
-import java.util.*;
+import java.util.Comparator;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
+
+import javaslang.Tuple2;
+import javaslang.Tuple3;
+import javaslang.control.Option;
 
 /**
  * An immutable {@code Map} interface.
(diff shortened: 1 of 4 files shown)


danieldietrich:
Also here I'm sorry for the merge conflicts!

src/main/java/javaslang/collection/Map.java
```diff
@@ -224,10 +229,10 @@ default boolean contains(Tuple2<K, V> element) {
         Objects.requireNonNull(unzipper, "unzipper is null");
         return iterator().unzip(unzipper).map(Stream::ofAll, Stream::ofAll);
     }
-    
+
     @Override
     default <T1, T2, T3> Tuple3<Seq<T1>, Seq<T2>, Seq<T3>> unzip3(
-    		Function<? super Tuple2<K,V>, Tuple3<? extends T1, ? extends T2, ? extends T3>> unzipper) {
```
danieldietrich: Please try not to reformat the code. Do you use Eclipse? For IntelliJ I exported the settings, located in `.ide/`.

v1ctor: I'm sorry. I've installed settings and reformat this.

v1ctor: Don't you want to setup checkstyle and findbugs target in travis to have static code analysis?

danieldietrich: Yes, do you have a hint how to do it?

I've mixed feelings with these tools because sometimes they seem to know more than the developer or produce false errors. However, these tools are great finding critical bugs. Using them should create value for us.

v1ctor: Yes I've some experience in it. If we will have such false positives we could setup the environment to ignore this errors.

danieldietrich:
Very nice, thank you Victor!

src/main/java/javaslang/collection/Map.java
danieldietrich: Great! Please update the travis config - I'm really looking forward to the checks :-)

I'm sure that there will be significant amount of TODO to fix the findings :-))

PR #863, opened 2015-11-26
HAMT: unnecessary Lazy objects removed

ruslansennov:
Now I'm playing with @netzwerg's [Profiler](https://github.com/netzwerg/paleo/blob/feature-9/paleo-core/src/test/java/ch/netzwerg/paleo/ProfilingTest.java). 
This commit (hope only the first in this investigation) reduces the number of Lazy objects twice and saving 24MB of memory.

### DIFF
--- a/src/main/java/javaslang/collection/HashArrayMappedTrie.java
+++ b/src/main/java/javaslang/collection/HashArrayMappedTrie.java
@@ -66,8 +66,6 @@ interface HashArrayMappedTrieModule {
         private static final int M2 = 0x33333333;
         private static final int M4 = 0x0f0f0f0f;
 
-        private final transient Lazy<Integer> hashCode = Lazy.of(() -> Traversable.hash(this));
-
         int bitCount(int x) {
             x = x - ((x >> 1) & M1);
             x = (x & M2) + ((x >> 2) & M2);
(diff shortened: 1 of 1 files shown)


danieldietrich:
travic-ci?

ruslansennov:
same question...

danieldietrich:
There was an incident today [Build processing delays for public projects (travis-ci.org)](https://www.traviscistatus.com/incidents/phvh6jslkhr7) - but should be fixed now. We will see...

danieldietrich:
We should question the whole idea of caching hashes. I mean, a lazy wrapper of a lambda/value is computed for every node of a data structure in general - also for primitive values like ints, which are commonly used. We need to profile the memory consumption of different scenarios and compare it with other frameworks, first of all Java standard lib.

We should answer the question how often the hash code is really needed in practice, e.g. how often collections are used as keys in hash-based collections. And what is the value count where caching really starts to matter. And what is more important: CPU or Memory. Many questions.

danieldietrich:
I tend to throw away the Lazy hash code of _all_ types for now. We can't argue a 10x or more memory consumption when we compare Javaslang with other collections.

ruslansennov:
but don't rush :smile:

danieldietrich:
Filed #864

PR #868, opened 2015-11-28
HAMT: Leaf like lightweight List implemented; Tuple2 objects removed

ruslansennov:
About 2 million objects removed

### DIFF
--- a/src/main/java/javaslang/collection/HashArrayMappedTrie.java
+++ b/src/main/java/javaslang/collection/HashArrayMappedTrie.java
@@ -66,7 +66,7 @@ interface HashArrayMappedTrieModule {
         private static final int M2 = 0x33333333;
         private static final int M4 = 0x0f0f0f0f;
 
-        int bitCount(int x) {
+        static int bitCount(int x) {
             x = x - ((x >> 1) & M1);
             x = (x & M2) + ((x >> 2) & M2);
             x = (x + (x >> 4)) & M4;
@@ -75,15 +75,15 @@ interface HashArrayMappedTrieModule {
             return x & 0x7f;
         }
 
-        int hashFragment(int shift, int hash) {
+        static int hashFragment(int shift, int hash) {
             return (hash >>> shift) & (BUCKET_SIZE - 1);
         }
 
-        int toBitmap(int hash) {
+        static int toBitmap(int hash) {
             return 1 << hash;
         }
 
-        int fromBitmap(int bitmap, int bit) {
+        static int fromBitmap(int bitmap, int bit) {
             return bitCount(bitmap & (bit - 1));
         }
 
@@ -167,7 +167,7 @@ interface HashArrayMappedTrieModule {
 
         @Override
         AbstractNode<K, V> modify(int shift, K key, Option<V> value) {
-            return value.isEmpty() ? this : new LeafNode<>(Objects.hashCode(key), key, value.get());
+            return value.isEmpty() ? this : new LeafSingleton<>(Objects.hashCode(key), key, value.get());
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


danieldietrich:
Sounds great! I want to understand more about this optimization. Maybe I can apply similar things to other parts of Javaslang.

What was the test case / how do you test it? You mentioned @netzwerg's test. Which tool do you use to profile it?

How do you get the number of objects used? Is it the overall number of objects in memory or are we able to measure the objects in a dependency graph?

src/main/java/javaslang/collection/HashArrayMappedTrie.java
```diff
@@ -66,7 +66,7 @@
         private static final int M2 = 0x33333333;
         private static final int M4 = 0x0f0f0f0f;
 
-        int bitCount(int x) {
+        static int bitCount(int x) {
```
danieldietrich: What's the benefit of making it static? Just asking. `static` is fine! It makes clear that we do not depend on any (mutable) state :-) 

We save the implicit `this`. Are there more benefits? I've read somewhere that "Memory is allocated for a method when the method is invoked. (And it doesn't matter whether the method is a static or instance method.". The bytecode for non-static methods is allocated only once and re-used by every instance.

src/main/java/javaslang/collection/HashArrayMappedTrie.java
```diff
@@ -212,80 +212,185 @@ private Object readResolve() {
...
 
-        private static final long serialVersionUID = 1L;
+        abstract AbstractNode<K, V> removeElement(K key);
+        abstract int hash();
+        abstract K key();
+        abstract V value();
 
-        private final int hash;
-        private final int size;
-        private final List<Tuple2<K, V>> entries;
```
danieldietrich: Aaah, I understand. Nice idea!! To save all the instances. Great!!

ruslansennov:
I use standard Oracle profiler <strong>Java VisualVM</strong>. Fortunately objects belonging to <code>Vector</code> are immediately visible in the list, because a lot of them.

Before my last two commits:
beforeopt

... and after
afteropt

It's clear that before optimizations total size of all Vector's object is about 200MB (like @netzwerg says), and after is about 115 MB :smile:

src/main/java/javaslang/collection/HashArrayMappedTrie.java
ruslansennov: > static is fine! It makes clear that we do not depend on any (mutable) state

Exactly! And I believe there is no other defference

danieldietrich:
Nice! Curious to see how memory consumption changes if we remove Lazy hashCode...

ruslansennov:
I'm afraid this change tragically decrease performance in some certain cases

danieldietrich:
Yes, it will. That's the trade between Memory and CPU.

Memory is more important when we take into account that huge Javaslang data structures are rarely used as Map keys. But Memory-Consumption is always present.

danieldietrich:
Theory and Practice...

src/main/java/javaslang/collection/HashArrayMappedTrie.java
```diff
@@ -212,80 +212,185 @@ private Object readResolve() {
...
+            this.hash = hash;
+            this.key = key;
+            this.value = value;
+            this.hashCode = Lazy.of(() -> Objects.hash(key, value));
         }
 
         @Override
         Option<V> lookup(int shift, K key) {
-            if (hash != key.hashCode()) {
+            if (key.equals(this.key)) {
```
danieldietrich: If we allow null keys, this has to be `Objects.equals(key, this.key)`, right?

ruslansennov: You're right.
And we need more, MORE, <strong>MORE</strong> tests :smile:

danieldietrich: Yes, tests will come. I'm on s.th. and am busy for a few days until I can fully focus on the tests...

the `equals` calls need to be changed all over the HAMT. we also compare `tuple._1.equals(key)` in LeafList etc.

danieldietrich:
Could you make a quick check how Lazy hashCode impacts our HAMT mem consumption? Should be a fast thing removing the Lazy instance var. Just to get a feeling...

danieldietrich:
And the `LeafSingleton.hash` field is removed, `hash() { return Objects.hashCode(key); }`

PR #872, opened 2015-11-28
Got Lazy right

### LINKED ISSUE #840: Reduce footprint of Lazy
Currently `Lazy` has an additional instance field which indicates if the lazy `Value` is empty or not. E.g. this is needed for `filter` because `Lazy` is single-valued.

We may omit this flag if we make Lazy an interface and provide two implementation, `Val` and `Empty`.

danieldietrich:
Fixes #840


danieldietrich:
(Note to myself: capture the state with _types_, not with _values_)

PR #883, opened 2015-11-29
Replace "\n" with platform-independent line separator

tkob:
This PR fixes the problem that some assertions fail on Windows platform because of hard-coded line separators.

### DIFF
--- a/src/test/java/javaslang/collection/AbstractTraversableTest.java
+++ b/src/test/java/javaslang/collection/AbstractTraversableTest.java
@@ -18,6 +18,7 @@ import java.util.*;
 import java.util.function.Function;
 import java.util.stream.Collector;
 
+import static java.lang.System.lineSeparator;
 import static javaslang.Serializables.deserialize;
 import static javaslang.Serializables.serialize;
 import static org.assertj.core.api.Assertions.fail;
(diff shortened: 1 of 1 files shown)


src/test/java/javaslang/collection/AbstractTraversableTest.java
```diff
@@ -1377,7 +1378,7 @@ public void shouldWriteToPrintWriter() {
         final StringWriter sw = new StringWriter();
         final PrintWriter out = new PrintWriter(sw);
         ofAll(1, 2, 3).out(out);
-        assertThat(sw.toString()).isEqualTo(ofAll(1, 2, 3).mkString("", "\n", "\n"));
+        assertThat(sw.toString()).isEqualTo(ofAll(1, 2, 3).mkString("", lineSeparator(), lineSeparator()));
```
danieldietrich: Great, I think that solves a problem we recently had (without any solution).

I ask myself, if it really should be

```java
mkString("", lineSeparator(), lineSeparator())
```

because we changed the argument order of `mkString` from `infix, prefix, suffix` to `prefix, infix, suffix`.

Therefore the following may be right:

```java
mkString(lineSeparator(), "", lineSeparator())
```

But if the tests pass, it probably is right as-is.

PR #885, opened 2015-11-29
Add 'accept' method to Option class

tkob:
This PR adds 'accept' method to Option class; the following expression is equal to "none" if `opt` is of type None and "some" if it is of type Some.

```
opt.accept(() -> "none", v -> "some")
```

The term 'accept" comes from GoF Visitor pattern. Another candidate would be 'match', since this can be seen as pattern match against union type.

### DIFF
--- a/src/main/java/javaslang/control/None.java
+++ b/src/main/java/javaslang/control/None.java
@@ -7,6 +7,8 @@ package javaslang.control;
 
 import java.io.Serializable;
 import java.util.NoSuchElementException;
+import java.util.function.Function;
+import java.util.function.Supplier;
 
 /**
  * None is a singleton representation of the undefined {@link javaslang.control.Option}. The instance is obtained by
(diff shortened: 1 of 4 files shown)


danieldietrich:
The

`opt.accept(() -> "none", v -> "some")`

is the same as

`opt.map(v -> "some").orElseGet(() -> "none")`

or

`opt.map(v -> "some").orElse("none")`

If the previous value is not needed, we can use one of the methods `ifDefined` or `ifEmpty`, inherited from `Value`. (In other words, all Values have them, e.g. Either, Try, Match, List, ...)

```java
opt.ifDefined("some", "none")

opt.ifDefined(() -> "some", () -> "none")

opt.ifEmpty("none", "some")

opt.ifEmpty(() -> "none", () -> "some")
```

---

To not blow up the API too much I will not pull this request. (Actually the first PR I rejected)

I hope you understand.

Thanks you for your ideas! Keep on PR-ing!

danieldietrich:
Please close the PR if you agree or ask if there are any questions!

tkob:
OK. Thank you for the reply.

PR #886, opened 2015-11-29
Add to Tuple2 overridden 'transform' method which takes BiFunction

tkob:
This PR add 'Tuple2.transform' method which takes BiFunction.

I think this is useful since you can write:

```
tuple.transform((fst, snd) -> ...)
```

instead of:

```
tuple.transform(t -> {
    T fst = t._1();
    U snd = t._2();
    ...
})

### DIFF
--- a/src-gen/main/java/javaslang/Tuple2.java
+++ b/src-gen/main/java/javaslang/Tuple2.java
@@ -135,6 +135,18 @@ public final class Tuple2<T1, T2> implements Tuple, Comparable<Tuple2<T1, T2>>, 
         return f.apply(this);
     }
 
+    /**
+     * Transforms this tuple to an arbitrary object (which may be also a tuple of same or different arity).
+     *
+     * @param f Transformation which takes elements of this tuple and return an object of type U
+     * @param <U> New type
+     * @return An object of type U
+     */
+    public <U> U transform(BiFunction<? super T1, ? super T2, U> f) {
+        Objects.requireNonNull(f, "f is null");
+        return f.apply(_1, _2);
+    }
+
     @Override
     public Seq<?> toSeq() {
         return List.ofAll(_1, _2);
--- a/src-gen/test/java/javaslang/Tuple2Test.java
+++ b/src-gen/test/java/javaslang/Tuple2Test.java
@@ -94,6 +94,13 @@ public class Tuple2Test {
     }
 
     @Test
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by danieldietrich** [comment]
This is a great idea!

I want to generalize it for all tuples. The Generator.scala has to be changed accordingly. I did this yesterday for `Monad.lift`. If you don't mind I will do this for Tuple.

Or you change the Generator.scala because it is your idea. What do you think?

**Comment by danieldietrich** [comment]
I've implemented in in Generator.scala. There is an ambiguity in Tuple1:

```java
// current method
<U> U transform(Function<? super Tuple1<T1>, U> f);

// newly added method - ambiguous after type erasure
<U> U transform(Function<? super T1, ? extends U> f);
```

Which one do we take?

* `Function<? super Tuple1<T1>, U>` is how transform is defined in general
* `Function<T1, U>` is easier to use

...?

ruslansennov:
Maybe we should implement it only for <code>arity() &gt;= 2</code> ?

danieldietrich:
You mean like so?

```java
Tuple0:
    U transform(Supplier<? extends U> f)

Tuple1:
    <U> U transform(Function<? super T1, ? extends U> f)

Tuple2:
    <U> U transform(Function<? super Tuple2<T1, T2>, U> f)
    <U> U transform(BiFunction<? super T1, ? super T2, ? extends U> f)

Tuple3:
    <U> U transform(Function<? super Tuple3<T1, T2, T3>, U> f)
    <U> U transform(Function3<? super T1, ? super T2, ? super T3, ? extends U> f)
...
```

ruslansennov:
yep

danieldietrich:
Ok - we need to go one way, then we will take this one.

danieldietrich:
Or we just leave out the whole `<U> U transform(Function<? super Tuple*, ? extends U> f)` methods. Makes no sense IMO.

danieldietrich:
Even one more time we see that Tuples are the most special types. It seems they follow no rules. Before it was not possible to implement the Value interface. Tuples are no objects on their own, they are defined only by the values they contain.

> Fascinating.

_Mr. Spock_

danieldietrich:
mind-explision

danieldietrich:
Implemented with #887, so this can be closed please.

tkob:
Thank you.

Issue #716, opened 2015-10-23
Add a transform method to all Javaslang types


danieldietrich:
... Functions excluded.

Signature:

```java
<U> U transform(Function<? super SELF_TYPE, ? extends U>)
```

Because Java classes have no 'self type' and generics of function args cannot be overridden, the `transform()` method cannot be defined in a general interface (like `Value`). Instead we have to decide where to put the `tranform()` method.

**Also double-check `Tuple*.transform()`. Which method signature fits best?**

```java
final class Tuple1<T1> {
/*1)*/    public <U> U transform(Function<Tuple1<T1>, U> f) { ... }
/*2)*/    public <U> U transform(Function<? super Tuple1<T1>, U> f) { ... }
/*3)*/    public <U> U transform(Function<? super Tuple1<T1>, ? extends U> f) { ... }
/*4)*/    public <U> U transform(Function<? super Tuple1<? super T1>, ? extends U> f) { ... }
}
```

Currently we have **2)**. Maybe **3)** fits better?

danieldietrich:
Copied from #748:

### Future.sequence()

```java
Stream.from(1)                                   // Stream(1, 2, 3, ...)
        .map(i -> Future.of(() -> getPerson(i))) // Stream(Future<Person>, Future<Person>, ...)
        .transform(Future::sequence)             // Future<Seq<Person>>
```

### List.flatten()

```java
List.of(1, 2, 3)                   // = List(1, 2, 3)
        .map(i -> List.of(i, i))   // = List(List(1, 1), List(2, 2), List(3, 3))
        .transform(List::flatten); // = List(1, 1, 2, 2, 3, 3)

// with...
interface List<T> {
    <T> List<T> flatten(Iterable<? extends Iterable<? extends T>> that) { ... }
}
```

danieldietrich:
Can't handle this until 2.0.0. Will move this to 2.0.1 which will come soon after...

danieldietrich:
We need a self-type `<U> U transform(Function<? super SELF_TYPE, ? extends U>)` in order to make this method inheritance compatible. This is not possible with Java except we add a generic type variable SELF_TYPE to all interfaces, which will not happen!

Therefore we will select the types that give us most value when adding the transform method.

In particular we will add `transform` to these `Traversable`s:

* **Seq:** `<U> U transform(Function<? super Seq<? super T>, ? extends U>)`
* **Set:** `<U> U transform(Function<? super Set<? super T>, ? extends U>)`
[...]

PR #887, opened 2015-11-29
Simplified Tuple*.transform signature

danieldietrich:
Generalizes #886
Towards #716


danieldietrich:
This change was suggested by @tkob.

PR #888, opened 2015-11-29
Added Option.some

danieldietrich:


mvh77:
No no, I think it should be Some.of, not Option.some, + Left.of, Right.of, etc.

danieldietrich:
Yes I know. Will take that into consideration. Have to think about it...

However, This method is needed, too, to be consistent with other types, e.g. `Try`

```java
Try.of(() -> x); // = Success(x) or Failure(...)
Try.success(x); // = Success(x)
Try.failure(x); // = Failure(x)
```

danieldietrich:
Please file an issue for the `of` methods...

PR #912, opened 2015-12-05
Make method to map Map into Seq

v1ctor:
I've added a method to map Map realizations to Seq.
I haven't invented good name for it, but maybe you have any ideas?

### DIFF
--- a/src/main/java/javaslang/collection/Map.java
+++ b/src/main/java/javaslang/collection/Map.java
@@ -39,11 +39,6 @@ public interface Map<K, V> extends Traversable<Tuple2<K, V>>, Function<K, V> {
 
     <U, W> Map<U, W> map(BiFunction<? super K, ? super V, ? extends Tuple2<? extends U, ? extends W>> mapper);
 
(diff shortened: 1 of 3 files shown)


danieldietrich:
Thx!

I don't get why the two Future tests fail now on every build and worked before... I'm currently investigating it.

danieldietrich:
Oh - works again...

src/main/java/javaslang/collection/Map.java
```diff
@@ -39,6 +39,11 @@ default V apply(K key) {
 
     <U, W> Map<U, W> map(BiFunction<? super K, ? super V, ? extends Tuple2<? extends U, ? extends W>> mapper);
 
+    default <U> Seq<U> flatten(BiFunction<K, V, ? extends U> mapper) {
```
danieldietrich: Great! Love this method.

Please rename it to `traverse` (same functionality as `Tree.traverse()`) and move it behind `int size();`.

Thx!

danieldietrich:
I will pull it. The error has nothing to do with the code changed here. Afterwards I will fix it with a separate PR.

Many thx!

PR #914, opened 2015-12-06
Broken travis-ci builds

danieldietrich:
This PR will be used to fix the CI builds.

It wasn't code changes that led to the broken builds. Local builds are passing. Then the cause of failure has to be the CI platform itself.

This is what changed recently:

> This job ran on our new platform for Legacy Precise builds. Please read [our blog post for more information](https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future).


danieldietrich:
The CI build does hang at the FutureTest. This might have two reasons:

* a deadlock happens
* the system runs out of resources when creating new threads. Maybe the fork-join thread pool waits for futures to be returned into the pool before creating new ones.

I will commit some debug output to investigate this on the CI server because we can't reproduce it locally on different machines.

PR #915, opened 2015-12-06
Add isMatch and isSome methods to Option

v1ctor:
I've added some useful methods at Option class. 
Correct me if there are such methods.

### DIFF
--- a/src/main/java/javaslang/control/Option.java
+++ b/src/main/java/javaslang/control/Option.java
@@ -178,6 +178,30 @@ public interface Option<T> extends Value<T> {
     }
 
     /**
+     *
+     * Returns {@code true} if this is a {@code Some} and the value equals the given value.
+     * Otherwise {@code false} is returned.
+     *
+     * @param value A value to check if it equals to Option's content
+     * @return true or false as specified
+     */
+    default boolean isSome(T value) {
+        return isMatch(o -> Objects.equals(o, value));
+    }
+
+    /**
+     *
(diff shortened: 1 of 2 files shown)


danieldietrich:
We already have `exists(predicate)` for all Values (Option, Try, Traversable, ...). This is the same as `isMatch(predicate)`.

This `isSome` looks like an _imperative_ smell to me from the viewpoint of functional programming. What is the use-case?

Instead of

```java
result = opt.isSome(value) ? a : b;
```

it is better to

```java
result = opt.map(...).ifDefined(a, b);
```

Please convince me that isSome is needed :)

danieldietrich:
@v1ctor I'm sure you have a use-case in mind or a code snippet. How would you use it in your application? Then I can see if we include it or express it with existing syntax.

v1ctor:
I do something like that:
```java
void setValue(Object value) {
    if (!this.option.isSome(value)) {
         //Update value in DB
    }
    //Do nothing
}
```

danieldietrich:
We will rename `boolean isSome(T element)` to `boolean contains(T element)` and pull the existing method from `Traversable` into `Value` (internal: `Iterable`).

I've done this with with #917. I hope you don't mind to close this PR. I'm very thankful for your great ideas! Keep on PR'ing!

v1ctor:
Alright! It looks the same :)

danieldietrich:
Thx!

PR #918, opened 2015-12-06
Seq.lastIndexOfSlice() refactored

ruslansennov:
* we shouldn't use recurrence here
* <code>IndexedSeq.tail()</code> is very expensive

### DIFF
--- a/src/main/java/javaslang/collection/IndexedSeq.java
+++ b/src/main/java/javaslang/collection/IndexedSeq.java
@@ -389,30 +389,12 @@ interface IndexedSeqModule {
             }
             final int maxPtr = t.length() - slice.length();
             while (p <= maxPtr) {
-                if(checkPrefix(t, p, slice)) {
+                if(t.startsWith(slice, p)) {
                     return p;
                 }
                 p++;
             }
             return -1;
-        }
-
-        private static <T> boolean checkPrefix(IndexedSeq<T> t, int p, IndexedSeq<T> prefix) {
-            if (prefix.isEmpty()) {
-                return true;
-            } else {
-                final int pLength = prefix.length();
-                final int tLength = t.length();
-                if(tLength < pLength) {
-                    return false;
-                }
-                for (int i = 0; i < pLength; i++) {
-                    if(!java.util.Objects.equals(t.get(i + p), prefix.get(i))) {
-                        return false;
-                    }
-                }
-                return true;
-            }
         }
     }
 }
--- a/src/main/java/javaslang/collection/IndexedSeq.java
+++ b/src/main/java/javaslang/collection/IndexedSeq.java
@@ -367,7 +367,8 @@ interface IndexedSeqModule {
             }
             int p = 0;
             int result = -1;
-            while (t.length() >= p + slice.length()) {
+            final int maxPtr = t.length() - slice.length();
+            while (p <= maxPtr) {
                 int r = findSlice(t, p, slice);
                 if (r < 0) {
                     return result;
@@ -386,7 +387,8 @@ interface IndexedSeqModule {
             if (t.isEmpty()) {
                 return slice.isEmpty() ? 0 : -1;
             }
-            while (t.length() >= p + slice.length()) {
+            final int maxPtr = t.length() - slice.length();
+            while (p <= maxPtr) {
                 if(checkPrefix(t, p, slice)) {
                     return p;
                 }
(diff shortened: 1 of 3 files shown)


src/main/java/javaslang/collection/IndexedSeq.java
```diff
@@ -346,3 +351,66 @@ default boolean startsWith(Iterable<? extends T> that, int offset) {
...
+            if (t.isEmpty()) {
+                return slice.isEmpty() ? 0 : -1;
+            }
+            if (slice.isEmpty()) {
+                int len = t.length();
+                return len < end ? len : end;
+            }
+            int p = 0;
+            int result = -1;
+            while (t.length() >= p + slice.length()) {
```
danieldietrich: Store somewhere

```java
final int length = t.length() - slice.length();
```

and then use

```java
while (p < length) {
    ...
}
```

src/main/java/javaslang/collection/IndexedSeq.java
```diff
@@ -346,3 +351,66 @@ default boolean startsWith(Iterable<? extends T> that, int offset) {
...
+                if(checkPrefix(t, p, slice)) {
+                    return p;
+                }
+                p++;
+            }
+            return -1;
+        }
+
+        private static <T> boolean checkPrefix(IndexedSeq<T> t, int p, IndexedSeq<T> prefix) {
```
danieldietrich: I think we don't need this method. Should be the same as `t.startsWith(prefix, p)`.

src/main/java/javaslang/collection/IndexedSeq.java
```diff
@@ -346,3 +351,66 @@ default boolean startsWith(Iterable<? extends T> that, int offset) {
...
+        static <T> int lastIndexOfSlice(IndexedSeq<T> t, IndexedSeq<T> slice, int end) {
+            if (end < 0) {
+                return -1;
+            }
+            if (t.isEmpty()) {
+                return slice.isEmpty() ? 0 : -1;
+            }
+            if (slice.isEmpty()) {
+                int len = t.length();
+                return len < end ? len : end;
```
danieldietrich: Is result = end when len > end? Why not len? Because the last index of the empty slice is the end of t?

Don't know, just a question...

danieldietrich: Ah - I think I've not understood/overseen `end` - we search until there. It is ok!

src/main/java/javaslang/collection/IndexedSeq.java
```diff
@@ -346,3 +351,66 @@ default boolean startsWith(Iterable<? extends T> that, int offset) {
...
+                    return result;
+                }
+            }
+            return result;
+        }
+
+        private static <T> int findSlice(IndexedSeq<T> t, int p, IndexedSeq<T> slice) {
+            if (t.isEmpty()) {
+                return slice.isEmpty() ? 0 : -1;
+            }
```
danieldietrich: we do not need to check this whole if - we've already done it in lastIndexOfSlice, right?

src/main/java/javaslang/collection/IndexedSeq.java
ruslansennov: Yep :smile:

src/main/java/javaslang/collection/IndexedSeq.java
```diff
@@ -346,3 +351,66 @@ default boolean startsWith(Iterable<? extends T> that, int offset) {
...
+                    result = r;
+                    p = r + 1;
+                } else {
+                    return result;
+                }
+            }
+            return result;
+        }
+
+        private static <T> int findSlice(IndexedSeq<T> t, int p, IndexedSeq<T> slice) {
```
danieldietrich: we need here only two args, p and length.

src/main/java/javaslang/collection/IndexedSeq.java
```diff
@@ -346,3 +351,66 @@ default boolean startsWith(Iterable<? extends T> that, int offset) {
...
+            if (t.isEmpty()) {
+                return slice.isEmpty() ? 0 : -1;
+            }
+            while (t.length() >= p + slice.length()) {
```
danieldietrich: while (p < length) {
    ...
}

src/main/java/javaslang/collection/IndexedSeq.java
ruslansennov: yes, you're right

src/main/java/javaslang/collection/IndexedSeq.java
ruslansennov: hmmm.... but we have <code>t.startsWith(slice, p)</code> inside. We can only move <code>while(...) {...}</code> from <code>findSlice</code> to main method

danieldietrich:
Thx!!

PR #924, opened 2015-12-08
Add ability to wrap a Java future with a JavaSlang future

enelson:
I didn't see this ability in the code, but I have occasions where I want to wrap my Java futures into JavaSlang futures to align with my other JavaSlang code. Please feel free to let me know if you agree with the ability to do this, and if you have feedback on the code itself. Thanks.


enelson:
The wrapping process itself is asynchronous, so it won't block during the conversion.

danieldietrich:
That looks wonderful :-)

Thank you!

Issue #921, opened 2015-12-08
Add factory method *Map.of(key1, val1, key2, val2, ...)


danieldietrich:
Example taken from http://www.javaadvent.com/2015/12/08:

```java
Map<String, Function2<Long, Long, Long>> functions = ImmutableMap.of(
        "add", (a, b) -> a + b,
        "mul", (a, b) -> a * b,
        "div", (a, b) -> a / b,
        "sub", (a, b) -> a - b);
```

bartolom:
Just as a comment: From a developer point of view, it would be nice if your library would use similar patterns to the upcoming Java 9 stuff in JEP 269. I think your factory method Map.of here does this already. Just mentioning, in case, you have not stumbled upon it yet.

http://openjdk.java.net/jeps/269
https://www.youtube.com/watch?v=OJrIMv4dAek "Young Pups: New Collections APIs for Java 9 by Stuart Marks"

danieldietrich:
@bartolom thank you, yes - I've seen it.

Our collections already have static factory methods `of(T)` and `of(T...)`. Map now has `of(Object...)`.

I opened the issue #933 to get sure we do not miss something.

Many thanks!

PR #928, opened 2015-12-09
*Map.of(key1, val1, key2, val2, ...) factory method #921

ruslansennov:


danieldietrich:
YAY! I loooove it! Simple yet great API addition.

Issue #933, opened 2015-12-10
Ensure the collection API aligns nicely to JEP 269


danieldietrich:
See also #921 and

* http://openjdk.java.net/jeps/269
* https://www.youtube.com/watch?v=OJrIMv4dAek "Young Pups: New Collections APIs for Java 9 by Stuart Marks"

danieldietrich:
We do not support `unmodifiableXxx` methods because our collections are immutable by design.

For what I can see, the only methods we currently do not have are `(Linked)HashMap.ofEntries(...)`. Instead we have (example: HashMap):

```java
public static <K, V> HashMap<K, V> ofAll(Tuple2<? extends K, ? extends V>... entries) {
public static <K, V> HashMap<K, V> ofAll(Iterable<? extends Tuple2<? extends K, ? extends V>> entries)
```

We should take into account renaming these methods to

```java
public static <K, V> HashMap<K, V> ofEntries(Tuple2<? extends K, ? extends V>... entries) {
public static <K, V> HashMap<K, V> ofEntries(Iterable<? extends Tuple2<? extends K, ? extends V>> entries)
```

and additionally provide

```java
public static <K, V> HashMap<K, V> ofEntries(java.util.Map.Entry<? extends K, ? extends V>... entries) {
public static <K, V> HashMap<K, V> ofEntries(Iterable<? extends java.util.Map.Entry<? extends K, ? extends V>> entries)
```

ruslansennov:
challenge accepted

danieldietrich:
great :-)

ruslansennov:
It's impossible to create these methods together because of *"both methods have same erasure"*
```java
public static <K, V> HashMap<K, V> ofEntries(Iterable<? extends Tuple2<? extends K, ? extends V>> entries)
public static <K, V> HashMap<K, V> ofEntries(Iterable<? extends java.util.Map.Entry<? extends K, ? extends V>> entries)
```

My proposal is: we should keep our methods as is
```java
public static <K, V> HashMap<K, V> ofAll(Tuple2<? extends K, ? extends V>... entries) {
public static <K, V> HashMap<K, V> ofAll(Iterable<? extends Tuple2<? extends K, ? extends V>> entries)
```
and additionally provide
```java
public static <K, V> HashMap<K, V> ofEntries(java.util.Map.Entry<? extends K, ? extends V>... entries) {
[...]

PR #934, opened 2015-12-10
Map.ofEntries(...) factories #933

ruslansennov:


danieldietrich:
Challenge accomplished :-) That was fast (as always).

PR #949, opened 2015-12-13
Add draw() method for trees

romildo:
This adds the `String draw()` method to the `Tree` interface. It pretty prints a tree, like `Data.Tree.drawTree` from Haskell, in a nice two dimensional way.


danieldietrich:
Very nice addition :) Thx!

Issue #862, opened 2015-11-26
Lack of Match.MatchFunction.When<T,R> then method which could consume matching object


krzysztofik:
It is quite frequent that we want to initiate some action on matching object and not exactly return some value. I know that it is not in agree with functional paradigm, but other way it is impossible to map many switch statements written in java to java.slang pattern matching. Adding methods which could consume matching object, for example	then(java.util.function.Consumer<? super T> consumer)  would solve this gap.

Example:
of(obj).
    whenType(Foo.class).then(out::println)

mariuszs:
:+1:

danieldietrich:
I see the use case.

In fact `Match` is the generalization of an `If`. In other words eventually someone may ask to implement an _else_ case:

```java
Match.of(...).when(...).then(v -> performSideEffects(v)).otherwise(v -> performOtherSideEffects(v));
```

`then` and `otherwise` need the following arg types:

* Consumer
* Runnable

Please not that it will be possible to create:

```java
Match.of(...)
    .when(...).then(function1)
    .when(...).then(function2)
    .when(...).then(consumer) // after this call only otherwise(Consumer) is possible
    .otherwise(consumer); // optional
```

krzysztofik:
It is great that You will include this use case in the upcoming release but Your proposition only partially solve presented (admittedly, poorly described) problem. What I am really talking about is such example scenario:
```
Match.of(msg)
    .whenType(Date.class).then(date -> println(date.getTime()))
    .whenType(String.class).then(str -> println(str.length()))
    .whenType(Number.class).then(num -> println(num.intValue()))
    .otherwise(obj -> println("Unknown " + obj));
```
Similarly 
```
when().then()
```
should be possible.
For more info take a look
https://github.com/nurkiewicz/typeof

danieldietrich:
Thanks for clarification, that is possible. I will do it.

krzysztofik:
Awesome :)

mariuszs:
Wow, great news :)

RobWin:
The question is when do you want to do that when I look at the issue list. Maybe it would be a nice PR or a :santa: gift. :) Looking at @krzysztofik

krzysztofik:
I have reserved nearest weekend for that. I will do my best :)

danieldietrich:
Cool to hear that, I will assist you!

Here come some first informations:

[...]

PR #951, opened 2015-12-14
Add pattern matching consumer and runnable to MatchFunction #862

krzysztofik:


krzysztofik:
It looks that there is a problem with potential ambiguity between overloaded method taking consumer and function :/
http://stackoverflow.com/questions/30584887/java-8-consumer-function-lambda-ambiguity

danieldietrich:
Hi Norbert, your additions look great. Now let's remove the ambiguities.

We will take other method names. I.e. instead of 'then'  we have to choose a different name. Here are some candidates:

* thenDo
* thenRun
* thenExec
* thenPerform
* thenConsume

I would suggest to also rename `then(Runnable)` to underline that it performs a side-effect. Instead of

* thenRun(Runnable)
* thenConsume(Consumer)

I would name both methods

* thenRun(Runnable)
* thenRun(Consumer)

What do you think?

danieldietrich:
There is a big PR by @ruslansennov in the pipeline. I will pull it first. I think it is easier if you merge the changes.

src/main/java/javaslang/control/Match.java
```diff
@@ -182,15 +182,31 @@ private WhenUntyped(Predicate<? super T> predicate) {
 
             public <R> When.Then<R> then(Supplier<? extends R> supplier) {
                 Objects.requireNonNull(supplier, "supplier is null");
-                return then(ignored -> supplier.get());
+                return then(ignored -> {return supplier.get();});
```
danieldietrich: if we removed the ambiguity, this can be reverted

src/main/java/javaslang/control/Match.java
```diff
@@ -182,15 +182,31 @@ private WhenUntyped(Predicate<? super T> predicate) {
...
             }
 
             public <R> When.Then<R> thenThrow(Supplier<? extends RuntimeException> supplier) {
                 Objects.requireNonNull(supplier, "supplier is null");
-                return then(ignored -> {
+                return then((Function<? super T, ? extends R>) ignored -> {
```
danieldietrich: dito

src/main/java/javaslang/control/Match.java
```diff
@@ -214,16 +230,32 @@ private When(Predicate<? super T> predicate, List<Case<R>> cases) {
 
             public Then<R> then(Supplier<? extends R> supplier) {
                 Objects.requireNonNull(supplier, "supplier is null");
-                return then(ignored -> supplier.get());
+                return then(ignored -> {return supplier.get();});
```
danieldietrich: dito

src/main/java/javaslang/control/Match.java
```diff
@@ -214,16 +230,32 @@ private When(Predicate<? super T> predicate, List<Case<R>> cases) {
...
             }
 
             public Then<R> thenThrow(Supplier<? extends RuntimeException> supplier) {
                 Objects.requireNonNull(supplier, "supplier is null");
-                return then(ignored -> {
+                return then((Function<? super T, ? extends R>) ignored -> {
```
danieldietrich: dito

krzysztofik:
Yes, I have exactly the same thought about it. But I was not sure if you would take to consider some additional method name. Your naming proposition is perfect. I will introduce changes tomorrow and of course remove this nasty explicit castings :).
Btw. in issue I described my consideration/problem related to introducing similar changes to MatchMonad. Could you look at it?

BR,
Norbert

danieldietrich:
Very nice, thx for your contribution!

Issue #945, opened 2015-12-12
List/Stream implementations visibility


ruslansennov:
@romildo [says](https://gitter.im/javaslang/javaslang/archives/2015/12/10)

> The best way of constructing a non-empty list given its head h and its tail l is "new List.Cons(h, t)"?

Maybe we should hide <code>Cons</code> and <code>Nil</code> from public?

danieldietrich:
I originally wanted it for pattern matching

```java
Match.of(list)
    .when(List(x, xs)).then(println("head: " + x + ", tail: " + xs))
    .otherwise(println("empty list"));
```

What other benefits could we have?

Ad long there is no decomposition in pattern matching I think we don't need them. Scala had them (for List only?).

danieldietrich:
Are there other types that have inner classes we don't want to expose?

danieldietrich:
I'm not sure if cyclops (@johnmcclean-aol) or derive4j (@jbgi) need access to Cons and Nil.

Does javaslang-jackson currently use them?

ruslansennov:
> Are there other types that have inner classes we don't want to expose?

IMO all implementations of these interfaces should be hidden:
* List
* Stream
* Option (in case if Option.of(null) will return Some(null))
* Lazy (already hidden)

> Does javaslang-jackson currently use them?

Of course no! :)

danieldietrich:
Basically we need the types for pattern matching, therefore they should stay with us.

```java
Supplier<R> s1 = Match.of(option)
    .whenType(List.Cons.class).then(some -> ...)
    .whenType(List.Nil.class).then(nil -> ...);

Supplier<R> s2 = Match.of(_try)
    .whenType(Success.class).then(success -> ...)
    .whenType(Failure.class).then(failure -> ...);
```

We could make the constructors (package) private to make clear how to obtain classes. There have to be factory methods for the subtypes. I began already with that. Example:

```java
<T> Option<T> of(T t);
```

danieldietrich:
* Move impl classes into their interfaces
* Hide constructors of impl classes
* Ensure that special factory methods for impl classes exist (but they return the interface)

ruslansennov:
Fixed by ff434724c11329d5e16e19eb7dbaf4aa90d5c5b7

PR #952, opened 2015-12-14
Inners #945

ruslansennov:


danieldietrich:
Awesome. Looks like a huge amount of work!

Big step towards 2.0.0
lightspeed

PR #954, opened 2015-12-15
add consumer and runnable to MatchMonad pattern matching

krzysztofik:
One open question is if we should add some more meaningful termination operation for monads with consumer/runnable. Currently we have get() which is a little strange in case when we have MatchMonad of Void. I will leave this decision to You.


danieldietrich:
Awesome, thx!

I asked myself the question, too. `get()` is mandatory because MatchMonad extends Value. I think we don't need to add other terminal operations because Consumer and Runnable will only perform side-effects - nothing is returned. It is ok as it is!

Issue #929, opened 2015-12-09
Let Seq, Map, Match, ... extends Function0/Function1 instead of Supplier/Function


danieldietrich:
this will

* enrich the functionality
* increase the interoperability

danieldietrich:
Example:

```java
// uses Function1#memoized()
Function1<Integer, String> cache = Map.of(1, "a", 2, "b", 3, "c").memoized();
```

PR #955, opened 2015-12-15
Function0/Function1 instead of Supplier/Function #929

ruslansennov:


src/main/java/javaslang/collection/Seq.java
```diff
@@ -90,7 +91,9 @@
  * @author Daniel Dietrich
  * @since 1.1.0
  */
-public interface Seq<T> extends Traversable<T>, IntFunction<T> {
+public interface Seq<T> extends Traversable<T>, Function1<Integer, T> {
```
ruslansennov: Here is the question: should we generate <code>IntFunctionN</code>, <code>LongFunctionN</code> etc. ?

danieldietrich: No - with Valhalla we will have `<any T>`. Until then we stay with Function*

danieldietrich:
Perfect.

PR #958, opened 2015-12-16
starting work on Validation

enelson:
This is an early look into my idea of the implementation. You'll notice that I have two unused interfaces: **Apply** and **Applicative**. I was trying to make a more generic version of Validation that could be used for other Applicatives, but with the lack of higher kinded types, it was proving to be a real pain. So I'd love some suggestions from you in that regard. For now, take a look at **Validation** and **ValidationBuilder**. The best examples of how they could be used are in the **ValidationTest** class, particularly in the `shouldBuildUpForSuccess()` and `shouldBuildUpForFailure()` methods. It seems to work fine, and I like the way it's allowing simple chaining of Validations and then applying a function at the end, very similar to how scalaz's implementation works. Again, a rough draft of my initial stab at it. If i'm completely off base from what you were thinking, please let me know. I'm using the local snapshoted version in another project to vet the API and see if it's usable and understandable.

### DIFF
--- a/src/main/java/javaslang/control/Applicative.java
+++ b/src/main/java/javaslang/control/Applicative.java
@@ -1,0 +1,17 @@
+/*     / \____  _    _  ____   ______  / \ ____  __    _ _____
+ *    /  /    \/ \  / \/    \ /  /\__\/  //    \/  \  / /  _  \   Javaslang
+ *  _/  /  /\  \  \/  /  /\  \\__\\  \  //  /\  \ /\\/  \__/  /   Copyright 2014-now Daniel Dietrich
+ * /___/\_/  \_/\____/\_/  \_/\__\/__/___\_/  \_//  \__/_____/    Licensed under the Apache License, Version 2.0
+ */
+package javaslang.control;
+
+public interface Applicative<T> extends Apply<T> {
+
+    Applicative<T> point(T t);
+
+    // Alias for point
+    default Applicative<T> pure(T t) {
+        return point(t);
+    }
+
+}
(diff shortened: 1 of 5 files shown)


enelson:
My **ValidationBuilder** only goes to ValidationBuilder4 right now, but it would go to ValidationBuilder8 to align with Function8.

danieldietrich:
Thank you, I will take a look this evening.

enelson:
Hi @danieldietrich, just wondering if you have any initial feedback or reaction to this initial stab at Validation?

danieldietrich:
Hi Eric, thank you for the prototype and sorry for being so silent - I was very busy with Christmas orga and family activity the last days...

Your solution looks well designed and simple. I like that. I see the scalaz-roots. My impression is that we can simplify the solution, because Java does not have higher-order kinds. Example: point aka pure constructs an instance - but we don't need that in Java. I also omitted it in Javaslang's Monad implementation.

When I read the code I had some more ideas. I will comment the prototype accordingly and hope that I'm able to afford the time soon...

enelson:
No worries Daniel, time with family is far more important. :) I look forward to your input when you get around to it. I see your point with no needing pure and point, I'll remove those and follow more closely to your pattern in other classes. Have a great Christmas and Holidays!

danieldietrich:
oh - you did close it? I wanted to spend time on it after vacation and release 2.0.0...

enelson:
@danieldietrich no, i didn't close it. although I did delete my validation branch and re-add it. I wonder if GitHub automatically closes a pull request if the branch is delete? If so, my bad. Do you need me to do a new PR?

danieldietrich:
Ok :) Yes, please do a new PR.

PR #960, opened 2015-12-17
Seq unused methods removed

ruslansennov:
both methods are overridden in <code>LinearSeq</code> and <code>IndexedSeq</code>


danieldietrich:
Thx!

PR #962, opened 2015-12-18
HAMT is almost ready

ruslansennov:


src/main/java/javaslang/collection/HashArrayMappedTrie.java
```diff
@@ -376,26 +355,49 @@ V value() {
                 if (action == REMOVE) {
                     return filtered;
                 } else {
-                    if (filtered.isEmpty()) {
-                        return new LeafSingleton<>(hash, key, value);
```
danieldietrich: I'm not as deep involved as you - but this catched my eye - do we need this singleton any more here?

ruslansennov: filtered can't be empty, because unfiltered (original LeafList) size >= 2

danieldietrich: ok

PR #970, opened 2015-12-22
RBTree signatures

ruslansennov:
it doesn't compile without these fixes
```java
public void kekeke() {
    RedBlackTree<Integer> tree = Stream.of(1, 2)
            .foldLeft(RedBlackTree.empty(), RedBlackTree::insert);
}
```


danieldietrich:
Ah, great. Will have an I on return types for other types...

PR #971, opened 2015-12-22
Added link to User guide and removed duplicate content.

RobWin:


danieldietrich:
Thanks, that helped :-))

PR #972, opened 2015-12-22
CharSeq refactored; tests added

ruslansennov:


src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -484,7 +494,7 @@ public CharSeq padTo(int length, Character element) {
         for (int i = 0; i < limit; i++) {
             sb.append(element);
         }
-        return sb.length() == 0 ? EMPTY : new CharSeq(sb.toString());
```
ruslansennov: See code above
```java
if (length <= back.length()) {
    return this;
}
```
This ensures that <code>sb.length() == 0</code>

danieldietrich: yes, right

danieldietrich:
thx!

Issue #969, opened 2015-12-22
Add Convertible.toLeft(R right) and Convertible.toRight(L left)


danieldietrich:
Currently we accept only Suppliers, e.g. `toLeft(Supplier<R>)` and `toRight(Supplier<L>)`.

There may be situations where it is unnecessary to create an additional lambda instance.

enelson:
@danieldietrich I'd be happy to take this on if it hasn't been started. Or if so, are there any un-started issues left in 2.0.0 milestone that I could help with?

danieldietrich:
@enelson great, it's yours! I haven't started to work on it.

The method declarations can be found in Monad.java (see Monad.Convertible), which is generated by Generator.scala.

The implementations reside in Value.java.

enelson:
@danieldietrich Thanks! I imagine that my changes for `Monad` will be made in `Generator.scala` instead of directly to `Monad.java`, yes? If so, how do I run the generator to create the changes in Monad.java?

danieldietrich:
@enelson yes, right. You run it with Maven.

* Command line (also IntelliJ IDEA):  `mvn generate-sources`, or better: `mvn clean test`
* Eclipse: Right click project / run / Maven build... then specify goals `generate-sources` or `clean test`

PR #974, opened 2015-12-23
Adding Convertible.toLeft(R right) and Convertible.toRight(L left)

enelson:
Code for issue #969.

### DIFF
--- a/generator/Generator.scala
+++ b/generator/Generator.scala
@@ -330,6 +330,17 @@ def generateMainClasses(): Unit = {
             <R> $EitherType<T, R> toLeft($SupplierType<? extends R> right);
 
             /**
+             * Converts this value to a {@link $EitherType}.
+             *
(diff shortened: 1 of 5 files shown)


ruslansennov:
Could you add tests?

enelson:
@ruslansennov certainly. I didn't see any tests for the existing Supplier versions, so I'll add those too.

enelson:
@ruslansennov and @danieldietrich, I added these tests to `AbstractTraversableTest`, but wondering if I should move those into `AbstractValueTest`?

ruslansennov:
I would add to <code>AbstractValueTest</code>

enelson:
@ruslansennov I agree, didn't feel right where I put it. Committing changes soon.

danieldietrich:
Great, and also thank you for adding the missing tests!

PR #979, opened 2015-12-25
Queue Tests

ruslansennov:


danieldietrich:
many thx!
shows again: untested code is broken code

Issue #767, opened 2015-11-05
Add sequence methods


danieldietrich:
We already have

```java
List<Future<T>> list = ...;
Future<Seq<T>> future = Future.sequence(list);
```

@talios suggested to do this In a similar way for all other `Value` types, also singleton types like Option, Try, Match etc.

danieldietrich:
this is very similar to #716

danieldietrich:
Can't handle this until 2.0.0. Will move this to 2.0.1 which will come soon after...

Example: Lazy.sequence

```java
    /**
     * Reduces many {@code Lazy} values into a single {@code Lazy} by transforming an
     * {@code Iterable<Lazy<? extends T>>} into a {@code Lazy<Seq<T>>}.
     *
     * @param lazyValues An iterable of lazy values.
     * @return A lazy sequence of values.
     */
    public Lazy<Seq<T>> sequence(Iterable<? extends Lazy<? extends T>> lazyValues) {
        Objects.requireNonNull(lazyValues, "lazyValues is null");
        return Lazy.of(() -> Stream.ofAll(lazyValues).map(v -> v.get()));
    }
```

enelson:
Thanks @danieldietrich. I'll start taking a look at this.

danieldietrich:
:+1:

danieldietrich:
We have finished:

* Lazy
* Future
* Option
* Try

We add sequence to:

* Validation

We will leave sequence for the following:

* Either/LeftProjection/RightProjection
* MatchValue
* all collections

PR #981, opened 2015-12-28
Adding 'sequence' to Try and Option

enelson:
Partial fulfillment of #767. Still looking at Either and Match.

### DIFF
--- a/src/main/java/javaslang/control/Option.java
+++ b/src/main/java/javaslang/control/Option.java
@@ -50,15 +50,22 @@ public interface Option<T> extends Value<T> {
      * {@code Iterable<Option<? extends T>>} into a {@code Option<Seq<T>>}. If any of
      * the Options are {@link Option.None}, then this returns {@link Option.None}.
      *
-     * @param values An {@code Iterable} of {@code Option}s.
-     * @param <T> type of the Options.
-     * @return An {@code Option} of a {@link Seq} of results.
+     * @param values An {@code Iterable} of {@code Option}s
+     * @param <T> type of the Options
+     * @return An {@code Option} of a {@link Seq} of results.\
+     * @throws NullPointerException if {@code values} is null
      */
     static <T> Option<Seq<T>> sequence(Iterable<? extends Option<? extends T>> values) {
-        final Option<Seq<T>> zero = Option.of(List.empty());
-        final BiFunction<Option<Seq<T>>, Option<? extends T>, Option<Seq<T>>> f =
-                (result, option) -> result.flatMap(sequence -> option.map(sequence::append));
-        return Iterator.ofAll(values).foldLeft(zero, f);
+        Objects.requireNonNull(values, "values is null");
+        List<T> list = List.empty();
+        for (Option<? extends T> value : values) {
+            if (value.isEmpty()) {
+                return Option.none();
+            }
+            list = list.prepend(value.get());
+        }
+
+        return Option.some(list.reverse());
     }
 
     /**
--- a/src/main/java/javaslang/control/Option.java
+++ b/src/main/java/javaslang/control/Option.java
@@ -7,15 +7,15 @@ package javaslang.control;
 
 import javaslang.Value;
 import javaslang.collection.Iterator;
+import javaslang.collection.List;
+import javaslang.collection.Seq;
+import javaslang.collection.Stream;
 
 import java.io.Serializable;
 import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.function.Consumer;
-import java.util.function.Function;
-import java.util.function.Predicate;
-import java.util.function.Supplier;
+import java.util.function.*;
 
 /**
  * Replacement for {@link java.util.Optional}.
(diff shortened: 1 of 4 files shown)


danieldietrich:
Hi Eric, great, thank you! I've already looked into it and will add some comments.

src/main/java/javaslang/control/Option.java
```diff
@@ -46,6 +46,22 @@
...
     /**
+     * Reduces many {@code Option}s into a single {@code Option} by transforming an
+     * {@code Iterable<Option<? extends T>>} into a {@code Option<Seq<T>>}. If any of
+     * the Options are {@link Option.None}, then this returns {@link Option.None}.
+     *
+     * @param values An {@code Iterable} of {@code Option}s.
+     * @param <T> type of the Options.
+     * @return An {@code Option} of a {@link Seq} of results.
+     */
+    static <T> Option<Seq<T>> sequence(Iterable<? extends Option<? extends T>> values) {
```
danieldietrich: We provide 'user friendly' messages on NullPointerException, here:

```java
```

src/main/java/javaslang/control/Option.java
```diff
@@ -46,6 +46,22 @@
     }
 
     /**
+     * Reduces many {@code Option}s into a single {@code Option} by transforming an
+     * {@code Iterable<Option<? extends T>>} into a {@code Option<Seq<T>>}. If any of
+     * the Options are {@link Option.None}, then this returns {@link Option.None}.
+     *
+     * @param values An {@code Iterable} of {@code Option}s.
+     * @param <T> type of the Options.
+     * @return An {@code Option} of a {@link Seq} of results.
```
danieldietrich: Accordingly

```java
* @throws NullPointerException if {@code values} is null
```

src/main/java/javaslang/control/Option.java
```diff
@@ -46,6 +46,22 @@
...
+     */
+    static <T> Option<Seq<T>> sequence(Iterable<? extends Option<? extends T>> values) {
+        final Option<Seq<T>> zero = Option.of(List.empty());
+        final BiFunction<Option<Seq<T>>, Option<? extends T>, Option<Seq<T>>> f =
+                (result, option) -> result.flatMap(sequence -> option.map(sequence::append));
+        return Iterator.ofAll(values).foldLeft(zero, f);
```
danieldietrich: foldLeft is the most concise solution here. But we cannot shortcut the loop if we visit the first None (see #779).

I would suggest to optimize the internal implementation similar to this:

```java
List<T> list = List.empty();
for (T value : values) {
    if (value.isEmpty) {
        return Option.none();
    }
    list = list.prepend(value.get());
}
return Option.some(list.reverse());
```

I know, from a functional perspective this code looks ugly. But it does perform better on the non-happy path.

src/main/java/javaslang/control/Option.java
```diff
@@ -46,6 +46,22 @@
...
+     * the Options are {@link Option.None}, then this returns {@link Option.None}.
+     *
+     * @param values An {@code Iterable} of {@code Option}s.
+     * @param <T> type of the Options.
+     * @return An {@code Option} of a {@link Seq} of results.
+     */
+    static <T> Option<Seq<T>> sequence(Iterable<? extends Option<? extends T>> values) {
+        final Option<Seq<T>> zero = Option.of(List.empty());
+        final BiFunction<Option<Seq<T>>, Option<? extends T>, Option<Seq<T>>> f =
+                (result, option) -> result.flatMap(sequence -> option.map(sequence::append));
```
danieldietrich: Btw - `sequence::append` internally does a `sequence.reverse().prepend(t).reverse()` for every element `t`. Generally use `prepend` instead for every element and finally a `reverse()`.

src/main/java/javaslang/control/Try.java
```diff
@@ -59,6 +62,22 @@
...
     /**
+     * Reduces many {@code Try}s into a single {@code Try} by transforming an
+     * {@code Iterable<Try<? extends T>>} into a {@code Try<Seq<T>>}. If any of
+     * the {@code Try}s are {@link Try.Failure}, then this returns a {@link Try.Failure}.
+     *
+     * @param values An {@link Iterable} of {@code Try}s.
+     * @param <T> type of the Trys.
+     * @return A {@code Try} of a {@link Seq} of results.
+     */
+    static <T> Try<Seq<T>> sequence(Iterable<? extends Try<? extends T>> values) {
```
danieldietrich: Also null-check and `@throws` in javadoc

src/main/java/javaslang/control/Try.java
```diff
@@ -59,6 +62,22 @@
...
+        final Try<Seq<T>> zero = Try.of(List::empty);
+        final BiFunction<Try<Seq<T>>, Try<? extends T>, Try<Seq<T>>> f =
+                (result, next) -> result.flatMap(sequence -> next.map(sequence::append));
+        return Iterator.ofAll(values).foldLeft(zero, f);
```
danieldietrich: I would rewrite it accordingly to `Option.sequence` as commented above

enelson:
@danieldietrich thanks for all the great feedback! I'll get this all done today.

src/main/java/javaslang/control/Option.java
enelson: Good to know. So out of curiosity, why do you use `seq::append` in the `sequence()` method for `Future`?

danieldietrich:
Awesome, I'm thankful for you help!

src/main/java/javaslang/control/Option.java
danieldietrich: It depends on the collection type. The Seq used in Future.sequence is a Stream, which has a lazy append operation taking O(1). It works like a caching iterator. Here we use List as Seq. It has O(1) prepend and O(n) append.

You find a good overview on the bottom of [this page](http://javaslang.com/javadoc/snapshot/javaslang/collection/package-summary.html).

enelson:
@danieldietrich Ok, changes committed for review.

danieldietrich:
Great, thank you!

Issue #659, opened 2015-10-02
Add factory method 'fill' and 'tabulate' to collections


danieldietrich:
The `fill` create a collection of a specific size and fills the values using a Supplier or Function:

```java
static <T> Xxx<T> fill(int size, Supplier<? extends T> supplier);

static <T> Xxx<T> tabulate(int size, Function<Integer, ? extends T> function);
```

Examples:

```java
List.fill(5, () -> "test");

Map.tabulate(5, i -> new Entry<>(i, "test");
```

_29.12.15: Renamed `fill(int, Function)` to 'tabulate(int, Function)'_

simonecarriero:
Hi,
about the static helper method in Collections we were talking about in #984 I think there are two options.
The first one is to return an array of elements, to be passed to the of(T...) method
```
static <T> T[] tabulate(int n, Function<? super Integer, ? extends T> f) {
    n = n < 0 ? 0 : n;
    @SuppressWarnings("unchecked")
    T[] elements = (T[]) new Object[n];
    for (int i = 0; i < n; i++) {
        elements[i] = f.apply(i);
    }
    return elements;
}
```

The second one is to return an iterator, to be passed to the ofAll(Iterable) method
```
static <T> Iterable<T> tabulate(int n, Function<? super Integer, ? extends T> f) {
    Objects.requireNonNull(f, "f is null");
    return new AbstractIterator<T>() {

        int i = 0;

        @Override
        public boolean hasNext() {
            return i < n;
        }

        @Override
        protected T getNext() {
            return f.apply(i++);
        }
    };
}
```

The bad thing of the first approach is that we have to iterate through the elements twice, one inside Collections.tabulate and one inside of(T...). With the second approach we iterate only once, inside ofAll(Iterator).

Thoughts?

danieldietrich:
Hi,

thank you! Looks great. I made some slight changes similar to the collection conversion methods in Value, which look like this:

```java
@Override
default Set<T> toSet() {
    return ValueModule.toTraversable(this, HashSet.empty(), HashSet::of, HashSet::ofAll);
}

@Override
default Stack<T> toStack() {
    return ValueModule.toTraversable(this, Stack.empty(), Stack::of, Stack::ofAll);
}

@Override
default Stream<T> toStream() {
    return ValueModule.toTraversable(this, Stream.empty(), Stream::of, Stream::ofAll);
}
```

where

```java
static <T extends Traversable<V>, V> T toTraversable(Value<V> value, T empty,
[...]

PR #984, opened 2015-12-29
Add factory method 'fill' to Seq hierarchy

simonecarriero:
WIP of #659, 'fill' should be added also to Map and Set hierarchies. Feedbacks are welcome :+1:

### DIFF
--- a/src/main/java/javaslang/collection/Array.java
+++ b/src/main/java/javaslang/collection/Array.java
@@ -213,7 +213,7 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
      * @return An Array consisting of elements {@code f(0),f(1), ..., f(n - 1)}
      * @throws NullPointerException if {@code n} or {@code f} are null
      */
-    public static <T> Array<T> fill(Integer n, Function<Integer, ? extends T> f) {
+    public static <T> Array<T> tabulate(Integer n, Function<Integer, ? extends T> f) {
         Objects.requireNonNull(n, "n is null");
         Objects.requireNonNull(f, "f is null");
         int nOrZero = java.lang.Math.max(n, 0);
@@ -235,7 +235,7 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
      * @throws NullPointerException if {@code n} or {@code s} are null
      */
     public static <T> Array<T> fill(Integer n, Supplier<? extends T> s) {
-        return fill(n, anything -> s.get());
+        return tabulate(n, anything -> s.get());
     }
 
 
--- a/src/main/java/javaslang/collection/Array.java
+++ b/src/main/java/javaslang/collection/Array.java
@@ -213,7 +213,7 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
      * @return An Array consisting of elements {@code f(0),f(1), ..., f(n - 1)}
      * @throws NullPointerException if {@code f} is null
      */
-    public static <T> Array<T> tabulate(int n, Function<Integer, ? extends T> f) {
+    public static <T> Array<T> tabulate(int n, Function<? super Integer, ? extends T> f) {
         Objects.requireNonNull(f, "f is null");
         int nOrZero = java.lang.Math.max(n, 0);
         @SuppressWarnings("unchecked")
--- a/src/main/java/javaslang/collection/Array.java
+++ b/src/main/java/javaslang/collection/Array.java
@@ -211,10 +211,9 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
      * @param n The number of elements in the Array
      * @param f The Function computing element values
      * @return An Array consisting of elements {@code f(0),f(1), ..., f(n - 1)}
-     * @throws NullPointerException if {@code n} or {@code f} are null
+     * @throws NullPointerException if {@code f} is null
      */
-    public static <T> Array<T> tabulate(Integer n, Function<Integer, ? extends T> f) {
-        Objects.requireNonNull(n, "n is null");
+    public static <T> Array<T> tabulate(int n, Function<Integer, ? extends T> f) {
         Objects.requireNonNull(f, "f is null");
         int nOrZero = java.lang.Math.max(n, 0);
         @SuppressWarnings("unchecked")
@@ -232,9 +231,9 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
      * @param n The number of elements in the Array
      * @param s The Supplier computing element values
      * @return An Array of size {@code n}, where each element contains the result supplied by {@code s}.
-     * @throws NullPointerException if {@code n} or {@code s} are null
+     * @throws NullPointerException if {@code s} is null
      */
-    public static <T> Array<T> fill(Integer n, Supplier<? extends T> s) {
+    public static <T> Array<T> fill(int n, Supplier<? extends T> s) {
         return tabulate(n, anything -> s.get());
     }
 
--- a/src/main/java/javaslang/collection/Array.java
+++ b/src/main/java/javaslang/collection/Array.java
@@ -216,8 +216,7 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
     public static <T> Array<T> tabulate(int n, Function<? super Integer, ? extends T> f) {
         Objects.requireNonNull(f, "f is null");
         int nOrZero = java.lang.Math.max(n, 0);
-        @SuppressWarnings("unchecked")
-        T[] elements = (T[]) new Object[nOrZero];
+        Object[] elements = new Object[nOrZero];
         for (int i = 0; i < n; i++) {
             elements[i] = f.apply(i);
         }
--- a/src/main/java/javaslang/collection/Array.java
+++ b/src/main/java/javaslang/collection/Array.java
@@ -215,8 +215,8 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
      */
     public static <T> Array<T> tabulate(int n, Function<? super Integer, ? extends T> f) {
         Objects.requireNonNull(f, "f is null");
-        int nOrZero = java.lang.Math.max(n, 0);
-        Object[] elements = new Object[nOrZero];
+        n = n < 0 ? 0 : n;
+        Object[] elements = new Object[n];
         for (int i = 0; i < n; i++) {
             elements[i] = f.apply(i);
         }
--- a/src/main/java/javaslang/collection/Array.java
+++ b/src/main/java/javaslang/collection/Array.java
@@ -203,6 +203,42 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
         return Array.ofAll(Iterator.ofAll(array));
     }
 
+    /**
+     * Returns an Array containing {@code n} values of a given Function {@code f}
+     * over a range of integer values from 0 to {@code n - 1}.
+     *
+     * @param <T> Component type of the Array
+     * @param n The number of elements in the Array
+     * @param f The Function computing element values
+     * @return An Array consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code n} or {@code f} are null
+     */
+    public static <T> Array<T> fill(Integer n, Function<Integer, ? extends T> f) {
+        Objects.requireNonNull(n, "n is null");
+        Objects.requireNonNull(f, "f is null");
+        int nOrZero = java.lang.Math.max(n, 0);
+        @SuppressWarnings("unchecked")
+        T[] elements = (T[]) new Object[nOrZero];
+        for (int i = 0; i < n; i++) {
+            elements[i] = f.apply(i);
+        }
+        return wrap(elements);
+    }
+
+    /**
+     * Returns an Array containing {@code n} values supplied by a given Supplier {@code s}.
+     *
+     * @param <T> Component type of the Array
+     * @param n The number of elements in the Array
+     * @param s The Supplier computing element values
+     * @return An Array of size {@code n}, where each element contains the result supplied by {@code s}.
+     * @throws NullPointerException if {@code n} or {@code s} are null
+     */
+    public static <T> Array<T> fill(Integer n, Supplier<? extends T> s) {
+        return fill(n, anything -> s.get());
+    }
+
+
     public static Array<Character> range(char from, char toExclusive) {
         return Array.ofAll(Iterator.range(from, toExclusive));
     }
--- a/src/main/java/javaslang/collection/CharSeq.java
+++ b/src/main/java/javaslang/collection/CharSeq.java
@@ -129,7 +129,7 @@ public final class CharSeq implements CharSequence, IndexedSeq<Character>, Seria
      * @return A CharSeq consisting of elements {@code f(0),f(1), ..., f(n - 1)}
      * @throws NullPointerException if {@code f} is null
      */
-    public static CharSeq tabulate(int n, Function<Integer, Character> f) {
+    public static CharSeq tabulate(int n, Function<? super Integer, Character> f) {
         Objects.requireNonNull(f, "f is null");
         StringBuilder sb = new StringBuilder();
         for (int i = 0; i < n; i++) {
--- a/src/main/java/javaslang/collection/CharSeq.java
+++ b/src/main/java/javaslang/collection/CharSeq.java
@@ -129,7 +129,7 @@ public final class CharSeq implements CharSequence, IndexedSeq<Character>, Seria
      * @return A CharSeq consisting of elements {@code f(0),f(1), ..., f(n - 1)}
      * @throws NullPointerException if {@code f} is null
      */
-    public static CharSeq tabulate(int n, Function<? super Integer, Character> f) {
+    public static CharSeq tabulate(int n, Function<? super Integer, ? extends Character> f) {
         Objects.requireNonNull(f, "f is null");
         StringBuilder sb = new StringBuilder();
         for (int i = 0; i < n; i++) {
@@ -146,7 +146,7 @@ public final class CharSeq implements CharSequence, IndexedSeq<Character>, Seria
      * @return A CharSeq of size {@code n}, where each element contains the result supplied by {@code s}.
      * @throws NullPointerException if {@code s} is null
      */
-    public static CharSeq fill(int n, Supplier<Character> s) {
+    public static CharSeq fill(int n, Supplier<? extends Character> s) {
         return tabulate(n, anything -> s.get());
     }
 
--- a/src/main/java/javaslang/collection/CharSeq.java
+++ b/src/main/java/javaslang/collection/CharSeq.java
@@ -121,6 +121,37 @@ public final class CharSeq implements CharSequence, IndexedSeq<Character>, Seria
     }
 
     /**
+     * Returns a CharSeq containing {@code n} values of a given Function {@code f}
+     * over a range of integer values from 0 to {@code n - 1}.
+     *
+     * @param n The number of elements in the CharSeq
+     * @param f The Function computing element values
+     * @return A CharSeq consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code n} or {@code f} are null
+     */
+    public static CharSeq fill(Integer n, Function<Integer, Character> f) {
+        Objects.requireNonNull(n, "n is null");
+        Objects.requireNonNull(f, "f is null");
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < n; i++) {
+            sb.append(f.apply(i));
+        }
+        return new CharSeq(sb.toString());
+    }
+
+    /**
+     * Returns a CharSeq containing {@code n} values supplied by a given Supplier {@code s}.
+     *
+     * @param n The number of elements in the CharSeq
+     * @param s The Supplier computing element values
+     * @return A CharSeq of size {@code n}, where each element contains the result supplied by {@code s}.
+     * @throws NullPointerException if {@code n} or {@code s} are null
+     */
+    public static CharSeq fill(Integer n, Supplier<Character> s) {
+        return fill(n, anything -> s.get());
+    }
+
+    /**
      * Creates a CharSeq starting from character {@code from}, extending to character {@code toExclusive - 1}.
      * <p>
      * Examples:
--- a/src/main/java/javaslang/collection/CharSeq.java
+++ b/src/main/java/javaslang/collection/CharSeq.java
@@ -127,10 +127,9 @@ public final class CharSeq implements CharSequence, IndexedSeq<Character>, Seria
      * @param n The number of elements in the CharSeq
      * @param f The Function computing element values
      * @return A CharSeq consisting of elements {@code f(0),f(1), ..., f(n - 1)}
-     * @throws NullPointerException if {@code n} or {@code f} are null
+     * @throws NullPointerException if {@code f} is null
      */
-    public static CharSeq tabulate(Integer n, Function<Integer, Character> f) {
-        Objects.requireNonNull(n, "n is null");
+    public static CharSeq tabulate(int n, Function<Integer, Character> f) {
         Objects.requireNonNull(f, "f is null");
         StringBuilder sb = new StringBuilder();
         for (int i = 0; i < n; i++) {
@@ -145,9 +144,9 @@ public final class CharSeq implements CharSequence, IndexedSeq<Character>, Seria
      * @param n The number of elements in the CharSeq
      * @param s The Supplier computing element values
      * @return A CharSeq of size {@code n}, where each element contains the result supplied by {@code s}.
-     * @throws NullPointerException if {@code n} or {@code s} are null
+     * @throws NullPointerException if {@code s} is null
      */
-    public static CharSeq fill(Integer n, Supplier<Character> s) {
+    public static CharSeq fill(int n, Supplier<Character> s) {
         return tabulate(n, anything -> s.get());
     }
 
--- a/src/main/java/javaslang/collection/CharSeq.java
+++ b/src/main/java/javaslang/collection/CharSeq.java
@@ -129,7 +129,7 @@ public final class CharSeq implements CharSequence, IndexedSeq<Character>, Seria
      * @return A CharSeq consisting of elements {@code f(0),f(1), ..., f(n - 1)}
      * @throws NullPointerException if {@code n} or {@code f} are null
      */
-    public static CharSeq fill(Integer n, Function<Integer, Character> f) {
+    public static CharSeq tabulate(Integer n, Function<Integer, Character> f) {
         Objects.requireNonNull(n, "n is null");
         Objects.requireNonNull(f, "f is null");
         StringBuilder sb = new StringBuilder();
@@ -148,7 +148,7 @@ public final class CharSeq implements CharSequence, IndexedSeq<Character>, Seria
      * @throws NullPointerException if {@code n} or {@code s} are null
      */
     public static CharSeq fill(Integer n, Supplier<Character> s) {
-        return fill(n, anything -> s.get());
+        return tabulate(n, anything -> s.get());
     }
 
     /**
--- a/src/main/java/javaslang/collection/CharSeq.java
+++ b/src/main/java/javaslang/collection/CharSeq.java
@@ -135,7 +135,7 @@ public final class CharSeq implements CharSequence, IndexedSeq<Character>, Seria
         for (int i = 0; i < n; i++) {
             sb.append(f.apply(i));
         }
-        return new CharSeq(sb.toString());
+        return of(sb);
     }
 
     /**
--- a/src/main/java/javaslang/collection/List.java
+++ b/src/main/java/javaslang/collection/List.java
@@ -288,7 +288,7 @@ public interface List<T> extends LinearSeq<T>, Stack<T> {
      * @return A List consisting of elements {@code f(0),f(1), ..., f(n - 1)}
      * @throws NullPointerException if {@code n} or {@code f} are null
      */
-    static <T> List<T> fill(Integer n, Function<Integer, ? extends T> f) {
+    static <T> List<T> tabulate(Integer n, Function<Integer, ? extends T> f) {
         Objects.requireNonNull(n, "n is null");
         Objects.requireNonNull(f, "f is null");
         int nOrZero = java.lang.Math.max(n, 0);
@@ -310,7 +310,7 @@ public interface List<T> extends LinearSeq<T>, Stack<T> {
      * @throws NullPointerException if {@code n} or {@code s} are null
      */
     static <T> List<T> fill(Integer n, Supplier<? extends T> s) {
-        return List.fill(n, anything -> s.get());
+        return List.tabulate(n, anything -> s.get());
     }
 
     static List<Character> range(char from, char toExclusive) {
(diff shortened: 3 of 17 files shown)


src/main/java/javaslang/collection/Array.java
```diff
@@ -203,6 +203,42 @@ private Array(Object[] back) {
...
+     * @param f The Function computing element values
+     * @return An Array consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code n} or {@code f} are null
+     */
+    public static <T> Array<T> fill(Integer n, Function<Integer, ? extends T> f) {
+        Objects.requireNonNull(n, "n is null");
+        Objects.requireNonNull(f, "f is null");
+        int nOrZero = java.lang.Math.max(n, 0);
+        @SuppressWarnings("unchecked")
+        T[] elements = (T[]) new Object[nOrZero];
```
ruslansennov: <code>final Object[] = ...</code>

danieldietrich:
Great thx!

src/main/java/javaslang/collection/Array.java
```diff
@@ -203,6 +203,42 @@ private Array(Object[] back) {
...
+     * Returns an Array containing {@code n} values of a given Function {@code f}
+     * over a range of integer values from 0 to {@code n - 1}.
+     *
+     * @param <T> Component type of the Array
+     * @param n The number of elements in the Array
+     * @param f The Function computing element values
+     * @return An Array consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code n} or {@code f} are null
+     */
+    public static <T> Array<T> fill(Integer n, Function<Integer, ? extends T> f) {
```
danieldietrich: * `n` should be `int`, then there's no need to check for null below
* We should  use `Function<? super Integer, ? extends T>` in order to allow functions that operate on `Number`

src/main/java/javaslang/collection/Stream.java
```diff
@@ -233,6 +233,41 @@ public T next() {
...
+    static <T> Stream<T> fill(Integer n, Function<Integer, ? extends T> f) {
+        Objects.requireNonNull(n, "n is null");
+        Objects.requireNonNull(f, "f is null");
+        int nOrZero = java.lang.Math.max(n, 0);
+        @SuppressWarnings("unchecked")
+        T[] elements = (T[]) new Object[nOrZero];
+        for (int i = 0; i < n; i++) {
+            elements[i] = f.apply(i);
+        }
+        return Stream.of(elements);
```
ruslansennov: it can be lazy?

src/main/java/javaslang/collection/Array.java
```diff
@@ -203,6 +203,42 @@ private Array(Object[] back) {
...
+        Objects.requireNonNull(n, "n is null");
```
danieldietrich: line can be deleted, see above

src/main/java/javaslang/collection/Stream.java
danieldietrich: yes, Stream methods should be lazy, s.th. similar to

```java
return Stream.ofAll(Iterator.fill(n, f));
```

src/main/java/javaslang/collection/Array.java
danieldietrich: I just realized that Scala has a different name for `fill(int, Function)`. It is `tabulate`:

```scala
scala> List.tabulate(5)(n => n * n)
res0: List[Int] = List(0, 1, 4, 9, 16)
```

We should rename the methods accordingly because we align to Scala syntax.

Note: The name `fill(int, Supplier)` remains.

src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -121,6 +121,37 @@ public static CharSeq ofAll(Iterable<? extends Character> elements) {
...
+     * @throws NullPointerException if {@code n} or {@code f} are null
+     */
+    public static CharSeq fill(Integer n, Function<Integer, Character> f) {
+        Objects.requireNonNull(n, "n is null");
+        Objects.requireNonNull(f, "f is null");
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < n; i++) {
+            sb.append(f.apply(i));
+        }
+        return new CharSeq(sb.toString());
```
ruslansennov: <code>return of(sb);</code>

we should return static instance <code>EMPTY</code> if <code>n == 0</code>

src/main/java/javaslang/collection/Array.java
```diff
@@ -203,6 +203,42 @@ private Array(Object[] back) {
...
+    /**
+     * Returns an Array containing {@code n} values supplied by a given Supplier {@code s}.
+     *
+     * @param <T> Component type of the Array
+     * @param n The number of elements in the Array
+     * @param s The Supplier computing element values
+     * @return An Array of size {@code n}, where each element contains the result supplied by {@code s}.
+     * @throws NullPointerException if {@code n} or {@code s} are null
+     */
+    public static <T> Array<T> fill(Integer n, Supplier<? extends T> s) {
```
danieldietrich: also `int` instead of `Integer`

src/main/java/javaslang/collection/Vector.java
```diff
@@ -100,6 +100,39 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
+     * @throws NullPointerException if {@code n} or {@code f} are null
+     */
+    public static <T> Vector<T> fill(Integer n, Function<Integer, ? extends T> f) {
+        Objects.requireNonNull(n, "n is null");
+        Objects.requireNonNull(f, "f is null");
+        HashArrayMappedTrie<Integer, T> trie = HashArrayMappedTrie.empty();
+        for (int i = 0; i <n; i++) {
+            trie = trie.put(trie.size(), f.apply(i));
+        }
+        return new Vector<>(trie);
```
ruslansennov: <code>return trie.isEmpty() ? empty() : new Vector<>(trie);</code>

src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -121,6 +121,37 @@ public static CharSeq ofAll(Iterable<? extends Character> elements) {
...
     /**
+     * Returns a CharSeq containing {@code n} values of a given Function {@code f}
+     * over a range of integer values from 0 to {@code n - 1}.
+     *
+     * @param n The number of elements in the CharSeq
+     * @param f The Function computing element values
+     * @return A CharSeq consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code n} or {@code f} are null
+     */
+    public static CharSeq fill(Integer n, Function<Integer, Character> f) {
```
danieldietrich: Also `int n, Function<? super Integer, ...`

src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -121,6 +121,37 @@ public static CharSeq ofAll(Iterable<? extends Character> elements) {
...
+        Objects.requireNonNull(n, "n is null");
```
danieldietrich: can be removed

src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -121,6 +121,37 @@ public static CharSeq ofAll(Iterable<? extends Character> elements) {
...
+
+    /**
+     * Returns a CharSeq containing {@code n} values supplied by a given Supplier {@code s}.
+     *
+     * @param n The number of elements in the CharSeq
+     * @param s The Supplier computing element values
+     * @return A CharSeq of size {@code n}, where each element contains the result supplied by {@code s}.
+     * @throws NullPointerException if {@code n} or {@code s} are null
+     */
+    public static CharSeq fill(Integer n, Supplier<Character> s) {
```
danieldietrich: Also `int n`.

We use `Supplier<? extends Character>` for the sake of completeness.

src/main/java/javaslang/collection/List.java
```diff
@@ -278,6 +278,41 @@
...
+     * Returns a List containing {@code n} values of a given Function {@code f}
+     * over a range of integer values from 0 to {@code n - 1}.
+     *
+     * @param <T> Component type of the List
+     * @param n The number of elements in the List
+     * @param f The Function computing element values
+     * @return A List consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code n} or {@code f} are null
+     */
+    static <T> List<T> fill(Integer n, Function<Integer, ? extends T> f) {
```
danieldietrich: Also `int n, Function<? super Integer, ...`

src/main/java/javaslang/collection/List.java
```diff
@@ -278,6 +278,41 @@
...
+        Objects.requireNonNull(n, "n is null");
```
danieldietrich: line can be removed

src/main/java/javaslang/collection/List.java
```diff
@@ -278,6 +278,41 @@
...
+    /**
+     * Returns a List containing {@code n} values supplied by a given Supplier {@code s}.
+     *
+     * @param <T> Component type of the List
+     * @param n The number of elements in the List
+     * @param s The Supplier computing element values
+     * @return A List of size {@code n}, where each element contains the result supplied by {@code s}.
+     * @throws NullPointerException if {@code n} or {@code s} are null
+     */
+    static <T> List<T> fill(Integer n, Supplier<? extends T> s) {
```
danieldietrich: also `int n`

src/main/java/javaslang/collection/Queue.java
```diff
@@ -238,6 +238,35 @@ private Queue(List<T> front, List<T> rear) {
...
+     * Returns a Queue containing {@code n} values of a given Function {@code f}
+     * over a range of integer values from 0 to {@code n - 1}.
+     *
+     * @param <T> Component type of the Queue
+     * @param n The number of elements in the Queue
+     * @param f The Function computing element values
+     * @return A Queue consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code n} or {@code f} are null
+     */
+    public static <T> Queue<T> fill(Integer n, Function<Integer, ? extends T> f) {
```
danieldietrich: Also `int n, Function<? super Integer, ...`

src/main/java/javaslang/collection/Queue.java
```diff
@@ -238,6 +238,35 @@ private Queue(List<T> front, List<T> rear) {
...
+        Objects.requireNonNull(n, "n is null");
```
danieldietrich: line can be removed

src/main/java/javaslang/collection/Queue.java
```diff
@@ -238,6 +238,35 @@ private Queue(List<T> front, List<T> rear) {
...
+    /**
+     * Returns a Queue containing {@code n} values supplied by a given Supplier {@code s}.
+     *
+     * @param <T> Component type of the Queue
+     * @param n The number of elements in the Queue
+     * @param s The Supplier computing element values
+     * @return An Queue of size {@code n}, where each element contains the result supplied by {@code s}.
+     * @throws NullPointerException if {@code n} or {@code s} are null
+     */
+    public static <T> Queue<T> fill(Integer n, Supplier<? extends T> s) {
```
danieldietrich: int n

src/main/java/javaslang/collection/Seq.java
```diff
@@ -118,6 +118,35 @@
...
+     * Returns a Seq containing {@code n} values of a given Function {@code f}
+     * over a range of integer values from 0 to {@code n - 1}.
+     *
+     * @param <T> Component type of the Seq
+     * @param n The number of elements in the Seq
+     * @param f The Function computing element values
+     * @return A Seq consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code n} or {@code f} are null
+     */
+    static <T> Seq<T> fill(Integer n, Function<Integer, ? extends T> f) {
```
danieldietrich: Also `int n, Function<? super Integer, ...`

src/main/java/javaslang/collection/Seq.java
```diff
@@ -118,6 +118,35 @@
...
+        Objects.requireNonNull(n, "n is null");
```
danieldietrich: can be removed

src/main/java/javaslang/collection/Seq.java
```diff
@@ -118,6 +118,35 @@
...
+    /**
+     * Returns a Seq containing {@code n} values supplied by a given Supplier {@code s}.
+     *
+     * @param <T> Component type of the Seq
+     * @param n The number of elements in the Seq
+     * @param s The Supplier computing element values
+     * @return A Seq of size {@code n}, where each element contains the result supplied by {@code s}.
+     * @throws NullPointerException if {@code n} or {@code s} are null
+     */
+    static <T> Seq<T> fill(Integer n, Supplier<? extends T> s) {
```
danieldietrich: int n

src/main/java/javaslang/collection/Stack.java
```diff
@@ -202,6 +202,35 @@
...
+     * Returns a Stack containing {@code n} values of a given Function {@code f}
+     * over a range of integer values from 0 to {@code n - 1}.
+     *
+     * @param <T> Component type of the Stack
+     * @param n The number of elements in the Stack
+     * @param f The Function computing element values
+     * @return A Stack consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code n} or {@code f} are null
+     */
+    static <T> Stack<T> fill(Integer n, Function<Integer, ? extends T> f) {
```
danieldietrich: Also `int n, Function<? super Integer, ...`

src/main/java/javaslang/collection/Stack.java
```diff
@@ -202,6 +202,35 @@
...
+        Objects.requireNonNull(n, "n is null");
```
danieldietrich: line can be removed

src/main/java/javaslang/collection/Stack.java
```diff
@@ -202,6 +202,35 @@
...
+    /**
+     * Returns a Stack containing {@code n} values supplied by a given Supplier {@code s}.
+     *
+     * @param <T> Component type of the Stack
+     * @param n The number of elements in the Stack
+     * @param s The Supplier computing element values
+     * @return A Stack of size {@code n}, where each element contains the result supplied by {@code s}.
+     * @throws NullPointerException if {@code n} or {@code s} are null
+     */
+    static <T> Stack<T> fill(Integer n, Supplier<? extends T> s) {
```
danieldietrich: int n

src/main/java/javaslang/collection/Stream.java
```diff
@@ -233,6 +233,41 @@ public T next() {
...
+     * Returns a Stream containing {@code n} values of a given Function {@code f}
+     * over a range of integer values from 0 to {@code n - 1}.
+     *
+     * @param <T> Component type of the Stream
+     * @param n The number of elements in the Stream
+     * @param f The Function computing element values
+     * @return A Stream consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code n} or {@code f} are null
+     */
+    static <T> Stream<T> fill(Integer n, Function<Integer, ? extends T> f) {
```
danieldietrich: Also `int n, Function<? super Integer, ...`

src/main/java/javaslang/collection/Stream.java
```diff
@@ -233,6 +233,41 @@ public T next() {
...
+        Objects.requireNonNull(n, "n is null");
```
danieldietrich: can be removed

src/main/java/javaslang/collection/Stream.java
```diff
@@ -233,6 +233,41 @@ public T next() {
...
+    /**
+     * Returns a Stream containing {@code n} values supplied by a given Supplier {@code s}.
+     *
+     * @param <T> Component type of the Stream
+     * @param n The number of elements in the Stream
+     * @param s The Supplier computing element values
+     * @return A Stream of size {@code n}, where each element contains the result supplied by {@code s}.
+     * @throws NullPointerException if {@code n} or {@code s} are null
+     */
+    static <T> Stream<T> fill(Integer n, Supplier<? extends T> s) {
```
danieldietrich: int n

src/main/java/javaslang/collection/Vector.java
```diff
@@ -100,6 +100,39 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
+     * Returns a Vector containing {@code n} values of a given Function {@code f}
+     * over a range of integer values from 0 to {@code n - 1}.
+     *
+     * @param <T> Component type of the Vector
+     * @param n The number of elements in the Vector
+     * @param f The Function computing element values
+     * @return A Vector consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code n} or {@code f} are null
+     */
+    public static <T> Vector<T> fill(Integer n, Function<Integer, ? extends T> f) {
```
danieldietrich: dito

src/main/java/javaslang/collection/Vector.java
```diff
@@ -100,6 +100,39 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
+        Objects.requireNonNull(n, "n is null");
```
danieldietrich: dito

src/main/java/javaslang/collection/Vector.java
```diff
@@ -100,6 +100,39 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
+    /**
+     * Returns a Vector containing {@code n} values supplied by a given Supplier {@code s}.
+     *
+     * @param <T> Component type of the Vector
+     * @param n The number of elements in the Vector
+     * @param s The Supplier computing element values
+     * @return A Vector of size {@code n}, where each element contains the result supplied by {@code s}.
+     * @throws NullPointerException if {@code n} or {@code s} are null
+     */
+    public static <T> Vector<T> fill(Integer n, Supplier<? extends T> s) {
```
danieldietrich: dito

src/test/java/javaslang/collection/AbstractSeqTest.java
```diff
@@ -59,6 +62,10 @@
     @Override
     abstract protected Seq<Short> ofAll(short[] array);
 
+    abstract protected <T> Seq<T> fill(Integer n, Function<Integer, ? extends T> f);
```
danieldietrich: dito

src/test/java/javaslang/collection/AbstractSeqTest.java
```diff
@@ -59,6 +62,10 @@
...
+
+    abstract protected <T> Seq<T> fill(Integer n, Supplier<? extends T> s);
```
danieldietrich: dito

src/test/java/javaslang/collection/ArrayTest.java
```diff
@@ -80,6 +82,16 @@
     }
 
     @Override
+    protected <T> Array<T> fill(Integer n, Function<Integer, ? extends T> f) {
```
danieldietrich: dito

src/test/java/javaslang/collection/ArrayTest.java
```diff
@@ -80,6 +82,16 @@
...
+        return Array.fill(n, f);
+    }
+
+    @Override
+    protected <T> Array<T> fill(Integer n, Supplier<? extends T> s) {
```
danieldietrich: dito

src/test/java/javaslang/collection/ListTest.java
```diff
@@ -87,6 +89,16 @@
     }
 
     @Override
+    protected <T> List<T> fill(Integer n, Function<Integer, ? extends T> f) {
```
danieldietrich: dito

src/test/java/javaslang/collection/ListTest.java
```diff
@@ -87,6 +89,16 @@
...
+        return List.fill(n, f);
+    }
+
+    @Override
+    protected <T> List<T> fill(Integer n, Supplier<? extends T> s) {
```
danieldietrich: dito

src/test/java/javaslang/collection/QueueTest.java
```diff
@@ -85,6 +87,16 @@
     }
 
     @Override
+    protected <T> Queue<T> fill(Integer n, Function<Integer, ? extends T> f) {
```
danieldietrich: dito

src/test/java/javaslang/collection/QueueTest.java
```diff
@@ -85,6 +87,16 @@
...
+        return Queue.fill(n, f);
+    }
+
+    @Override
+    protected <T> Queue<T> fill(Integer n, Supplier<? extends T> s) {
```
danieldietrich: dito

src/test/java/javaslang/collection/StackTest.java
```diff
@@ -80,6 +82,16 @@
     }
 
     @Override
+    protected <T> Stack<T> fill(Integer n, Function<Integer, ? extends T> f) {
```
danieldietrich: dito

src/test/java/javaslang/collection/StackTest.java
```diff
@@ -80,6 +82,16 @@
...
+        return Stack.fill(n, f);
+    }
+
+    @Override
+    protected <T> Stack<T> fill(Integer n, Supplier<? extends T> s) {
```
danieldietrich: dito

src/main/java/javaslang/collection/Vector.java
simonecarriero: why this? empty() is implemented as new Vector<>(HashArrayMappedTrie.empty())

src/test/java/javaslang/collection/StreamTest.java
```diff
@@ -86,6 +88,16 @@
     }
 
     @Override
+    protected <T> Stream<T> fill(Integer n, Function<Integer, ? extends T> f) {
```
danieldietrich: dito

src/test/java/javaslang/collection/StreamTest.java
```diff
@@ -86,6 +88,16 @@
...
+        return Stream.fill(n, f);
+    }
+
+    @Override
+    protected <T> Stream<T> fill(Integer n, Supplier<? extends T> s) {
```
danieldietrich: dito

src/test/java/javaslang/collection/VectorTest.java
```diff
@@ -81,6 +83,16 @@
     }
 
     @Override
+    protected <T> Vector<T> fill(Integer n, Function<Integer, ? extends T> f) {
```
danieldietrich: dito

src/test/java/javaslang/collection/VectorTest.java
```diff
@@ -81,6 +83,16 @@
...
+        return Vector.fill(n, f);
+    }
+
+    @Override
+    protected <T> Vector<T> fill(Integer n, Supplier<? extends T> s) {
```
danieldietrich: dito

src/main/java/javaslang/collection/Vector.java
danieldietrich: It returns the singleton instance:

```java
public static <T> Vector<T> empty() {
    return (Vector<T>) EMPTY;
}
```

src/main/java/javaslang/collection/Array.java
```diff
@@ -203,6 +203,41 @@ private Array(Object[] back) {
...
+     * @param <T> Component type of the Array
+     * @param n The number of elements in the Array
+     * @param f The Function computing element values
+     * @return An Array consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code f} is null
+     */
+    public static <T> Array<T> tabulate(int n, Function<Integer, ? extends T> f) {
+        Objects.requireNonNull(f, "f is null");
+        int nOrZero = java.lang.Math.max(n, 0);
+        @SuppressWarnings("unchecked")
```
ruslansennov: and this annotation can be removed

simonecarriero: I get an unchecked cast warning without this annotation (required: T[], found: java.lang.Object[])

ruslansennov: <code>Array::wrap()</code> accept <code>Object[]</code> as parameter, so you can do this:
```java
Object[] elements = new Object[nOrZero];
```

src/main/java/javaslang/collection/Vector.java
simonecarriero: Ok, got it. So is ofAll(Iterable) broken with respect to this semantic? The following test is red
```@Test
public void t() {
    Object[] c = new Object[0];
    Vector<Object> v = Vector.ofAll(Array.of(c));
    assertThat(v).isSameAs(Vector.empty());
}
```

ruslansennov: Yes, <code>Array::ofAll(Iterable<? extends T> elements)</code> is broken :)

src/main/java/javaslang/collection/List.java
```diff
@@ -278,6 +278,40 @@
...
+     */
+    static <T> List<T> tabulate(int n, Function<? super Integer, ? extends T> f) {
+        Objects.requireNonNull(f, "f is null");
+        int nOrZero = java.lang.Math.max(n, 0);
+        @SuppressWarnings("unchecked")
+        T[] elements = (T[]) new Object[nOrZero];
+        for (int i = 0; i < n; i++) {
+            elements[i] = f.apply(i);
+        }
+        return List.of(elements);
```
danieldietrich: Here we can optimize like this:

```java
int nOrZero = Math.max(n, 0);
List<T> list = List.empty();
for (int i = nOrZero - 1; i >= 0; i--) {
    list = list.prepend(f.apply(i));
}
return list;
```

src/main/java/javaslang/collection/List.java
```diff
@@ -278,6 +278,40 @@
...
+     * @param f The Function computing element values
+     * @return A List consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code f} is null
+     */
+    static <T> List<T> tabulate(int n, Function<? super Integer, ? extends T> f) {
+        Objects.requireNonNull(f, "f is null");
+        int nOrZero = java.lang.Math.max(n, 0);
+        @SuppressWarnings("unchecked")
+        T[] elements = (T[]) new Object[nOrZero];
+        for (int i = 0; i < n; i++) {
```
danieldietrich: oh, we use `n` instead of nOrZero. Please double-check all other places.

Ruslan would overwrite the given method arg like this: `n = Math.max(n, 0);` ;-)

src/main/java/javaslang/collection/Seq.java
```diff
@@ -118,6 +118,34 @@
...
+     * @param n The number of elements in the Seq
+     * @param s The Supplier computing element values
+     * @return A Seq of size {@code n}, where each element contains the result supplied by {@code s}.
+     * @throws NullPointerException if {@code s} is null
+     */
+    static <T> Seq<T> fill(int n, Supplier<? extends T> s) {
+        return tabulate(n, anything -> s.get());
+    }
+
+    /**
```
danieldietrich: We should remove all static methods from Seq because it is not obvious to users which implementation is selected. If we want a Seq we have to choose it directly according to the underlying use-case. This gives the user more control about what happens.

(Note: Also the existing Seq.ofAll, I will move it in another PR to the internal class Collections)

src/main/java/javaslang/collection/List.java
ruslansennov:
```java
n = n < 0 ? 0 : n;
```
:smile:

danieldietrich: :smile: :smile:

src/main/java/javaslang/collection/Stream.java
danieldietrich: Please make Stream.tabulate lazy: `return Stream.ofAll(Iterator.tabulate(n, f));`

You may add these methods to `javaslang.collection.Iterator`:

```java
    static <T> Iterator<T> fill(int n, Supplier<? extends T> s) {
        return tabulate(n, anything -> s.get());
    }

    static <T> Iterator<T> tabulate(int n, Function<? super Integer, ? extends T> f) {

            int i = 0;

            @Override
            public boolean hasNext() {
                return i < n;
            }

            @Override
            protected T getNext() {
                return f.apply(i++);
            }
        };
    }
```

src/main/java/javaslang/collection/List.java
simonecarriero: It's OK to use n it the loop condition, if it's negative the condition is always false. The nOrZero is necessary only to create the array, to avoid a NegativeArraySizeException.
Anyway, I didn't reassigned the parameter because I'm used to get a checkstyle error in that case :smile: but in this case it's pretty clear, I'll go with n = n < 0 ? 0 : n :+1:

src/main/java/javaslang/collection/Stream.java
simonecarriero: Sure, I'll do it

danieldietrich: Thx!

src/main/java/javaslang/collection/List.java
danieldietrich: @simonecarriero you don't need to go with it. Please replace the method with the optimization below.

src/main/java/javaslang/collection/List.java
simonecarriero: I avoided this on purpose, my point is that doing in this way we call f in the reverse order, from f(n-1) to f(0). In case of pure functions this is fine, but if f is a weird stateful function, we get some inconsistencies between different implementations of Seq.
For example, given the following function
```
java.util.LinkedList<Integer> ints = new java.util.LinkedList<>(Arrays.asList(0, 1, 2));
Function<Integer, Integer> f = i -> ints.remove();
```
Array.fill(3, f) returns Array(0, 1, 2) but List.fill(3, f) returns List(2, 1, 0)

danieldietrich: Note: In the solution above we assume that application of f is stateless. We apply it in reverse order of i (n-1..0) instead of (0..n-1). I hope this does not produce some unexpected behavior because we cannot say if f is really stateless at runtime...

danieldietrich: @simonecarriero You're right - please do it as you suggested.

simonecarriero:
Iterator.tabulate and Iterator.fill are not covered by tests yet. Once i implement tabulate and fill for Maps and Sets I can move tests up in the hierarchy (from AbstractSeqTest to AbstractTraversableTest) so also Iterator will be covered

danieldietrich:
Looks great, thx!
Now that we have Iterator.fill and tabulate, Set and Map should be straight forward (similar to Stream.tabulate)

simonecarriero:
Yes but...at this point I could also refactor all the other tabulate (List, Vector, Array, ...) to just call XXX.ofAll(Iterator.tabulate(n,f)).
What do you think?

danieldietrich:
I think a static helper method in [Collections]() fits better. Ruslan did s.th. similar in ValueModule.toTraversable: https://github.com/javaslang/javaslang/pull/988/files
This will reduce duplicate code.
The Iterator.tabulate method should be used only in the lazy case.

simonecarriero:
Ok I'll take a look! But you said to use Iterator.tabulate for Maps and Sets. In that case it's not lazy right?

danieldietrich:
Yes, it is not lazy. If we create a helper method In the Collections class it would be better to use that, also for Maps and Sets. For SortedMap and SortedSet we need additional fill and tabulate methods having a Comparator argument, you will see it.

Issue #803, opened 2015-11-14
Add Validation Applicative Functor


danieldietrich:
See

* [learning Scalaz — Validation](http://eed3si9n.com/learning-scalaz/Validation.html)
* [Scalaz Validation](http://docs.typelevel.org/api/scalaz/nightly/index.html#scalaz.Validation)
* [A Monad in Practicality: First-Class Failures](http://robotlolita.me/2013/12/08/a-monad-in-practicality-first-class-failures.html)

Note: As [John stated](https://gitter.im/javaslang/javaslang?at=5640848fc712fe074e4e9eac), a list of errors can also be collected with the `sequence` function (which will be rolled out in the broad with #767) by turning a `List<Either>` to `Either<List>`.

danieldietrich:
@enelson The `Validation` type should be located in `javaslang.control`. I would use `Try` as reference impl (some function args will be re-visited soon with #865).

We should keep an eye on the functional interfaces Try.Checked*. We use them also in Future and maybe in Validation. There was a recent discussion in #865 if it would make sense to move them outside of Try. But I'm not convinced yet.

enelson:
@danieldietrich Thanks for the info. I'll do some initial work in another branch and see what you think about it. I want to make sure its something that feels right for this project based on your ideas and goals.

danieldietrich:
:+1:

danieldietrich:
We need to take care that Validation is no monad, it is an applicative functor.

See also http://robotlolita.me/2013/12/08/a-monad-in-practicality-first-class-failures.html

enelson:
Yep, exactly what I was going for. If it was a Monad, it would short circuit after the first error, and we want an aggregation of all errors.

danieldietrich:
Great, I will clear the way (see #947)

enelson:
Thanks @danieldietrich.

enelson:
@danieldietrich I would assume that you'd like to keep your current pattern of putting classes with multiple permutations (ie. Function0, Function1, etc) in src-gen, instead of embedding each one as static classes in one Java file? I'm working on a ValidationBuilder class that will have up to ValidationBuilder8 (coinciding with Function8).

danieldietrich:
@enelson Yes, we generate redundant code. However, it better to first write a stable 'prototype' which later is moved to the generator (because writing generator code is not really fun) :-)

[...]

PR #986, opened 2015-12-29
starting work on Validation

enelson:


danieldietrich:
Thanks! I have some initial thoughts on it - comments come in a minute

enelson:
👍

src/main/java/javaslang/control/Applicative.java
```diff
@@ -0,0 +1,17 @@
+/*     / \____  _    _  ____   ______  / \ ____  __    _ _____
+ *    /  /    \/ \  / \/    \ /  /\__\/  //    \/  \  / /  _  \   Javaslang
+ *  _/  /  /\  \  \/  /  /\  \\__\\  \  //  /\  \ /\\/  \__/  /   Copyright 2014-now Daniel Dietrich
+ * /___/\_/  \_/\____/\_/  \_/\__\/__/___\_/  \_//  \__/_____/    Licensed under the Apache License, Version 2.0
+ */
+package javaslang.control;
+
+public interface Applicative<T> extends Apply<T> {
```
danieldietrich: `point` and `pure` are not needed, if the `ap` method is abstract. Instead implementations will use the actual constructor.

Then we do not need to separate Applicative and Apply. I suggest to merge them and name the type Applicative.

src/main/java/javaslang/control/Apply.java
```diff
@@ -0,0 +1,15 @@
...
+ * /___/\_/  \_/\____/\_/  \_/\__\/__/___\_/  \_//  \__/_____/    Licensed under the Apache License, Version 2.0
+ */
+package javaslang.control;
+
+import javaslang.Function1;
+import javaslang.algebra.Functor;
+
+public interface Apply<T> extends Functor<T> {
+
+    <U> Apply<U> ap(Apply<Function1<T,U>> f);
```
danieldietrich: In Javaslang we choose readable methods. Otherwise our code gets too cryptic. Let's call it `apply` instead of `ap`.

Instead of Function1 we use Function in the case of method parameter type (Function1 extends Function).

As you mentioned already, here we have the same problem as with Monad.flatMap because of the lack of higher-kinded types / type constructors. We circumvented this in Monad.flatMap by having a Function arg that returns Iterator.

Here I would give this signature a try:

```java
<U> Applicative<U> apply(Applicative<? extends Function<? super T, ? extends U>> f);
```

Question: How will implementations deal with an `f` that has a different Applicative type than the present?

Alternative: We could add a _recursive_ self-type `Applicative<SELF extends Applicative<SELF, ?>, T>` which is not a good idea.

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,253 @@
...
+ * @param <E> Value type in the case of failure.
+ * @param <T> Value type in the case of success.
+ * @author Eric Nelson
+ * @since 2.0.1
+ * @see <a href="http://eed3si9n.com/learning-scalaz/Validation.html">Validation</a>
+ */
+public interface Validation<E,T> extends Functor<T> {
+
+    static <E,T> Validation<E,T> success(T value) {
+        return new Success<>(value);
```
danieldietrich: Note: We recently refactored Try implementations and made Success and Failure inner types of Try to clear the way for Validation.Success and Validation.Failure.

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    static <E,T> Validation<E,T> success(Supplier<T> supplier) {
```
danieldietrich: Supplier<? extends T>

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,253 @@
...
+        return new Success<>(supplier.get());
+    }
+
+    static <E,T> Validation<E,T> failure(E error) {
+        return new Failure<>(error);
+    }
+
+    static <E,T> Validation<E,T> failure(Supplier<E> supplier) {
```
danieldietrich: `Supplier<? extends E>`

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,253 @@
...
+        return new Failure<>(supplier.get());
+    }
+
+    boolean isSuccess();
+
+    boolean isFailure();
+
+    Object get();
+
+    void foreach(Consumer<T> f);
```
danieldietrich: We use `forEach` instead of `foreach`. We will get it automatically when implementing `Value`. However, before implementing common Javaslang interfaces we should create a working _spike_ in order to keep the complexity low. We don't care for now.

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,253 @@
...
+    @Override
+    boolean equals(Object o);
+
+    @Override
+    int hashCode();
+
+    @Override
+    String toString();
+
+    default <U> U fold(Function<? super E,? extends U> fail, Function<? super T,? extends U> success) {
```
danieldietrich: Note to myself: Should be added also to `Either<L, R>`.

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,253 @@
...
+        } else {
+            return failure(() -> {
+                List<E> e  = ((Failure<List<E>,Function1<T,U>>) vv).get();
+                E  e2 = ((Failure<E,T>) this).get();
+                return e.append(e2);
+            });
+        }
+    }
+
+    default <U> ValidationBuilder<E,T,U> bld(Validation<E,U> v) {
```
danieldietrich: Without having dived into it, my feeling says that maybe there is a way to compose Applicatives without the need for builders... Need to think about it...

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,253 @@
...
+        if(isFailure()) {
+            E v = ((Failure<E,T>) this).get();
+            return Validation.failure(f.apply(v));
+        } else {
+            T v = ((Success<E,T>) this).get();
+            return Validation.success(v);
+        }
+    }
+
+    default <U> Validation<List<E>,U> ap(Validation<List<E>,Function1<T,U>> vv) {
```
danieldietrich: We need to override Applicative.apply and relax the signature. Impl needs to be figured out then.

src/main/java/javaslang/control/Applicative.java
enelson: Well, `ap` (or `apply`) is not related to `pure` or `point`. `Pure` take a value `T` and lifts it into an `Applicative` context. Whereas `apply` takes a value already in such context, and a function also in an `Applicative` context, and applies the function to the value, returning its result still in the context. 

So I believe we need both as they perform different functions. As to whether we merge `Apply` into `Applicative` is your call. Both scalaz and cats (https://github.com/non/cats) keep the two separate. It seems cleaner to keep the conventions that both libraries use, and that would be my vote, but you tell me what you'd like to do and I'll do it. Here are links to the 2 traits in both libraries so you can take a look and decide.

Apply:
https://github.com/non/cats/blob/master/core/src/main/scala/cats/Apply.scala
https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Apply.scala

Applicative:
https://github.com/non/cats/blob/master/core/src/main/scala/cats/Applicative.scala
https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Applicative.scala

danieldietrich: Thank you, I will read a bit to get a clear picture.

enelson: You know, the more I play with this, the more I think I'm on the wrong path. `Apply` and `Applicative` are both generic traits over higher kinded types: `Apply[F[_]]`. I can't get the types to work right in Java since you can't do this. Do you think that you'll have any other `Applicative` types in Javaslang? Because if not, then it might make sense to abandon my approach with `Apply` and `Applicative` and just keep all of this behavior in `Validation`. What are your thoughts?

danieldietrich: Yes, I know, Java's type system is frustrating sometimes. On the other hand its simplicity is also its strength.

According to the Scalaz docs, `Apply` is ""

```scala
```

enelson: Thanks for your feedback Daniel. I'm not saying that Javaslang should be just like other libraries, it was just a suggestion based on other examples. But I'm totally fine with your ideas on this. I see your point that `pure` and `point` are just constructors. 

Here's where I'm really stuck though, and why in my previous post I suggested possibly doing away with the `Applicative` interface altogether. I wasn't just complaining about Java's type system, I know it's strengths and weaknesses. I was simply pointing out that because of its lack of higher kinded types, I'm not sure how to make the `Applicative` interface work for different kinds of applicative types. I hope I can make sense in what I'm trying to say, please let me know if I'm not.

Looking at scalaz, they have `trait Apply[F[_]]`. And then ap is defined as `def ap[A,B](fa: => F[A])(f: => F[A => B]): F[B]`. `Apply`'s type parameter describes `Validation` (or any other implementation of Apply), not the type held by `Validation`. 

We can't model this in Java. Currently my interface is defined as `public interface Applicative<T>`, and apply as `<U> Applicative<U> apply(Applicative<? extends Function<? super T,? extends U>> f)`. But this doesn't work. Looking at `Validation`, it's defined as `public interface Validation<E,T> extends Applicative<Validation<E,T>>`. That means that the overridden apply method would look like `<U> Applicative<U> apply(Applicative<? extends Function<? super Validation<E, T>, ? extends U>> f)`. But this isn't right. I need it to take a function of `Function<T,U>`, not `Function<Validation<E,T>, U>`. This is all messed up because of lack of higher kinded types. 

I may be totally missing the mark here, which is why I'm asking for your input. I'm not seeing how to make the `Applicative` interface generic enough for any kind of type that wants to implement it. Which is why I was suggesting that if we don't really think we'll add any other type to Javaslang that would implement `Applicative`, we just get rid of it complete and pull all of our applicative login into `Validation`. I really hope I'm making sense here. :) Thanks for your patience.

enelson: If I change the `Validation` definition to `public interface Validation<E,T> extends Applicative<T>`, then my apply looks like this `<U> Applicative<U> apply(Applicative<? extends Function<? super T, ? extends U>> f)`. Which is what I was looking for. But this isn't right because we're saying that the type `T` is the `Applicative`, which it is not. `Validation` is the `Applicative`.

danieldietrich: This is the first shot using my infamous `Kind` type constructor fake. It is a direct translation of Scalaz's Validation Applicative (more or less). The list of errors is handled internally by the Validation, i.e. we only make the type E explicit and most probably there will be a method `getFailures()` or something similar which returns a `List<E>` if `isFailure() == true`.

```java
// trait Apply[F[_]] extends Functor[F]
public interface Applicative<F extends Kind<F, ?>, A> extends Functor<A> {

    // def ap[A,B](fa: => F[A])(f: => F[A => B]): F[B]
    <B> Applicative<F, B> ap(Kind<F, ? extends Function<? super A, ? extends B>> f);

    @Override
    <B> Applicative<F, B> map(Function<? super A, ? extends B> f);
}

// type constructor: TYPE<T> is split in TYPE<?> and T
interface Kind<TYPE extends Kind<TYPE, ?>, T> {
}

interface Validation<E, A> extends Kind<Validation<E, ?>, A> {

    // def ap[EE >: E, B](x: => Validation[EE, A => B])(implicit E: Semigroup[EE]): Validation[EE, B]
    <B> Validation<E, B> ap(Validation<E, ? extends Function<? super A, ? extends B>> f);

    <B> Validation<E, B> map(Function<? super A, ? extends B> f);

    // implicit def ValidationApplicative[L: Semigroup]: Applicative[Validation[L, ?]]
    final class ValidationApplicative<E, A> implements Applicative<Validation<E, ?>, A> {

        final Validation<E, A> fa;

        public ValidationApplicative(Validation<E, A> fa) {
            this.fa = fa;
        }

        // def ap[A, B](fa: => Validation[L, A])(f: => Validation[L, A => B]) = fa ap f
        @Override
        public <B> ValidationApplicative<E, B> ap(Kind<Validation<E, ?>, ? extends Function<? super A, ? extends B>> f) {
            return new ValidationApplicative<>(fa.ap((Validation<E, ? extends Function<? super A, ? extends B>>) (Object) f));
        }

        // override def map[A, B](fa: Validation[L, A])(f: A => B) = fa map f
        @Override
        public <B> Applicative<Validation<E, ?>, B> map(Function<? super A, ? extends B> f) {
            return new ValidationApplicative<>(fa.map(f));
        }
    }
}
```

Next we could get rid of the class ValidationApplicative because Java has no implicits. However, as you already said, I don't if it is worth to make all this effort just to mark Validation as Applicative.

danieldietrich: This is the shortened version I talked about (got rid of implicit workaround class):

```java
// trait Apply[F[_]] extends Functor[F]
public interface Applicative<F extends Kind<F, ?>, A> extends Functor<A> {

    // def ap[A,B](fa: => F[A])(f: => F[A => B]): F[B]
    <B> Applicative<F, B> ap(Kind<F, ? extends Function<? super A, ? extends B>> f);

    @Override
    <B> Applicative<F, B> map(Function<? super A, ? extends B> f);
}

// type constructor: TYPE<T> is split in TYPE<?> and T
interface Kind<TYPE extends Kind<TYPE, ?>, T> {
}

interface Validation<E, A> extends Kind<Validation<E, ?>, A>, Applicative<Validation<E, ?>, A> {

    // def ap[EE >: E, B](x: => Validation[EE, A => B])(implicit E: Semigroup[EE]): Validation[EE, B]
    <B> Validation<E, B> ap(Validation<E, ? extends Function<? super A, ? extends B>> f);

    @Override
    <B> Validation<E, B> map(Function<? super A, ? extends B> f);

    // def ap[A, B](fa: => Validation[L, A])(f: => Validation[L, A => B]) = fa ap f
    @Override
    default <B> Validation<E, B> ap(Kind<Validation<E, ?>, ? extends Function<? super A, ? extends B>> f) {
        return ap((Validation<E, ? extends Function<? super A, ? extends B>>) (Object) f);
    }
}
```

enelson: Thank your for all of this great input, and spending some time with me on this. Very interesting approach. I'm plugged in this shortened code to play with it, but the result isn't coming out quite right. The signature for the overridden `ap()` method isn't what you wrote above, but `<B> Applicative<Validation<E, ?>, B> ap(Kind<Validation<E, ?>, ? extends Function<? super T, ? extends B>> f)`. If I force it to `<B> Validation<E, B> ap(Validation<E, ? extends Function<? super T, ? extends B>> f)`, that signature works, but doesn't work as the overridden  `Applicative` method `ap()`.

I'm still struggling with whether this is all worth it. This is only worth all this work if we feel we could create other Applicative types. If not, then we should go the simpler route for now and keep all our implemented code in `Validation`, which I already have a working version of. I'm curious about your thoughts on all of this and if the complexity is worth it? Thanks so much for your help on this Daniel.

enelson: This is a shortened version of my implementation of your code. The `@Override` method is the only one that satisfies the `Applicative` method.

```
public interface Validation<E,T> extends Kind<Validation<E,?>, T>, Applicative<Validation<E, ?>, T> {

    @Override
    default <B> Applicative<Validation<E, ?>, B> ap(Kind<Validation<E, ?>, ? extends Function<? super T, ? extends B>> f) {
        return null;
    }

    default <U> Validation<E, U> ap(Validation<E, ? extends Function<? super T, ? extends U>> v) {
        return null;
    }

    default <U> Validation<E, U> map(Function<? super T, ? extends U> f) {
        return null;
    }
    
}
```

src/main/java/javaslang/control/Validation.java
enelson: I'd be curious as to your ideas without the builder. The reason I went with the builder is that it encapsulates all the logic for combining multiple `Validation`s and takes it out of `Validation.java`. This seems cleaner. I started messing around with combining validations without a builder, and it got messy. Each time you combine 2 validations, you need to somehow combine their success values. To do this, you either need to supply a combiner function each time, or combine them according to Semigroup rules. But using semigroup doesn't make much sense since it limits how you can combine values. Plus, you won't always have like types in each `Success`, so Semigroup won't work. The builder allows you to accumulate all `Success`s or `Failure`s and then apply all the success values to a function provided after all accumulation has occurred. Doing this inside of `Validation` itself would require some intermediate accumulator in side each `Validation` instance.

enelson: Still assuming we use the Builder, I have an alternative way of accumulation for consideration. Some might not like the pattern:

``` java
Validation<List<String>,TestValidation> result  = v1.combine(v2).apply(TestValidation::new);
Validation<List<String>,TestValidation> result2 = v1.combine(v2).combine(v3).apply(TestValidation::new);
Validation<List<String>,TestValidation> result3 = v1.combine(v2).combine(v4).apply(TestValidation::new);
Validation<List<String>,String> result4 = v1.combine(v2).combine(v3).apply((p1, p2, p3) -> p1+":"+p2+":"+p3.orElse("none"));
```

So I've toyed around with this as an alternative way (we could leave both ways in to give choice on accumulation style):

``` java
Validation<List<String>,TestValidation> result = Validation.map2(v1, v2).apply(TestValidation::new);
Validation<List<String>,TestValidation> result2 = Validation.map3(v1, v2, v3).apply(TestValidation::new);
Validation<List<String>,String> result3 = Validation.map3(v1, v2, v3).apply((p1, p2, p3) -> p1+":"+p2+":"+p3.orElse("none"));
```

src/main/java/javaslang/control/Applicative.java
danieldietrich: Thanks for all your efforts, Eric. The progress is great. We will fix the last details.

> The signature for the overridden ap() method isn't what you wrote above, but <B> Applicative<Validation<E, ?>, B> ap(Kind<Validation<E, ?>, ? extends Function<? super T, ? extends B>> f).

I can't reproduce it. When I copy-paste your code into IntelliJ IDEA (15.0.2 on a Mac with jdk 1.8.0_66) _and_ change the return type of the overridden `ap` method to `Validation<E, B>` it does compile fine (within IDE and with Maven on the command line), especially the `@Override` does also compile. (Note: `map` also may have a `@Override`)

Maybe our Applicative interfaces differ? Could you please check it?

```java
public interface Applicative<F extends Kind<F, ?>, T> extends Functor<T> {

    <U> Applicative<F, U> ap(Kind<F, ? extends Function<? super T, ? extends U>> f);

    @Override
    <U> Applicative<F, U> map(Function<? super T, ? extends U> f);
}
```

And Kind looks like this:

```java
package javaslang;

public interface Kind<TYPE extends Kind<TYPE, ?>, T> {
}
```

The Applicative interface would be a great addition to the algebra package. I questioned the whole package several times and tended to leave out these types because of several problems. It has taken many iterations but Monad does fit now very well into the library. If we do not give up too early we will achieve the same with Applicative.

src/main/java/javaslang/control/Validation.java
danieldietrich: Thank you for the insight. We need both accumulation styles:

* `v1.combine(v2)...combine(vn)` may be used in fold/reduce.
* The static `Validation.combine(v1, v2, ..., vn)` is a great shortcut. I had this in mind, too.

Do Scalaz/Cats also use a builder? Do you have a link to source-code?

I ask myself if it sufficient if we use tuples or functions of arity 1..8 for accumulated results. If we use validation for a html form for example, the form field will probably exceed the limit of 8. It is not a good idea to blow up the arity of tuples and functions (Scala has 22).

On the other hand it is not possible to use an IndexedSeq instead, holding the accumulated results, because the combined Validations have different Success component types.

[HList](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/ops/hlists.scala) to the rescue?

Example:

```scala
def validate(musician: Musician): StringValidation[Musician] = {
    import scalaz._
    import scalaz.Scalaz._

    def validCurrentBand(band: Option[Band]): StringValidation[Option[Band]] = ???

    def validName(name: String): StringValidation[String] = ???

    def validateAge(born: LocalDate): StringValidation[LocalDate] =
      if (born.isAfter(LocalDate.now().minusYears(12))) "too young".failure
      else born.success

    def validInstrument(instruments: Seq[Instrument]): StringValidation[Seq[Instrument]] = ???

    (validName(musician.name)
      |@| atLeast12(musician.born)
      |@| validInstrument(musician.instruments)
      |@| validCurrentBand(musician.currentBand))(_ => musician)
}
```
(taken from https://www.innoq.com/de/blog/validate-your-domain-in-scala/)

danieldietrich: Let's move the discussion to the issue :) #803

src/main/java/javaslang/control/Applicative.java
enelson: Agreed Daniel, I think it would be a good addition to the algebra package. I'll keep trying this. I've looked over my code again and it is the same as yours, but I'm still getting the errors on the override method for `ap`. I'm also using IntelliJ 15.0.2 on a Mac with JDK  1.8.0_66. Not sure what the discrepancy is, but I'll keep at it. You've got some good ideas on how to make this work, so I think it will be good once it's done.

enelson: Never mind, I was looking at part of this wrong. I think I have it working. Let me play with it a little more.

danieldietrich: Alright!

enelson:
@danieldietrich Some new code ready for review. I was able to get things working with your help on `Kind`. My next thing to tackle is to have `ValidationBuilder`'s `ap()` method return something like `Validation<List<E>,R>` or `Validation<E extends Semigroup,R`. You see what I'm trying to do. I need it to have an error type this is able to accumulate the errors.

danieldietrich:
@enelson Thank you! This looks great.

Regarding the List/Semigroup I think we should not try to get the same generics and method signatures in Javaslang as in Scalaz.

I would keep it as simple as possible. The only purpose of Validation is to accumulate errors. Otherwise we could just flatMap a Try. IMO the order of error occurrence is important. That means we need at least a Seq. If we want to give the user a hint of the performance characteristics of the result we could choose LinearSeq or IndexedSeq. Because the number of subsequent validations is currently bounded (I think) to the number of tuples we have, I would use a LinearSeq. I don't expect millions of errors for one validation, therefore linear time complexity for access is ok. In practice the errors are Iterated. Personally I prefer List over LinearSeq. That's what it is. It does not need to be generic.

Try has two methods to access the results: `get()` and `failed()`. We need something similar for Validation. `get()` will throw NoSuchElementException in the case of Invalid. We need to rename the factory methods `success(T)` to `valid(T)` and `failure(E)` to `invalid(E)`. Also we need `isValid()` and `isInvalid()` instead of `isSuccess()` and `isFailure()` to be consistent. Try has `failed()` to access the Throwable of a Failure. Resp. we should have a method `invalid()` or `errors()`. I would prefer `invalid` because it is more general. We don't know the use case, especially we don't know if the invalid case describes errors.

This leads to:

```java
interface Validation<E, T> {

    static <E, T> Validation<E, T> valid(T t) { ... }

    static <E, T> Validation<E, T> invalid(E... es) { ... }

    boolean isValid();

    boolean isInvalid();

    T get();

    List<E> invalid();

}
```

The List does not need to be part of the Validation type signature because it is fixed. We just need to know the component type E.

What do you think. Does this sound reasonable? Or is there a use case where it makes sense to be more generic?

enelson:
Thanks @danieldietrich. This is great feedback. I can't believe I forgot to switch the `success()/failure()` methods to `valid()/invalid()`. :) Thanks for catching that. 

I made the change to `T get()`, but I'm not sure that I like the idea of naming the counter part as `invalid()`. I think that naming it `error()` is more descriptive about what the method is returning. I'm curious what you mean by 'We don't know the use case, especially we don't know if the invalid case describes errors.' I would argue that the very point of the `Invalid` class is to store validation "errors". The type of `Invalid` may not always be of type `Exception` or `Throwable`, but even if it's a string, it still describes a validation error. But you tell me how you feel about it. Am I missing something in what you were trying to explain?

Concerning my comments about List/Semigroup, I agree. The sole purpose of Validation is to accumulate errors. But the way I was looking at it, each instance of `Validation` may not contain a list of errors. For instance, you may have multiple invalids of type `Validation<String,?>`. Then the purpose of the `ValidationBuilder` is to chain them all and return the final validation with combined errors as `Validation<List<E>,?>`. Does that make sense? So instead of making the `Validation.errors()` (or whatever we call it) method return List<E>, it returns E. Then we get the list after we've combined the instances together. This is consistent with other Validation implementations. My only though about using Semigroup was that instead of hardcoding the accumulation vehicle to List, you could use any type of Semigroup. But I think that's going to get too ugly, with little benefit. So I agree with using List. 

I've toyed with making `Invalid.errors()` return a List<E>, and it gets ugly. For instance, the implementation for `Validation.swap()` now tries to stuff a List<E> into a Valid<List<E>>, which doesn't work. It just seems much cleaner to keep Valid<T> and Invalid<E>. I'd suggest that we push the List<E> code to `Applicative.ap()` and `ValidationBuilder`. This keeps Valid and Invalid clean. Thoughts?

Thanks again for being so proactive on working with me on this. It's been a fun piece of code to work on. :)

danieldietrich:
Yes, it is fun for me, too. Your suggestions are thoughtful.

I'm fine naming it `error()`.You're right, when a validation fails it is an error.

And you're also right - we need `validation.swap().swap()` to be the identity, namely `validation`. That's the only constraint. If we achieve this by moving the relevant parts to the Builder, it is the right choice.

It is good to have you on board :)

enelson:
Ok, I finally have some working code! I had to rework `Kind` and `Applicative` a little, in order to expose the error type so that I could return `List<E>` from `ap()`. Take a look, as well at changes made according to your other suggestions as well. I think we're getting close! I've updated my tests which are all passing right now.

danieldietrich:
Awesome!!
We will include it in 2.0.0. I plan a new release candidate soon, 2.0.0-RC3. I need some slight modifications in order to pull it. Please see my comments.

src/main/java/javaslang/control/Kind.java
```diff
@@ -0,0 +1,5 @@
+package javaslang.control;
+
+public interface Kind<TYPE extends Kind<TYPE, ?, ?>, E, T> {
```
danieldietrich: We call it Kind2 and move out of the `javaslang/control` package into the `javaslang`package.
Maybe I will later generate Kind1..Kindn but that's another story.

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+        return new ValidationBuilder.ValidationBuilder3<>(v1, v2, v3);
+    }
+
+    boolean isValid();
+
+    boolean isInvalid();
+
+    T get();
+
+    E error();
```
danieldietrich: Let's name it `getError()`. It is conform to `getCause()` of Try.

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+ *
+ * @param <E> Value type in the case of invalid.
+ * @param <T> Value type in the case of valid.
+ * @author Eric Nelson
+ * @since 2.0.1
+ * @see <a href="http://eed3si9n.com/learning-scalaz/Validation.html">Validation</a>
+ */
+public interface Validation<E,T> extends Kind<Validation<?,?>, E, T>, Applicative<Validation<?,?>, E, T> {
+
+    static <E,T> Validation<E,T> valid(T value) {
```
danieldietrich: ok, value may be null!

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+        return new Valid<>(value);
+    }
+
+    static <E,T> Validation<E,T> valid(Supplier<? extends T> supplier) {
```
danieldietrich: Please add a check `Objects.requireNonNull(supplier, "supplier is null");`

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+        return new Valid<>(supplier.get());
+    }
+
+    static <E,T> Validation<E,T> invalid(E error) {
```
danieldietrich: Please add a check `Objects.requireNonNull(error, "error is null");`

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+        return new Invalid<>(error);
+    }
+
+    static <E,T> Validation<E,T> invalid(Supplier<? extends E> supplier) {
```
danieldietrich: Please add a check `Objects.requireNonNull(supplier, "supplier is null");`

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+        return new Invalid<>(supplier.get());
+    }
+
+    static <E,T1,T2> ValidationBuilder<E,T1,T2> map2(Validation<E,T1> v1, Validation<E,T2> v2) {
```
danieldietrich: Please add checks
```java
Objects.requireNonNull(v1, "v1 is null");
Objects.requireNonNull(v2, "v2 is null");
```

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+        return new ValidationBuilder<>(v1, v2);
+    }
+
+    static <E,T1,T2,T3> ValidationBuilder.ValidationBuilder3<E,T1,T2,T3> map3(Validation<E,T1> v1, Validation<E,T2> v2, Validation<E,T3> v3) {
```
danieldietrich: Please add null-checks for v1, v2, v3

src/main/java/javaslang/control/Validation.java
danieldietrich: Please add null-checks for fail and success

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+            E v = this.error();
+            return Validation.valid(v);
+        } else {
+            T v = this.get();
+            return Validation.invalid(v);
+        }
+    }
+
+    @Override
+    default <U> Validation<E,U> map(Function<? super T, ? extends U> f) {
```
danieldietrich: Please null-check f

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+        if(isInvalid()) {
+            return (Invalid<E,U>) this;
+        } else {
+            T v = this.get();
+            return Validation.valid(f.apply(v));
+        }
+    }
+
+    default <U,R> Validation<U,R> bimap(Function<? super E,? extends U> fail, Function<? super T,? extends R> success) {
```
danieldietrich: please null-check success

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+        if(isInvalid()) {
+            E v = this.error();
+            return Validation.invalid(fail.apply(v));
+        } else {
+            T v = this.get();
+            return Validation.valid(success.apply(v));
+        }
+    }
+
+    default <U> Validation<U,T> leftMap(Function<? super E,? extends U> f) {
```
danieldietrich: please null-check f

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+        if(isInvalid()) {
+            E v = this.error();
+            return Validation.invalid(f.apply(v));
+        } else {
+            return (Valid<U,T>) this;
+        }
+    }
+
+    default <U> Validation<List<E>,U> ap(Validation<List<E>, ? extends Function<? super T,? extends U>> v) {
```
danieldietrich: please null-check v

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+            return invalid(() -> {
+                List<E> errors = v.error();
+                E e = this.error();
+                return errors.append(e);
+            });
+        }
+    }
+
+    @Override
+    default <U> Validation<List<E>,U> ap(Kind<Validation<?, ?>, List<E>, ? extends Function<? super T, ? extends U>> f) {
```
danieldietrich: please null-check f

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+        return ap((Validation<List<E>,? extends Function<? super T,? extends U>>) ((Object) f));
+    }
+
+    default <U> ValidationBuilder<E,T,U> combine(Validation<E,U> v) {
```
danieldietrich: please null-check v

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+ * </code>
+ * </pre>
+ *
+ * @param <E> Value type in the case of invalid.
+ * @param <T> Value type in the case of valid.
+ * @author Eric Nelson
+ * @since 2.0.1
+ * @see <a href="http://eed3si9n.com/learning-scalaz/Validation.html">Validation</a>
+ */
+public interface Validation<E,T> extends Kind<Validation<?,?>, E, T>, Applicative<Validation<?,?>, E, T> {
```
danieldietrich: A general thing: Please let us choose full variable names, e.g. `validation` instead of `v`, same for `v1`, `v2`. I think it increases the readability of our API. It is ok that the lines get slightly longer.

danieldietrich: We stay with `f` as abbr. for `function`. It is very common and well known from math and FP.

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+            return true;
+        }
+
+        @Override
+        public boolean isInvalid() {
+            return false;
+        }
+
+        @Override
+        public void forEach(Consumer<? super T> f) {
```
danieldietrich: please null-check `f`

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+
+    final class Invalid<E,T> implements Validation<E,T> {
+
+        private final E error;
+
+        /**
+         * Construct an {@code Invalid}
+         * @param error The value of this error
+         */
+        public Invalid(E error) {
```
danieldietrich: Please make the constructor `private` to be consistent with the other controls. Instances are created via the Validation API.

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+
+    final class Valid<E,T> implements Validation<E,T> {
+
+        private final T value;
+
+        /**
+         * Construct a {@code Valid}
+         * @param value The value of this success
+         */
+        public Valid(T value) {
```
danieldietrich: Please make the constructor private (see Invalid).

src/main/java/javaslang/control/Validation.java
```diff
@@ -0,0 +1,276 @@
...
+            return false;
+        }
+
+        @Override
+        public boolean isInvalid() {
+            return true;
+        }
+
+        @Override
+        public void forEach(Consumer<? super T> f) {
```
danieldietrich: I would pull that method out of Valid and Invalid into the Validation interface:

```java
if (isValid()) {
    f.accept(get());
}
```

src/main/java/javaslang/control/ValidationBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+ * /___/\_/  \_/\____/\_/  \_/\__\/__/___\_/  \_//  \__/_____/    Licensed under the Apache License, Version 2.0
+ */
+package javaslang.control;
+
+import javaslang.Function2;
+import javaslang.Function3;
+import javaslang.Function4;
+import javaslang.collection.List;
+
+public class ValidationBuilder<E,T1,T2> {
```
danieldietrich: Please

* move the ValidationBuilder* into the Validation interface. I see them as '2nd level API', they should not be accessed directly, we use Validation API instead (some kind of 'single point of contact').
* 'flatten' the ValidationBuilders, i.e. Validation.ValidationBuilder, Validation.ValidationBuilder3, ...
* make the classes final
* make the constructors private

danieldietrich:
I'm really looking forward to that pull!

danieldietrich:
Decide yourself if method params are renamed. They do not to necessarily. It isn't consistent in Javaslang.

src/main/java/javaslang/control/ValidationBuilder.java
enelson: Does it makes sense to put ValidationBuilder into a `ValidationModule` interface inside of `Validation`, like is done in `Match` and other places?

danieldietrich: No, it needs to be public/accessible for callers of the API. We place it in the Validation interface.

This is similar to all the nested Match interfaces. We use this pattern in many places (Try.Success/Failure, Option.Some/None, ...). It can be seen as some kind of internal DSL. The Builders are needed to technically ensure that we get our types right but in first place they are not needed by our _business domain_ (which is also technical here but that's only a matter of abstraction).

danieldietrich: I would flatten the builders for reasons of simplicity. We may decide in future to increase the number of Tuples. This would lead to an increased depth of Builder classes which is not practical (javadoc for example).

enelson:
Okay, more changes to review. I'm not 100% sure what you're envisioning when you say to 'flatten the builders'. Can you send me an example of what you're thinking? Thanks Daniel.

enelson:
I'm going to be working on adding much more javadocs, just want to finalize the code and methods first.

danieldietrich:
With flattening I meant:

```java
interface Valitation {
   final class ValidationBuilder {
   }
   final class ValidationBuilder3 {
   }
   final class ValidationBuilder4 {
   }
   ...
}
```

instead of

```java
interface Valitation {
   final class ValidationBuilder {
       final class ValidationBuilder3 {
           final class ValidationBuilder4 {
               ...
           }
       }
   }
}
```

I think we then also can simplify the names:

```java
interface Valitation {
   final class Builder {
   }
   final class Builder3 {
   }
   final class Builder4 {
   }
   ...
}
```

danieldietrich:
Looks pretty polished already! I added two comments (accidentally to the commit on your branch I think)

enelson:
Good suggestions @danieldietrich. Those changes have been made.

src/test/java/javaslang/control/ValidationTest.java
```diff
@@ -0,0 +1,359 @@
...
+
+    @Test
+    public void shouldProcessFunctionInForEach() {
+        Validation<String,String>  v1 = Validation.valid("Eric Nelson");
+        Validation<String,String>  v2 = Validation.invalid("error");
+
+        // Not sure best way to test a side-effect only function?
+        v1.forEach(System.out::println);
+        v2.forEach(System.out::println);
+        assertThat(true).isTrue();
```
ruslansennov: I would suggest something like this:

```java
        final java.util.List<String> accumulator = new ArrayList<>();

        accumulator.clear();
        v1.forEach(accumulator::add);
        assertThat(accumulator.size()).isEqualTo(1);
        assertThat(accumulator.get(0)).isEqualTo("Eric Nelson");

        accumulator.clear();
        v2.forEach(accumulator::add);
        assertThat(accumulator.size()).isEqualTo(0);
```

enelson: @ruslansennov That's a better suggestion. :) Thanks. I'll update that test.

enelson:
@danieldietrich At this point, I'm just looking for any feedback on what will keep you from merging this PR. Are there still any refinements or documentation that you'd like to see before merging? Thanks Daniel.

danieldietrich:
@enelson Thank you for this huge effort! There is nothing to say other than I really enjoyed working on this with you and that I'm super-happy to press the Merge button in a second :-)

I will integrate the monadic stuff into Validation, I've been working on this the last days...

enelson:
@danieldietrich yay! Thanks so much for your help on this. You made me stick with the Kind2 and Validation work, even though it was more difficult. And it was the right decision. So thanks for all your help. And I'm happy to continue helping on whatever I can to help your push to 2.0.0. I'm very excited for this upcoming release.

danieldietrich:
Thank you! I will have a look the next days what can be done in order to get 2.0.0 right.

danieldietrich:
@enelson I confused Monad with Value... next, Validation needs to implement Value. It would be great if you could do that. It will be right-biased, like Either, e.g.

```java
interface Validation<E, T> extends Value<T> { ... }
```

danieldietrich:
Oh - and just saw that we wanted to rename the static `map2`, ..., `map8` methods all to just `combine` to be consistent with the non-static `validation.combine(Validation)` which returns a Builder. Could you do that, please?

enelson:
Yep, I'll rename those and submit. Thanks for the reminder.

PR #990, opened 2015-12-31
Add match() function to Value as syntactic sugar for Match.of(value)

### LINKED ISSUE #968: Tighten pattern matching of Values
Currently we have

```java
// match-monad
Match.of(value).when(...).then(...).get();

// match-function
Match.when(...).then(...).apply(value);
```

Scala offers

```scala
value match {
  case ... => ...
}
```

Javaslang has a `Convertible` interface to equip Values with conversion methods. We need to add a `match()` method as convenience method for creating a match-monad `Match.of(...)`.

```java
value.match()
    .when(...).then(...)
    .get();
```

This is syntactic sugar for

```java
Match.of(value)
    .when(...).then(...)
    .get();
```

and especially great when we write fluent APIs and the expression to compute `value` is big.

mbarbieri:
Attempt to fix #968 

The only test is `@Ignore` right now, waiting for @danieldietrich to refactor `Match` provinding the `equals` method.

Feedbacks are welcome of course

### DIFF
--- a/src/main/java/javaslang/Value.java
+++ b/src/main/java/javaslang/Value.java
@@ -8,7 +8,10 @@ package javaslang;
 import javaslang.algebra.Foldable;
 import javaslang.algebra.Monad;
 import javaslang.collection.*;
-import javaslang.control.*;
+import javaslang.control.Either;
+import javaslang.control.Match;
+import javaslang.control.Option;
+import javaslang.control.Try;
 
 import java.io.PrintStream;
 import java.io.PrintWriter;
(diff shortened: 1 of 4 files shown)


danieldietrich:
Nice, thanks Matteo!

src/main/java/javaslang/Value.java
```diff
@@ -510,6 +513,12 @@ default CharSeq toCharSeq() {
         return ValueModule.toTraversable(this, Vector.empty(), Vector::of, Vector::ofAll);
     }
 
+    @Override
+    @SuppressWarnings("unchecked")
+    default Match.MatchMonad.Of<T> match() {
+        return Match.of((T) this);
```
danieldietrich: There is a reason that `(T) this` does not check. It is `Convertible<T>` and cannot be cast to `T`.

I reopened #968 and already fixed it locally. PR comes soon...

Issue #976, opened 2015-12-23
Redesigning the side-effecting branch of the Match API


danieldietrich:
Match language. The internal DSL is a state-machine. We should ensure that in the case of side effects via `when(...).thenRun(action)` not other non-side-effects take place, e.g. `whenApplicable(function).thenApply()`. It does not make sense to return a value once the return value is `Void`.

We accomplish this by modeling additional states of the state machine, which does not allow us to leak back to the non-side-effecting Match API.

I've done this already for MatchFunction in a local fork, it looks like this:

![match](https://cloud.githubusercontent.com/assets/743833/11986841/bb2ab1bc-a9d4-11e5-8a6f-58112a3a9038.png)

danieldietrich:
The side-effecting-only parts of the Match DSL where separated from the parts that result in a value. This provides a clean API and helps the user not to mess up program logic.

![match](https://cloud.githubusercontent.com/assets/743833/12076274/5a84e7d8-b1a4-11e5-8447-3a941e6b8e2a.png)

danieldietrich:
When only side-effects are performed, we need not to throw a MatchError when not case matched.

Example:

```java
// nothing happens
Match.whenIs(1)
          .thenRun(() -> System.out.println("matched!"))
          .accept(0);
```

danieldietrich:
Finished, tests will be added with #477

PR #992, opened 2016-01-03
Re-organized Match types, removed whenRunnable(Runnable)…

danieldietrich:
…in favor of otherwiseRun(Runnable)

Towards #976


danieldietrich:
Fixes #976

Issue #864, opened 2015-11-26
Measure the impact of using Lazy hash codes in Traversables


danieldietrich:
We should question the whole idea of caching hashes. I mean, a lazy wrapper of a lambda/value is computed for every node of a data structure in general - also for primitive values like ints, which are commonly used. We need to profile the memory consumption of different scenarios and compare it with other frameworks, first of all Java standard lib.

We should answer the question how often the hash code is really needed in practice, e.g. how often collections are used as keys in hash-based collections. And what is the value count where caching really starts to matter. And what is more important: CPU or Memory.

RobWin:
Most important point for real production usage of collections.

ruslansennov:
It seems we should remove all lazy hash codes from sequences, but keep these in certain cases like in <code>HAMT.ArrayNode</code> (it significantly reduced performance)

danieldietrich:
Let's do it like you say. If the memory footprint is still too big compared with ordinary collections, we can still remove the cache.

ruslansennov:
Now we have lazy hash code here:
* <code>Queue</code> In case of non-empty rear list it reduces performance. We can add hashCode to constructor, but this makes code more (and more) difficult
* <code>HAMT.ArrayNode</code> Each node changing made by <code>System.arrayCopy()</code> which is fast. But additional loop for hashCode calculation significantly reduces performance.
* <code>HashMap</code>, <code>HashSet</code>, <code>Vector</code> because of <code>HAMT</code>
* <code>Stream</code> must be lazy
* <code>Array</code> same problem like in <code>HAMT.ArrayNode</code>

In all these cases (except <code>HAMT.ArrayNode</code>) we have exactly one instance of <code>Lazy</code>, so it is not problem. For example, <code>List</code> produces a lot of lazy objects and it was a bad.
In case <code>HAMT.ArrayNode</code> it is not problem because number of this kind of node in trie is small (something about 3%)

I believe this issue can be closed

danieldietrich:
Ok, sounds reasonable. We will have an eye on it when benchmarking the collections in depth.

PR #995, opened 2016-01-05
reduce HAMT memory usage

ruslansennov:
Removing lazy objects from a leaf nodes is enough to reduce memory usage twice. It couldn't have any change to CPU usage.
Now we have 48M for HAMT size 1M. 
See also #864


danieldietrich:
Great!

PR #1006, opened 2016-01-10
Added Either.fold. Refactored fold. Fixes #1005

### LINKED ISSUE #1005: Add fold to Either (similar to Validation.fold)
Validation has

```java
default <U> U fold(Function<? super E,? extends U> fail, Function<? super T,? extends U> success)
```

we need s.th. similar for Either.

danieldietrich:


danieldietrich:
(pressed the wrong button on my smartphone)

PR #1007, opened 2016-01-10
659 tabulate fill

simonecarriero:


src/main/java/javaslang/collection/Tree.java
```diff
@@ -100,6 +100,22 @@
...
      * Creates a Tree of the given elements.
+     *
+     * @param <T> Component type of the List.
+     * @param nodes Zero or more nodes.
+     * @return A Tree containing the given nodes.
+     * @throws NullPointerException if {@code nodes} is null
+     */
+    @SuppressWarnings("varargs")
+    @SafeVarargs
+    static <T> Tree<T> of(T... nodes) {
```
danieldietrich: Let's rename `nodes` to `values`

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -399,20 +399,7 @@ public Short getNext() {
...
-            public boolean hasNext() {
-                return i < n;
-            }
-
-            @Override
-            protected T getNext() {
-                return f.apply(i++);
-            }
-        };
+        return Collections.tabulate(n, f, Iterator.empty(), Iterator::of);
```
danieldietrich: here we should you the lazy `Collections.tabulate(n, f)` method

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -425,7 +412,8 @@ protected T getNext() {
      * @throws NullPointerException if {@code s} is null
      */
     static <T> Iterator<T> fill(int n, Supplier<? extends T> s) {
-        return tabulate(n, anything -> s.get());
+        Objects.requireNonNull(s, "s is null");
+        return Collections.fill(n, s, Iterator.empty(), Iterator::of);
```
danieldietrich: here we should also use the lazy `Collections.tabulate(n, f)` method

danieldietrich:
Hi Simone! Well done. Please find my comments above.
Many thanks for all your effort!

danieldietrich:
Great, thx!

PR #1014, opened 2016-01-12
Added right-biased mode to Either (making addition of Xor unnecessary)

### LINKED ISSUE #882: Let Either.Left/RightProjection return Either instead of Left/RightProjection
We changed the behavior of Either compared to Scala in the way, that we _stay_ in the projection Layer. 

In other word, currently we need to call toEither explicitly.

We should change that, operations of the projections should return Either instances.

Examples:

```java
Either<String, String> good = Either.right("good");
Either<String, String> bad = Either.right("bad");

// status quo
Either<String, Tuple2<String, String>> result = good.right()
    .flatMap(g -> bad.right().map(b -> Tuple.of(g,b)))
    .toEither();

// goal
Either<String, Tuple2<String, String>> result = good.right()
    .flatMap(g -> bad.right().map(b -> Tuple.of(g,b)));
```

danieldietrich:
Fixes #882

### DIFF
--- a/src/main/java/javaslang/Convertible.java
+++ b/src/main/java/javaslang/Convertible.java
@@ -19,7 +19,7 @@ import java.util.function.Supplier;
  *
  * @param <T> Component type.
  * @author Daniel Dietrich
- * @since 2.0.0
+ *
  */
 public interface Convertible<T> {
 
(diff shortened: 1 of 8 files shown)


enelson:
Nice work!

danieldietrich:
Thank you :)

PR #1017, opened 2016-01-13
renaming map(n) functions to combine

enelson:


danieldietrich:
Thank you Eric!

enelson:
You bet! BTW, do you ever sleep? :)

danieldietrich:
Haven't slept well today, I have a cold and coughed all the night. But it is getting better :)

PR #1019, opened 2016-01-13
Factory methods of* return the singleton empty

simonecarriero:


src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -99,7 +99,7 @@ public static CharSeq of(char... characters) {
         Objects.requireNonNull(characters, "characters is null");
         final char[] chrs = new char[characters.length];
         System.arraycopy(characters, 0, chrs, 0, characters.length);
-        return new CharSeq(new String(chrs));
+        return characters.length == 0 ? empty() : new CharSeq(new String(chrs));
```
ruslansennov: Hi Simone!
I would made this check before 
```
final char[] chrs = ...
```
to avoid instantiate empty char[]

src/main/java/javaslang/collection/Iterator.java
```diff
@@ -141,6 +141,7 @@ public T getNext() {
     @SafeVarargs
     static <T> Iterator<T> of(T... elements) {
         Objects.requireNonNull(elements, "elements is null");
+        if (elements.length == 0) return Iterator.empty();
```
ruslansennov: please use brackets
```java
if (elements.length == 0) {
    return Iterator.empty();
}
```

src/test/java/javaslang/collection/AbstractMapTest.java
```diff
@@ -99,6 +99,11 @@
         return IntMap.of(emptyMap());
     }
 
+    @Override
+    protected boolean emptyShouldBeSingleton() {
+        return false;
```
danieldietrich: For non-sorted Maps, e.g. HashMap, it is a singleton (I think).
For sorted Maps, e.g. TreeMap, it is no singleton because of the Comparator.
Maybe we can add a `abstract class AbstractSortedMapTest extends AbstractMapTest` to overwrite it for sorted Maps?

danieldietrich:
Very good! That saves memory/GC cycles.

src/test/java/javaslang/collection/AbstractMapTest.java
ruslansennov: I believe now is enough to override it in TreeMapTest

danieldietrich: You are right.

simonecarriero: Yes they are singleton. The tests defined in AbstractTraversableTest don't work for Maps because AbstractMapTest::empty() returns an IntMap. I add another method emptyMapShouldBeSingleton to manage on which subclasses of Map the tests should be run.

danieldietrich: Can't we just overwrite the method `emptyShouldBeSingleton` and let `AbstractMapTest` return `true` and `TreeMapTest` return `false`? Then the tests should run, right?

simonecarriero: No, the tests defined in `AbstractTraversableTest` would not pass with subclasses of `Map`, because `AbstractMapTest::empty()` returns `IntMap.of(emptyMap())`. The `Map` hierarchy has its own set of tests using `emptyMap()` instead!

danieldietrich: Good!

danieldietrich:
Thank you very much!

PR #1026, opened 2016-01-15
reduce*Option fixed #1021

### LINKED ISSUE #1021: When reduce* returns null reduce*Option must return Some(null) instead of None
The semantics of reduceLeftOption, reduceRightOption is to return None, if the underlying Value/Traversable is empty. If the operation successfully combines one or more elements, the result has to be Some.

ruslansennov:


danieldietrich:
Right!

But... I have completely removed the methods reduceRightOption, reduceLeftOption from Traversable. They are now implemented as default in Value. There where also other unnecessary impls I think.

This PR can be closed. I urgently need to push my changes. They get more and more while refactoring this Monad stuff. Please wait until I've pushed them today...

danieldietrich:
(I've also fixed that bug already)

danieldietrich:
This is what I've done so far. Too much for one PR I think O_o
https://github.com/javaslang/javaslang/compare/master...danieldietrich:master

ruslansennov:
> Too much for one PR I think O_o

yep, very bad PR :)

danieldietrich:
I broke my new-years resolution to create local branches instead of working on my master. Needs to get better...

Issue #1029, opened 2016-01-15
Fix bounds of generic argument parameters in methods


danieldietrich:
I did this already for Tuples which are invariant because they are final, e.g. instead of

```java
interface Either<L, R> {
    <L2, R2> Either<L2, R2> bimap(Function<? super Tuple2<? super L, ? super R>, ? extends Tuple2<? extends L2, ? extends R2>> f);
}
```

we have

```java
interface Either<L, R> {
    <L2, R2> Either<L2, R2> bimap(Function<Tuple2<L, R>, Tuple2<L2, R2>> f);
}
```

- easier to read and no need for casting. However, the bounds in `Tuple2<? super L, ? super R>` where wrong though.

There are different cases that have to be hardened. E.g. there is the question, if a Monoid or Semigroup is covariant in general. E.g. can a `Monoid<C>` be used to combine `B`'s if `C extends B`?

We find these cases in Foldable for example:

```java
// can go wrong IMO, if the Monoid requires properties of U extends T, which T does not have
default T foldRight(Monoid<? extends T> monoid)

// should be like this IMO
default T foldRight(Monoid<T> monoid)
```

This is just an example. There are more method signatures which have to be fixed...

danieldietrich:
Here is another example of a bound that is currently suspicious:

```java
interface Try<T> {
    <U> Try<U> flatMap(Function<? super T, ? extends Try<? extends U>> mapper);
    <U> Try<U> flatMapTry(CheckedFunction<? super T, ? extends Try<? extends U>> mapper);
}
```

When we follow the types (in the sense that `@SuppressWarnings("unchecked")` disappear, then this seems to be a better solution:

```java
interface Try<T> {
    <U> Try<U> flatMap(Function<? super T, ? extends Try<U>> mapper);
    <U> Try<U> flatMapTry(CheckedFunction<? super T, ? extends Try<U>> mapper);
}
```

danieldietrich:
We had a similar issue before, please take a look at this:

[Fix result bound of function argument of higher order functions #352](https://github.com/javaslang/javaslang/issues/352)

danieldietrich:
I think have we are fine for now.

Issue #1028, opened 2016-01-15
Test monadic behavior using MonadTestMixin


danieldietrich:
The upcoming Monad and BiMonad interfaces are not any more implemented by Value.

```
         Monad    Value
            \      /
             Option


MonadTestMixin    AbstractValueTest
           \        /
           OptionTest
```

We need a way to annotate default methods of interfaces with junit's `@Test` annotation and get these executed when implemented by classes.

danieldietrich:
Especially check if algebraic properties are correctly inherited, e.g.

* Set is no Functor
* Map is Functor on values but not on key and therefore no BiFunctor
* Tuple2 is a BiFunctor

danieldietrich:
We will not do that since the property test component is in its own module now.

Issue #1027, opened 2016-01-15
Remove Tuple*.transform in favor of Tuple*.map that returns a generic value


danieldietrich:
```java
// Tuple0
<U> U map(Supplier<U> supplier);

// Tuple2
<U> U map(BiFunction<? super T1, ? super T2, ? extends U>);
```

danieldietrich:
Driving idea:

```java
<U> U transform(Function<Tuple2<T1, T2>, ? extends U> f);
```

is basically the same as

```java
<U> U map(BiFunction<? super T1, ? super T2, ? extends U>);
```

**Pros of map:**

* We need map. Removing transform simplifies the API.
* We have the contained values at hand and don't need to extract them from the Tuple. The tuple itself is not needed for mapping the values.
* The target type may be a Tuple, therefore we don't need an additional `map(Function<Tuple,Tuple>)`

**Cons:** (none)

danieldietrich:
Because Tuple0 has no components, we need a `<U> Tuple0.map(Supplier<? extends U>)` that replaces `transform`.

danieldietrich:
There are situations a `map(Function, Function, ..., Function)` is handy, e.g. when used in combination with method references. We already have it. Example:

```java
// Map.unzip3(Function<Tuple2, Tuple3>) return Tuple3<Seq, Seq, Seq>
iterator().unzip3(unzipper).map(Stream::ofAll, Stream::ofAll, Stream::ofAll);
```

However, when having this kind of map function and the map function mentioned in the comments above, Tuple1 would have:

```java
<U> Tuple1<U> map(Function<? super T, ? extends U> f);
<U> U map(Function<? super T, ? extends U> f);
```

which raises ambiguities.

Therefore we take the current approach, which is conform to the Functor contract:

```java
<U> Tuple1<U> map(Function<? super T, ? extends U> f);
```

but simplify our `transform` in the way that it takes Supplier, Function, BiFunction, Function3 instead of Function of Tuple0, Tuple1, Tuple2, Tuple3, ...

Issue #1025, opened 2016-01-15
Add isolated component mappings map1, map2, ... to each Tuple


danieldietrich:
Example:

```java
Tuple.of("a", "b", "c").map2(b -> "test"); // = (a, test, c)
```

danieldietrich:
This is really nice when mapping just a few components:

```java
// (1, 2, 3, 4, 5, 6, 7, 8) -> (1, "two", 3, 4, "five", 6, 7, "last")
tuple8.map2(i -> "two")
      .map5(i -> "five")
      .map8(i -> "last");

// the same as above but can get cumbersome when mapping gets more complex
tuple8.map((i1, i2, i3, i4, i5, i6, i7, i8) -> Tuple.of(i1, "two", i3, i4, "five", i6, i7, "last"));
```

PR #1031, opened 2016-01-17
Revisited Monad

### LINKED ISSUE #1018: Keep Monad clean, fix flatMap
In order to ensure that we keep monadic behavior when composing Monads (e.g. by flatMapping), a re-design is needed. Example: We need it when composing AutoClosables that have to be all closed when an exception occurs.

* separate Monad.flatMap from specific flatMap methods by renaming Monad.flatMap to Monad.flatMapM ✓
* Monad needs to extend recursive self-type Kind1 ✓
* remove all methods from Monad other than map and flatMapM, especially filter ✓
* remove Value.getOption ✓
* re-implement flatMap of all Value impls in a clean way
* ~~remove filter from all Value impls other than Traversable, Gen and Arbitrary~~
* ...

### LINKED ISSUE #1021: When reduce* returns null reduce*Option must return Some(null) instead of None

### LINKED ISSUE #1022: Unify Either.bimap(Function,Funtion) vs. Functor2.map(Function, Function)
(Functor2 will be there soon)

We should use map instead of bimap.

danieldietrich:
This PR is - _eeehh_ - a little bit bigger than expected...

Pro-tip: Never ever try to re-design monadic stuff throughout the whole code-base.

profile

**Code-reviews are welcome!** I will leave this PR open for a few hours...

* The MonadTests are deactivated now. I created a MonadTestMixin, which also extends MonadLaws. We need to solve #1028 in order to enable default interface method unit tests. There are libs out there to achieve this. Maybe junit-lambda (SNAPSHOT).
* I pulled the 'dirty' flatMap code into a new method `Monad.unit`. This might come handy later when a Monad needs to cleanup system resources.
* The Type conversion stuff now resides in Value. It should not be within the algebra layer.
* You can ignore the changes within Tuple*. Tuple.transform and .map will be re-visited tomorrow, see #1025, #1027.
* I started to fix some generic bounds in method signatures on my way through the monadic code. This should be done for other method signatures also. See #1029 
* I added BiFunctor which has `bimap(Function, Function)`, implemented now by Validation, Map, Either, Either.LeftProjection, Either.RightProjection
* I re-arranged some methods in alphabetic order (yes, I know - never re-format code in a PR...)
* ...

Fixes #1018 "Keep Monad clean, fix flatMap"
Fixes #1021 "When reduce* returns null reduce*Option must return Some(null) instead of None"
Fixes #1022 "Unify Either.bimap(Function,Funtion) vs. Functor2.map(Function, Function)"
Towards #1029 "Fix bounds of generic argument parameters in methods"


danieldietrich:
Note: The Monad tests are temporarily deactivated. This probably is the reason for the decreased code coverage.

ruslansennov:
> This probably is the reason for the decreased code coverage.

Probably?! Realy?! :)

danieldietrich:
@ruslansennov I try to be diplomatic :smile: - they will be re-activated soon.

I had to deactivate them because Value is no Monad any more. This is because our new Validation (which will extend Value soon) is an Applicative.

I will pull these things now to go on with the other issues. If you have any remarks, please comment on the code anyway...

Issue #1044, opened 2016-01-18
CHANGES TO THE PROJECT STRUCTURE


danieldietrich:
Within the last months again and again problems raised based on Java's type system.

* there is no way to define a general flatMap function for controls, traversables and monads
* trying to define a generic flatMap method raises new problems of white spots/undefined return values
* now that we will implement orElse, known from Scala, there raise new problems similar to flatMap
* inheriting from algebraic classes is not a good idea because it does not reflect the variety of possibilities
* and much more

There were also design decisions to uniform things which appear to be artificial.

* all types have to be a Value, e.g. support get() of a single element etc. that makes not much sense for traversables
* on the other hand all single-valued types, e.g. controls, have to be foldable, which is only practicable for multi valued types
* other properties currently have no common interface like getOrElse(Function), recover, recoverWith

I've thoroughly created a solution which does fit Java's type system. However, it does need us to slice and modularize our project and split up some interfaces.

**We need to do these changes before final 2.0.0.**

1) Modularize the Maven project into
  - javaslang-parent (pom)
  - javaslang-algebra (jar) containing the javaslang.algebra package
  - javaslang-test (jar) containing the javaslang.test package
  - javaslang (jar) containing the current classes minus the above packages

2) Split the Value class into a portion inherited by controls & concurrent (single-valued) types and into a part inherited by (multi-valued) traversables
  - remove map, flatMap, filter from Value
  - maybe move also peek into traversable, it is the same as ifDefined, ifEmpty
  - move the conversion methods out of Value and let Value and Traversable inherit from it
  - move the equality check and iterable extensions out of Value and let Value and Traversable inherit from it
  - move the output methods (stdout, out) out of Value and remove them completely or let Value and Traversavle inherit them
  - because the signature of the map & flatMap methods differs for for Seq/Set/Tree and Map, we need  to remove them from Traversable
  - we need to move Foldable from the algebra package to the javaslang module. It is not algebraic but an abstraction over inherently recursive data structures. It depends on Monoid. We will turn these methods inside-out and move them into the algebra module into Monoid, e.g. `Foldable.foldLeft(Monoid)` will become `Monoid.foldLeft(Foldable f) { return f.foldLeft(this.zero(), this.combine()); }`.
  - we also remove all other dependencies from the javaslang module to javaslang-algebra. We do not inherit from Functor or Monad. Instead we will provide contructors `Monad.from(List)`, `Monad.from(Option)`, `Applicable.from(Validation)`.
[...]

PR #1048, opened 2016-01-20
Changed Try and Future methods to take checked functions

### LINKED ISSUE #865: Specific Try (and Future) methods should take Checked functions
The original question was: "What about renaming Try.recoverWith to tryRecover which takes a CheckedFunction?"

It is important that it is consistent. I started to think about Function vs CheckedFunction args in Try methods when implementing common interfaces like Value. There we have map, flatMap but also orElseGet, orElseThrow et. al. These use Function. We expect all Values to behave the same.
We could make all Try-specific methods take Checked-versions.

This is also the reason why we have mapTry and flatMapTry. They can't overwrite the Value.map and Value.flatMap, which is a pitty. Similar cases we have with Map.map(BiFunction) and Map.flatMap(BiFunction) - they are special.

**Q:** Why do we need function anyway? Why not always a CheckedFuntion?

**A:** It is a matter of interoperability with Java 8. We need interfaces to the standard lib - otherwise Javaslang will be an Alien living on another Planet or the code gets too verbose because of the need to convert things...

**Q:** What about changing recoverWith to recoverTry?

**A:** I will take it into account. Deeper integration of Try outside of it / within other types might make sense. But we should keep the API thin but deep. In other words we should not create many words in our language for the same thing but compose our sentences from a set of few words.

Future has also Try capabilities. Maybe Try and Future need a common interface.

danieldietrich:
Fixes #865
Towards #1044

### DIFF
--- a/src/main/java/javaslang/Lazy.java
+++ b/src/main/java/javaslang/Lazy.java
@@ -5,7 +5,6 @@
  */
 package javaslang;
 
-import javaslang.algebra.Monad;
 import javaslang.collection.Iterator;
 import javaslang.collection.List;
 import javaslang.collection.Seq;
--- a/src/main/java/javaslang/collection/List.java
+++ b/src/main/java/javaslang/collection/List.java
@@ -1022,11 +1022,10 @@ public interface List<T> extends Kind1<List<?>, T>, LinearSeq<T>, Stack<T> {
         return found ? result.reverse() : this;
     }
 
-    @SuppressWarnings("unchecked")
     @Override
     default List<T> removeAll(Iterable<? extends T> elements) {
         Objects.requireNonNull(elements, "elements is null");
-        final List<T> removed = List.ofAll(elements).distinct();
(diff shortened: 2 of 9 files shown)


danieldietrich:
@RobWin Would you please review it and take a critical look?

Do the changes make sense? I think we need to still change onSuccess and onFailure.

```java
myTry
    .onSuccess(CheckedConsumer) // <-- currently changes the result if consumer fails, which is wrong IMO
    .onFailure(CheckedConsumer)
```

Same for Future.

RobWin:
Ok.I have a look at it.
My first thought is that the Checked* interfaces work fine, if you use lamdas. But if you want to use existing Supplier, Predicates or Functions, it does not work anymore.
Having two methods like map and mapTry does not introduce that disadvantage.

danieldietrich:
Thank you Robert, that helped. I re-added the *Try() methods after testing a bit with futures. It makes life easier to have both flavors - checked and unchecked function args.

Issue #1024, opened 2016-01-14
Switch from int to long when it comes to size of collections.


danieldietrich:
Examples are: take, drop, zipWithIndex, ...

A return type of `Tuple2<T, Long>` instead of `Tuple2<T, Integer>` affects compatibility.

Java 8 does it, too. See https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html

PR #1053, opened 2016-01-23
Switch from int to long indices #1024

ruslansennov:


danieldietrich:
Thanks, that was fast! Should be able to create RC3 at the end of this weekend. Then only tests and bug fixes follow. 2.0.0 final is near :)

PR #1064, opened 2016-01-27
HAMT performance

ruslansennov:


danieldietrich:
Thank you! Did you profile it?

ruslansennov:
> Did you profile it?

Yes :)
It was surprise for me that caching <code>Objects.hashCode(...)</code> can significantly increase performance. Also I removed <code>List</code> from <code>IndexedNode</code> because of very slow <code>List.get()</code> method.

danieldietrich:
Great!!

danieldietrich:
In theory, theory equals practice. Or as Lenin said "Trust but verify" (or something like this) :-))

Issue #1065, opened 2016-01-27
Upper-bounded Traversable.contains behaves unexpected


danieldietrich:
We need to change the `contains(T)` and `*indexOf(T)` methods to `contains(Object)` and `*indexOf(Object)`.

Here are some example of code that currently does not work:

```java
List<Double> slangDoubles = List.of(42d, 99d);
List<? extends Number> slangNumbers = slangDoubles;
slangNumbers.contains(42d); // DOES NOT COMPILE

java.util.List<Double> javaDoubles = Arrays.asList(42d, 99d);
java.util.List<? extends Number> javaNumbers = javaDoubles;
javaNumbers.contains(42d); // Compiles fine...
```

However, `javaNumbers.add(42d)` does also not work. But it is quite eligible and also intuitive to say

```java
// = false
listOfAnimals.contains(stone);
```

Another example:

```java
 class SimpleComposition {

            private final List<? extends Number> numbers;

            private SimpleComposition(List<? extends Number> numbers) {
                this.numbers = numbers;
            }

            boolean contains(Number number) {
                return this.numbers.contains(number); // DOES NOT COMPILE
            }
}

List<Double> slangDoubles = List.of(2d, 9d);
new SimpleComposition(slangDoubles).contains(2d);
```

Using `<U extends T> boolean contains(U that)` does also not work.

As @RobWin says

Producer extends and Consumer super "PECS" is from the collection's point of view. If you are only pulling items from a generic collection, it is a producer and you should use extends; if you are only stuffing items in, it is a consumer and you should use super.
(from stackoverflow)

@netzwerg great finding!!

RobWin:
PECS does not help here :(
Only an Object argument helps.

danieldietrich:
Moving back contains from Value to Traversable...

danieldietrich:
Maybe this could affect even more methods, e.g. let's take a look at `java.util.Map`:

```java
    boolean containsKey(Object key);
    boolean containsValue(Object value);
    V get(Object key);
    V put(K key, V value);
    V remove(Object key);
    ...
```

danieldietrich:
@ruslansennov @netzwerg @RobWin I think this goes in the wrong direction. We loose type safety, which is no good idea. In Rahels example we just need to cast the `List<Double>` to a `List<Number>` and all is fine. But Java says we can only get a `List<? extends Number>`, which is useless. Even if `contains` is fixed, we still have no working `prepend` etc.

[...]

PR #1072, opened 2016-01-29
relax signatures of indexOf(), contains(), etc #1065

ruslansennov:


javaslang/src/main/java/javaslang/collection/TreeSet.java
```diff
@@ -505,9 +505,10 @@
         }
     }
 
+    @SuppressWarnings("unchecked")
     @Override
-    public boolean contains(T element) {
-        return tree.contains(element);
+    public boolean contains(Object element) {
+        return tree.contains((T) element);
```
ruslansennov: I'm not sure about that...
This line can produce <code>ClassCastException</code> if <code>Comparator</code> do not check this.

javaslang/src/main/java/javaslang/Value.java
```diff
@@ -109,16 +109,6 @@
...
-     * Shortcut for {@code exists(e -> Objects.equals(e, element))}, tests if the given {@code element} is contained.
-     *
-     * @param element An Object of type A, may be null.
-     * @return true, if element is contained, false otherwise.
-     */
-    default boolean contains(T element) {
-        return exists(e -> Objects.equals(e, element));
-    }
-
-    /**
```
danieldietrich: Yes, very good to move it to Traversable again. I did it, too, in a local branch.
The javadoc line `<li>{@link #contains(Object)}</li>` needs to be removed in class `Value`

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -1003,8 +1003,8 @@ public Character get(int index) {
     }
 
     @Override
-    public int indexOf(Character element, int from) {
-        return back.indexOf(element, from);
+    public int indexOf(Object element, int from) {
```
danieldietrich: I think `element` can stay of type `Character` because `Character` is a final class. The we do not need the cast in the line below.

javaslang/src/main/java/javaslang/collection/HashArrayMappedTrie.java
```diff
@@ -117,13 +117,14 @@ static int fromBitmap(int bitmap, int bit) {
 
         abstract AbstractNode<K, V> modify(int shift, int keyHashCode, K key, V value, Action action);
 
+        @SuppressWarnings("unchecked")
         @Override
-        public Option<V> get(K key) {
-            return lookup(0, Objects.hashCode(key), key);
+        public Option<V> get(Object key) {
+            return lookup(0, Objects.hashCode(key), (K) key);
```
danieldietrich: The cast `(K) key` may fail, right? But a check with `key instanceof K` does also not work because of type erasure.

javaslang/src/main/java/javaslang/collection/HashArrayMappedTrie.java
```diff
@@ -132,9 +133,10 @@ public boolean containsKey(K key) {
             return modify(0, Objects.hashCode(key), key, value, PUT);
         }
 
+        @SuppressWarnings("unchecked")
         @Override
-        public HashArrayMappedTrie<K, V> remove(K key) {
-            return modify(0, Objects.hashCode(key), key, null, REMOVE);
+        public HashArrayMappedTrie<K, V> remove(Object key) {
+            return modify(0, Objects.hashCode(key), (K) key, null, REMOVE);
```
danieldietrich: cast here also unsafe

javaslang/src/main/java/javaslang/collection/Map.java
```diff
@@ -186,9 +186,11 @@ default V apply(K key) {
     @Override
     Map<K, V> clear();
 
+    @SuppressWarnings("unchecked")
     @Override
-    default boolean contains(Tuple2<K, V> element) {
-        return get(element._1).map(v -> Objects.equals(v, element._2)).getOrElse(false);
+    default boolean contains(Object element) {
+        Tuple2<K, V> tuple = (Tuple2<K, V>) element;
```
danieldietrich: This cast is also unsafe. Tuple2 is final, therefore we don't need to use Object here I think.

javaslang/src/main/java/javaslang/collection/TreeMap.java
```diff
@@ -420,14 +420,15 @@ private TreeMap(RedBlackTree<Tuple2<K, V>> entries) {
         return createTreeMap(entries.comparator(), iterator().dropWhile(predicate));
     }
 
+    @SuppressWarnings("unchecked")
     @Override
-    public boolean containsKey(K key) {
+    public boolean containsKey(Object key) {
         final V ignored = null;
-        return entries.contains(new Tuple2<>(key, ignored));
+        return entries.contains(new Tuple2<>((K) key, ignored));
```
danieldietrich: cast `(K) key` is unsafe

javaslang/src/main/java/javaslang/collection/TreeMap.java
```diff
@@ -456,10 +457,11 @@ public boolean containsValue(V value) {
         return iterator().foldRight(zero, f);
     }
 
+    @SuppressWarnings("unchecked")
     @Override
-    public Option<V> get(K key) {
+    public Option<V> get(Object key) {
         final V ignored = null;
-        return entries.find(new Tuple2<>(key, ignored)).map(Tuple2::_2);
+        return entries.find(new Tuple2<>((K) key, ignored)).map(Tuple2::_2);
```
danieldietrich: also unsafe cast

javaslang/src/main/java/javaslang/collection/TreeSet.java
danieldietrich: Yes.

I've looked into Java 8's `TreeMap` methods `getEntry(Object key)` and `getEntryUsingComparator(Object key)`. (TreeMap is internally used by TreeSet). There they do also unchecked cases like `(K) key` or `(Comparable<? super K>) key`.

Knowing this, the above class casts in this PR are fine, maybe...

danieldietrich:
Hi Ruslan, thank you! I think, currently we cannot get around most of the casts, Java 8 does it also in TreeMap etc. However, I think we can use contains(T) if we know that T is final, like for Character or Tuple2.

javaslang/src/main/java/javaslang/collection/Map.java
ruslansennov: but <code>contains(Tuple2<K, V> element)</code> not compiles because it override <code>contains(Object element)</code>

danieldietrich: ok, no problem, then contains(object)

javaslang/src/main/java/javaslang/collection/HashArrayMappedTrie.java
ruslansennov: Yes. We can relax signatures or make unsafe cast, but not both.
I don't like this PR, but I see no other possibility

danieldietrich:
See comments on the original issue:

* https://github.com/javaslang/javaslang/issues/1065#issuecomment-177153894
* https://github.com/javaslang/javaslang/issues/1065#issuecomment-177167063

danieldietrich:
Uh, oh - that was fast. Do you think the cast solution is better than using Objects?

ruslansennov:
I don't know :) I'll keep this branch undeleted

danieldietrich:
ok

PR #1081, opened 2016-02-02
Added narrow() to Values

danieldietrich:
Towards #1065

### DIFF
--- a/javaslang/src/main/java/javaslang/Lazy.java
+++ b/javaslang/src/main/java/javaslang/Lazy.java
@@ -60,6 +60,20 @@ public final class Lazy<T> implements Value<T>, Supplier<T>, Serializable {
     }
 
     /**
+     * Narrows a widened {@code Lazy<? extends T>} to {@code Lazy<T>}
+     * by performing a type safe-cast. This is eligible because immutable/read-only
+     * collections are covariant.
+     *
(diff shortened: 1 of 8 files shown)


danieldietrich:
Note: The following classes do _not_ have a narrow() method yet:

* Either.LeftProjection, Either.RightProjection
* Match.MatchValue, Match.MatchFunction

These are intended to be used as intermediate objects. narrow() may be added later for these if needed.

catostrophe:
@danieldietrich why did you choose such a name (`narrow`) for this operation?

In `scalaz` it's `widen`:
https://github.com/scalaz/scalaz/blob/80ba9d879b4f80f0175b5f904ac4587b02400251/core/src/main/scala/scalaz/Functor.scala#L92-L97

In `cats` it's `widen`:
https://github.com/typelevel/cats/blob/88ccb7a67b884eecea8d41e7fa735d23f5788b5c/core/src/main/scala/cats/Functor.scala#L37-L58

In `arrow-kt` it's `widen`:
https://github.com/arrow-kt/arrow-core/blob/b9bac5a03562a8e792b95c976182e9094b8ae425/arrow-core-data/src/main/kotlin/arrow/typeclasses/Functor.kt#L227-L247

PR #1086, opened 2016-02-06
Match tests

ruslansennov:


danieldietrich:
Thank you Ruslan. I'm currently investigating structural-pattern-matching - a real stroke of luck! See #1087

PR #1088, opened 2016-02-06
Add search to Seq

netzwerg:
The **binary search** implementation in `IndexedSeq` is roughly ported from `java.util.Collections.binarySearch` (1.8.0_66). It deviates in the following points:
* Naming (seq vs. list, element vs. key)
* Code duplication avoided by abstracting over element comparison
* Fail fast upon `null` comparator

The **linear search** implementation in `LinearSeq` is roughly ported from `scala.collection.Searching.SearchImpl.linearSearch` (2.11.7). It uses an abstracted `comparison` in place of Scala's implicit `Ordering`.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/IndexedSeq.java
+++ b/javaslang/src/main/java/javaslang/collection/IndexedSeq.java
@@ -5,6 +5,7 @@
  */
 package javaslang.collection;
 
+import javaslang.Function1;
 import javaslang.Tuple2;
 import javaslang.Tuple3;
 import javaslang.control.Match;
@@ -366,6 +367,23 @@ public interface IndexedSeq<T> extends Seq<T> {
     @Override
     IndexedSeq<Tuple2<T, Long>> zipWithIndex();
 
+    @SuppressWarnings("unchecked")
+    default int search(T key) {
+        Function1<Integer, Integer> midToCmp = mid -> {
+            Comparable<? super T> midVal = (Comparable<? super T>) get(mid);
+            return midVal.compareTo(key);
+        };
+        return IndexedSeqModule.Search.binarySearch(this, midToCmp);
+    }
+
+    default int search(T key, Comparator<? super T> comparator) {
+        Function1<Integer, Integer> midToCmp = mid -> {
+            T midVal = get(mid);
+            return comparator.compare(midVal, key);
+        };
+        return IndexedSeqModule.Search.binarySearch(this, midToCmp);
+    }
+
 }
 
 interface IndexedSeqModule {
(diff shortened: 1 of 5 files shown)


danieldietrich:
Hi Rahel, mir fehlen die Worte im Englischen... Perfekt ausgeführt und klasse dokumentiert! Ich würde dich sofort einstellen, wenn ich eine Firma hätte :-))
Vielen Dank!

netzwerg:
Wow, danke, das freut mich sehr! :-)

javaslang/src/main/java/javaslang/collection/IndexedSeq.java
```diff
@@ -409,4 +452,24 @@ default boolean startsWith(Iterable<? extends T> that, int offset) {
             return -1;
         }
     }
+
+    interface Search {
+        static <T> int binarySearch(IndexedSeq<T> seq, Function1<Integer, Integer> comparison) {
```
danieldietrich: I've searched the API docs and think we can do a little tweak - replacing the `Function1<Integer, Integer>` with `IntUnaryOperator`. In our special case it would prevent auto-boxing/unboxing of int values.

javaslang/src/main/java/javaslang/collection/LinearSeq.java
```diff
@@ -348,5 +388,20 @@ default int segmentLength(Predicate<? super T> predicate, int from) {
             return null;
         }
     }
+    interface Search {
+        static <T> int linearSearch(LinearSeq<T> seq, Function1<T, Integer> comparison) {
```
danieldietrich: Accordingly this could be replaced by `ToIntFunction`. I think it is a only a slight improvement but then we got the most out of it.

I was recently asked why I do not always use Javaslang Functions myself (within Javaslang). For consumers I try to use the most general function, mostly JDK API, like Function, BiFunction, etc. Producers return the most specfic types, E.g. Seq extends `Function1<Integer, T>` instead of `Function<Integer, T>`. Could be also `IntFunction`. These @FunctionalInterfaces drive me crazy :-) I wished there were no primitive types in Java and that Function1 and Function were the same.

javaslang/src/main/java/javaslang/collection/IndexedSeq.java
netzwerg: Good catch!

It's a little embarrassing: I quickly thought of this but couldn't find an adequate interface, moved on, and forgot – `IntUnaryOperator`, what a stupid name?!

javaslang/src/main/java/javaslang/collection/LinearSeq.java
netzwerg: :+1:

PR #1089, opened 2016-02-07
Prevent auto-(un)boxing in Seq.search

netzwerg:
Improves #1088


danieldietrich:
Good morning & thank you!

Issue #1103, opened 2016-02-13
Wrong generics on Tree.transform()


lukaseder:
The current `Tree.transform()` method has a wrong generic type signature:

```java
// Is
<U> U transform(Function<? super Tree<? super T>, ? extends U> f);

// Should be
<U> U transform(Function<? super Tree<T>, ? extends U> f);
```

The following code currently doesn't compile, but would compile if you adapted the signature:

```java
Node<Integer> tree = Tree.of(1, Tree.of(2, Tree.of(3), Tree.of(4)), Tree.of(5, Tree.of(6)));

Function<Tree<Integer>, Tree<String>> toStringInt = i -> i.map(Objects::toString);
Function<Tree<? extends Number>, Tree<String>> toStringNumber = i -> i.map(Objects::toString);
Function<Tree<?>, Tree<String>> toStringAny = i -> i.map(Objects::toString);

System.out.println(tree.transform(toStringInt)); // Doesn't compile
System.out.println(tree.transform(toStringNumber)); // Doesn't compile
System.out.println(tree.transform(toStringAny));
```

I'm not sure what value the `? super T` currently adds... Is there any specific use-case you've had in mind?

I haven't checked other `transform()` methods, but the same thing probably applies.

danieldietrich:
Good morning!
Yes thanks, is saw that yesterday and filed #1100 - the current signature makes no sense.
I think in our case we can even replace `? extends U` with `U`, because method return types are co-variant in Java - but I'm not 100% sure.

lukaseder:
> I think in our case we can even replace ? extends U with U, because method return types are co-variant in Java - but I'm not 100% sure.

Those are differemt beasts. Right now, I can write:

```java
Node<Integer> tree = Tree.of(1, Tree.of(2, Tree.of(3), Tree.of(4)), Tree.of(5, Tree.of(6)));
Function<Tree<?>, Tree<Object>> toStringAny = i -> i.map(Objects::toString);
//                     ^^^^^^ Not String
System.out.println(tree.transform(toStringAny));
```

This wouldn't be possible if you change `? extends U` to `U`

danieldietrich:
Yep, you are right!

danieldietrich:
Duplicate to #1100 but contains additional information

danieldietrich:
Fixed with #1104

Issue #1100, opened 2016-02-12
Fix transform methods signatures


danieldietrich:
Our transform methods are currently not usable.

Example:

```java
// defined in Seq
default <U> U transform(Function<? super Seq<? super T>, ? extends U> f)

// defined in Option
default <U> U transform(Function<? super Option<? super T>, ? extends U> f)

...
```

**1st fix:** All transform methods have to be changed to this signature:

```java
// defined in Seq
default <U> U transform(Function<? super Seq<T>, ? extends U> f)

// defined in Option
default <U> U transform(Function<? super Option<T>, ? extends U> f)

...
```

**2nd fix:** move the transform of _Traversables_(!) to the end-points of the inheritance tree:

```java
// defined in Stream, List, Array, LinkedHashMap, HashMap, Tree, ...
default <U> U transform(Function<? super Stream<T>, ? extends U> f) { ... }
default <U> U transform(Function<? super List<T>, ? extends U> f) { ... }
...
```

![collections](https://cloud.githubusercontent.com/assets/743833/13020493/fe27d692-d1d5-11e5-8d4a-2fec2ac266ff.png)

/cc @ruslansennov I updated `U` to `? extends U` above. See #1103

danieldietrich:
Ruslan, would be great if you could do that!

I'm finishing meanwhile the structural pattern matching.

(Currently feeling like:
![double_espresso](https://cloud.githubusercontent.com/assets/743833/13020889/22d025ba-d1d8-11e5-9e8d-45d2b137aa51.png)
)

danieldietrich:
Fixed with #1104

PR #1104, opened 2016-02-13
transform() signatures #1100 #1103

ruslansennov:


javaslang/src/main/java/javaslang/collection/Stack.java
```diff
@@ -719,6 +719,8 @@
     @Override
     Stack<T> takeWhile(Predicate<? super T> predicate);
 
+    <U> U transform(Function<? super List<T>, ? extends U> f);
```
ruslansennov: Here we should use <code>List</code> in generic

danieldietrich:
Awesome! Thx

spock_0

PR #1107, opened 2016-02-14
Skip evaluation of N+1 when taking N elements of a Stream

jmdesprez:
Hello,

I've made a little change in the take method of Stream to skip the evaluation of tail() if the current element is the last one (a.k.a. N == 1).

Evaluation of tail() can cause performance issue if the Stream is complicated to generate. 

It also can cause infinite recursion in some case. For example when you filter a stream and take all of the N elements corresponding to that filter then take(N) will try to find a N+1 element which doesn't exist (https://projecteuler.net/problem=37 is a good example as there only 11 Truncatable primes and take(11) try to find a 12th truncable prime)

Disclaimer: I can't get javaslang to compile on my computer so I've test the modification on a side project.

Jean-Marc


danieldietrich:
Awesome, thank you!

PR #1124, opened 2016-02-20
Performing numberic operations of non-fraction numbers on Long

danieldietrich:


danieldietrich:
Sum and product have to take place using Long values for Integer, Short and Byte. Switching to Long revealed a bug in our Euler08 solution (right solution also here: http://blog.dreamshire.com/project-euler-8-solution/).

Example: Let's multiply 5 byte numbers

```java
255 * 255 * 255 * 255 * 2 = 8_456_501_250
```

Issue #1132, opened 2016-02-24
reduce common code


ruslansennov:
This work fine for <code>Map.partition(Predicate)</code>. It seems we can do it for million <code>Map</code> methods (and maybe for <code>Traversable</code>)
```java
/*package*/ abstract class AbstractMap<K, V, M extends AbstractMap<K, V, ?>> implements Map<K, V> {

    abstract M wrapEntries(Iterable<? extends Tuple2<? extends K, ? extends V>> entries);

    @Override
    public Tuple2<M, M> partition(Predicate<? super Tuple2<K, V>> predicate) {
        Objects.requireNonNull(predicate, "predicate is null");
        final Tuple2<Iterator<Tuple2<K, V>>, Iterator<Tuple2<K, V>>> p = iterator().partition(predicate);
        return Tuple.of(wrapEntries(p._1), wrapEntries(p._2));
    }
}

public final class HashMap<K, V> extends AbstractMap<K, V, HashMap<K, V>> implements Serializable {
    @Override
    HashMap<K, V> wrapEntries(Iterable<? extends Tuple2<? extends K, ? extends V>> entries) {
        return HashMap.ofEntries(entries);
    }

//    @Override
//    public Tuple2<HashMap<K, V>, HashMap<K, V>> partition(Predicate<? super Tuple2<K, V>> predicate) {
//        Objects.requireNonNull(predicate, "predicate is null");
//        final Tuple2<Iterator<Tuple2<K, V>>, Iterator<Tuple2<K, V>>> p = iterator().partition(predicate);
//        return Tuple.of(HashMap.ofEntries(p._1), HashMap.ofEntries(p._2));
//    }
}

public final class TreeMap<K, V> extends AbstractMap<K, V, TreeMap<K, V>> implements SortedMap<K, V>, Serializable {
    @Override
    TreeMap<K, V> wrapEntries(Iterable<? extends Tuple2<? extends K, ? extends V>> it) {
        return createTreeMap(entries.comparator(), it);
    }

//    @Override
//    public Tuple2<TreeMap<K, V>, TreeMap<K, V>> partition(Predicate<? super Tuple2<K, V>> predicate) {
//        Objects.requireNonNull(predicate, "predicate is null");
//        final Tuple2<Iterator<Tuple2<K, V>>, Iterator<Tuple2<K, V>>> p = iterator().partition(predicate);
//        final TreeMap<K, V> treeMap1 = createTreeMap(entries.comparator(), p._1);
[...]

PR #1136, opened 2016-02-24
introducing of AbstractMap #1132

ruslansennov:


danieldietrich:
Awesome! I will pull this in as refactoring, no comments here.

I put a special attention on those methods in abstract Map that don't use the generic type parameter M. But all methods fit there well. I had to think a while about the `distinct()`. Isn't it really bound to the specific Map implementation? But no - keys have to be a (Sorted)Set., so entries are distinct by default. However, the distinct method can't be moved as default method to the Map interface because the specific Map type isn't known there. Well done!

From another viewpoint than refactoring I identified minor changes that I will apply afterwards. E.g. a `isTraversableAgain()` should be moved to the Map interface because a Map would be pretty useless, if I could ask it only once for a value by a key.

Many thanks for this big effort!

PR #1146, opened 2016-02-27
Abstract remove

ruslansennov:


javaslang/src/main/java/javaslang/collection/AbstractMap.java
```diff
@@ -41,6 +41,24 @@
...
     @SuppressWarnings("unchecked")
     @Override
+    public M put(Tuple2<? extends K, ? extends V> entry) {
+        Objects.requireNonNull(entry, "entry is null");
+        return (M) put(entry._1, entry._2);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public M removeAll(Iterable<? extends K> keys) {
```
danieldietrich: I would not move this method to AbstractMap. For each remove corresponds to a `wrap(delegate.remove(element))`. But we should first `delegate.remove(element)` for all elements and finally wrap the result (once). This saves n intermediate instances.

ruslansennov: You're right. I'll prepare another PR

danieldietrich: Thanks. I know - a lot of work with the repetitions in the collection classes o_O

PR #1155, opened 2016-02-29
Added generator for enum and array types.

malduarte:
I've been using enum/array based generators and thought they might fit into the main library.

Other than that, I'd like to thank all javaslang contributors for this beautiful library.

PR #1158, opened 2016-03-01
Added a choose from iterable.

malduarte:
Also, renamed generic parameters introduced in https://github.com/javaslang/javaslang/pull/1155 to T for consistency.


danieldietrich:
looks nice, thx!

PR #1159, opened 2016-03-01
Fixed status indicators.

malduarte:
- changed package to com.javaslang to match the real groupid in maven
- changed icon types to .png (png is more widely available).


danieldietrich:
Thank you - right, the current readme should always provide working links.

Issue #1147, opened 2016-02-27
Classes should implement Kind1/Kind2 where applicable


danieldietrich:
This does not change the functionality but gives more flexibility in designing APIs on top of Javaslang classes.

danieldietrich:
Not relevant for 2.0.0, will move it to 2.0.1

danieldietrich:
Kind* will be removed in 3.0, see #1493

PR #1160, opened 2016-03-01
Collections implements Kind1/Kind2 #1147

ruslansennov:


javaslang/src/main/java/javaslang/collection/Stack.java
```diff
@@ -38,7 +39,7 @@
  * @author Daniel Dietrich
  * @since 2.0.0
  */
-public interface Stack<T> extends LinearSeq<T> {
+public interface Stack<T> extends Kind1<List<?>, T>, LinearSeq<T> {
```
ruslansennov: @danieldietrich :cry:

danieldietrich: I will make Stack an ordinary interface having only a few Stack-related methods and which don't extends LinearSeq. A few people discussed about that on gitter and I made a poll. Scala has removed it entirely but having a simple interface should be ok. I will take care of the javadoc in order not to confuse users whether to to use prepend or push on a List...

ruslansennov: I voted against the stack :)

danieldietrich: haha, I was also tempted to remove it entirely... but we will go with a thin interface.

PR #1165, opened 2016-03-03
suggestion for static sorted method on Seq

mvh77:
What do you think of replacing Seq#sorted() with a static method like this? Reasoning is that it's not a good idea to have a runtime ClassCastException if the elements of the Seq are not comparable, since we can't do implicit type constraints like in Scala. This one method would work for all subclasses of Seq as long as their sorted(Comparator) method returns the same instance as they are.

If you think it's good I'll remove the other one, rename this one properly and transform the test cases. Are there other methods that would benefit from a similar treatment (that throw ClassCastException)?

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Seq.java
+++ b/javaslang/src/main/java/javaslang/collection/Seq.java
@@ -109,6 +109,11 @@ public interface Seq<T> extends Traversable<T>, Function1<Integer, T> {
         return (Seq<T>) seq;
     }
 
+    @SuppressWarnings("unchecked")
+    static <T extends Comparable<? super T>, S extends Seq<T>> S sorted2(S seq) {
+        return (S) seq.sorted(Comparator.naturalOrder());
+    }
+
     /**
      * A {@code Seq} is a partial function which returns the element at the specified index by calling
      * {@linkplain #get(int)}.


danieldietrich:
Hi Martti,

it was a hard decision to make some of the API unsafe. I've thoroughly though about the safety topic while designing the API. There are too many of these methods. All numeric operations in collections (product, average, sum). All candidates xxx that have an xxxOption alternative might also fail, e.g. head, init, last, tail, reduce. The get method of all Values (e.g. Option, Try, ...) might throw NoSuchElementException - partially defined functions and so on. I bet there are more examples.

I chose an API design that is practicable in terms of conciseness. The unsafety of ordered collections is already from java.util.SortedSet. But there are also examples that were too unsafe and I removed them again. An example is the flatten() method which assumed that a List has type `List<List<T>>` for example. A static flatten method is possible but we don't have it right now.

These static methods are hard to maintain. They follow not the rules of inheritance. Every interface/class has to define their own. Regarding the increasing amount of collection impls this is really cumbersome work.

Let's keep it as it is - not optimal but practicable. There are other libs for Java that try to emulate real functional languages. But I think this far from being practicable. The code is hard to write and hard to read.

Are you fine with this or does it feel wrong?

mvh77:
Hi,

Ok, I'm not suggesting changing every method that can throw an exception, just maybe the ones that throw ClassCastException. This is because changing a type shouldn't lead to runtime exceptions. Of course Option#get() has to throw an exception if the option is not defined, this is the way it's in Scala too. But if you feel there are too many of these and it makes it difficult, let's keep it as it is.

mvh77:
One option is to just remove the sorted() method since you can do this and it's type safe:
```java
Vector<Integer> vec = Vector.of(1,2,3);
vec.sorted(Comparator.naturalOrder());
```

If you change the type of vec to Object, the sorted call doesn't compile anymore.

danieldietrich:
You are right, that would be safe. But I want to provide the same API as introduced with Java 8: [Stream.sorted()](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#sorted--)

From the Java 8 API docs:

> For ordered streams, the sort is stable. For unordered streams, no stability guarantees are made.

We have stated this, too, at [our interface](http://static.javadoc.io/com.javaslang/javaslang/2.0.0-RC4/javaslang/collection/Seq.html#sorted--):

> throws ClassCastException - if this elements are not Comparable

danieldietrich:
Let's leave this behind us. It is not worth to put too much effort in it. Java needs inference - otherwise all looks just as a workaround, the one or the other way.

I will close this ticket.

mvh77:
Ok, no problem. I don't think following Java8 api's too closely is a very good idea, but I understand your point. :wink:

danieldietrich:
:+1:

Issue #1087, opened 2016-02-06
Structural Pattern Matching


danieldietrich:
---

**Note:** I know that this will be a big change, directly before the release of 2.0.0. But Scala-like structural pattern matching for Java will be a _real game changer_ if done right! We need to do this - now. Changing the Match API directly after 2.0.0 is no viable solution.

---

```java
List<Tuple3<String, Integer, Double> list = ...;

//
// x  is of type Tuple3<String, Integer, Double>
// xs is of type List<Tuple3<String, Integer, Double>>
//
list.match()
    .when(List($(), $()))
    .then((x, xs) -> ...);

// Match.of(list) also works

//
// s is of type String
//
list.match()
    .when(List(Tuple3($("begin"), _, _), _))
    .then(s -> ...);
```

**The basic syntax**

* `Match.when(pattern).then(function)`
* There will be _atomic patterns_,
  - named any-value: `$()`, e.g. `when(List($(), $())).then((x, xs) -> ...)`
  - named specifc-value: `$("test")`, e.g. `when(List($(1), List.nil())).then(one -> ...)`
  - ignoring any-matcher: `_`, e.g. `when(List(_, $())).then(tail -> ...)`
* There will be _patterns_ for types, e.g. `List(...)`, all upper-case
* There will be _matchers_ like `is(...)`, `isIn(...)`, all lower-case

**Changes to the current Match API:**

* The Match implementation will get simpler. The core will fit on one page + generated duplication for arity 0..8 of pattern return values, all methods literally one-liners.
* MatchFunction will pass away. Maybe we can replace it by `o -> Match.of(o).when(...).then(...)` or by `Match.of(o).when(...).then(...).toFunction()` if a function is needed.
* Never liked the fact that Match grew to an internal DSL, like whenIs, whenIsIn, whenType, whenThis, whenThat. We see similar things in a wide range of other libs, like when().is(), when().isIn(), ... Instead we will _compose_ the _match expression_ with other _functions_ like this:

    ```java
    list.match()
        .when(is(val))
        .then(v -> ...);

    list.match()
        .when(isIn(val1, val2, val3))
        .then(v -> ...);
    ```

**Match will gain importance**

* it will move to the first level package `javaslang`
* it will be as easy to use as 1-2-3.

    ```java
    import static javaslang.Match.*; // standard patterns $_, $(), $(val), is(), isIn(), type(), ...
[...]

PR #1167, opened 2016-03-04
Made structural pattern matching practicable...

danieldietrich:
...by reducing the amount of generated methods (per unapply method) from `11^n` to `1`, where `n` is the number of unapplied fields of an object.

Now, object trees of arbitrary(!) depth can be matched but the decomposition takes place in the first level of the object tree (the given object is level 0).

Before we could select arbitrary nodes of the object tree. That is not possible any more because it requires the patterns to return the actual arity using tuples - and therefore all possible return combinations had to be generated. This is simply not possible resp. practicable in a real-world application and soon exceeds the class file size. However, this is solved with this PR!!!


danieldietrich:
Towards #1087

PR #1168, opened 2016-03-05
<w00t>Extracting extended types of decomposed parts.</w00t>

danieldietrich:


danieldietrich:
Towards #1087

danieldietrich:
`<unw00t>Subtypes still not extracted right for deeper nested examples.</unw00t>`

PR #1175, opened 2016-03-07
Added some tests for predicates and options

kmarwood:
Not sure what the expected behaviour is when calling transform on None.

### DIFF
--- a/javaslang/src/test/java/javaslang/control/OptionTest.java
+++ b/javaslang/src/test/java/javaslang/control/OptionTest.java
@@ -397,10 +397,7 @@ public class OptionTest extends AbstractValueTest {
 
     @Test
     public void shouldHandleTransformOnNone() {
-        // TODO: What is the expected behavior when transform is called on None given return type may or may not be an Option?
-        // Calling None.get() will throw NoSuchElementException
-        // Should it be left to the caller to decide how to handle a value of None in their transform function?
-        // If not, should the default implementation of transform be overridden in None so that it is handled consistently?
+        assertThat(Option.none().<String>transform(self -> self.isEmpty() ? "ok" : "failed")).isEqualTo("ok");
     }
 
     // -- iterator
--- a/javaslang/src/test/java/javaslang/control/OptionTest.java
+++ b/javaslang/src/test/java/javaslang/control/OptionTest.java
@@ -11,6 +11,7 @@ import javaslang.collection.Seq;
 import org.junit.Test;
 
 import java.util.*;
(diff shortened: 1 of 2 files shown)


danieldietrich:
Very nice, thang you :)

javaslang/src/test/java/javaslang/control/OptionTest.java
```diff
@@ -379,6 +380,29 @@ public void shouldNotConsumeAnythingOnPeekWhenValueIsNotDefined() {
...
+        Function<Option<Integer>, String> f = o -> o.get().toString().concat("-transformed");
+        assertThat(option.<String>transform(f)).isEqualTo("1-transformed");
+    }
+
+    @Test
+    public void shouldHandleTransformOnNone() {
+        // TODO: What is the expected behavior when transform is called on None given return type may or may not be an Option?
+        // Calling None.get() will throw NoSuchElementException
+        // Should it be left to the caller to decide how to handle a value of None in their transform function?
+        // If not, should the default implementation of transform be overridden in None so that it is handled consistently?
```
danieldietrich: `transform` is the swiss-army-knife of the methods. It allows us to return an object that is derived from the current object in a fluent manner. Not more, not less. It is up to the caller what type of object is returned.

**Example:**

Given an `Option<?> o = Option.none();`, instead of

```java
// non-fluent / imperative API
final String s;
if (o.isEmpty()) {
    s = "None";
} else {
    s = "Some";
}
String result = s.toUpperCase();
```

we write

```java
// fluent API
String result = o.transform(self -> self.isEmpty() ? "None" : "Some").toUpperCase();
```

danieldietrich: @kmarwood A good test for transform on None would be

```java
assertThat(Option.none().transform(self -> self.isEmpty() ? "ok" : "failed")).isEqualTo("ok");
```

Could you add that, please?

kmarwood:
@danieldietrich Thanks for the feedback and explanation! :-)

I will update the test as per your suggestion.

danieldietrich:
Thank you :-)

kmarwood:
Would your preference be for the new commit containing the additional test to be squashed into my previous commit to keep the commit history clean, or would you prefer a separate commit?

danieldietrich:
As simple as possible here because it is just a test. You may choose.

danieldietrich:
Thx!

kmarwood:
My pleasure!

PR #1180, opened 2016-03-07
Try.isEmpty() returns false even when it's a Success with null value

ggalmazor:
It should return true to comply with: 
```Option.of(null).isDefined == Try.of(()->null).toOption().isDefined()```

This is also broken on 1.x branch. I can prepare a patch for that as well

Also, should ```Try.isEmpty()``` check for types that have an ```isEmpty()``` (like String, List, etc.) implementation and delegate the results? Problem is that there is no common hierarchic ancestor that guarantees the presence of such method. We could just make the "best effort" with known types.

### DIFF
--- a/javaslang/src/main/java/javaslang/control/Try.java
+++ b/javaslang/src/main/java/javaslang/control/Try.java
@@ -29,896 +29,896 @@ import java.util.function.Supplier;
  */
 public interface Try<T> extends Value<T> {
 
-  /**
-   * Creates a Try of a CheckedSupplier.
-   *
-   * @param supplier A checked supplier
-   * @param <T>      Component type
-   * @return {@code Success(supplier.get())} if no exception occurs, otherwise {@code Failure(throwable)} if an
-   * exception occurs calling {@code supplier.get()}.
-   */
-  static <T> Try<T> of(CheckedSupplier<? extends T> supplier) {
-    try {
-      return new Success<>(supplier.get());
-    } catch (Throwable t) {
-      return new Failure<>(t);
-    }
-  }
-
-  /**
-   * Creates a Try of a CheckedRunnable.
-   *
-   * @param runnable A checked runnable
-   * @return {@code Success(null)} if no exception occurs, otherwise {@code Failure(throwable)} if an exception occurs
-   * calling {@code runnable.run()}.
-   */
-  static Try<Void> run(CheckedRunnable runnable) {
-    try {
-      runnable.run();
-      return new Success<>(null); // null represents the absence of an value, i.e. Void
-    } catch (Throwable t) {
-      return new Failure<>(t);
-    }
-  }
-
-  /**
-   * Reduces many {@code Try}s into a single {@code Try} by transforming an
-   * {@code Iterable<Try<? extends T>>} into a {@code Try<Seq<T>>}. If any of
-   * the {@code Try}s are {@link Try.Failure}, then this returns a {@link Try.Failure}.
-   *
-   * @param values An {@link Iterable} of {@code Try}s
-   * @param <T>    type of the Trys
-   * @return A {@code Try} of a {@link Seq} of results
-   * @throws NullPointerException if {@code values} is null
-   */
-  static <T> Try<Seq<T>> sequence(Iterable<? extends Try<? extends T>> values) {
-    Objects.requireNonNull(values, "values is null");
-    List<T> list = List.empty();
-    for (Try<? extends T> value : values) {
-      if (value.isFailure()) {
-        return Try.failure(value.getCause());
-      }
-      list = list.prepend(value.get());
-    }
-
-    return Try.success(list.reverse());
-  }
-
-  /**
-   * Creates a {@link Success} that contains the given {@code value}. Shortcut for {@code new Success<>(value)}.
-   *
-   * @param value A value.
-   * @param <T>   Type of the given {@code value}.
-   * @return A new {@code Success}.
-   */
-  static <T> Try<T> success(T value) {
-    return new Success<>(value);
-  }
-
-  /**
-   * Creates a {@link Failure} that contains the given {@code exception}. Shortcut for {@code new Failure<>(exception)}.
-   *
-   * @param exception An exception.
-   * @param <T>       Component type of the {@code Try}.
-   * @return A new {@code Failure}.
-   */
-  static <T> Try<T> failure(Throwable exception) {
-    return new Failure<>(exception);
-  }
-
-  /**
-   * Narrows a widened {@code Try<? extends T>} to {@code Try<T>}
-   * by performing a type safe-cast. This is eligible because immutable/read-only
-   * collections are covariant.
-   *
-   * @param t   A {@code Try}.
-   * @param <T> Component type of the {@code Try}.
-   * @return the given {@code t} instance as narrowed type {@code Try<T>}.
-   */
-  @SuppressWarnings("unchecked")
-  static <T> Try<T> narrow(Try<? extends T> t) {
-    return (Try<T>) t;
-  }
-
-  /**
-   * Shortcut for {@code andThenTry(consumer::accept)}, see {@link #andThenTry(CheckedConsumer)}.
-   *
-   * @param consumer A consumer
-   * @return this {@code Try} if this is a {@code Failure} or the consumer succeeded, otherwise the
-   * {@code Failure} of the consumption.
-   * @throws NullPointerException if {@code consumer} is null
-   */
-  default Try<T> andThen(Consumer<? super T> consumer) {
-    Objects.requireNonNull(consumer, "consumer is null");
-    return andThenTry(consumer::accept);
-  }
-
-  /**
-   * Passes the result to the given {@code consumer} if this is a {@code Success}.
-   * <p>
-   * The main use case is chaining checked functions using method references:
-   * <p>
-   * <pre>
-   * <code>
-   * Try.of(() -&gt; 100)
-   *    .andThen(i -&gt; System.out.println(i));
-   *
-   * </code>
-   * </pre>
-   *
-   * @param consumer A checked consumer
-   * @return this {@code Try} if this is a {@code Failure} or the consumer succeeded, otherwise the
-   * {@code Failure} of the consumption.
-   * @throws NullPointerException if {@code consumer} is null
-   */
-  default Try<T> andThenTry(CheckedConsumer<? super T> consumer) {
-    Objects.requireNonNull(consumer, "consumer is null");
-    if (isFailure()) {
-      return this;
-    } else {
-      try {
-        consumer.accept(get());
-        return this;
-      } catch (Throwable t) {
-        return new Failure<>(t);
-      }
-    }
-  }
-
-  /**
-   * Shortcut for {@code andThenTry(runnable::run)}, see {@link #andThenTry(CheckedRunnable)}.
-   *
-   * @param runnable A runnable
-   * @return this {@code Try} if this is a {@code Failure} or the runnable succeeded, otherwise the
-   * {@code Failure} of the run.
-   * @throws NullPointerException if {@code runnable} is null
-   */
-  default Try<T> andThen(Runnable runnable) {
-    Objects.requireNonNull(runnable, "runnable is null");
-    return andThenTry(runnable::run);
-  }
-
-  /**
-   * Runs the given runnable if this is a {@code Success}, otherwise returns this {@code Failure}.
-   * <p>
-   * The main use case is chaining runnables using method references:
-   * <p>
-   * <pre>
-   * <code>
-   * Try.run(A::methodRef).andThen(B::methodRef).andThen(C::methodRef);
-   * </code>
-   * </pre>
-   * <p>
-   * Please note that these lines are semantically the same:
-   * <p>
-   * <pre>
-   * <code>
-   * Try.run(this::doStuff)
-   *    .andThen(this::doMoreStuff)
-   *    .andThen(this::doEvenMoreStuff);
-   *
-   * Try.run(() -&gt; {
-   *     doStuff();
-   *     doMoreStuff();
-   *     doEvenMoreStuff();
-   * });
-   * </code>
-   * </pre>
-   *
-   * @param runnable A checked runnable
-   * @return this {@code Try} if this is a {@code Failure} or the runnable succeeded, otherwise the
-   * {@code Failure} of the run.
-   * @throws NullPointerException if {@code runnable} is null
-   */
-  default Try<T> andThenTry(CheckedRunnable runnable) {
-    Objects.requireNonNull(runnable, "runnable is null");
-    if (isFailure()) {
-      return this;
-    } else {
-      try {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by danieldietrich** [comment]
Hi Guillermo,

no. :-) That is contrary to the whole idea of the Try monad. The state of Try only depends on a success or a failure of a computation, reflected by the types Success and Failure. `null` has nothing to do with that state. Success(null).toOption() is Some(null), which is also allowed in Scala, see below.

But after performing some checks on the REPL If recognized that the name `isEmpty` might be confusing. Scala's Try does not have it.

```java
scala> import scala.util.Try                                                    
import scala.util.Try                                                           
                                                                                
scala> Try(null)                                                                
res1: scala.util.Try[Null] = Success(null)                                      
                                                                                
scala> Try(null).toOption                                                       
res2: Option[Null] = Some(null)                                                 
                                                                                
scala> Try(null).toOption.isDefined                                             
res3: Boolean = true                                                            
                                                                                
scala> Try(null).isEmpty                                                        
<console>:9: error: value isEmpty is not a member of scala.util.Try[Null]       
              Try(null).isEmpty                                                 
                        ^                                                       
                                                                                
scala> Try(null).toOption.isEmpty                                               
res5: Boolean = false                                                           
```

I introduced Value.isEmpty() in Javaslang to have common property of all Value types (Lazy, Option, Future, Try, Either,... and collections) that reflects a 'negative' state, not to be confused with _undefined_.

Here are some example what isEmpty looks like in different type:

* Option.isEmpty() == (option == Option.None.instance())
* Either.isLeft() == Either.isEmpty()
* Try.isFailure() == Try.isEmpty()
* Collection.isEmpty() == (Collection.length() == 0)
* Validation.isEmpty() == Validation.isInvalid()
* Lazy.isEmpty() == false // by definition

Additionally all Values have a method Value.isDefined() == !Value.isEmpty().

Now I see, that the name `isEmpty` might be confusing. But I think it is correct. Lets look at the meanings:

A Value wraps an instance, the so called _underlying value_. The javadoc of `Value.isEmpty()` says:

> Checks, this Value is empty, i.e. if the underlying value is absent.

* If a Try computation failed, we have no resulting underlying value. We can say that the resulting Try instance is empty. Instead we have an exception.
* If a Validation result is invalid, the Applicative did not apply a function to produce an underlying value, the result is empty. Instead we have errors of a different type.
* A lazy value (Lazy) is always defined. The evaluation of that value is deferred.
* A collection containing no values is empty.
* A Future that failed is empty because it contains no result. Instead it contains the exception that occurred.
* etc.

If we want to model the empty state as property, it needs a name. It might differ from Scala, which has no common Value type but why should we not choose the name `isEmpty`?

Does this make sense to you?

ggalmazor:
Hi Daniel! I understand and next time I'll check Scala for reference :) I'm closing the PR.

Anyway, my problem is not with ```isEmpty()```. My pull request modified it only because it's used by ```toOption()```.

My problem is that I have a computation that runs inside a ```Try``` which can return a null. I can't get rid of that because I'm integrating with a third party here. I want to wrap the result in an ```Option``` for further use of the result but then I may have a ```Some(null)``` which can produce runtime bugs in my app. 

I know that Scala's ```Try(null).toOption``` can give a ```Some(null)```, but this is kind of wrong, isn't it? The fix is an easy one: we could make ```Value.toOption()``` to go through ```Option.of()``` factory method when it's not empty to solve the problem:

```
    default Option<T> toOption() {
        if (this instanceof Option) {
            return (Option<T>) this;
        } else {
            return isEmpty() ? Option.none() : Option.of(get());
        }
    }
```

Researching for this reply I found this thread in StackOverflow that [explains why Some(null) is an acceptable value in Scala](http://stackoverflow.com/questions/5796616/why-somenull-isnt-considered-none) TLDR: It's just to achieve interoperability with Java (calls to methods that return null, accessors in Lists, Maps, etc.).

If this is true, we can decide to produce a ```None``` instead a ```Some(null)``` in ```Value.toOption()``` if the value is not empty. 

I mean... I never **ever** want a ```Some(null)``` in my code. 

If we don't fix this, I can't trust that the value of a ```Some``` is not null and we're back to the ```if (something == null)``` and that's a situation that I'd like to avoid (again).

ggalmazor:
Even better: We could:

```
    static <T> Option<T> some(T value) {
        return of(value);
    }
```

This would prevent a ```Some(null)``` from any source.

(Also, remove ```some(T value)``` factory)

danieldietrich:
We should not do that. I had a use case where I needed `Some<Void>`. `null` is the only instance here of `Void`. As long as null is part of Java, we can't completely get rid of it without breaking the logical structure or making look things artificial.

You could solve your use-case by using `transform` instead of `toOption`:

```java
Try(...).transform(self -> self.isSuccess() ? Option.of(self.get()) : Option.empty());
```

or write a helper for that

```java
Try(...).transform(FancyName::toOption);

class FancyName {
    static Function<? super Try<T>, ? extends Option<T>> toOption() {
        return it -> it.isSuccess() ? Option.of(it.get()) : Option.empty());
    }
}
```

ggalmazor:
Ok... That's what I'm doing right now.

I'd want to understand why you'd need a Some(null). Not trying to be rude... really :) Could you provide an example?

danieldietrich:
Yes, it is an important topic and it needs to be clarified, also in the docs.

I will provide an example later today, currently I only have a mobile phone at hand :)

The main point is that we need to preserve type-consistency. null is just a value, not more, not less. If we lift it into the context of a type (/state), conversion methods need to preserve this state.

image

We are not allowed to change this behavior. It wouldn't be logically (and algebraically?) correct any more. The special behavior you suggested would need to be implemented for ecery type in order to be consistent again. But that would not be natural any more. We would break the canonical way and add our own business logic (in a technical domain) because we think it does fit better for us.

I will give the example later...

danieldietrich:
These 'laws' come to my mind that should be obeyed by the toXxx conversion methods:

For all Javaslang Value type A, B, C and a &isin; A:

1) `a.toA()` == `a`
2) `a.toC()` equals `a.toB().toC()`
3) `a` equals `a.toB().toA()` // follows through 1) and 2)

or at least

1) `typeOf(a.toA())` == `typeOf(a)`
2) `typeOf(a.toC())` equals `typeOf(a.toB().toC())`

So let's imagine, we change the behavior of Try.toOption() in the way that Success(value).toOption() returns Option.of(value).

Then `Try.of(() -> null).toOption()` is `None`.

`Try.of(() -> null).toEither().toOption()` is `Right(null).toOption()`, which is `Some(null)`.

We could now change the behavior of toOption() for every Value implementation according to your suggestion. But then we would have

// singularity in the space-time continuum :-)
`Success(null).toOption().toTry()` == `Failure(NoSuchElementException)`

This might not be, what other users expect.

danieldietrich:
**Example:**

In this example I assume that there exists 3rd party code that puts null into maps. Or at least it is not prohibited to do so.

What Java currently looks like:

```java
Map<Integer, String> map = new HashMap<>();

// == false
map.containsKey(1);

// == null
map.get(1);

map.put(1, null);

// == true
map.containsKey(1);

// == null
map.get(1);
```

How we can do better with the use of `Some(null)`:

```java
Map<Integer, String> map = new HashMap<>();

// = None
safeGet(map, 1);

map.put(1, null);

// = Some(null)
safeGet(map, 1);
```

with

```java
static <K, V> Option<V> safeGet(Map<K, V> map, K key) {
    return map.containsKey(key) ? Option.some(map.get(key)) : Option.none();
}
```

Right?

See also ["Your codebase looks like this"](http://ghost-danieldietrich.rhcloud.com/your-codebase-looks-like-this/)

Another example is [an early version](https://github.com/javaslang/javaslang/blob/1f35cb8ba038453b1903908a83b7797466e5133e/javaslang/src-gen/main/java/javaslang/Match.java#L374) of the new structural pattern matching. However, I changed this.

danieldietrich:
And now let's imagine we add a new Value implementation that handles specific values in a specific way (like Option handles null in a specific way).

We would have to pull that specific behavior into all conversion methods `toXxx` to meet our special requirements. This does not feel right.

If we act like shown in my hand-drawing, there is no room for subjective interpretation of values. It is just as it is - a value put into a context of a type.

danieldietrich:
The key to my thoughts is:

I think it we cannot assume that `null` stands for 'absent'  or 'undefined' in every (use-)case because Java lets room for interpretation of that value.

That is a billion dollar mistake that is irreversible.

ruslansennov:
bender-applause_medium

danieldietrich:
:-))

danieldietrich:
This is the theory behind `Value`: [Dual category theory](https://en.wikipedia.org/wiki/Dual_%28category_theory%29)

ggalmazor:
(I haven't read the last links yet, sorry about that)

I've prepared [a test](https://github.com/ggalmazor/javaslang/blob/master/javaslang/src/test/java/javaslang/InteroperabilityTest.java) representing transformations between Option, Try and Either to reflect on what you've explained.

The test covers all possible transformations between these types according to the current implementation:

interoperability of some javaslang types

So, arrows and arrow points work as usual. Notice that None to Either gets inverted (in red) for obvious reasons.

Some things from the top of my head that I don't understand quite yet:

 - Sorry to throw your own quotation back at you but (I need to understand this):

> I think it we cannot assume that null stands for 'absent' or 'undefined' in every (use-)case because Java lets room for interpretation of that value.

Isn't the same problem to asume that None => Failure? (There is no line between None and Failure) Why is this different?

 - Left.toOption() doesn't do the same as Left.left().toOption() which is weird. [Is this a bug](https://github.com/ggalmazor/javaslang/blob/master/javaslang/src/test/java/javaslang/InteroperabilityTest.java#L92)?

 - Try.run().toOption() gives a Some(null) [which makes everything more strange](https://github.com/ggalmazor/javaslang/blob/master/javaslang/src/test/java/javaslang/InteroperabilityTest.java#L157)

danieldietrich:
> Sorry to throw your own quotation back at you but (I need to understand this):

Aaarrgh! Please, don't throw things at me. :-)

> I think it we cannot assume that null stands for 'absent' or 'undefined' in every (use-)case because Java lets room for interpretation of that value.

A simple example is the only instance `null` of the type `Void`. I needed such thing as valid extension of a specific domain in pattern matching. E.g.

* `Pattern2<T1, T2>` unapplies an object to `Option<T1, T2>` (None, if the pattern is not applicable)
* `Pattern1<T>` extracts `Option<T>` out of an object (we don't need the wrapper Tuple1 to save an instance)
* `Pattern0` extracts `Void` out of an object (theoretically it could be also Tuple0). The singleton instance `Some<Void>` is obtained by `Option.nothing()` in Javaslang. Formally an optional instance of nothing exists.

I flatMapped over these patterns and pattern0 needed to return `Some<Nothing>` to make it work...

> If we don't fix this, I can't trust that the value of a Some is not null and we're back to the if (something == null) and that's a situation that I'd like to avoid (again).

As long as null exists we can't even trust that the Some itself is not a null reference.

```java
Some<T> some = null; // ooops!
```

> Isn't the same problem to asume that None => Failure?

No, because we should not start to interpret things. The conversion needs to take place within the same space of types (the isEmpty == true space).

--

Either is a very special type. It is a _disjoint union type_. It consists of two Monad types, all other Values consist only of one Monad type! **I just saw that my picture is terribly WRONG.** The Left type is in the wrong place, sorry. Either itself is not a Monad. I recently changed it to be a **right-biased** Either. That means we treat a Right value as _defined_ and a Left value as _empty_. But in an unbiased version both Left and Right are defined and none is empty.

This is the reason for having toLeft(R alternative) and toRight(L alternative) conversion methods. They work [like in Scala](http://www.scala-lang.org/api/2.11.7/index.html#scala.None$) (you need to scroll down to see the methods). 

A second thing is, when we call left() or right() on an Either, we transform it to new types: LeftProjection and RightProjection. These act like the right-biased Either. From the viewpoint of a LeftProjection a Right is empty. From the viewpoint of a RightProjection a Left is empty.

Your questions:

1) left.toOption()

left is the empty state of our right-biased either, so left.toOption() is null.

2) left.left().toOption()

left.left() is a LeftProjection of a Left, which is defined (and not empty) in our understanding. So left.left().toOption() is Some.

Trust me, it works like in Scala. This projections are not every-man's stuff but it is correct.

talios:
On 8 Mar 2016, at 22:13, Guillermo Gutérrez wrote:

> I'd want to understand why you'd need a Some(null). Not trying to be rude... really :) Could you provide an example?

When pulling values out of a database - a missing row might be a `None` but a row containing a null value would be `Some(null)` - Just like when dealing with some result sets when you `group()` or `left outer join` them you might end up with a `List(1,2,3,null,4,null,6)`.

Double wrapping _everything_ as `List(Some(1),Some(2),Some(3),None,Some(4),None)` itself kinda feels horrible and wrong. Tho in the list scenario it might be better as `List(List(1),List(2),List(3),Nil,List(4),Nil)` and `flatten` the list to drop the `Nils`.

It is probably _rare_ you want `Some(null)` but I find sometimes it's useful to have. Esp. if what you are saying is just that "this is defined, and it is null" and you're dealing with legacy data.


-- 
Mark Derricutt
http://www.theoryinpractice.net
http://www.chaliceofblood.net
http://plus.google.com/+MarkDerricutt
http://twitter.com/talios
http://facebook.com/mderricutt

ggalmazor:
Thanks a lot @danieldietrich. I've learned a lot these three days. You have a lot of patience :)

Thanks for the example too, @talios. In our apps we force modeling nullable properties into Options, with the convention that a ```null``` is always mapped to a ```None```. It makes sense because when a query returns a ```null``` value, then that column is an optional value that is not present. Also it provides nice semantics when reasoning about our domain model and a nice API when working with those objects.

We don't have queries which have rows filled completely with nulls because that wouldn't make any sense in our app.

We do have some queries that aggregate values from different sources which can produce some groups of columns with null values (doing left joins that you mention) but then again, we model that as an optional property in an aggregate domain. 

We try to forbid null values in our apps. We have clearly defined domain boundaries where data is produced (web port for user input, database port, filesystem port, messaging port for queues, etc.) and that let's us to have a central code realm that follows our own rules. One of those rules is to not ever have a null and have ```None```s instead.

Thanks to this, we can protect ourselves from my original problem of ```Try.of(()->null).toOptional()``` being a ```Some(null)``` and transform it into a ```None``` in one of those boundaries where that happens.

I guess that the bottom line for me is: does it make sense to forbid null values in your business logic? (assuming that I can control all data flow into it) 

We've been making this (among other things, naturally) for almost 2 years now, in a big codebase involving domains of operational applications, management tools, data analytics and big data. It feels right and produces clean, understandable, less buggy code. 

I'd be very happy to hear about any "horror story" from you :)

danieldietrich:
Thanks @talios for the example.

@ggalmazor You're welcome. It was fun to me having a birds-view on Javaslang and reconsolidate some concepts.

I think also that it is the right way to forbid null values, at least in the business domain. In low, technical layers it might help to optimize code / performance when saving additional instances. But that should be the job of library designers and this implementation detail should not leak to the outside.

@all Have a great day!

Issue #725, opened 2015-10-28
Create collection benchmarks


danieldietrich:
We could use this as basis:

https://github.com/CodeFX-org/lab-java8streamperformancebenchmark

ruslansennov:
btw <code>benchmark.collection.MapSetBench</code> is wrong.
You can try to modify <code>main(String[] args)</code> on something like this
```java
    public static void main(String[] args) {
        benchHashSet_put();
        benchHashSet_put();
        benchHashSet_put();
        ...
```
and output will be
```
Java HashMap.put bench(1000000) took 0.05 sec.
HashMap.put bench(1000000) took 5.98 sec.
Java HashMap.put bench(1000000) took 0.041 sec.
HashMap.put bench(1000000) took 3.293 sec.
Java HashMap.put bench(1000000) took 0.081 sec.
HashMap.put bench(1000000) took 3.15 sec.
Java TreeSet.add bench(1000000) took 0.317 sec.
```

danieldietrich:
thx! not good. we should delete our little benchmark framework and create a separate `javaslang-benchmark` project containing something more sophisticated.

danieldietrich:
Just wanted to save this snippet here:

```java
public class Bench {

    static final int COUNT = 100000;
    static final int ITERATIONS = 100;
    static final int WARMUP = 10;

    public static void main(String[] args) {

        // Average time after 100 iterations: 6.490000 ms
        timed(() -> scala.collection.immutable.Stream.range(0, COUNT, 1, new scala.math.Numeric.IntIsIntegral$()).length());

        // Average time after 100 iterations: 7.070000 ms (using Lazy with synchronized)
        // Average time after 100 iterations: 5.500000 ms (using Lazy without synchronized)
        timed(() -> javaslang.collection.Stream.range(0, COUNT).length());

        // Average time after 100 iterations: 0.600000 ms
        timed(() -> java.util.stream.IntStream.range(0, COUNT).boxed().count());
    }

    static void timed(Runnable stuff) {
        long vs = 0;
        for (int i = 0; i < WARMUP; i++) {
            System.gc(); // try to get rid of potential GC pauses
            stuff.run();
        }
        for (int i = 0; i < ITERATIONS; i++) {
            System.gc(); // try to get rid of potential GC pauses
            long t = System.currentTimeMillis();
            stuff.run();
            vs += System.currentTimeMillis() - t;
        }
[...]

PR #1187, opened 2016-03-13
Added javaslang-benchmark module with microbenchmarks.

malduarte:
Progress towards #725: Micro benchmarks implemented for:
* head()
* tail()
* get(int)
* update(int, T)
* prepend(T)
* append(T)

For the following Javaslang colections:
* List
* Stream
* Vector
* Queue


danieldietrich:
Hi Miguel, that's really great! I'm thrilled to let the tests run this evening an I'm really looking forward to the comparison with Scala. That will be worth a nice blog post, maybe as guest post to reach more people :-)

malduarte:
@danieldietrich I'm happy to contribute to the blog post. That being said I wouldn't be comfortable to make a blog post with the current use cases - they're a start but we need more insights before we try to have a meaningful discussion. Posts on benchmarks are likely to fuel flame wars :). I would prefer to add composite operations (say, groupBy, distinct elements etc) also. What do you think?

danieldietrich:
@malduarte yes, we should get a clear picture. Yesterday I also thought about adding more complex measures, e.g. combined operations like adding and removing elements in one test. Also it makes sense to add real-world use-cases like solving project Euler problems. I'm looking forward to have side-by-side measures with Scala. We will take our time to understand the results.

PR #1198, opened 2016-03-18
Stack definition fixed

ruslansennov:


danieldietrich:
It was ok before, Stack isn't a Seq any more. Otherwise also the return types need to be overwritten. It is just a convenience interface to augment another collection with Stack operations.

This PR can be closed.

In the case of object serialization it is sufficient if List is serializable.

ruslansennov:
OK
Then I can completely remove <code>Stack</code> from jackson

danieldietrich:
Yep, thx!

Issue #1194, opened 2016-03-15
Should Map have withDefault and withDefaultValue as in Scala


mvh77:
Or has this been studied already?

danieldietrich:
Hi Martti,

I haven't considered it yet. Looking at the Scala source, there is one wrapper `WithDefault`. For Scala it is possible to have only one wrapper because the higher-kinded types allow to preserve the Map type when wrapping the actual Map instance. In Java this is not possible. We would need a wrapper for each Map implementation.

We need to further investigate, how an implementation could look like. We should target the following:

* Don't leak additional types (wrappers) as public API to the outside
* Minimize the number of additional types. Currently I see HashMap.WithDefault, LinkedHashMap.WithDefault, TreeMap.WithDefault, HashMultimap.WithDefault, LinkedHashMultimap.WithDefault, TreeMultimap.WithDefault. These are necessary because of the next point:
* This change should not affect the existing Map implementation in the way that additional instances are added if not asked for (i.e. if withDefault was not called).

mvh77:
Yes you're right, it's a bit complex. Also the behaviour is not necessarily obvious since _some methods (get, contains, iterator, keys, etc) are not affected and invoking transformer methods (e.g. map) will not preserve the default value_. The only real advantage is you can transform a map from a partial function to a full function. This could be solved with a static method on Map that would return a wraper that only affects the apply method...

danieldietrich:
Currently Map implements Function1 which throws if the key is not present:

```java
interface Map<K, V> extends Traversable<Tuple2<K, V>>, Function1<K, V> {
    @Override
    default V apply(K key) {
        return get(key).getOrElseThrow(NoSuchElementException::new);
    }
}
```

It would be better to reflect that a Map is a partial function by returning `Option<V>` instead of `V`:

```java
interface Map<K, V> extends Traversable<Tuple2<K, V>>, Function1<K, Option<V>> {
    @Override
    default Option<V> apply(K key) {
        return get(key);
    }
}
```

Then `withDefault` could just return a total function instead of a Map:

```java
interface Map<K, V> extends Traversable<Tuple2<K, V>>, Function1<K, Option<V>> {
    @Override
    default Function1<K, V> withDefault(V defaultValue) {
        return k -> get(key).getOrElse(defaultValue);
    }
}
```

(Is this signature right compared to Scala? There withDefault and withDefaultValue are distinguished...)

mvh77:
[...]

PR #1200, opened 2016-03-19
lift, withDefault & withDefaultValue for Map & Seq

mvh77:
Title says it.


danieldietrich:
Very nice, thank you!

danieldietrich:
Towards #1194

PR #1202, opened 2016-03-19
widened argument bound on zip & zipAll

mvh77:
for next patch release, backwards compatible.


danieldietrich:
Thanks! Interesting that we missed that before. I will prepare a patch-branch and cherry-pick this commit.

PR #1214, opened 2016-03-21
IntelliJ Idea 2016.1 failure fixed

ruslansennov:
It seems there is a bug in Idea 2016.1
```
Class 'TreeMap' must either be declared abstract or implement abstract method 
'merge(Map<? extends K, U>, BiFunction<? super V, ? super U, ? extends V>)' in 'SortedMap'
```


danieldietrich:
Thanks! This is evil. We should report that and remove the workaround.

ruslansennov:
What do you think will be enough to give a link to this PR? :)

danieldietrich:
No :) I think the issue will be best resolved near time when we try to distill a minimal example (one file, using nested classes and interfaces if necessary). If there is no simple solution to do this now, we should open an issue for that.

ruslansennov:
[IDEA-153421](https://youtrack.jetbrains.com/issue/IDEA-153421)

PR #1215, opened 2016-03-22
added missing quote

otrosien:
Just something I found while reading the docs.


otrosien:
This is probably not the correct place to fix the issue, so I also opened https://github.com/javaslang/javaslang-docs/pull/16

danieldietrich:
Thank you! Both places are right.

PR #1228, opened 2016-03-25
minor fixes based on findBugs feedback

mvh77:
Ran findBugs on javaslang and it found a few things, I suggest you run it too and take a look. Iterator.next is supposed to throw NoSuchElementException, and there was this Tuple creation that didn't do anything with it + other things.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Stream.java
+++ b/javaslang/src/main/java/javaslang/collection/Stream.java
@@ -267,11 +267,7 @@ public interface Stream<T> extends Kind1<Stream<?>, T>, LinearSeq<T> {
 
             @Override
             public T next() {
-                if(hasNext()) {
-                    return elements[i++];
-                } else {
-                    throw new NoSuchElementException();
-                }
+                return elements[i++];
             }
         });
     }
--- a/javaslang/src/main/java/javaslang/collection/Stream.java
+++ b/javaslang/src/main/java/javaslang/collection/Stream.java
@@ -267,7 +267,11 @@ public interface Stream<T> extends Kind1<Stream<?>, T>, LinearSeq<T> {
 
             @Override
             public T next() {
-                return elements[i++];
+                if(hasNext()) {
+                    return elements[i++];
+                } else {
+                    throw new NoSuchElementException();
+                }
             }
         });
     }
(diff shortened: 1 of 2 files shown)


javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1095,7 +1095,7 @@ public int length() {
             init = init.put(init.size(), t);
             if (predicate.test(t)) {
                 if (init.size() == length()) {
-                    Tuple.of(this, empty());
+                    return Tuple.of(this, empty());
```
danieldietrich: Good finding! As far as I can see the semantic does not change because in the `==` case the loop ended and the fall-back return of the method does the same.

danieldietrich:
Thank you, I will run it. Best would be to automate it as part of the CI build.

The Iterator.next method intentionally does not throw. The iterator is used internally and there is no chance that it ever will escape to the outside or throw a NoSuchElementException. Usage:

```java
    interface StreamFactory {

        static <T> Stream<T> create(java.util.Iterator<? extends T> iterator) {
            return iterator.hasNext() ? Stream.cons(iterator.next(), () -> create(iterator)) : Empty.instance();
        }
    }
```

That's one thing that drives me crazy about FindBugs, PMD et al. These tools are good to spot blockers and major bugs. But when it comes to code optimizations they are a real pain in the ass.

javaslang/src/main/java/javaslang/collection/Stream.java
```diff
@@ -267,7 +267,11 @@ public boolean hasNext() {
 
             @Override
             public T next() {
-                return elements[i++];
+                if(hasNext()) {
+                    return elements[i++];
+                } else {
+                    throw new NoSuchElementException();
+                }
```
danieldietrich: This is a valid optimization. Please revert it according to my [comment](https://github.com/javaslang/javaslang/pull/1228#issuecomment-201316395).

danieldietrich:
Thank you Martti!
I ran FindBugs in IntelliJ and it helped!

danieldietrich:
I investigated all FindBugs findings. All should work fine with the exception of one finding which has to be fixed - after deserializing a HashMap/HashSet the hashCode function will fail. It will be fixed with 2.0.2.

PR #1234, opened 2016-03-26
Add isNotNull() and isNull() predicates

### LINKED ISSUE #1221: Add more predicates to javaslang.Predicates
We need additional predicates for null-checks:

* isNull
* isNotNull

pivovarit:
This closes #1221

### DIFF
--- a/javaslang/src/main/java/javaslang/Predicates.java
+++ b/javaslang/src/main/java/javaslang/Predicates.java
@@ -15,6 +15,7 @@ import java.util.function.Predicate;
  * {@link javaslang.API.Match}.
  *
  * @author Daniel Dietrich
+ * @author Grzegorz Piwowarek
  * @since 2.0.0
  */
 public final class Predicates {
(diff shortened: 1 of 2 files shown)


javaslang/src/main/java/javaslang/Predicates.java
```diff
@@ -118,4 +119,23 @@ private Predicates() {
...
+    }
+
+    /**
+     * Creates a {@code Predicate} that tests, if an object is not null
+     * @param <T> value type
+     * @return A new {@code Predicate}
+     */
+    public static <T> Predicate<T> isNotNull() {
+        return Objects::nonNull;
+    }
```
danieldietrich: Very nice :-)
Please move the new methods above line 72 `// -- Predicate combinators` in order we do not confuse it with the predicate combinators.
Thanks!

pivovarit: @danieldietrich fixed

danieldietrich: Thank you Grzegorz.
What is your opinion about the names isNotNull vs. nonNull?
I like that isNotNull has the prefix 'is', which is typical for predicates. Personally I'm used to it.
I like about nonNull that it is a Java standard and that it is shorter. Personally I'm not used to it.
Once we defined the predicates, they should stay untouched to keep the API stable.

pivovarit: Both seem right to me, but I would stick to "isNotNull" since it seems to follow a certain convention.

danieldietrich: Thanks, that helps. Naming :-) nonNull can be added later, if needed.

Issue #1131, opened 2016-02-23
Create a Multimap


danieldietrich:
We have to make a decision about the type of container that holds values. Scala uses Set but I would use Traversable like in Guava. This allows us to store the same values in containers in certain multimap implementations. On the other side, using the Set as container is much easier to implement and much shorter API

danieldietrich:
Hi Ruslan,
yes, that's a tough decision to make. I like the idea to have different value container types. But in order to benefit from it, the `values()` method should return a Traversable of the correct type. This is only possible, if the Multimap is aware of the type by specifiying an additional generic:

```java
interface MultiMap<K, V, TYPE extends Traversable<V>> {
    TYPE values();
}
```

We could have factory methods for the default-value container Set:

```java
static <...> Multimap<..., Set<V>> empty();
static <...> Multimap<..., Set<V>> ofEntries(...);
...
```

Because we need appropriate create and add methods for construction and element addition, we could pre-define known Traversable types that can be used as value-containers. These enum _knows_ all necessary operations.

```java
interface Multimap {

    enum Type {

        INDEXED_SEQ(Vector::new, Vector::append),
        LINEAR_SEQ(...),
        SET(...),
        SORTED_SET(...)

        Type(...) {
            ...
        }
    }

    static Multimap<..., Set<V>> empty(Type type);
    static Multimap<..., Set<V>> ofEntries(Type type, ...);
    ...
}
```

To have the most flexibility, we could provide also a factory method that allows to create arbitrary value containers:

```java
static <...> Multimap<..., Set<V>> empty(Supplier<TYPE> unit, BiFunction<TYPE, V, TYPE> add);

// maybe we omit all other factory methods for the general case to keep it simple?
// static <...> Multimap<..., Set<V>> ofEntries(...);
...
```

ruslansennov:
> interface MultiMap&lt;K, V, TYPE extends Traversable&lt;V&gt;&gt;

Yes, I thought about it.
OK, let's make the big Zoo of Multimap's implementations :)

danieldietrich:
Cool, one Multimap will rule all :-)

talios:
My original naive implementation was https://gist.github.com/talios/97c27fe577894e96cc46 - which was basically doing the bare minimum of what I was wanting. And when I say naive - I mean incredibly so :)

ruslansennov:
partly blocked by #1132

ruslansennov:
| Backed by | Container | Multimap |
[...]

PR #1240, opened 2016-03-28
Multimap #1131

ruslansennov:


javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
```diff
@@ -0,0 +1,455 @@
...
+ */
+ abstract class AbstractMultimap<K, V, M extends Multimap<K, V>> implements Multimap<K, V> {
+
+    interface SerializableSupplier<T> extends Supplier<T>, Serializable {
+    }
+
+    private static final long serialVersionUID = 1L;
+
+    final Map<K, Traversable<V>> back;
+    final SerializableSupplier<Traversable<?>> emptyContainer;
```
danieldietrich: Hi Ruslan,
yeah, Multimap is coming :)
Why are `back` and `emptyContainer` package-private? Can they be made protected?

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
```diff
@@ -0,0 +1,455 @@
...
+        this.back = back;
+        this.containerType = containerType;
+        this.emptyContainer = emptyContainer;
+    }
+
+    abstract <K2, V2> Map<K2, V2> emptyMapSupplier();
+
+    abstract <K2, V2> Multimap<K2, V2> emptyInstance();
+
+    abstract <K2, V2> Multimap<K2, V2> createFromMap(Map<K2, Traversable<V2>> back);
```
danieldietrich: These three methods maybe could also be protected.

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
```diff
@@ -0,0 +1,455 @@
...
+
+    @SuppressWarnings("unchecked")
+    private <K2, V2> Multimap<K2, V2> createFromEntries(Iterable<? extends Tuple2<? extends K2, ? extends V2>> entries) {
```
danieldietrich: I think it can be `Tuple2` instead of `? extends Tuple2`.

For K2 and V2 I'm not sure but probably we do not need the `? extends` because this method is private.

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
```diff
@@ -0,0 +1,455 @@
...
+    public M retainAll(Iterable<? extends Tuple2<K, V>> elements) {
+        Objects.requireNonNull(elements, "elements is null");
+        return (M) createFromEntries(back.flatMap(t -> t._2.map(v -> Tuple.of(t._1, v))).retainAll(elements));
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public M scan(Tuple2<K, V> zero, BiFunction<? super Tuple2<K, V>, ? super Tuple2<K, V>, ? extends Tuple2<K, V>> operation) {
+        Objects.requireNonNull(operation, "operation is null");
+        return (M) Collections.scanLeft(this, zero, operation, Queue.<Tuple2<K, V>>empty(), Queue::append, this::createFromEntries);
```
danieldietrich: I had to look two times but I think using Queue is correct. The rear List is implicitly reversed when the finisher is called. Ok!

javaslang/src/test/java/javaslang/collection/AbstractMultimapTest.java
```diff
@@ -0,0 +1,770 @@
...
+    public void shouldFillTheSeqWith0Elements() {
+        assertThat(mapFill(0, () -> new Tuple2<>(1, 1))).isEqualTo(empty());
+    }
+
+    @Test
+    public void shouldFillTheSeqWith0ElementsWhenNIsNegative() {
+        assertThat(mapFill(-1, () -> new Tuple2<>(1, 1))).isEqualTo(empty());
+    }
+
+    /////////////////////////////////////////////////////////////////
```
danieldietrich: nice ascii art separator ;-)

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
```diff
@@ -0,0 +1,455 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public M takeWhile(Predicate<? super Tuple2<K, V>> predicate) {
+        Objects.requireNonNull(predicate, "predicate is null");
+        final Multimap<K, V> taken = createFromEntries(iterator().takeWhile(predicate));
+        return taken.length() == length() ? (M) this : (M) taken;
+    }
+
```
danieldietrich: Do we not need also to @Override the equals method here?

javaslang/src/main/java/javaslang/collection/Multimap.java
```diff
@@ -0,0 +1,477 @@
...
+        );
+
+        final BiFunction<Traversable<?>, Object, Traversable<?>> add;
+        final BiFunction<Traversable<?>, Object, Traversable<?>> remove;
+
+        ContainerType(BiFunction<Traversable<?>, Object, Traversable<?>> add,
+                              BiFunction<Traversable<?>, Object, Traversable<?>> remove) {
+            this.add = add;
+            this.remove = remove;
+        }
```
danieldietrich: We should make the instance variables `add` and `remove` and the constructor `ContainerType` private.

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
```diff
@@ -0,0 +1,455 @@
...
+                if(emptyContainer.get() instanceof Set) {
+                    return new java.util.HashSet<>();
+                } else if(emptyContainer.get() instanceof SortedSet) {
+                        return new java.util.TreeSet<>();
+                } else {
+                    return new java.util.ArrayList<>();
+                }
+            }).add(t._2);
+        }
+        return javaMap;
```
danieldietrich: I would move the value-provider to the outside of the loop. Then we make the type check only once.

```java
final java.util.Map<K, Collection<V>> javaMap = new java.util.HashMap<>();
final Supplier<Collection<V>> javaContainerSupplier;
if (containerType == SEQ) {
    javaContainerSupplier = java.util.ArrayList::new;
} else if (containerType == SET) {
    javaContainerSupplier = java.util.HashSet::new;
} else if (containerType == SORTED_SET) {
    javaContainerSupplier = java.util.TreeSet::new;
} else {
    throw new IllegalStateException("Unknown ContainerType: " + containerType);
}
for (Tuple2<K, V> t : this) {
    javaMap.computeIfAbsent(t._1, k -> javaContainerSupplier.get()).add(t._2);
}
return javaMap;
```

Don't know if the throw is possible there or if the compiler will complain that `javaContainerSupplier` is uninitialized. If so, we could create a private method getJavaContainerSupplier()...

javaslang/src/main/java/javaslang/collection/HashMultimap.java
```diff
@@ -0,0 +1,159 @@
...
+            for (java.util.Map.Entry<? extends K, ? extends V2> entry : entries) {
+                result = result.put(entry.getKey(), entry.getValue());
+            }
+            return (HashMultimap<K, V2>) result;
+        }
+
+        @SuppressWarnings("unchecked")
+        public <K, V2 extends V> HashMultimap<K, V2> tabulate(int n, Function<? super Integer, ? extends Tuple2<? extends K, ? extends V2>> f) {
+            Objects.requireNonNull(f, "f is null");
+            return ofEntries(Collections.tabulate(n, (Function<? super Integer, ? extends Tuple2<K, V2>>) f));
```
danieldietrich: I think we can write `Tuple2` instead of `? extends Tuple2`. Tuples are final.

javaslang/src/main/java/javaslang/collection/HashMultimap.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        @SuppressWarnings("unchecked")
+        public <K, V2 extends V> HashMultimap<K, V2> fill(int n, Supplier<? extends Tuple2<? extends K, ? extends V2>> s) {
```
danieldietrich: Dito

javaslang/src/main/java/javaslang/collection/LinkedHashMultimap.java
```diff
@@ -0,0 +1,162 @@
...
+        private Builder(ContainerType containerType, SerializableSupplier<Traversable<?>> emptyContainer) {
+            this.containerType = containerType;
+            this.emptyContainer = emptyContainer;
+        }
+
+        public <K, V2 extends V> LinkedHashMultimap<K, V2> empty() {
+            return new LinkedHashMultimap<>(LinkedHashMap.empty(), containerType, emptyContainer);
+        }
+
+        public <K, V2 extends V> LinkedHashMultimap<K, V2> ofEntries(Iterable<? extends Tuple2<? extends K, ? extends V2>> entries) {
```
danieldietrich: Dito

javaslang/src/main/java/javaslang/collection/LinkedHashMultimap.java
```diff
@@ -0,0 +1,162 @@
...
+            Objects.requireNonNull(entries, "entries is null");
+            Multimap<K, V2> result = empty();
+            for (java.util.Map.Entry<? extends K, ? extends V2> entry : entries) {
+                result = result.put(entry.getKey(), entry.getValue());
+            }
+            return (LinkedHashMultimap<K, V2>) result;
+        }
+
+        @SuppressWarnings("unchecked")
+        public <K, V2 extends V> LinkedHashMultimap<K, V2> tabulate(int n, Function<? super Integer, ? extends Tuple2<? extends K, ? extends V2>> f) {
```
danieldietrich: Dito... I will not mark the other ? extends Tuple occurrences. I'm sure I'll miss one. Would be best to make search-replace.

javaslang/src/test/java/javaslang/collection/LinkedHashMultimapTest.java
```diff
@@ -0,0 +1,131 @@
...
+    protected <T1, T2> Multimap<T1, T2> emptyMap() {
+        switch (containerType) {
+            case SEQ:
+                return LinkedHashMultimap.withSeq().empty();
+            case SET:
+                return LinkedHashMultimap.withSet().empty();
+            case SORTED_SET:
+                return LinkedHashMultimap.withSortedSet(TreeSetTest.toStringComparator()).empty();
+        }
+        throw new RuntimeException();
```
danieldietrich: Better `throw new IllegalStateException("Unknown container type: " + containerType);`, also in all other places.

danieldietrich:
Super-awesome work!!!

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
ruslansennov: I would keep it as is. This signature simplifies the code in some methods like <code>distinctBy(Comparator<? super Tuple2<K, V>> comparator)</code>

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
ruslansennov: Yes, thank you. I believe <code>this.corresponds(that)</code> should work

javaslang/src/main/java/javaslang/collection/HashMultimap.java
ruslansennov: We should do it everywhere. Let's do it in separate issue :)

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
ruslansennov: Great, it works

javaslang/src/main/java/javaslang/collection/HashMultimap.java
ruslansennov: I mean everywhere in Javaslang. Not only in Multimap

javaslang/src/main/java/javaslang/collection/Multimap.java
ruslansennov: OK for fields. But 'private' modifier for enum constructor is redundant

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
danieldietrich: ok

javaslang/src/main/java/javaslang/collection/HashMultimap.java
danieldietrich: Yes we can do it in a separate issue. In many places I've already done it I think. See #1241

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
danieldietrich: We can do that. I'm working on a simplification of equals/hashcode, see #1242. Before merging it I will take a look again at the MM equals/hashCode impls.

PR #1242 is still work in progress.
Currently it is not fully working because of the order of elements in HAMT equals.
Also the HAMP hashCode needs to be improved. Currently XOR accumulates the hashCodes because it is symmetric (e.g. a XOR b == b XOR a). That is important because a HAMT represents a Set, which has no order of elements. The hashCode always has to be equal.

javaslang/src/main/java/javaslang/collection/Multimap.java
danieldietrich: Ok

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
ruslansennov: XOR is bad function for hashCode. For example, HAMT of 1000 integers all between 0 and 16 will also between 0 and 16.

danieldietrich: Yes, in my tests I had collisions. Maybe we both could take at equals and hashCode impl in HAMT...

danieldietrich:
Aaaand... merged :)
Many thanks, great feature!

Issue #1235, opened 2016-03-27
Remove lazy hashCode and hashCode calculation on initialization


danieldietrich:
This will reduce the memory footprint (slightly?) and reduce the initialization time on object construction.

Especially collections are rarely used as HashMap/HashSet keys. I think it will be more balanced to remove lazy hashCode instance variables of collection nodes.

Caution: Take a look at the equals methods. There the laziness of hashCode was used sometimes. Hashcode should not be used in equals impls.

danieldietrich:
It seems HashArrayMappedTrie hashCode and equals are not symmetric. I noticed it when simplifying the equals/hashCode impls of collections.

It makes sense to define that two HAMT's are equal, if their structure and their contained values are equal. That means that I will update the equals method of HAMT and take care that no collection relies on comparing their backends by HAMT.equals.

The same applies for RedBlackTrees (needs to be checked).

danieldietrich:
HAMT equals and hashCode should be expressed independent of the order of elements. In other words, it does not matter how HAMT(and RedBlackTree) structures store values. Only the values matter.

* HAMT hashCode should be expressed by applying XOR to the hashCode of all entries. XOR is commutative and equally distributed (50% 0, 50% 1). See https://stackoverflow.com/questions/5889238/why-is-xor-the-default-way-to-combine-hashes
* HAMT equals should be expressed by checking key sets for equality and then check if values are congruent. size()-check may be also used for optimization, it is O(1)

This strategy allows us to delegate equals and hashCode to an underlying HAMT backend

danieldietrich:
Fixed

PR #1242, opened 2016-03-29
better equals/hashcode impls

danieldietrich:
**DO NOT MERGE, YET**

I created this PR to take a sneak peak on the diff. After finishing it, it can be merged.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Tree.java
+++ b/javaslang/src/main/java/javaslang/collection/Tree.java
@@ -5,7 +5,9 @@
  */
 package javaslang.collection;
 
-import javaslang.*;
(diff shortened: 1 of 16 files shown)


danieldietrich:
Towards #1235

javaslang/src/main/java/javaslang/collection/Tree.java
```diff
@@ -714,8 +716,6 @@ default String stringPrefix() {
 
         private final T value;
         private final List<Node<T>> children;
-        private final Lazy<Integer> size;
```
danieldietrich: Add intance variable `int size`, which is calculated at construction time.

Issue #1238, opened 2016-03-27
Add leftPadTo(int, T)


danieldietrich:
We already have [`Seq<T> padTo(int, T)`](https://github.com/javaslang/javaslang/blob/master/javaslang/src/main/java/javaslang/collection/Seq.java#L487).
Now we want to add `Seq<T> leftPadTo(int, T)` in a similar way.

Motivated by [this article](http://arstechnica.com/information-technology/2016/03/rage-quit-coder-unpublished-17-lines-of-javascript-and-broke-the-internet/).

mvh77:
Meaning Javaslang won't be found on NPM after all? :stuck_out_tongue_winking_eye:

danieldietrich:
Successfully finished.

PR #1245, opened 2016-03-30
issue-1238: Add leftPadTo(int, T) to Seq.

nicholasren:
Hi Daniel, 

I implemented the `leftPadTo(int, T)` for `Seq` which is mentioned in [issue 1238](https://github.com/javaslang/javaslang/issues/1238). please help to review it and merge.

Thanks

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/CharSeq.java
+++ b/javaslang/src/main/java/javaslang/collection/CharSeq.java
@@ -494,20 +494,24 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
 
     @Override
     public CharSeq padTo(int length, Character element) {
-        if (length <= back.length()) {
+        final int actualLength = back.length();
+        if (length <= actualLength) {
             return this;
+        } else {
+            final StringBuilder sb = new StringBuilder(back).append(padding(element, length - actualLength));
+            return new CharSeq(sb.toString());
         }
-        final StringBuilder sb = new StringBuilder(back).append(padding(element, length - back.length()));
-        return new CharSeq(sb.toString());
     }
 
     @Override
     public CharSeq leftPadTo(int length, Character element) {
-        if (length <= back.length()) {
+        final int actualLength = back.length();
+        if (length <= actualLength) {
             return this;
+        } else {
+            final StringBuilder sb = padding(element, length - actualLength).append(back);
+            return new CharSeq(sb.toString());
         }
-        final StringBuilder sb = padding(element, length - back.length()).append(back);
-        return new CharSeq(sb.toString());
     }
 
     private static StringBuilder padding(Character element, int limit) {
--- a/javaslang/src/main/java/javaslang/collection/CharSeq.java
+++ b/javaslang/src/main/java/javaslang/collection/CharSeq.java
@@ -498,11 +498,26 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
             return this;
         }
         final StringBuilder sb = new StringBuilder(back);
-        final int limit = length - back.length();
-        for (int i = 0; i < limit; i++) {
-            sb.append(element);
-        }
+        sb.append(padding(element, length - back.length()));
         return new CharSeq(sb.toString());
+    }
+
+    @Override
+    public CharSeq leftPadTo(int length, Character element) {
+        if (length <= back.length()) {
+            return this;
+        }
+        final StringBuilder sb = new StringBuilder(padding(element, length - back.length()));
+        sb.append(back);
+        return new CharSeq(sb.toString());
+    }
+
+    private static StringBuilder padding(Character element, int limit) {
+        final StringBuilder padding = new StringBuilder();
+        for (int i = 0; i < limit; i++) {
+            padding.append(element);
+        }
+        return padding;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -498,13 +498,28 @@ public CharSeq padTo(int length, Character element) {
...
+    public CharSeq leftPadTo(int length, Character element) {
+        if (length <= back.length()) {
+            return this;
         }
+        final StringBuilder sb = new StringBuilder(padding(element, length - back.length()));
+        sb.append(back);
         return new CharSeq(sb.toString());
     }
 
+    private static StringBuilder padding(Character element, int limit) {
```
ruslansennov: let's change signature to 
```java
private static void padding(StringBuilder sb, Character element, int limit)
```
to avoid sreating new StringBuilder instance

ruslansennov: no, this doens't work, sorry

nicholasren: I made the `padding` method to return a new `StringBuilder`  intentionally,  pay a small cost to make it more readable.

Also, I just pushed another commit, change to a fluent style.

danieldietrich:
Hi Nicholas,
very nice - well done! I have one question - please see the review comment.
Many thanks!
Daniel

Btw: I saw your project euler github project. We also have projecteuler tests for the case you are interested in adding ones: https://github.com/javaslang/javaslang/tree/master/javaslang/src/test/java/javaslang/collection/euler

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -497,14 +497,27 @@ public CharSeq padTo(int length, Character element) {
...
+        final StringBuilder sb = new StringBuilder(back).append(padding(element, length - back.length()));
+        return new CharSeq(sb.toString());
+    }
+
+    @Override
+    public CharSeq leftPadTo(int length, Character element) {
+        if (length <= back.length()) {
+            return this;
         }
+        final StringBuilder sb = new StringBuilder(padding(element, length - back.length())).append(back);
```
danieldietrich: Is it the same as this?

```java
final StringBuilder sb = padding(element, length - back.length()).append(back);
```

javaslang/src/main/java/javaslang/collection/Stream.java
```diff
@@ -980,6 +980,17 @@ default int length() {
     }
 
     @Override
+    default Stream<T> leftPadTo(int length, T element) {
+        if (length <= 1) {
```
ruslansennov: this test fails
```java
assertThat(empty().leftPadTo(1, 1)).isEqualTo(of(1));
```
I believe <code>if (length <= 1)</code> should be removed

nicholasren: the `if (length <= 1)` is required as it is the termination condition of recursion.

it has to be `length <= 1` instead of  `length <= 0` because I've already prepend an `element` before the recursion call.

I'll try to make it more explicit.

javaslang/src/main/java/javaslang/collection/CharSeq.java
nicholasren: Yep, changed.

javaslang/src/main/java/javaslang/collection/Stream.java
ruslansennov: what do you think about this implementation?
```java
    @Override
    default Stream<T> leftPadTo(int length, T element) {
        final int thisLen = length();
        if (length <= thisLen) {
            return this;
        } else {
            return Stream.ofAll(Iterator.continually(element).take(length - thisLen)).appendAll(this);
        }
    }
```

nicholasren: I was trying to avoid using `length()` on Stream as `Stream` is an _lazy_ collection, while calling `length()` on a `Stream` object will trigger a full evaluation.

But looks like both `padTo` and `leftPadTo` have to full evaluate the whole stream any  way, you solution works pretty well.

Shall we make it explicit that these methods should be carefully used on `Stream`?

javaslang/src/main/java/javaslang/collection/CharSeq.java
nicholasren: Pretty happy to work on `projecteuler` stuff.

javaslang/src/main/java/javaslang/collection/Stream.java
ruslansennov: > Shall we make it explicit that these methods should be carefully used on Stream?

we can mention this in the javadoc of this method

danieldietrich: +1 for using `length <= thisLen` and adding a javadoc comment to Seq.java#leftPadTo, sth. like:

```java
/**
 ...
 * <p>
 * Please note that lazily-evaluated Seq implementations need to process all elements in order to gather the overall length.
 ...
 */
```

---

A second thing: We should also unify the if-clauses in all methods by using an else branch instead of an if-guard and evaluate the length only once by introducing a local variable.

```java
final int thisLength = length();
if (length <= thisLength) {
    ...
} else {
    ...
}
```

(In all classes)

Many thanks!

nicholasren:
Hi guys, 

is there any further action on this PR?

ruslansennov:
Hi
we are waiting for our leader :)

nicholasren:
No problem,

I was wondering if I missed something  :smile: 

Cool, I'll leave it to our leader

danieldietrich:
Hi Nicholas, sorry for being so silent - I was on vacation with my family and had no internet - only mountains, air, snow and sun :-)
Your PR looks great. Many thanks!

@ruslansennov maybe I need a deputy... I will write up some thoughts in an email tomorrow.

PR #1256, opened 2016-03-31
fixed in: IntelliJ IDEA 2016.1.1 Build #IU-145.597

ruslansennov:
reverts #1214
see [IDEA-153421](https://youtrack.jetbrains.com/issue/IDEA-153421)


danieldietrich:
Thx!

PR #1258, opened 2016-04-02
Improvements to Arbitrary / Gen

talios:
* Add support for char[] based generators
* Add an intersperse method on Arbitrary/Gen to blend two Arbitrary/Gen instances into a new one
* Added Arbitrary.of() convenience factory method
* Added Arbitrary.distinct() and distinctBy() methods

### DIFF
--- a/javaslang-test/src/main/java/javaslang/test/Gen.java
+++ b/javaslang-test/src/main/java/javaslang/test/Gen.java
@@ -5,12 +5,15 @@
  */
 package javaslang.test;
 
+import javaslang.Tuple;
 import javaslang.Tuple2;
 import javaslang.collection.Iterator;
+import javaslang.collection.List;
 import javaslang.collection.Stream;
 
 import java.util.Objects;
 import java.util.Random;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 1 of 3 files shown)


javaslang-test/src/main/java/javaslang/test/Gen.java
```diff
@@ -277,6 +310,17 @@
...
     }
 
+  /**
+   * Intersperse values from this generator instance with those of another.
+   *
+   * @param other another T generator to accept values from.
+   * @return A new T generator
+   */
+    default Gen<T> intersperse(Gen<T> other) {
+        AtomicReference<Tuple2<Gen<T>, Gen<T>>> interspersedReference = new AtomicReference<>(Tuple.of(other, this));
```
talios: Using an `AtomicReference` works, but I feel somewhat dirty using it. But due to immutability and `final` references from inside lambdas, I wanted something I could update an order swapped tuple.

javaslang-test/src/main/java/javaslang/test/Gen.java
```diff
@@ -277,6 +310,17 @@
...
+        return random -> interspersedReference.updateAndGet(generators -> Tuple.of(generators._2, generators._1))._1.apply(random);
```
talios: I also don't overly like creating new `Tuple` instances on every iteration, do we have any notion of a "rotate" or "shift" in javaslang that could maintain the underlying structure?

javaslang-test/src/test/java/javaslang/test/GenTest.java
```diff
@@ -44,6 +44,14 @@ public void shouldCreateGenOfSeedAndFunction() {
         assertThat(gen.apply(RANDOM)).isEqualTo(3);
     }
 
+    @Test
+    public void shouldCreateGenOfFixedValues() {
+        final Gen<Integer> gen = Gen.fixed(1,2,3);
```
talios: Any suggestions on how I would even _approach_ writing tests for arbitrary values....

PR #1259, opened 2016-04-04
Remove "throws Throwable" clause from partial application in CheckedFunction

### LINKED ISSUE #1257: Throwable in signature for partial application of CheckedFunctions
I do not really understand why the signature of methods for partial application of CheckedFunctions expects a Throwable, e.g.:

```java
default CheckedFunction1<T2, R> apply(T1 t1) throws Throwable {
    return (T2 t2) -> apply(t1, t2);
}
```

Partial application of checked functions become very unhandy under this circumstances and from my point of view it's unncessary to expect a Throwable at this point. But I think you see a sense behind this signature :-) so I would like to know the reason your design decision ;-)

pivovarit:
This closes #1257

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -721,9 +721,8 @@ def generateMainClasses(): Unit = {
                    *
                    ${(1 to j).gen(k => s"* @param t$k argument $k")("\n")}
                    * @return a partial application of this function
-                   * ${checked.gen("@throws Throwable if something goes wrong partially applying this function to the given arguments")}
                    */
-                  default $name${i - j}<$resultFunctionGenerics, R> apply($partialApplicationArgs)${checked.gen(" throws Throwable")} {
+                  default $name${i - j}<$resultFunctionGenerics, R> apply($partialApplicationArgs) {
                       return ($resultFunctionArgs) -> apply($fixedApplyArgs, $variableApplyArgs);
                   }
                 """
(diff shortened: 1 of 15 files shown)


ruslansennov:
Hi Grzegorz 
You should edit <code>Generator</code> but not <code>src-gen/*</code> classes

pivovarit:
Oh, sh*t. I'm on it

pivovarit:
@ruslansennov I updated my commit

ruslansennov:
Sorry, I was not accurate in my comment :)
You should 
1. manually change only Generator
2. run <code>mvn test</code> (then all <code>src-gen/</code> files will be changed automatically)
3. all changes push to GitHub

pivovarit:
So you expect me to push both Generator.scala and updated generated files?

ruslansennov:
Yes!

pivovarit:
What's the point of keeping track of the generator and stuff it generates during compilation?

ruslansennov:
> What's the point of keeping track of the generator and stuff it generates during compilation?

Both points of view (include/exclude generated files) makes sense. I prefer exclude it from my projects,  It seems Daniel has another opinion :smile:

pivovarit:
I prefer this approach too. Anyway, I updated my commit

ruslansennov:
Thank you
Now we are waiting for Daniel

talios:
One good thing of including generated code in version control (IMHO) comes during code review, being able to see what code is being generated is often handy, and may often change the opinion of the code review.  i.e. your change may _look_ small - but it might have far further reaching impact than you, or an unfamiliar reviewer may expect.

That being said, care should be taken to generate the code as far away from "real code" as possible - and _never_ in the same directory :)

pivovarit:
@talios well, what if commiter is true evil and commits fake generated files? ;)

danieldietrich:
@pivovarit good point! I will open an issue to enhance the ci-build to detect evil commits of generated sources.

It is as @talios said - I want to see the changes in code reviews. And I want to browse the actual code online :-)

danieldietrich:
LGTM! Thx

PR #1260, opened 2016-04-04
Improvements to Arbitrary / Gen

talios:
* Added support for char[] based generators
* Added support for vararg based generator construction
* Add an intersperse method on Arbitrary/Gen to blend two Arbitrary/Gen instances into a new one
* Added Arbitrary.forAll() convenience factory method
* Added Arbitrary.fixed() convenience factory method
* Added Arbitrary.distinct() and distinctBy() methods

### DIFF
--- a/javaslang-test/src/main/java/javaslang/test/Arbitrary.java
+++ b/javaslang-test/src/main/java/javaslang/test/Arbitrary.java
@@ -9,6 +9,7 @@ import javaslang.collection.Iterator;
 import javaslang.collection.List;
 import javaslang.collection.Stream;
 
+import java.util.Comparator;
 import java.util.Objects;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


javaslang-test/src/main/java/javaslang/test/Arbitrary.java
```diff
@@ -117,7 +163,30 @@
...
+
+    /**
+     * Generates an arbitrary value from a given generator
+     * @param generator A generator to produce arbitrary values
+     * @param <U> Type of generator value
+     * @return A new generator
+     */
+    static <U> Arbitrary<U> forAll(Gen<U> generator) {
+      return size -> generator;
+    }
```
danieldietrich: Hi @talios, finally I'm able to answer... Thanks - the code looks great.
Instead of `fixed` and `forAll` I would name the methods `of(T...)` and `ofAll(Gen)` to follow the naming scheme of the collections. What do you think?
We already have `gen.arbitrary()` but I think having `Arbitrary.ofAll(Gen)` is an intuitive place to look for a factory.

javaslang-test/src/main/java/javaslang/test/Gen.java
```diff
@@ -277,6 +293,17 @@
...
+  /**
+   * Intersperse values from this generator instance with those of another.
+   *
+   * @param other another T generator to accept values from.
+   * @return A new T generator
+   */
+    default Gen<T> intersperse(Gen<T> other) {
+        AtomicReference<Tuple2<Gen<T>, Gen<T>>> interspersedReference = new AtomicReference<>(Tuple.of(other, this));
+        return random -> interspersedReference.updateAndGet(generators -> Tuple.of(generators._2, generators._1))._1.apply(random);
+    }
```
danieldietrich: Your questions of the prev PR:
- We do not have an in-place swap. This is not possible for Tuple2 because of the generic types T1 != T2 (wlog)
- Lets implement it as iterator with a boolean `switch = !switch` state to avoid new Tuple instances:

```java
Iterator<Gen<T>> iter = new Java.util.Iterator<>() {
    boolean switch = false;
    hasNext() = true
    next() = {
        gen = switch ? ... : ...;
        seitch = !switch;
        return gen;
    }
}
return rnd -> iter.next().apply(rnd);
```

javaslang-test/src/main/java/javaslang/test/Arbitrary.java
talios: Sure, I'd used `forAll` to follow the usage from `Property.def('').forAll(...)` and since `of()` somewhat implies a fixed ( or more, known ) set of values.  One usage of `fixed` made it nicer when static importing it, tho consistency with the rest of the API is probably better.

javaslang-test/src/main/java/javaslang/test/Gen.java
talios: Sounds good.

danieldietrich:
Awesome! Thx for improving the test-framework :)

PR #1266, opened 2016-04-10
Add Node size computation at construction time

### LINKED ISSUE #1248: Calculate Tree size at construction time
Add intance variable `int size`, which is calculated at construction time.

We already do it in List, HashArrayMappedTrie, ...

pivovarit:
This closes #1248 

I messed up your wildcard imports, but for me it's a good change

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Tree.java
+++ b/javaslang/src/main/java/javaslang/collection/Tree.java
@@ -11,12 +11,32 @@ import javaslang.Tuple3;
 import javaslang.collection.List.Nil;
 import javaslang.collection.Tree.Empty;
 import javaslang.collection.Tree.Node;
-import javaslang.collection.TreeModule.*;
+import javaslang.collection.TreeModule.FlatMap;
+import javaslang.collection.TreeModule.Replace;
+import javaslang.collection.TreeModule.Traversal;
+import javaslang.collection.TreeModule.Unzip;
+import javaslang.collection.TreeModule.Zip;
+import javaslang.collection.TreeModule.ZipAll;
 import javaslang.control.Option;
 
-import java.io.*;
-import java.util.*;
-import java.util.function.*;
+import java.io.IOException;
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.BinaryOperator;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
 import java.util.stream.Collector;
 
 import static javaslang.collection.Tree.Order.PRE_ORDER;
@@ -26,6 +46,7 @@ import static javaslang.collection.Tree.Order.PRE_ORDER;
  *
  * @param <T> component type of this Tree
  * @author Daniel Dietrich
+ * @author Grzegorz Piwowarek
  * @since 1.1.0
  */
 public interface Tree<T> extends Traversable<T> {
@@ -222,13 +243,6 @@ public interface Tree<T> extends Traversable<T> {
     }
 
     /**
-     * Returns the number of nodes (including root and leafs).
-     *
-     * @return The size of the tree.
-     */
-    int size();
-
-    /**
      * Transforms this {@code Tree}.
      *
      * @param f   A transformation
@@ -334,11 +348,7 @@ public interface Tree<T> extends Traversable<T> {
      * @return The number of nodes of this tree.
      */
     default int nodeCount() {
-        if (isEmpty()) {
-            return 0;
-        } else {
-            return 1 + getChildren().foldLeft(0, (count, child) -> count + child.nodeCount());
-        }
+        return length();
     }
 
     // -- Methods inherited from Traversable
@@ -487,12 +497,7 @@ public interface Tree<T> extends Traversable<T> {
         return values().iterator();
     }
 
-    @Override
-    default int length() {
-        return size();
-    }
-
-    @Override
+        @Override
     default <U> Tree<U> map(Function<? super T, ? extends U> mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
         return isEmpty() ? Empty.instance() : TreeModule.Map.apply((Node<T>) this, mapper);
@@ -717,6 +722,8 @@ public interface Tree<T> extends Traversable<T> {
         private final T value;
         private final List<Node<T>> children;
 
+        private final int size;
+
         /**
          * Constructs a rose tree branch.
          *
@@ -729,6 +736,7 @@ public interface Tree<T> extends Traversable<T> {
             Objects.requireNonNull(children, "children is null");
             this.value = value;
             this.children = children;
+            this.size = 1 + this.children.foldLeft(0, (acc, child) -> acc + child.size);
         }
 
         @Override
@@ -747,13 +755,13 @@ public interface Tree<T> extends Traversable<T> {
         }
 
         @Override
-        public boolean isLeaf() {
-            return children.isEmpty();
+        public int length() {
+            return size;
         }
 
         @Override
-        public int size() {
-            return 1 + children.foldLeft(0, (acc, child) -> acc + child.length());
+        public boolean isLeaf() {
+            return size == 1;
         }
 
         @Override
@@ -937,13 +945,13 @@ public interface Tree<T> extends Traversable<T> {
         }
 
         @Override
-        public boolean isLeaf() {
-            return false;
+        public int length() {
+            return 0;
         }
 
         @Override
-        public int size() {
-            return 0;
+        public boolean isLeaf() {
+            return false;
         }
 
         @Override
--- a/javaslang/src/main/java/javaslang/collection/List.java
+++ b/javaslang/src/main/java/javaslang/collection/List.java
@@ -5,15 +5,36 @@
  */
 package javaslang.collection;
 
-import javaslang.*;
+import javaslang.Function1;
+import javaslang.Kind1;
+import javaslang.Tuple;
+import javaslang.Tuple2;
+import javaslang.Tuple3;
 import javaslang.collection.List.Nil;
 import javaslang.collection.ListModule.Combinations;
 import javaslang.collection.ListModule.SplitAt;
 import javaslang.control.Option;
 
-import java.io.*;
-import java.util.*;
-import java.util.function.*;
+import java.io.IOException;
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.ListIterator;
+import java.util.NavigableSet;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.BinaryOperator;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
 import java.util.stream.Collector;
 
 /**
--- a/javaslang/src/test/java/javaslang/collection/TreeTest.java
+++ b/javaslang/src/test/java/javaslang/collection/TreeTest.java
@@ -243,6 +243,18 @@ public class TreeTest extends AbstractTraversableTest {
     public void shouldNotGetValueOfNonNil() {
         assertThat(tree.getValue()).isEqualTo(1);
     }
+    
+    // -- size
+
+    @Test
+    public void shouldCalculateSizeOfALeaf() {
+        assertThat($(0).size()).isEqualTo(1);
+    }
+    
+    @Test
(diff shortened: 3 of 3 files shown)


javaslang/src/main/java/javaslang/collection/Tree.java
```diff
@@ -729,6 +751,7 @@ public Node(T value, List<Node<T>> children) {
             Objects.requireNonNull(children, "children is null");
             this.value = value;
             this.children = children;
+            this.size = children.size() + 1;
```
ruslansennov: children.foldLeft(...)

pivovarit: /**
     * Returns the number of nodes (including root and leafs).
     *
     * @return The size of the tree.
     */

So this should be applied recursively?

ruslansennov: children.foldLeft(...) will be recursively (see current implementation of Tree.size() method)

javaslang/src/main/java/javaslang/collection/Tree.java
```diff
@@ -729,6 +751,7 @@ public Node(T value, List<Node<T>> children) {
             Objects.requireNonNull(children, "children is null");
             this.value = value;
             this.children = children;
+            this.size = nodeCount();
```
paplorinc: hmm, but this makes it eager, what's the advantage of that?

javaslang/src/main/java/javaslang/collection/Tree.java
```diff
@@ -748,12 +771,12 @@ public boolean isEmpty() {
...
         @Override
         public boolean isLeaf() {
-            return children.isEmpty();
+            return size == 1;
         }
 
         @Override
         public int size() {
-            return 1 + children.foldLeft(0, (acc, child) -> acc + child.length());
+            return size;
```
paplorinc: maybe `memoize`, if you think this is an expensive, often called method

javaslang/src/main/java/javaslang/collection/Tree.java
```diff
@@ -11,12 +11,32 @@
 import javaslang.collection.List.Nil;
 import javaslang.collection.Tree.Empty;
 import javaslang.collection.Tree.Node;
-import javaslang.collection.TreeModule.*;
+import javaslang.collection.TreeModule.FlatMap;
```
paplorinc: In my opinion you should follow the author's style for imports instead

javaslang/src/main/java/javaslang/collection/Tree.java
pivovarit: That it does not need to be recalculated each time for each leaf recursively

paplorinc: I think I answered that below

javaslang/src/main/java/javaslang/collection/Tree.java
pivovarit: It's not really my idea. Check #1248 
Memoization is imo worth thinking about, thanks!

javaslang/src/main/java/javaslang/collection/Tree.java
pivovarit: I analyzed another classes and found out that there is no really any "common denominator" when considering importing strategy, @danieldietrich am I right?

javaslang/src/main/java/javaslang/collection/Tree.java
paplorinc: I'm wondering why didn't `size` use `nodeCount` before? (there seem to be other related size calculations here also, maybe unify them also?)

By `memoization` I meant lazy init in this case, sorry.

pivovarit: Because someone omitted this when writing those classes.

I think lazy init might be not the best case with recursive operations. The cost of calculating size when creating a single Node is nothing, but when you need to traverse whole tree and then compute size recursively for every Node it might be painful.

Anyway, the best argument I can come up with is:
Authors seem to have followed the same principle when creating other classes eg. `List`

paplorinc: I'm not sure I can follow, it seems to me that the two calculations are exactly the same (recursive iteration in both cases).

* The current impl is faster, if you call `size` 0 times
* The new imp is faster if you call it `> 1` times

With lazy init would be the best of both worlds.

In the constructor (or rather after the field)
```Java
this.size = Lazy.of(() -> nodeCount());
```

and use it as:
```Java
public int size() { return size.get(); }
```

pivovarit: I agree. This is what I meant by "memoization". Making it lazy initialized at the beginning and then remembering the result. I guess this issue is worth bringing up and applying globally in Javaslang

paplorinc: @danieldietrich, should we convert all computationally expensive operations (like size) to memoized `Lazy`?

danieldietrich: @paplorinc Sorry for the late response - I returned from holidays.

I recently removed all lazy types from internal collection instance vars like `size` and `hash`. For now please do not use Lazy for size. We trade memory for cpu cycles. If needed we may change it later without the need to change public API or Tree semantics.

javaslang/src/main/java/javaslang/collection/Tree.java
danieldietrich: Currently most of us use IntelliJ IDEA's "organize imports" feature. The IDAE settings are included in the repo (see folder `.ide`). I have no strong opinion regarding imports. The generated sources' import section differs too.

javaslang/src/main/java/javaslang/collection/Tree.java
danieldietrich: `nodeCount()` and `size()` should return `this.size`.

`this.size()` is initialized within the constructor (using fold...)

paplorinc: Hmmm... I don't see why a `Lazy` should be a real compromise, `get`-ing its value should be a lot faster than recalculating a tree/map/list size.

If it is, maybe we should change the double-check locking in `Lazy#get` (which many consider an anti-pattern anyway, especially since you're not checking against a `volatile`) to a static singleton.

I will experiment with it, let's see what the performance is of both :)

ruslansennov: > I will experiment with it, let's see what the performance is of both :)

Also <code>Lazy</code> requires 48 bytes per instance.

danieldietrich: @paplorinc @ruslansennov ok, for trees with a huge branching factor it may be worth saving the time - let's experiment with it. I'm mainly interested in saving the memory. Ruslan, the 48 bytes are just for the reference, not for the whole Lazy object, aren't they? We also have a Supplier within the Lazy etc.

ruslansennov: > the 48 bytes are just for the reference, not for the whole Lazy object, aren't they?

It is size of object. I believe that reference's size is smaller :)

paplorinc: `Lazy` is pretty fast already, but can slightly be optimized by minimizing the volatile reads, like:
```Java
@Override
public T get() {
    T result = value; // volatile field is accessed fewer times this way
    if (result == null) {
        synchronized (this) {
            result = value;
            if (result == null) {
                result = value = supplier.get();
                supplier = null; // free mem
            }
        }
    }
    return result;
}
```
though that will imply that the `Supplier` may not return `null`.
The speed advantages would be something like:
```
10000000 Lazys, 100 iterations (lot of GC)

// Before 
eagerGetDuration: 7755
initDuration: 19958
calculationDuration: 379
lazyGetDuration: 15146

vs.

// After
eagerGetDuration: 7307
initDuration: 28623
calculationDuration: 428
lazyGetDuration: 11356
```

or
```
1000000 Lazys, 1000 iterations (less GC)

Before
eagerGetDuration: 8866
initDuration: 60
calculationDuration: 35
lazyGetDuration: 11731


```

danieldietrich: @ruslansennov 

> It is size of object. I believe that reference's size is smaller :)

sounds good

@pivovarit 

Many thanks for your investigation! for now we should not optimize Lazy - the current impl is bulletproof and easy. @jbgi also made some basic benchmarks comparing different methods for lazy evaluation (also used in scalaz) and it turned out that the current impl is as fast as the other methods.

I'm not happy to increase the GC significantly. GC implicitly decreases the overall time-performance of the application. So the time-benefit of using Lazy instances may be reduced by having more time-consuming GCs.

I think it depends on the use-case what is the best solution. Also: How probable is it that size() is called? If the internal implementation of the collection methods will depend on size() calls (we've done that for optimizations) then it comes handy to have size already computed (we save memory and GC). Also when size() is not called at all in some applications then the Lazy instances are unnecessary.

Having a Lazy size is an optimization. Maybe it is premature because we currently have no reason to optimize the implementation. I like to implement the simplest solution possible and optimize when necessary.

danieldietrich:
@pivovarit The current impl of the `size` method and the Node constructor look good to me! Let's get the PR through the door by cleaning up a bit. We have three similar methods (which is ok): `nodeCount`, `size` and `length`. Let's do the following:

1) remove `size()` method in Tree, Node and Empty because it is implemented in Traversable as follows:

```java
default int size() {
    return length();
}
```

2) unify `nodeCount()` and `length()` methods by changing the default impl in Tree as follows:

```java
interface Tree<T> ... {
    default int nodeCount() {
        return length();
    }
}
```

3) remove Tree.length() default impl and implement Node.length() and Empty.length() the same way Node.size() and Empty.size() was implemented before:

```java
interface Tree<T> {
    class Node<T> implements Tree<T> {
        int length() { return size; }
    }
    class Empty<T> implements Tree<T> {
        int length() { return 0; }
    }
}
```

Many thanks!

pivovarit:
@danieldietrich I updated my changes

javaslang/src/main/java/javaslang/collection/Tree.java
```diff
@@ -334,11 +348,7 @@ default int leafCount() {
      * @return The number of nodes of this tree.
      */
     default int nodeCount() {
-        if (isEmpty()) {
-            return 0;
-        } else {
-            return 1 + getChildren().foldLeft(0, (count, child) -> count + child.nodeCount());
-        }
+        return length();
```
paplorinc: Kinda' unrelated, but why is this a default method exactly?

javaslang/src/main/java/javaslang/collection/Tree.java
```diff
@@ -937,13 +950,13 @@ public boolean isEmpty() {
         }
 
         @Override
-        public boolean isLeaf() {
-            return false;
+        public int length() {
+            return 0;
```
paplorinc: hmmm, length of 0?

javaslang/src/main/java/javaslang/collection/Tree.java
danieldietrich: This is a very good question! In 2013 (I think) I attended the Coursera course _progfun_ by Martin Odersky/EPFL. There was a assignment where we had to implement some data structure using traits (I think). In my first submission I missed some points because I didn't implemented a method similar to this default method.

I thought about it and now I design the interfaces/classes the way that only methods that are atomic / cannot be deduced by other methods are implemented within the implementing classes (like List.Nil, List.Cons). Other methods are implemented as default method within the interface. I think in general this makes it more easy to provide different implementations for one interface - all reduces to the most important methods.

javaslang/src/main/java/javaslang/collection/Tree.java
```diff
@@ -747,13 +755,18 @@ public boolean isEmpty() {
...
-            return children.isEmpty();
+            return size == 1;
         }
 
         @Override
-        public int size() {
-            return 1 + children.foldLeft(0, (acc, child) -> acc + child.length());
+        public int nodeCount() {
+            return size;
         }
```
danieldietrich: just one commit away from merging the PR :-) we do not need this method here because Tree has a default nodeCount() method.

javaslang/src/main/java/javaslang/collection/Tree.java
paplorinc: Hmm, do you have an article which explains this in more detail, sounds interesting. If not, you should write one (maybe even in the Javaslang documentation, if you think it adds functional value to Java) :)

pivovarit: Well, this idea is pretty simple. Default methods are basically a redundant helper methods. You can obtain the same results by manipulating abstract methods

paplorinc: So you're simply using it for multiple implementation inheritance (traits style), right?

pivovarit: I believe an answer to your question is: yes, if I understood it correctly :)

pivovarit: @danieldietrich I am not sure if you get notifications when I force push an updated commit. I applied your last suggestions

danieldietrich: > So you're simply using it for multiple implementation inheritance (traits style), right?

Yes. But there is one more reason: High cohesion. It is nice to see the behavior of a function (/method) in one place. In our example we have just two distinct impls (Empty and Node) but there are situations where we might have even more (e.g. Root, Inner, Leaf). Spreading the behavior of one method over the classes makes an algorithm hard to read.

An Interface/Trait is like a blueprint of a type, we see how it basically works. By looking at the impls we see the specific characteristics. Those are often one-pagers.

danieldietrich: @pivovarit yep, you're right - I recieved only a notification of your comment.

Great work, thx!

Issue #1252, opened 2016-03-31
CharSeq.split() returns an array & not a Seq


danieldietrich:
In general, returning native arrays `[]` is not very helpful. We should revise our API.

sajit:
Hello there. first time foray into trying to contribute to OSS project so please forgive if this sounds stupid. Just that I understand the requirement you want this method to return Seq<Character>

https://github.com/javaslang/javaslang/blob/master/javaslang/src/main/java/javaslang/collection/Seq.java

danieldietrich:
@sajit In general, the split result will be further processed.

We want to write:

```java
CharSeq.of(csvLine).split(";").map(String::trim);
```

but having a result of type `CharSeq[]` we currently are forced to write

```java
Vector.of(CharSeq.of(csvLine).split(";")).map(String::trim);
```

danieldietrich:
Thanks guys! Seems @gde-vt also fixed this issue on a local branch :)

danieldietrich:
Hey guys,

I've sleepless nights because this little change will break backward compatibility which may be a major issue regarding transitive library dependencies (e.g. two maven libs, one depends on the other, both use different Javaslang versions that are not compatible).

Such a change as this should

* either be made in a major release
* or we need to add an additional method without changing the original one

So we could defer this change to Javaslang 3.0.0 _or_ introduce new methods, e.g. `splitSeq`.

What do you think @sajit, @ruslansennov, @paplorinc ?

ruslansennov:
my vote for a new method splitSeq

emmanueltouzery:
Maybe `splitSeq` and `@Deprecated` on `split`?

danieldietrich:
I'm fighting with myself - targeting the most simple API possible, the new method's name would be `split`.

If we name it `splitSeq` and remove `split` in 3.0.0 we still have the problem that a third party lib that uses a 2.x version of Javaslang will not work any more if we use both in a new application, the third party lib and Javaslang 3.0.0. Maven for example will resolve the dependency conflict and choose Javaslang 3.0.0 over 2.x. The third party lib will then be forced to use Javaslang 3.0.0 at runtime, even if it is compiled against 2.x.

Now we are in backward-compatibility hell, like Java. We could keep the current `split` instead of deprecating and removing it. But from a birds-view this is a big problem. It would be like a stagnation in the evolution of the API and we would keep mistakes forever, duplicate similar methods like in this case and so on.

[...]

PR #1267, opened 2016-04-12
Issue 1252

sajit:
An attempt to address issue https://github.com/javaslang/javaslang/issues/1252

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/CharSeq.java
+++ b/javaslang/src/main/java/javaslang/collection/CharSeq.java
@@ -2132,16 +2132,13 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      */
     public Seq<CharSeq> split(String regex, int limit) {
         final String[] javaStrings = back.split(regex, limit);
-        final IndexedSeq<CharSeq> result = Vector.empty();
-
-
-        for (int i = 0; i < javaStrings.length; i++) {
-            result.append(CharSeq.of(javaStrings[i]));
-
+        final CharSeq[] strings = new CharSeq[javaStrings.length];
+        for (int i = 0; i < strings.length; i++) {
+            strings[i] = of(javaStrings[i]);
         }
-
-        return result;
+        return Vector.of(strings);
     }
+
     /**
      * Splits this string around matches of the given <a
      * href="../util/regex/Pattern.html#sum">regular expression</a>.
--- a/javaslang/src/main/java/javaslang/collection/CharSeq.java
+++ b/javaslang/src/main/java/javaslang/collection/CharSeq.java
@@ -2125,7 +2125,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      *
      * @param regex the delimiting regular expression
      * @param limit the result threshold, as described above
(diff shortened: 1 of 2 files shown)


javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -2130,13 +2130,13 @@ public CharSeq replace(CharSequence target, CharSequence replacement) {
      * @throws PatternSyntaxException if the regular expression's syntax is invalid
      * @see java.util.regex.Pattern
      */
-    public CharSeq[] split(String regex, int limit) {
+    public Seq<CharSeq> split(String regex, int limit) {
```
ruslansennov: Could you also change javadoc?

danieldietrich:
Hi @sajit, thanks for your contribution! The code looks good to me. I will merge it when the javadoc is updated. I think it is sufficient to update the `@return` tag like this `s/array/Seq/`:

```java
/**
 * ...
 * @return the Seq of strings computed by splitting this string
 * around matches of the given regular expression
 * ...
```

paplorinc:
Please name the commits to represent logical actions: they should make the review as simple as possible.
You can `rebase` and `force push` them back, squashing related commits and separating unrelated modifications.
After push, please comment the code to explain `why` you changed certain things (that might not be immediately obvious).
Thanks :)

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -2130,7 +2130,7 @@ public CharSeq replace(CharSequence target, CharSequence replacement) {
      * @throws PatternSyntaxException if the regular expression's syntax is invalid
      * @see java.util.regex.Pattern
      */
-    public IndexedSeq<CharSeq> split(String regex, int limit) {
+    public Seq<CharSeq> split(String regex, int limit) {
```
paplorinc: please amend your previous commit instead, in case you made a mistake.

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -2132,16 +2132,13 @@ public CharSeq replace(CharSequence target, CharSequence replacement) {
      */
     public Seq<CharSeq> split(String regex, int limit) {
         final String[] javaStrings = back.split(regex, limit);
-        final IndexedSeq<CharSeq> result = Vector.empty();
-
-
-        for (int i = 0; i < javaStrings.length; i++) {
-            result.append(CharSeq.of(javaStrings[i]));
-
+        final CharSeq[] strings = new CharSeq[javaStrings.length];
```
paplorinc: You might want to do it in a functional way instead.
The error in your previous code is probably that you the result of the append is ignored.
Please only commit, when all tests are passing.

```Java
public Seq<CharSeq> split(String regex, int limit) {
    Array<String> wrap = Array.wrap(back.split(regex, limit));
    return wrap.map(CharSeq::of);
}
```

danieldietrich:
@sajit thanks for your contribution and for updating the docs!
@paplorinc you're right - naming commits appropriately helps. I'm also a candidadate to take more care here ;-)

PR #1269, opened 2016-04-14
Option cleanup

paplorinc:
Fixes https://github.com/javaslang/javaslang/issues/1272

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/AbstractMap.java
+++ b/javaslang/src/main/java/javaslang/collection/AbstractMap.java
@@ -68,11 +68,11 @@ abstract class AbstractMap<K, V, M extends AbstractMap<K, V, M>> implements Map<
     public M drop(long n) {
         if (n <= 0) {
             return (M) this;
-        }
-        if (n >= length()) {
+        } else if (n >= length()) {
             return emptyInstance();
+        } else {
+            return createFromEntries(iterator().drop(n));
         }
-        return createFromEntries(iterator().drop(n));
     }
 
     @SuppressWarnings("unchecked")
@@ -80,11 +80,11 @@ abstract class AbstractMap<K, V, M extends AbstractMap<K, V, M>> implements Map<
     public M dropRight(long n) {
         if (n <= 0) {
             return (M) this;
-        }
-        if (n >= length()) {
+        } else if (n >= length()) {
             return emptyInstance();
+        } else {
+            return createFromEntries(iterator().dropRight(n));
         }
-        return createFromEntries(iterator().dropRight(n));
     }
 
     @Override
@@ -123,20 +123,20 @@ abstract class AbstractMap<K, V, M extends AbstractMap<K, V, M>> implements Map<
         return sliding(size, size);
     }
 
-    @SuppressWarnings("unchecked")
+    @Override
+    public abstract M init();
+
     @Override
     public Option<M> initOption() {
-        return isEmpty() ? Option.none() : Option.some((M) init());
+        return isEmpty() ? Option.none() : Option.some(init());
     }
 
-    @SuppressWarnings("unchecked")
+    @Override
+    public abstract M tail();
+
     @Override
     public Option<M> tailOption() {
-        if (isEmpty()) {
-            return Option.none();
-        } else {
-            return Option.some((M) tail());
-        }
+        return isEmpty() ? Option.none() : Option.some(tail());
     }
 
     @SuppressWarnings("unchecked")
@@ -227,7 +227,7 @@ abstract class AbstractMap<K, V, M extends AbstractMap<K, V, M>> implements Map<
     public M replace(Tuple2<K, V> currentElement, Tuple2<K, V> newElement) {
         Objects.requireNonNull(currentElement, "currentElement is null");
         Objects.requireNonNull(newElement, "newElement is null");
-        return containsKey(currentElement._1) ? (M) remove(currentElement._1).put(newElement) : (M) this;
+        return (M) (containsKey(currentElement._1) ? remove(currentElement._1).put(newElement) : this);
     }
 
     @Override
--- a/javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
+++ b/javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
@@ -24,7 +24,7 @@ import java.util.function.*;
  * @author Ruslan Sennov
  * @since 2.0.0
  */
- abstract class AbstractMultimap<K, V, M extends Multimap<K, V>> implements Multimap<K, V> {
+abstract class AbstractMultimap<K, V, M extends Multimap<K, V>> implements Multimap<K, V> {
 
     interface SerializableSupplier<T> extends Supplier<T>, Serializable {
     }
@@ -81,7 +81,7 @@ import java.util.function.*;
     @Override
     public <K2, V2> Multimap<K2, V2> flatMap(BiFunction<? super K, ? super V, ? extends Iterable<Tuple2<K2, V2>>> mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
-        return foldLeft(this.<K2, V2>emptyInstance(), (acc, entry) -> {
+        return foldLeft(this.emptyInstance(), (acc, entry) -> {
             for (Tuple2<? extends K2, ? extends V2> mappedEntry : mapper.apply(entry._1, entry._2)) {
                 acc = acc.put(mappedEntry);
             }
@@ -102,7 +102,7 @@ import java.util.function.*;
     @Override
     public <K2, V2> Multimap<K2, V2> map(BiFunction<? super K, ? super V, Tuple2<K2, V2>> mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
-        return foldLeft(this.<K2, V2>emptyInstance(), (acc, entry) -> acc.put(mapper.apply(entry._1, entry._2)));
+        return foldLeft(this.emptyInstance(), (acc, entry) -> acc.put(mapper.apply(entry._1, entry._2)));
     }
 
     @Override
@@ -116,7 +116,7 @@ import java.util.function.*;
     public M put(K key, V value) {
         final Traversable<V> values = back.get(key).getOrElse((Traversable<V>) emptyContainer.get());
         final Traversable<V> newValues = containerType.add(values, value);
-        return newValues == values ? (M) this : (M) createFromMap(back.put(key, newValues));
+        return (M) (newValues == values ? this : createFromMap(back.put(key, newValues)));
     }
 
     @Override
@@ -128,7 +128,7 @@ import java.util.function.*;
     @SuppressWarnings("unchecked")
     @Override
     public M remove(K key) {
-        return back.containsKey(key) ? (M) createFromMap(back.remove(key)) : (M) this;
+        return (M) (back.containsKey(key) ? createFromMap(back.remove(key)) : this);
     }
 
     @SuppressWarnings("unchecked")
@@ -136,14 +136,20 @@ import java.util.function.*;
     public M remove(K key, V value) {
         final Traversable<V> values = back.get(key).getOrElse((Traversable<V>) emptyContainer.get());
         final Traversable<V> newValues = containerType.remove(values, value);
-        return newValues == values ? (M) this : newValues.isEmpty() ? (M) createFromMap(back.remove(key)): (M) createFromMap(back.put(key, newValues));
+        if (newValues == values) {
+            return (M) this;
+        } else if (newValues.isEmpty()) {
+            return (M) createFromMap(back.remove(key));
+        } else {
+            return (M) createFromMap(back.put(key, newValues));
+        }
     }
 
     @SuppressWarnings("unchecked")
     @Override
     public M removeAll(Iterable<? extends K> keys) {
         Map<K, Traversable<V>> result = back.removeAll(keys);
-        return result == back ? (M) this : (M) createFromMap(result);
+        return (M) (result == back ? this : createFromMap(result));
     }
 
     @Override
@@ -159,11 +165,7 @@ import java.util.function.*;
     @SuppressWarnings("unchecked")
     @Override
     public M distinct() {
-        if(containerType == ContainerType.SEQ) {
-            return (M) createFromEntries(iterator().distinct());
-        } else {
-            return (M) this;
-        }
+        return (M) (containerType == ContainerType.SEQ ? createFromEntries(iterator().distinct()) : this);
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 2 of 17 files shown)


javaslang/src/main/java/javaslang/collection/AbstractMap.java
```diff
@@ -21,238 +21,238 @@
  *
  * @param <K> Key type
  * @param <V> Value type
- * @param <M> Map type
+ * @param <R> Map type
  * @author Ruslan Sennov
  * @since 2.0.0
  */
-abstract class AbstractMap<K, V, M extends AbstractMap<K, V, M>> implements Map<K, V> {
+abstract class AbstractMap<K, V, R extends AbstractMap<K, V, R>> implements Map<K, V> {
```
paplorinc: **M** = **M**ap
**R** = **R**esult

we will probably need it for non-maps also (e.g. `Queues`)

javaslang/src/main/java/javaslang/collection/AbstractMap.java
```diff
@@ -21,238 +21,238 @@
...
-    public M drop(long n) {
+    public R drop(long n) {
         if (n <= 0) {
-            return (M) this;
-        }
-        if (n >= length()) {
+            return (R) this;
+        } else if (n >= length()) {
             return emptyInstance();
+        } else {
```
paplorinc: `else if` instead of `if` signals to the user that the branches are mutually exclusive

javaslang/src/main/java/javaslang/collection/AbstractMap.java
```diff
@@ -21,238 +21,238 @@
...
     }
 
     @SuppressWarnings("unchecked")
     @Override
-    public M replace(Tuple2<K, V> currentElement, Tuple2<K, V> newElement) {
+    public R replace(Tuple2<K, V> currentElement, Tuple2<K, V> newElement) {
         Objects.requireNonNull(currentElement, "currentElement is null");
         Objects.requireNonNull(newElement, "newElement is null");
-        return containsKey(currentElement._1) ? (M) remove(currentElement._1).put(newElement) : (M) this;
+        return (R) (containsKey(currentElement._1) ? remove(currentElement._1).put(newElement) : this);
```
paplorinc: single cast

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
```diff
@@ -113,37 +113,43 @@ public ContainerType getContainerType() {
...
     }
 
     @SuppressWarnings("unchecked")
     @Override
-    public M remove(K key, V value) {
+    public R remove(K key, V value) {
         final Traversable<V> values = back.get(key).getOrElse((Traversable<V>) emptyContainer.get());
         final Traversable<V> newValues = containerType.remove(values, value);
-        return newValues == values ? (M) this : newValues.isEmpty() ? (M) createFromMap(back.remove(key)): (M) createFromMap(back.put(key, newValues));
+        if (newValues == values) {
```
paplorinc: confusing ternary-in-ternary

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
```diff
@@ -258,7 +260,7 @@ public M init() {
 
     @SuppressWarnings("unchecked")
     @Override
-    public Option<Multimap<K, V>> initOption() {
+    public Option<R> initOption() {
```
paplorinc: if `init` is `R`, than `initOption` should be `Option<R>`

javaslang/src/main/java/javaslang/Lazy.java
```diff
@@ -127,7 +127,7 @@ private Lazy(Supplier<? extends T> supplier) {
 
     public Option<T> filter(Predicate<? super T> predicate) {
         T v = get();
-        return predicate.test(v) ? Option.some(v) : Option.none();
+        return Option.when(predicate.test(v), () -> v);
```
paplorinc: let's eat our own dogfood :)

javaslang/src/main/java/javaslang/collection/AbstractMap.java
```diff
@@ -122,21 +122,19 @@ public R filter(Predicate<? super Tuple2<K, V>> predicate) {
     public Iterator<R> grouped(long size) {
         return sliding(size, size);
     }
+    public abstract R init();
```
paplorinc: this is better than casting

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -641,11 +640,6 @@ public T head() {
     }
 
     @Override
-    public Option<T> headOption() {
```
paplorinc: this doesn't return `M`, therefore it can be implemented in the parent, where which has `T` already.
Note: reducing the number of method decreased the code coverage slightly.

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -858,11 +854,7 @@ public CharSeq tail() {
 
     @Override
     public Option<CharSeq> tailOption() {
-        if (isEmpty()) {
-            return Option.none();
-        } else {
-            return Option.some(CharSeq.of(back.substring(1)));
```
paplorinc: logic was copy pasted

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1746,12 +1738,10 @@ default String stringPrefix() {
...
     @Override
     default Option<Iterator<T>> tailOption() {
-        if (hasNext()) {
+        return Option.when(hasNext(), () -> {
             next();
-            return Option.some(this);
-        } else {
-            return Option.none();
-        }
+            return this;
```
paplorinc: this seems very mutable, but that might be on purpose

javaslang/src/main/java/javaslang/collection/Seq.java
```diff
@@ -437,7 +437,7 @@ default int lastIndexWhere(Predicate<? super T> predicate) {
      * this sequence in a Some if the index is within bounds, otherwise a None.
      */
     default Function1<Integer, Option<T>> lift() {
-        return i -> (i >= length() || i < 0) ? Option.none() : Option.some(apply(i));
+        return i -> Option.when(i < length() && i >= 0, () -> apply(i));
```
paplorinc: inverted the condition

javaslang/src/main/java/javaslang/collection/Traversable.java
```diff
@@ -490,15 +491,23 @@ default T get() {
      *
      * @return {@code Some(traversable)} or {@code None} if this is empty.
      */
-    Option<? extends Traversable<T>> initOption();
+    default Option<? extends Traversable<T>> initOption() {
+        return whenNotEmpty(this::init);
+    }
+
+    default <R> Option<R> whenNotEmpty(Supplier<R> supplier) {
```
paplorinc: if we don't want to expose this (i.e. no `protected` in interfaces), maybe we could import it statically (though we won't be able to access `isEmpty` then, and there would be no advantage of having this method...

javaslang/src/main/java/javaslang/collection/AbstractMap.java
danieldietrich: yes, using `else if` is correct here. in other places we use if-guards without else branch but that pattern does not apply here.

javaslang/src/main/java/javaslang/collection/AbstractMap.java
danieldietrich: Classical naming question - highly opinionated. I would stay with M because it is specific to AbstractMap which captures commonalities between Map impls. The M definitely denotes a Map type. Technically it is used as a result type but it does not need necessarily to. M may be (theoretically) used in arbitrary places, e.g. return type, local var etc...

If we need such param for an AbstractQueue it will be Q. That type will not be related to AbstractMap.

javaslang/src/main/java/javaslang/collection/AbstractMap.java
danieldietrich: looks good

javaslang/src/main/java/javaslang/collection/AbstractMap.java
paplorinc: in my opinion they will represent the same thing.
Will rename back, if you don't agree :)

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
danieldietrich: yes, more readable now!

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
```diff
@@ -229,8 +231,8 @@ public M filter(Predicate<? super Tuple2<K, V>> predicate) {
         return foldLeft(HashMap.empty(), (map, entry) -> {
             final C key = classifier.apply(entry);
             final Multimap<K, V> values = map.get(key)
-                    .map(entries -> entries.put(entry._1, entry._2))
-                    .getOrElse(createFromEntries(Iterator.of(entry)));
+                                             .map(entries -> entries.put(entry._1, entry._2))
+                                             .getOrElse(createFromEntries(Iterator.of(entry)));
```
danieldietrich: please try not to reformat existing code. Do you use Eclipse? Most of us use IntelliJ IDEA. The repo includes the IDE settings...

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
danieldietrich: yes, sounds reasonable!

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
paplorinc: No, I'm using `Idea`, with the imported settings, I just have `Align when multiline` enabled everyhere.
Would you like me to revert this part?

javaslang/src/main/java/javaslang/collection/AbstractMap.java
danieldietrich: does it need an `@Override`

paplorinc: indeed, thanks!

javaslang/src/main/java/javaslang/collection/List.java
```diff
@@ -142,6 +142,8 @@
         return Nil.instance();
     }
 
+    boolean isEmpty();
```
paplorinc: this was needed to unapply the default method.
TODO: this needs an `@Override` also

javaslang/src/main/java/javaslang/collection/Array.java
danieldietrich: This needs to be implemented in `Traversable` (`lastOption` actually is)

paplorinc: yup, that's what I did

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: An iterator actually internally is mutable, like in Scala. Therefore calling multiple intermediate operations might produce unexpected results (see javadoc).

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
danieldietrich: yes, please revert, we all should use the same settings

paplorinc: Done, pushed a new version, squashing the changes (as the rename isn't separated anymore).

javaslang/src/main/java/javaslang/collection/Traversable.java
```diff
@@ -475,7 +474,9 @@ default T get() {
      *
      * @return {@code Some(element)} or {@code None} if this is empty.
      */
-    Option<T> headOption();
+    default Option<T> headOption() {
```
paplorinc: made it symmetric with `lastOption`

javaslang/src/main/java/javaslang/collection/Seq.java
danieldietrich: Good!
Let's switch the conditions, too: `Option.when(i >= 0 && i < length(), ...)`. Then we omit the `length()` call if `i < 0`.

paplorinc: good call :)

paplorinc: good call, done :)!

javaslang/src/main/java/javaslang/Lazy.java
danieldietrich: Ohh sh... - the implementation of Option.when is broken: if the supplier returns a null we create currently no Some instance, see https://github.com/javaslang/javaslang/blob/master/javaslang/src/main/java/javaslang/control/Option.java#L137

Option.when needs to be fixed to call some(...) instead of of(...)

danieldietrich: See #1272

javaslang/src/main/java/javaslang/collection/Traversable.java
danieldietrich: Yes, we should hide it. A good place is within javaslang.collection.Collections as static method.

danieldietrich:
Hi Pap,
I have thought about this PR a while. Your optimizations are well thought through. The code size reduces by using whenIsNotEmpty *but* we create an additional Supplier instance. This changes the runtime behavior of our collection. I'm afraid, therefore we can't apply that specific change of this PR.

paplorinc:
I changed the implementation of one of my applications from Java 8 api to Javaslang and it got 5x slower.

I think omitting these kind of changes -- that simplify the design -- are premature optimizations and aren't the bottleneck of the application.

paplorinc:
I don't have to agree to accommodate, I pushed a new version with a fix for nullable supplier and with non-supplier based Option creation :)

javaslang/src/main/java/javaslang/control/Option.java
```diff
@@ -134,7 +134,7 @@
      */
     static <T> Option<T> when(boolean condition, Supplier<? extends T> supplier) {
         Objects.requireNonNull(supplier, "supplier is null");
-        return condition ? of(supplier.get()) : none();
+        return condition ? some(supplier.get()) : none();
```
paplorinc: fixes https://github.com/javaslang/javaslang/issues/1272

danieldietrich:
Pap, I thank you! That was really fast. Your ideas are great. I think together we can bring Javaslang forward in the right direction.
I'm really looking forward to BitSet and PriorityQueue!

PR #1281, opened 2016-04-15
implemented alternative Try::filter method that accepts supplier for …

### LINKED ISSUE #1273: Try::filter and custom throwable (supplier)
What do think about extending Try with an additional filter method that expects a throwable or throwable supplier? I have a use case where I would like to concretize a problem by a custom Throwable that is more meaningful than a NoSuchElementException if a specific filter is applied. This makes further processing and analyzing of the Try instance easier and more flexible.

If also think that this makes sense I could do a PR.

manu-m:
…providing custom throwable if predicate evaluates to false. fixes #1273

### DIFF
--- a/javaslang/src/main/java/javaslang/control/Try.java
+++ b/javaslang/src/main/java/javaslang/control/Try.java
@@ -238,6 +238,19 @@ public interface Try<T> extends Value<T> {
     }
 
     /**
+     * Shortcut for {@code filterTry(predicate::test, throwableSupplier)}, see {@link #filterTry(CheckedPredicate)}}.
+     *
+     * @param predicate A predicate
+     * @return a {@code Try} instance
+     * @throws NullPointerException if {@code predicate} or {@code throwableSupplier} is null
+     */
+    default Try<T> filter(Predicate<? super T> predicate, Supplier<? extends Throwable> throwableSupplier) {
+        Objects.requireNonNull(predicate, "predicate is null");
+        Objects.requireNonNull(predicate, "throwableSupplier is null");
+        return filterTry(predicate::test, throwableSupplier);
+    }
+
+    /**
      * Shortcut for {@code filterTry(predicate::test)}, see {@link #filterTry(CheckedPredicate)}}.
      *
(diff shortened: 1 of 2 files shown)


javaslang/src/main/java/javaslang/control/Try.java
```diff
@@ -238,6 +238,19 @@
...
     /**
+     * Shortcut for {@code filterTry(predicate::test, throwableSupplier)}, see {@link #filterTry(CheckedPredicate)}}.
+     *
+     * @param predicate A predicate
+     * @return a {@code Try} instance
+     * @throws NullPointerException if {@code predicate} or {@code throwableSupplier} is null
+     */
+    default Try<T> filter(Predicate<? super T> predicate, Supplier<? extends Throwable> throwableSupplier) {
+        Objects.requireNonNull(predicate, "predicate is null");
+        Objects.requireNonNull(predicate, "throwableSupplier is null");
```
paplorinc: should rather be
```Java
Objects.requireNonNull(throwableSupplier, "throwableSupplier is null");
```

paplorinc:
Nice! :)

If I may suggest, If we've opened up `Try` already, please consider creating some tests for the uncovered methods also https://codecov.io/github/javaslang/javaslang/javaslang/src/main/java/javaslang/control/Try.java?ref=196914af043e5ba2b9860c0aae05ad8a04be2d2a :)

javaslang/src/main/java/javaslang/control/Try.java
danieldietrich: good catch

manu-m: Oh, a little oversight. Thank you for the hint

manu-m:
I will take look at the unit tests in more detail within the next days. I think can do some contribution, although I don't like javaslang's style of unit test at all. I will do a new PR for these contributions.

danieldietrich:
Great, many thanks!

> although I don't like javaslang's style of unit test

Let's improve it where possible...

Issue #1276, opened 2016-04-15
Simplify complexity of `removeAll`/`retainAll`/`containsAll` in `List`/`Stream`/`Vector`


paplorinc:
Doing a `contains` on
```Java
final List<T> removed = List.ofAll(elements).distinct();
```
is linear (`O(n)`), changing it simply to
```Java
final Set<T> removed = HashSet.ofAll(elements);
```
would simplify it to constat time (`O(1)`).

Should be investigated for the whole codebase.

danieldietrich:
Nice! Also the `removed.contains(element)` with be `O(log n)` instead of `O(n)` then.

PR #1285, opened 2016-04-16
Unified 'removeAll' and `retainAll` for `Traversable`s

paplorinc:
This should improve complexity (speed-wise and code-wise) significantly in many cases.

Coverage is reduced, because there are fewer lines :).

Fixes https://github.com/javaslang/javaslang/issues/1276

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Array.java
+++ b/javaslang/src/main/java/javaslang/collection/Array.java
@@ -11,7 +11,6 @@ import javaslang.control.Option;
 
 import java.io.Serializable;
 import java.util.*;
-import java.util.HashSet;
 import java.util.function.*;
 import java.util.stream.Collector;
 
@@ -593,7 +592,9 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
                 list.add(t);
             }
         }
-        if (list.size() == back.length) {
+        if (list.isEmpty()) {
+            return empty();
+        } else if (list.size() == size()) {
             return this;
         } else {
             return list.isEmpty() ? empty() : wrap(list.toArray());
@@ -899,38 +900,12 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
 
     @Override
     public Array<T> removeAll(T element) {
-        final java.util.List<T> list = new ArrayList<>();
-        for (int i = 0; i < length(); i++) {
-            T value = get(i);
-            if (!element.equals(value)) {
-                list.add(value);
-            }
-        }
-        if (list.size() == length()) {
-            return this;
-        } else {
-            return wrap(list.toArray());
-        }
+        return Collections.removeAll(this, element);
     }
 
     @Override
     public Array<T> removeAll(Iterable<? extends T> elements) {
-        final java.util.Set<T> removed = new HashSet<>();
-        for (T element : elements) {
-            removed.add(element);
-        }
-        final java.util.List<T> list = new ArrayList<>();
-        for (int i = 0; i < length(); i++) {
-            T value = get(i);
-            if (!removed.contains(value)) {
-                list.add(value);
-            }
-        }
-        if (list.size() == length()) {
-            return this;
-        } else {
-            return wrap(list.toArray());
-        }
+        return Collections.removeAll(this, elements);
     }
 
     @Override
@@ -971,23 +946,7 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
 
     @Override
     public Array<T> retainAll(Iterable<? extends T> elements) {
-        Objects.requireNonNull(elements, "elements is null");
-        final java.util.Set<T> kept = new HashSet<>();
-        for (T element : elements) {
-            kept.add(element);
-        }
-        final java.util.List<T> list = new ArrayList<>();
-        for (int i = 0; i < length(); i++) {
-            T value = get(i);
-            if (kept.contains(value)) {
-                list.add(value);
-            }
-        }
-        if (list.size() == length()) {
-            return this;
-        } else {
-            return wrap(list.toArray());
-        }
+        return Collections.retainAll(this, elements);
     }
 
     @Override
--- a/javaslang/src/main/java/javaslang/collection/CharSeq.java
+++ b/javaslang/src/main/java/javaslang/collection/CharSeq.java
@@ -9,8 +9,7 @@ import javaslang.*;
 import javaslang.collection.CharSeqModule.Combinations;
 import javaslang.control.Option;
 
-import java.io.Serializable;
-import java.io.UnsupportedEncodingException;
+import java.io.*;
 import java.nio.charset.Charset;
 import java.util.*;
 import java.util.function.*;
@@ -359,7 +358,13 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
                 sb.append(ch);
             }
         }
-        return sb.length() == 0 ? EMPTY : sb.length() == length() ? this : of(sb);
+        if (sb.length() == 0) {
+            return EMPTY;
+        } else if (sb.length() == length()) {
+            return this;
+        } else {
+            return of(sb);
+        }
     }
 
     @Override
@@ -650,28 +655,12 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
 
     @Override
     public CharSeq removeAll(Character element) {
-        final StringBuilder sb = new StringBuilder();
-        for (int i = 0; i < length(); i++) {
-            final char c = get(i);
-            if (c != element) {
-                sb.append(c);
-            }
-        }
-        return sb.length() == 0 ? EMPTY : sb.length() == length() ? this : of(sb);
+        return Collections.removeAll(this, element);
     }
 
     @Override
     public CharSeq removeAll(Iterable<? extends Character> elements) {
-        Objects.requireNonNull(elements, "elements is null");
-        final Set<Character> distinct = HashSet.ofAll(elements);
-        final StringBuilder sb = new StringBuilder();
-        for (int i = 0; i < length(); i++) {
-            final char c = get(i);
-            if (!distinct.contains(c)) {
-                sb.append(c);
-            }
-        }
-        return sb.length() == 0 ? EMPTY : sb.length() == length() ? this : of(sb);
+        return Collections.removeAll(this, elements);
     }
 
     @Override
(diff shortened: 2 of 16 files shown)


javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -899,38 +898,12 @@ public int length() {
...
-            if (!element.equals(value)) {
-                list.add(value);
-            }
-        }
-        if (list.size() == length()) {
-            return this;
-        } else {
-            return wrap(list.toArray());
-        }
+        return (Array<T>) IndexedSeq.super.removeAll(element);
```
paplorinc: cast is needed, as return value is the same as the subclass, which isn't available in the parent class.
Also, in case of trait-like inheritance, it seems that only the directly declared parents are accessible (hence all the delegating methods)

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -899,38 +898,12 @@ public int length() {
 
     @Override
     public Array<T> removeAll(T element) {
-        final java.util.List<T> list = new ArrayList<>();
-        for (int i = 0; i < length(); i++) {
```
paplorinc: as a bonus, this always results in the same `empty()` instance (if that's the case), with the new implementation

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1558,7 +1547,7 @@ default T reduceRight(BiFunction<? super T, ? super T, ? extends T> op) {
             throw new NoSuchElementException("reduceRight on Nil");
         } else {
             Stream<T> reversed = Stream.ofAll(this).reverse();
-            return reversed.tail().foldLeft(reversed.head(), (xs, x) -> op.apply(x, xs));
+            return reversed.reduceLeft((xs, x) -> op.apply(x, xs));
```
paplorinc: unrelated, but useful

javaslang/src/main/java/javaslang/collection/Seq.java
```diff
@@ -585,7 +584,11 @@ default int prefixLength(Predicate<? super T> predicate) {
      * @return a Seq containing all elements of this but none of the given elements.
      * @throws NullPointerException if {@code elements} is null
      */
-    Seq<T> removeAll(Iterable<? extends T> elements);
+    default Seq<T> removeAll(Iterable<? extends T> elements) {
+        Objects.requireNonNull(elements, "elements is null");
+        final Set<T> removed = HashSet.ofAll(elements);
+        return filter(e -> !removed.contains(e));
```
paplorinc: this is where the magic happens :)

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -593,7 +592,9 @@ public T get(int index) {
                 list.add(t);
             }
         }
-        if (list.size() == back.length) {
+        if (list.isEmpty()) {
```
paplorinc: changed most `filter` methods to return empty, if possible.
This affects many other methods also.

javaslang/src/main/java/javaslang/collection/Traversable.java
```diff
@@ -205,10 +201,9 @@
      * @return true, if this List contains all given elements, false otherwise.
      * @throws NullPointerException if {@code elements} is null
      */
-
     default boolean containsAll(Iterable<? extends T> elements) {
-        Objects.requireNonNull(elements, "elements is null");
-        return List.ofAll(elements).distinct().find(e -> !this.contains(e)).isEmpty();
+        HashSet<T> uniqueElements = HashSet.ofAll(elements);
+        return toSet().intersect(uniqueElements).size() == uniqueElements.size();
```
paplorinc: this is another exciting change :)

E.g. for a `LinkedList` it had a complexity of `O(elements.size * size())`.
Now it's `O(elements.size + size())`. For sets it is reduced to the size of the smaller set.

javaslang/src/test/java/javaslang/collection/AbstractTraversableTest.java
```diff
@@ -1326,34 +1326,45 @@ public void shouldReplaceAllElementsOfNonNilUsingCurrNew() {
...
+        Traversable<Integer> src = of(1, 2, 1, 2, 2);
+        Traversable<Integer> expected = of(1, 2, 1, 2, 2);
+        Traversable<Integer> actual = src.retainAll(of(1, 2));
         if (useIsEqualToInsteadOfIsSameAs()) {
-            assertThat(of(1, 2, 1, 2, 2).retainAll(of(1, 2))).isEqualTo(of(1, 2, 1, 2, 2));
+            assertThat(actual).isEqualTo(expected);
         } else {
-            final Traversable<Integer> src = of(1, 2, 1, 2, 2);
-            assertThat(src.retainAll(of(1, 2))).isSameAs(src);
+            assertThat(actual).isSameAs(src);
```
paplorinc: `isSameAs(src)` and not `isSameAs(expected)` since not all `Traversable`s are consumable multiple times

javaslang/src/main/java/javaslang/collection/HashSet.java
```diff
@@ -593,7 +596,13 @@ public T head() {
         if (isEmpty() || elements.isEmpty()) {
             return empty();
         } else {
-            return retainAll(elements);
+            int size = size();
```
paplorinc: this reduces the complexity of `intersect` to the size of the smaller set (used for `containsAll` also)

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: yep

javaslang/src/main/java/javaslang/collection/Seq.java
danieldietrich: This is a great idea! However, we decide not to have delegating default implementations with return type casts in our interfaces. Instead let's add the method(s) to our `javaslang.collection.Collections` class:

```java
static <C extends Traversable<T>, T> C removeAll(C collection, Iterable<? extends T> elements) {
    final Set<T> removed = HashSet.ofAll(elements);
    return (C) collection.filter(e -> !removed.contains(e));
}
```

(We don't need to check if the given `collection` is null because it is for internal use only and the calling collections will always provide `this`.)

Bonus points: We isolate the unsafe cast.

javaslang/src/main/java/javaslang/collection/Array.java
danieldietrich: good!

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -360,7 +358,13 @@ public CharSeq filter(Predicate<? super Character> predicate) {
...
             }
         }
-        return sb.length() == 0 ? EMPTY : sb.length() == length() ? this : of(sb.toString());
+        if (sb.length() == 0) {
+            return EMPTY;
+        } else if (sb.length() == length()) {
+            return this;
+        } else {
+            return of(sb);
+        }
```
danieldietrich: Sometimes I use a special indentation to reduce cyclomatic complexity by having only one return statement:

```java
return sb.length() == 0        ? EMPTY :
       sb.length() == length() ? this  : of(sb.toString());
```

But no formatter will handle this sort of whitespace :)

Using if/else if is ok!

javaslang/src/main/java/javaslang/collection/HashMap.java
```diff
@@ -331,7 +327,14 @@ public boolean isEmpty() {
...
         }
-        return result.size() == trie.size() ? this : wrap(result);
+
+        if (result.isEmpty()) {
+            return empty();
+        } else if (result.size() == trie.size()) {
+            return this;
+        } else {
+            return wrap(result);
+        }
```
danieldietrich: yes, good!

javaslang/src/main/java/javaslang/collection/List.java
```diff
@@ -749,8 +754,8 @@ default int indexOf(T element, int from) {
     @Override
     default List<T> intersperse(T element) {
         return isEmpty()
-                ? Nil.instance()
-                : foldRight(empty(), (x, xs) -> xs.isEmpty() ? xs.prepend(x) : xs.prepend(element).prepend(x));
+               ? Nil.instance()
+               : foldRight(empty(), (x, xs) -> xs.isEmpty() ? xs.prepend(x) : xs.prepend(element).prepend(x));
```
danieldietrich: I ask myself if we could omit the isEmpty check, like this:

```java
if (isEmpty()) {
    return this;
} else {
    final List<T> that = reverse();
    return that.tail().foldLeft(of(head()), (x, xs) -> xs.prepend(element).prepend(x));
}
```

javaslang/src/main/java/javaslang/collection/List.java
```diff
@@ -812,11 +817,15 @@ default int lastIndexOf(T element, int end) {
...
-            (predicate.test(t) ? left : right).add(t);
+            if (predicate.test(t)) {
+                left = left.append(t);
+            } else {
+                right = right.append(t);
+            }
         }
-        return Tuple.of(List.ofAll(left), List.ofAll(right));
+        return Tuple.of(left.toList(), right.toList());
     }
```
danieldietrich: We should save even more instances by directly prepending to Lists instead of appending to Queues:

```java
List<T> left = empty(), right = empty();
for (T t : this) {
    if (predicate.test(t)) {
        left = left.prepend(t);
    } else {
        right = right.prepend(t);
    }
}
return Tuple.of(left.reverse(), right.reverse());
```

javaslang/src/main/java/javaslang/collection/Traversable.java
danieldietrich: Nice!!

javaslang/src/main/java/javaslang/collection/List.java
paplorinc: it seems a lot more logical to me to append to queues and convert those to lists at the end - i.e. only moving forward, without reversing everything twice.

javaslang/src/main/java/javaslang/collection/List.java
paplorinc: how about
```Java
return ofAll(iterator().intersperse(element));
```

javaslang/src/main/java/javaslang/collection/CharSeq.java
paplorinc: you could turn the formatter off before and on after ... but ...

javaslang/src/main/java/javaslang/collection/Seq.java
paplorinc: Good point, pushing a change now

javaslang/src/main/java/javaslang/collection/List.java
danieldietrich: yes, nice solution!

javaslang/src/main/java/javaslang/collection/List.java
danieldietrich: Yep, Queue has two Lists inside and we only use the rear list (which finally is effectively reversed), the front List is wasted.

Best would be to use the code snippet of my last comment (and not java.util.ArrayList).

javaslang/src/main/java/javaslang/collection/Seq.java
paplorinc: pushed

danieldietrich:
Awesome, great work!

PR #1296, opened 2016-04-24
Deprecated `CharSeq`'s `split`, replacing it with `splitSeq`

paplorinc:
fixes https://github.com/javaslang/javaslang/issues/1252#issuecomment-213848494

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/CharSeq.java
+++ b/javaslang/src/main/java/javaslang/collection/CharSeq.java
@@ -13,6 +13,7 @@ import java.io.*;
 import java.nio.charset.Charset;
 import java.util.*;
 import java.util.function.*;
+import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collector;
 
@@ -130,7 +131,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      */
     public static CharSeq tabulate(int n, Function<? super Integer, ? extends Character> f) {
         Objects.requireNonNull(f, "f is null");
-        StringBuilder sb = new StringBuilder();
+        final StringBuilder sb = new StringBuilder();
         for (int i = 0; i < n; i++) {
             sb.append(f.apply(i));
         }
@@ -546,7 +547,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
         final StringBuilder left = new StringBuilder();
         final StringBuilder right = new StringBuilder();
         for (int i = 0; i < length(); i++) {
-            Character t = get(i);
+            final Character t = get(i);
             (predicate.test(t) ? left : right).append(t);
         }
         if (left.length() == 0) {
@@ -607,7 +608,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
         final StringBuilder sb = new StringBuilder();
         boolean found = false;
         for (int i = 0; i < length(); i++) {
-            char c = get(i);
+            final char c = get(i);
             if (!found && c == element) {
                 found = true;
             } else {
@@ -850,7 +851,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
         Objects.requireNonNull(predicate, "predicate is null");
         final StringBuilder sb = new StringBuilder();
         for (int i = 0; i < length(); i++) {
-            char c = get(i);
+            final char c = get(i);
             if (!predicate.test(c)) {
                 break;
             }
@@ -922,8 +923,8 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
     public <U> IndexedSeq<Tuple2<Character, U>> zip(Iterable<? extends U> that) {
         Objects.requireNonNull(that, "that is null");
         IndexedSeq<Tuple2<Character, U>> result = Vector.empty();
-        Iterator<Character> list1 = iterator();
-        java.util.Iterator<? extends U> list2 = that.iterator();
+        final Iterator<Character> list1 = iterator();
+        final java.util.Iterator<? extends U> list2 = that.iterator();
         while (list1.hasNext() && list2.hasNext()) {
             result = result.append(Tuple.of(list1.next(), list2.next()));
         }
@@ -934,8 +935,8 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
     public <U> IndexedSeq<Tuple2<Character, U>> zipAll(Iterable<? extends U> that, Character thisElem, U thatElem) {
         Objects.requireNonNull(that, "that is null");
         IndexedSeq<Tuple2<Character, U>> result = Vector.empty();
-        Iterator<Character> list1 = iterator();
-        java.util.Iterator<? extends U> list2 = that.iterator();
+        final Iterator<Character> list1 = iterator();
+        final java.util.Iterator<? extends U> list2 = that.iterator();
         while (list1.hasNext() || list2.hasNext()) {
             final Character elem1 = list1.hasNext() ? list1.next() : thisElem;
             final U elem2 = list2.hasNext() ? list2.next() : thatElem;
@@ -987,7 +988,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
         }
         final StringBuilder left = new StringBuilder();
         for (int i = 0; i < length(); i++) {
-            Character t = get(i);
+            final Character t = get(i);
             if (!predicate.test(t)) {
                 left.append(t);
             } else {
@@ -1005,7 +1006,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
         }
         final StringBuilder left = new StringBuilder();
         for (int i = 0; i < length(); i++) {
-            Character t = get(i);
+            final Character t = get(i);
             left.append(t);
             if (predicate.test(t)) {
                 break;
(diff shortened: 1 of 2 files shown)


javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -130,7 +130,7 @@ public static CharSeq ofAll(Iterable<? extends Character> elements) {
      */
     public static CharSeq tabulate(int n, Function<? super Integer, ? extends Character> f) {
         Objects.requireNonNull(f, "f is null");
-        StringBuilder sb = new StringBuilder();
+        final StringBuilder sb = new StringBuilder();
```
paplorinc: unrelated, but requested last time by @danieldietrich :)

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -2161,8 +2161,12 @@ public CharSeq replace(CharSequence target, CharSequence replacement) {
      * @throws PatternSyntaxException if the regular expression's syntax is invalid
      * @see java.util.regex.Pattern
      */
-    public Seq<CharSeq> split(String regex, int limit) {
-        Array<String> split = Array.wrap(back.split(regex, limit));
+    @Deprecated(/* Use splitSeq instead, will be removed in 3.0.0 */)
```
paplorinc: this is the actual fix, replaced the impl with the new one (added back old tests also)

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -1331,10 +1332,10 @@ public boolean contentEquals(CharSequence cs) {
      * <li> The two characters are the same (as compared by the
      * {@code ==} operator)
      * <li> Applying the method {@link
-     * java.lang.Character#toUpperCase(char)} to each character
+     * Character#toUpperCase(char)} to each character
```
paplorinc: unrelated also: organized imports for JavaDoc also

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -2159,10 +2160,14 @@ public CharSeq replace(CharSequence target, CharSequence replacement) {
      * @return the Seq of strings computed by splitting this string
      * around matches of the given regular expression
      * @throws PatternSyntaxException if the regular expression's syntax is invalid
-     * @see java.util.regex.Pattern
+     * @see Pattern
      */
-    public Seq<CharSeq> split(String regex, int limit) {
-        Array<String> split = Array.wrap(back.split(regex, limit));
+    @Deprecated(/* Use splitSeq instead, will be removed in 3.0.0 */)
```
paplorinc: this is the essence - used new function for old impl also

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -2193,16 +2198,20 @@ public CharSeq replace(CharSequence target, CharSequence replacement) {
...
      * around matches of the given regular expression
      * @throws PatternSyntaxException if the regular expression's syntax is invalid
-     * @see java.util.regex.Pattern
+     * @see Pattern
      */
-    public Seq<CharSeq> split(String regex) {
-        return split(regex, 0);
+    @Deprecated(/* Use splitSeq instead, will be removed in 3.0.0 */)
+    public CharSeq[] split(String regex) {
+        return splitSeq(regex, 0).toJavaArray(CharSeq.class);
```
paplorinc: this is the essence - used new function for old impl also

javaslang/src/test/java/javaslang/collection/StringTest.java
```diff
@@ -216,12 +216,16 @@ public void replace() {
 
     @Test
     public void split() {
-        assertThat(CharSeq.of("123").split("2")).isEqualTo(Array.of(CharSeq.of("1"), CharSeq.of("3")));
+        assertThat(CharSeq.of("123").split("2")).isEqualTo(new CharSeq[] { CharSeq.of("1"), CharSeq.of("3") }); // TODO deprecated
```
paplorinc: added tests back for the deprecated methods too

javaslang/src/test/java/javaslang/collection/StringTest.java
```diff
@@ -216,12 +216,18 @@ public void replace() {
 
     @Test
     public void split() {
-        assertThat(CharSeq.of("123").split("2")).isEqualTo(Array.of(CharSeq.of("1"), CharSeq.of("3")));
+        //noinspection deprecation
+        assertThat(CharSeq.of("123").split("2")).isEqualTo(new CharSeq[] { CharSeq.of("1"), CharSeq.of("3") }); // TODO deprecated
```
paplorinc: tests for old methods, with disabled deprecation warning (the build won't pass otherwise)

javaslang/src/test/java/javaslang/collection/StringTest.java
```diff
@@ -214,14 +214,20 @@ public void replace() {
         assertThat((Iterable<Character>) CharSeq.of("123").replace("1", "#")).isEqualTo(CharSeq.of("#23"));
     }
 
+    @SuppressWarnings("deprecation")
```
paplorinc: wouldn't compile otherwise ...

javaslang/src/test/java/javaslang/collection/StringTest.java
```diff
@@ -214,14 +214,20 @@ public void replace() {
...
     @Test
     public void split() {
-        assertThat(CharSeq.of("123").split("2")).isEqualTo(Array.of(CharSeq.of("1"), CharSeq.of("3")));
+        assertThat(CharSeq.of("123").split("2")).isEqualTo(new CharSeq[] { CharSeq.of("1"), CharSeq.of("3") }); // TODO deprecated
```
paplorinc: tests for old behavior also

danieldietrich:
Let's just adjust the return type in a `3.0.0` branch as described [here](https://github.com/javaslang/javaslang/issues/1252#issuecomment-213935402). That is a major design decision for the future development of Javaslang.

paplorinc:
The `3.0` branch seems to be up to date already (i.e. a single `split`, returning a a `Seq`), only the `2.1` should be downgraded -- done in this commit.

Opinions @danieldietrich, @sajit, @ruslansennov, @jest?

danieldietrich:
@paplorinc deprecating the one and adding the other is ok for me. It is a good hint that it will change. 3.0 will clear it up. If no-one has another opinion I will merge (when I return to the keyboard in two hours or so)

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -2159,10 +2160,14 @@ public CharSeq replace(CharSequence target, CharSequence replacement) {
...
+     * @see Pattern
      */
-    public Seq<CharSeq> split(String regex, int limit) {
-        Array<String> split = Array.wrap(back.split(regex, limit));
+    @Deprecated(/* Use splitSeq instead, will be removed in 3.0.0 */)
+    public CharSeq[] split(String regex, int limit) {
+        return splitSeq(regex, limit).toJavaArray(CharSeq.class);
+    }
+    public Seq<CharSeq> splitSeq(String regex, int limit) {
+        final Seq<String> split = Array.wrap(back.split(regex, limit));
```
paplorinc: should we worry that the regex is recompiled every time?
@danieldietrich

danieldietrich: No, we do not need to worry, this is how Java 8's impl of `java.lang.String.split` works:

```java
    public String[] split(String regex, int limit) {
        /* fastpath if the regex is a
         (1)one-char String and this character is not one of the
            RegEx's meta characters ".$|()[{^?*+\\", or
         (2)two-char String and the first char is the backslash and
            the second is not the ascii digit or ascii letter.
         */
        char ch = 0;
        if (((regex.value.length == 1 &&
             ".$|()[{^?*+\\".indexOf(ch = regex.charAt(0)) == -1) ||
             (regex.length() == 2 &&
              regex.charAt(0) == '\\' &&
              (((ch = regex.charAt(1))-'0')|('9'-ch)) < 0 &&
              ((ch-'a')|('z'-ch)) < 0 &&
              ((ch-'A')|('Z'-ch)) < 0)) &&
            (ch < Character.MIN_HIGH_SURROGATE ||
             ch > Character.MAX_LOW_SURROGATE))
        {
            int off = 0;
            int next = 0;
            boolean limited = limit > 0;
            ArrayList<String> list = new ArrayList<>();
            while ((next = indexOf(ch, off)) != -1) {
                if (!limited || list.size() < limit - 1) {
                    list.add(substring(off, next));
                    off = next + 1;
                } else {    // last one
                    //assert (list.size() == limit - 1);
                    list.add(substring(off, value.length));
                    off = value.length;
                    break;
                }
            }
            // If no match was found, return this
            if (off == 0)
                return new String[]{this};

            // Add remaining segment
            if (!limited || list.size() < limit)
                list.add(substring(off, value.length));

            // Construct result
            int resultSize = list.size();
            if (limit == 0) {
                while (resultSize > 0 && list.get(resultSize - 1).length() == 0) {
                    resultSize--;
                }
            }
            String[] result = new String[resultSize];
            return list.subList(0, resultSize).toArray(result);
        }
        return Pattern.compile(regex).split(this, limit);
    }
```

(example taken from Java's original source code)

paplorinc: yeah, I checked it also, but `Pattern.compile(regex)` doesn't cache, AFAIK

danieldietrich: Yep, it doesn't cache. I would do the same and not cache (for now).

sajit:
Sorry about the unresponsiveness.. Had some family emergency.

danieldietrich:
Thx for reply

PR #1299, opened 2016-04-26
Added default helper methods to `SerializableComparator`

paplorinc:


danieldietrich:
Hi Pap, where do we use these methods? Does another PR follow? Generally we should only add internal code if we use it - otherwise it is dead code.

paplorinc:
Yes, it's used in the other PR.

PR #1302, opened 2016-04-27
Random fixes

paplorinc:


javaslang/src/main/java/javaslang/collection/Comparators.java
```diff
@@ -47,7 +47,11 @@ private Comparators() {
         long serialVersionUID = 1L;
 
         static <T> SerializableComparator<T> of(Comparator<T> comparator) {
-            return comparator::compare;
+            if (comparator instanceof SerializableComparator) {
```
paplorinc: this caused the stack overflow and 2300x slowdown in my other PR

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -467,52 +467,16 @@ public Short getNext() {
...
-                        if (curr == prev) {
-                            hasNext = false;
-                        }
-                    }
-                    return next;
-                }
-            };
-        }
+        final double toInclusive = (step > 0) ? Math.nextDown(toExclusive) : Math.nextUp(toExclusive);
+        return rangeClosedBy(from, toInclusive, step);
```
paplorinc: make it inclusive by incrementing it with a `ULP`

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -556,8 +520,8 @@ public Double getNext() {
      * @throws IllegalArgumentException if {@code step} is zero
      */
     static Iterator<Integer> rangeBy(int from, int toExclusive, int step) {
-        final int one = step > 0 ? 1 : -1;
-        return Iterator.rangeClosedBy(from, toExclusive - one, step);
+        final int toInclusive = toExclusive - (step > 0 ? 1 : -1);
```
paplorinc: unified the rest to match the `double` one

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -658,36 +622,45 @@ public Double getNext() {
         } else if (step == 0) {
             throw new IllegalArgumentException("step cannot be 0");
         } else if (from == toInclusive) {
-            return Iterator.of(from);
-        } else if (step * (from - toInclusive) > 0) {
-            return Iterator.empty();
+            return of(from);
+        } else if (Math.signum(step) == Math.signum(from - toInclusive)) {
```
paplorinc: it's simpler this way ...

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -658,36 +622,45 @@ public Double getNext() {
...
+            return empty();
         } else {
-            return new AbstractIterator<Double>() {
-
-                double prev = Double.NaN;
-                double curr = from;
-                boolean hasNext = true;
+            if (step > 0) {
+                return new AbstractIterator<Double>() {
```
paplorinc: Instead of checking for constant conditions inside the loop, let's make two versions, and ascending and a descending -- simplified implementation significntly

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1426,11 +1416,6 @@ default T head() {
     }
 
     @Override
-    default Option<T> headOption() {
```
paplorinc: same as `super`

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1569,11 +1553,11 @@ public boolean hasNext() {
                 @Override
                 public T getNext() {
                     final T elem = that.next();
-                    if (done || !Objects.equals(currentElement, elem)) {
-                        return elem;
-                    } else {
-                        done = true;
+                    if (isFirst && Objects.equals(currentElement, elem)) {
```
paplorinc: inverted to match similar method below

javaslang/src/main/java/javaslang/collection/List.java
```diff
@@ -535,23 +535,23 @@
...
     default List<T> append(T element) {
-        return foldRight(List.of(element), (x, xs) -> xs.prepend(x));
+        return foldRight(of(element), (x, xs) -> xs.prepend(x));
     }
 
     @Override
     default List<T> appendAll(Iterable<? extends T> elements) {
         Objects.requireNonNull(elements, "elements is null");
-        return foldRight(List.ofAll(elements), (x, xs) -> xs.prepend(x));
+        return ofAll(elements).prependAll(this);
```
paplorinc: same complexity, but easier to understand

javaslang/src/main/java/javaslang/collection/List.java
```diff
@@ -1056,7 +1056,7 @@ default T peek() {
 
     @Override
     default List<T> reverse() {
-        return isEmpty() ? this : foldLeft(empty(), List::prepend);
+        return (length() <= 1) ? this : foldLeft(empty(), List::prepend);
```
paplorinc: a one element list is the same as its reverse also

javaslang/src/test/java/javaslang/collection/AbstractTraversableRangeTest.java
```diff
@@ -547,29 +547,4 @@ public void shouldProhibitDoubleRangeByFromEqualNaN() {
     public void shouldProhibitDoubleRangeByStepEqualNaN() {
         rangeBy(0.0, 10.0, Double.NaN);
     }
-
```
paplorinc: Moved to `Iterator`, because now iteration is possible for elements where the `ulp > step` (i.e. when `i == i+1`)

javaslang/src/test/java/javaslang/collection/IteratorTest.java
```diff
@@ -405,6 +401,30 @@ public void shouldNonNilGroupByEqual() {
         assertThat(actual).isEqualTo(expected);
     }
 
+    @Test
+    public void shouldCreateDoubleRangeByFromInfinity() {
```
paplorinc: semantics have slightly changed, please validate them agains the old tests

javaslang/src/test/java/javaslang/collection/IteratorTest.java
```diff
@@ -405,6 +401,30 @@ public void shouldNonNilGroupByEqual() {
...
+
+    @Test
+    public void shouldCreateDoubleRangeByFromMaxToInfinity() {
+        assertThat(rangeBy(Double.MAX_VALUE, Double.POSITIVE_INFINITY, 3E307)).isEqualTo(of(Double.MAX_VALUE));
+        assertThat(rangeBy(-Double.MAX_VALUE, Double.NEGATIVE_INFINITY, -3E307)).isEqualTo(of(-Double.MAX_VALUE));
+    }
+
+    @Test
+    public void shouldCreateDoubleRangeClosedByFromMaxToInfinity() {
+        assertThat(rangeClosedBy(Double.MAX_VALUE, Double.POSITIVE_INFINITY, 3E307)).isEqualTo(of(Double.MAX_VALUE));
```
paplorinc: no closed infinite range

javaslang/src/test/java/javaslang/collection/AbstractTraversableRangeTest.java
ruslansennov: d578cdfeb7dc93ace77021e1a0034ae2

paplorinc: Try this in your `groovy`/`java` console:
```Groovy
double i = 10000000000000000
println i == i + 1                 // true
println Math.nextUp(i) - i  // 2.0
```

javaslang/src/main/java/javaslang/collection/Comparators.java
danieldietrich: This protection is necessary. But I ask myself why it has been called with a SerializableComparator as argument. Once a plain Comparator is changed to be a SerializableComparator it should not be necessary to call SerializableComparator.of again. At least I didn't have to when implementing the RedBlackTree...

paplorinc: I needed this the existing `comparator` to create new instances of the `PriorityQueue`, but wanted it to be callable for normal `Comparator`s also, so the constructor accepted the Java one and converted in inside.

danieldietrich: Ok. We hide the constructors by making them private or package private (if needed).

paplorinc: they are package private:
```Java
    PriorityQueueBase(Comparator<T> comparator, List<Node<Ranked<T>>> forest) {
        this.comparator = SerializableComparator.of(comparator);
        this.forest = forest;
    }
```

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -658,36 +622,45 @@ public Double getNext() {
...
+                    }
 
-                @Override
-                public boolean hasNext() {
-                    return hasNext;
-                }
+                    @Override
+                    public Double getNext() {
+                        final double next = i;
+                        i = Math.max(i + step, Math.nextUp(i));
```
danieldietrich: Doesn't break this semantics? E.g. for `range(.0d, Double.MAX_VALUE, Double.MIN_VALUE)`. The steps will be greater and greater... But we expect constant steps.

How does Scala behave here?

paplorinc: Scala seems to use BigDecimal under the hood - the problem is with the CPU floating point representation, I think we should try to make it as intuitive as possible.

@danieldietrich, @ruslansennov: What would a user find more intuitive:
* a loop that stops before the end because consecutive numbers have a greater distance than the step size
* if it stepped bigger in the last few steps
* if it's slower (uses `BigDecimal`) and has a few consecutive values that are equal?

danieldietrich: Java's IntStream has range and rangeClosed, Java's DoubleStream does not have them. Maybe this is the reason.

I have not a good answer but when in doubt we should always align to Scala. We have users that create issues if there is a different behavior than in Scala. So this is a good chance to align things.

danieldietrich: We should create a new issue because this change is important and more than a random change :)

paplorinc: Maybe, but these things will only manifest for extremely big and small numbers together.

javaslang/src/test/java/javaslang/collection/IteratorTest.java
danieldietrich: I like that you shortened/simplified the range code. But I'm not sure about the changes regarding included and excluded INFINITY because of ULP. Also here I would take a look how Scala behaves.

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: Maybe, but it is a problem when using floating points. We need an appropriate solution. At $work team members implemented a calculation core for financial products and there were error cases because of the step problem.

javaslang/src/test/java/javaslang/collection/IteratorTest.java
paplorinc: I think that people shouldn't use `Double` in these extreme ranges, if they don't know how it works.
Floating point arithmetic is a bitch, there are no good calls here in my opinion, just compromises.

danieldietrich:
(Off topic: Again it is time to sleep for me - I will go on tomorrow. Today I started reading about PriorityQueues, next is the paper you referenced...)

javaslang/src/main/java/javaslang/collection/Iterator.java
paplorinc: `Double` should never be used for money :)

danieldietrich: When you predict the price of your assets for the next 30 years it is ok to have an error. Nobody knows the future :)

When you calculate those prices for your whole portfolio, let's say for hundreds of thousands of trades, and you have to calculate scenarios and to shift values for hundreds of days back in time, then the calculation has to be *really* fast. We have 90 nodes in our calculation grid and the net time is 250 hours calculation when calculating the risk for example!

paplorinc: I know, but isn't an `int` / `long` (with a fixed or stored scale) faster for that?

danieldietrich: I've also seen such impls. But the calc cores I've seen so far use double. Some implemented in Java, some in C/C++.

paplorinc: https://books.google.ro/books?id=ka2VUBqHiWkC&pg=PA219#v=onepage&q&f=false

danieldietrich: thx for the link

paplorinc:
@danieldietrich, @ruslansennov, separated the controversial `Iterator` modifications to a separate PR.

javaslang/src/main/java/javaslang/collection/Iterator.java
paplorinc: https://github.com/javaslang/javaslang/issues/1309

danieldietrich:
Thank you Pap, looks great.

paplorinc:
@danieldietrich, Pap is my family name, feel free to call me Lőrinc :)

danieldietrich:
Haha, will do :-) Thanks for the hint, Lőrinc!

PR #1303, opened 2016-04-27
New static method Stream.ofString

mind-blowing:
Right now for creating Stream object from String object we need to use something like this:
```
Stream.ofAll(s.toCharArray());
```
But it may be a little bit easier with one new static method like this:
```
Stream.ofString(s);
```

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Stream.java
+++ b/javaslang/src/main/java/javaslang/collection/Stream.java
@@ -351,6 +351,17 @@ public interface Stream<T> extends Kind1<Stream<?>, T>, LinearSeq<T> {
     }
 
     /**
+     * Creates a Stream based on the string characters.
+     *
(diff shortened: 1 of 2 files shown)


danieldietrich:
Hi Masich, many thanks for your PR!

I think Stream is not the right place to add such a special method. We already have `CharSeq` (similar to Scala's `String`) for exactly that use-case!

Example:

```java
CharSeq.of("Hello");
```

Additionally `CharSeq` has much more functionality tailored to char sequences than `Stream`.

Therefore I will close this PR.

mind-blowing:
Hi Daniel,

Thanks for suggestion, didn't know about `CharSeq.of("Hello");`, next time I will ask in Gitter before creating PR :smile:.

danieldietrich:
Looking forward to your next PR! 😄

Issue #1306, opened 2016-04-28
Add `sorted` and `sortBy` to `Set`


paplorinc:
Similarly to `List`'s methods:
```java
    @Override
    default List<T> sorted() {
        return isEmpty() ? this : toJavaStream().sorted().collect(List.collector());
    }

    @Override
    default List<T> sorted(Comparator<? super T> comparator) {
        Objects.requireNonNull(comparator, "comparator is null");
        return isEmpty() ? this : toJavaStream().sorted(comparator).collect(List.collector());
    }

    @Override
    default <U extends Comparable<? super U>> List<T> sortBy(Function<? super T, ? extends U> mapper) {
        return sortBy(U::compareTo, mapper);
    }

    @Override
    default <U> List<T> sortBy(Comparator<? super U> comparator, Function<? super T, ? extends U> mapper) {
        final Function<? super T, ? extends U> domain = Function1.of(mapper::apply).memoized();
        return toJavaStream()
                .sorted((e1, e2) -> comparator.compare(domain.apply(e1), domain.apply(e2)))
                .collect(collector());
    }
```

The `Set` interface should also provide these.
I think all of them can simply return a new `TreeSet`, like:
```java
    default SortedSet<T> sorted() {
        return sorted(naturalComparator());
    }

    default SortedSet<T> sorted(Comparator<? super T> comparator) {
        Objects.requireNonNull(comparator, "comparator is null");
        return TreeSet.ofAll(comparator, this);
    }

    default <U extends Comparable<? super U>> SortedSet<U> sortBy(Function<? super T, ? extends U> mapper) {
        return sortBy(U::compareTo, mapper);
    }

    default <U> SortedSet<U> sortBy(Comparator<? super U> comparator, Function<? super T, ? extends U> mapper) {
        return TreeSet.ofAll(comparator, this.map(mapper));
    }
```

ruslansennov:
> I think all of them can simply return a new TreeSet

I don't like this idea.
I believe we should return the same type of Set like origin.
In case if someone want to have TreeSet he can do TreeSet.ofAll(...)

paplorinc:
the idea of providing a `sorted` method, like in `List`?

ruslansennov:
> providing a sorted method, like in List

why not :smile:

paplorinc:
Not sure I follow :/
You don't agree with the issue, or the suggested method signature or the internal impl?

ruslansennov:
Sorry, I thought for a moment
[...]

PR #1310, opened 2016-04-28
Added `toSortedSet` to `Value`

paplorinc:
Fixes https://github.com/javaslang/javaslang/issues/1306


danieldietrich:
Great, thank you!

Offtopic: I ask myself why we are not checking if the current Value is of the type we convert. E.g. `Set.toSet()` should return `this`. But this is should be done in another issue...

Issue #1316, opened 2016-05-03
Add `isNotEmpty` to `Traversable`


paplorinc:
Similarly to how `Option` has
```java
default boolean isDefined() {
    return !isEmpty();
}
```
we may add the opposite of `isEmpty`, as `Java`'s `!` operator is hard to see (especially before an `i`)
```java
/**
 * Checks if this Traversable is not empty.
 *
 * @return whether this Traversable contains any elements.
 */
default boolean isNotEmpty() {
    return !isEmpty();
}
```

danieldietrich:
All `Value`s have `isDefined()`. `isNotEmpy()` would be an alias, similar to `length()` and `size()`. Just saying, but it makes sense for collections to have that name.

valery1707:
@danieldietrich, `Traversable` already have `nonEmpty` that have same behavior as discussed `isNotEmpy()`.
Is this issue need to be closed?

paplorinc:
Unfortunatly @danieldietrich thinks this complicates the API: https://github.com/javaslang/javaslang/pull/1320/files#r62437967

valery1707:
@paplorinc, method `Traversable#nonEmpty` was introduced in `2016-02-16` by @danieldietrich.
This discussion took place at `2016-05-08`, when method already exists.

If this issue stil actual - I don't understand what need to be done.

danieldietrich:
@valery1707 you are completely right. This issue can be closed.

Value has isEmpty and isDefined. The latter is meant to reflect the non-empty state, especially for single-valued types.

Traversable inherits isEmpty and isDefined and adds nonEmpty for multi-valued types.

These method names are common in Scala.

@paplorinc I have to admit that I forgot to mention that we have nonEmpty already. Sorry! The point in the comment you are referring is, that we do not need additional API (with non-empty semantics) within the internal, non-Traversable collections like HashArrayMappedTrie or RedBlackTree.

Thank you for your engagement and keep up the great work!

PR #1320, opened 2016-05-07
Added `isNotEmpty` to collections

paplorinc:
Fixes https://github.com/javaslang/javaslang/issues/1316

### DIFF
--- a/javaslang-benchmark/src/test/java/javaslang/benchmark/BenchmarkResultAggregator.java
+++ b/javaslang-benchmark/src/test/java/javaslang/benchmark/BenchmarkResultAggregator.java
@@ -18,7 +18,7 @@ public class BenchmarkResultAggregator {
             for (Tuple2<String, Array<RunResult>> group : getGroupsBasedOnBenchmarkPrefix(runResults)) {
                 final Multimap<Integer, Tuple2<String, Double>> results = aggregate(group._2);
                 final Multimap<String, Double> ratios = getRatios(results);
-                if (!ratios.isEmpty()) {
+                if (ratios.isNotEmpty()) {
                     System.out.println("Group '" + group._1 + "':");
 
                     final Seq<String> output = getFormattedRatios(ratios);
@@ -29,7 +29,7 @@ public class BenchmarkResultAggregator {
     }
 
     private static void printHeader(Array<RunResult> list) {
-        if (!list.isEmpty()) {
+        if (list.isNotEmpty()) {
             final Array<Integer> containerSizes = list.map(r -> getContainerSize(r.getAggregatedResult())).distinct().sorted();
             final String header = formatList(containerSizes);
             System.out.println("\nRatios for: " + header);
--- a/javaslang/src/main/java/javaslang/Value.java
+++ b/javaslang/src/main/java/javaslang/Value.java
@@ -328,9 +328,9 @@ public interface Value<T> extends Iterable<T> {
     }
 
     /**
-     * Checks, this {@code Value} is empty, i.e. if the underlying value is absent.
+     * Checks whether this {@code Value} is empty, i.e. if the underlying value is absent.
      *
-     * @return false, if no underlying value is present, true otherwise.
+     * @return true, if no underlying value is present, false otherwise.
      */
     boolean isEmpty();
 
--- a/javaslang/src/main/java/javaslang/Value.java
+++ b/javaslang/src/main/java/javaslang/Value.java
@@ -43,6 +43,7 @@ import java.util.stream.StreamSupport;
  * <li>{@link #getOrElse(Supplier)}</li>
  * <li>{@link #getOrElseThrow(Supplier)}</li>
  * <li>{@link #isEmpty()}</li>
+ * <li>{@link #isNotEmpty()}</li>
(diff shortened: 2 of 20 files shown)


javaslang/src/main/java/javaslang/Value.java
```diff
@@ -328,9 +328,9 @@ default T getOrElseTry(Try.CheckedSupplier<? extends T> supplier) {
     }
 
     /**
-     * Checks, this {@code Value} is empty, i.e. if the underlying value is absent.
+     * Checks whether this {@code Value} is empty, i.e. if the underlying value is absent.
      *
-     * @return false, if no underlying value is present, true otherwise.
+     * @return true, if no underlying value is present, false otherwise.
```
paplorinc: I think it was reversed

javaslang/src/main/java/javaslang/collection/Stack.java
```diff
@@ -26,14 +26,14 @@
...
- *
+ * <p>
  * See Okasaki, Chris: <em>Purely Functional Data Structures</em> (p. 7 ff.). Cambridge, 2003.
  *
  * @param <T> component type
  * @author Daniel Dietrich
  * @since 2.0.0
  */
-public interface Stack<T> {
+public interface Stack<T> extends Traversable<T> {
```
paplorinc: @danieldietrich, is this ok? We need to inherit the default `isNotNull`...

javaslang-benchmark/src/test/java/javaslang/benchmark/BenchmarkResultAggregator.java
```diff
@@ -18,7 +18,7 @@ public static void displayRatios(Collection<RunResult> runResultsCollection) {
             for (Tuple2<String, Array<RunResult>> group : getGroupsBasedOnBenchmarkPrefix(runResults)) {
                 final Multimap<Integer, Tuple2<String, Double>> results = aggregate(group._2);
                 final Multimap<String, Double> ratios = getRatios(results);
-                if (!ratios.isEmpty()) {
+                if (ratios.isNotEmpty()) {
```
paplorinc: changed all the code references to take advantage of the new API method

javaslang/src/main/java/javaslang/collection/RedBlackTree.java
```diff
@@ -496,10 +505,10 @@ private boolean isLeaf() {
         private static <T> Node<T> balanceLeft(Color color, int blackHeight, RedBlackTree<T> left, T value,
                                                RedBlackTree<T> right, Empty<T> empty) {
             if (color == BLACK) {
-                if (!left.isEmpty()) {
+                if (left.isNotEmpty()) {
                     final Node<T> ln = (Node<T>) left;
                     if (ln.color == RED) {
-                        if (!ln.left.isEmpty()) {
+                        if (ln.left.isNotEmpty()) {
```
paplorinc: nice example of why I recommended introducing this method

javaslang/src/main/java/javaslang/Value.java
danieldietrich: yep

javaslang/src/main/java/javaslang/collection/Stack.java
danieldietrich: We should not extend Traversable. If we need the method, it has to be duplicated (without inheritance).

There was a long discussion some time ago whether we should remove the Stack interface or not. We decided to keep it as rudimentary standalone interface. It is not a collection but a contract for Stack-like collections. List is currently the only implementation.

javaslang/src/main/java/javaslang/collection/List.java
```diff
@@ -941,7 +941,7 @@ default T peek() {
         final Deque<T> preceding = new ArrayDeque<>(size());
         List<T> result = this;
         boolean found = false;
-        while (!found && !result.isEmpty()) {
+        while (!found && result.isNotEmpty()) {
```
danieldietrich: Generally I'm used to see `!someVar` or `!someMethodCall` and I would not have changed all occurreces of `!isEmpty` to `isNotEmpty`. `!` is part of the Java language and it is used everywhere. We can't live without it.

This line is a good example. We still have `!found`. We could rename it `notFound` and invert the content but that seems too complicated. We had to rewrite well-known algorithms to fit the new notation. Also it is not possible to eliminate `!` in algorithms that use both `someVar` and `!someVar`. So it is the best to accept `!`.

For the health of Javaslang it is better not to apply highly opinionated changes all over the code repository (a change may invent errors). It is like formatting. Some like 2 spaces indentation, some like 4 spaces or even tabs. Some prefer to use `!someVar`, some not.

Adding the additional method to `HashArrayMappedTrie` (HAMT) makes not much sense. We add code to an internal class just for a flavor of readability.

--

With this change we should add the method `isNotEmpty` to `Value` in order to offer the users of the _public_ API different ways to express themselves. But we should not change the whole code-base.

danieldietrich: @paplorinc sorry for having ruined this PR! You are very fast with refactorings in the broad - it is hard for me to follow and spend enough time to think about changes. Keep up the awesome work! You already have made significant changes to Javaslang!

PR #1321, opened 2016-05-07
Added benchmarks for `List` and `Array`

paplorinc:
Contributes to https://github.com/javaslang/javaslang/issues/725 and resolves https://github.com/javaslang/javaslang/issues/1322.

Included `PCollections` and `Functional Java` also

The results are interesting for adding and iterating elements:
```Groovy
Ratios for: [10, 1000, 100000]

Group 'AddAll':
slang_persistent/fjava_persistent      : [1.37, 1.87, 2.14]
slang_persistent/java_mutable          : [1.08, 1.07, 0.81]
slang_persistent/pcollectins_persistent: [0.79, 0.88, 1.14]
slang_persistent/scala_mutable         : [1.02, 1.10, 1.40]
slang_persistent/scala_persistent      : [0.80, 1.05, 1.48]

Group 'Iterate':
slang_persistent/fjava_persistent      : [1.14, 1.01, 1.62]
slang_persistent/java_mutable          : [0.76, 0.39, 0.68]
slang_persistent/pcollectins_persistent: [0.81, 1.01, 1.62]
slang_persistent/scala_mutable         : [1.11, 1.07, 1.57]
slang_persistent/scala_persistent      : [1.08, 1.00, 1.26]
```

* `fjava` seems to be always slower
* Mutable and persistent Scala seem to be slower in most cases
* `pcollection` seems to be faster for a few elements (maybe they're delegating everything to an array for small sizes, will investigate)
* `ArrayList` is always faster for iteration, but often slower for insertion

### DIFF
--- a/javaslang-benchmark/pom.xml
+++ b/javaslang-benchmark/pom.xml
@@ -33,6 +33,21 @@
             <version>7.3.0-M1</version>
         </dependency>
         <dependency>
+            <groupId>org.pcollections</groupId>
+            <artifactId>pcollections</artifactId>
+            <version>2.1.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.functionaljava</groupId>
+            <artifactId>functionaljava</artifactId>
+            <version>4.5</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>15.0</version>
+        </dependency>
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <scope>test</scope>
--- a/javaslang-benchmark/src/test/java/javaslang/benchmark/ArrayBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/benchmark/ArrayBenchmark.java
@@ -1,0 +1,129 @@
+package javaslang.benchmark;
+
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.results.RunResult;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.options.*;
+
+import java.util.*;
+import java.util.concurrent.TimeUnit;
+
+import static javaslang.benchmark.BenchmarkResultAggregator.displayRatios;
+
+public class ArrayBenchmark {
+    public static void main(String... args) throws Exception { /* main is more reliable than a test */
+        final Options opts = new OptionsBuilder()
+                .include(ArrayBenchmark.class.getSimpleName())
+                .shouldDoGC(false)
+                .shouldFailOnError(true)
+                .build();
+
+        final Collection<RunResult> results = new Runner(opts).run();
+        displayRatios(results, "^slang.+$");
+    }
+
+    @State(Scope.Benchmark)
+    @BenchmarkMode(Mode.Throughput)
+    @OutputTimeUnit(TimeUnit.SECONDS)
+    @Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+    @Measurement(iterations = 2, time = 1, timeUnit = TimeUnit.SECONDS)
+    @Fork(value = 1, jvmArgsAppend = { "-XX:+UseG1GC", "-Xss100m", "-Xms1g", "-Xmx1g", "-disableassertions" }) /* set fork to 0 if you want to debug */
+    public static class Base {
+        @Param({ "10", "100", "1000" })
+        public int CONTAINER_SIZE;
+
+        public Integer[] ELEMENTS;
+
+        @Setup
+        public void setup() {
+            final Random random = new Random(0);
+
+            ELEMENTS = new Integer[CONTAINER_SIZE];
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                final int value = random.nextInt(CONTAINER_SIZE) - (CONTAINER_SIZE / 2);
+                ELEMENTS[i] = value;
+            }
+        }
+
+        protected static <T> void assertEquals(T a, T b) {
+            if (!Objects.equals(a, b)) {
+                throw new IllegalStateException(a + " != " + b);
+            }
+        }
+    }
+
+    public static class AddAll extends Base {
+        @Benchmark
+        @SuppressWarnings("ManualArrayCopy")
+        public void java_mutable() {
+            final Integer[] values = new Integer[ELEMENTS.length];
+            for (int i = 0; i < ELEMENTS.length; i++) {
+                values[i] = ELEMENTS[i];
+            }
+            assertEquals(values.length, CONTAINER_SIZE);
+        }
+
+        @Benchmark
+        public void slang_persistent() {
+            javaslang.collection.Array<Integer> values = javaslang.collection.Array.empty();
+            for (Integer element : ELEMENTS) {
+                values = values.append(element);
+            }
+            assertEquals(values.size(), CONTAINER_SIZE);
+        }
+    }
+
+    public static class Iterate extends Base {
+        @State(Scope.Thread)
+        public static class Initialized {
+            Integer[] javaMutable;
+
+            int expectedAggregate = 0;
+            javaslang.collection.Array<Integer> slangPersistent = javaslang.collection.Array.empty();
+
+            @Setup(Level.Invocation)
+            public void initializeMutable(Base state) {
+                assertEquals(javaMutable, null);
+                javaMutable = state.ELEMENTS.clone();
+                assertEquals(javaMutable.length, state.CONTAINER_SIZE);
+
+                if (expectedAggregate == 0) {
+                    for (Integer element : state.ELEMENTS) {
+                        expectedAggregate ^= element;
+                    }
+
+                    assertEquals(slangPersistent.size(), 0);
+                    for (Integer element : state.ELEMENTS) {
+                        slangPersistent = slangPersistent.prepend(element);
+                    }
+                    assertEquals(slangPersistent.size(), state.CONTAINER_SIZE);
+                }
+            }
+
+            @TearDown(Level.Invocation)
+            public void tearDown() {
+                javaMutable = null;
+            }
+        }
+
+        @Benchmark
+        @SuppressWarnings("ForLoopReplaceableByForEach")
+        public void java_mutable(Initialized state) {
+            int aggregate = 0;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                aggregate ^= state.javaMutable[i];
+            }
+            assertEquals(aggregate, state.expectedAggregate);
+        }
+
+        @Benchmark
+        @SuppressWarnings("ForLoopReplaceableByForEach")
+        public void slang_persistent(Initialized state) {
+            int aggregate = 0;
+            for (javaslang.collection.Array<Integer> values = state.slangPersistent; !values.isEmpty(); values = values.tail()) {
+                aggregate ^= values.head();
+            }
+            assertEquals(aggregate, state.expectedAggregate);
+        }
+    }
+}
--- a/javaslang-benchmark/src/test/java/javaslang/benchmark/ListBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/benchmark/ListBenchmark.java
@@ -1,0 +1,249 @@
+package javaslang.benchmark;
+
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.results.RunResult;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.options.*;
+
+import java.util.*;
+import java.util.concurrent.TimeUnit;
+
+import static javaslang.benchmark.BenchmarkResultAggregator.displayRatios;
+
+public class ListBenchmark {
+    public static void main(String... args) throws Exception { /* main is more reliable than a test */
+        final Options opts = new OptionsBuilder()
+                .include(ListBenchmark.class.getSimpleName())
+                .shouldDoGC(false)
+                .shouldFailOnError(true)
+                .build();
+
+        final Collection<RunResult> results = new Runner(opts).run();
+        displayRatios(results, "^slang.+$");
+    }
+
+    @State(Scope.Benchmark)
+    @BenchmarkMode(Mode.Throughput)
+    @OutputTimeUnit(TimeUnit.SECONDS)
+    @Warmup(iterations = 20, time = 1, timeUnit = TimeUnit.SECONDS)
+    @Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+    @Fork(value = 1, jvmArgsAppend = { "-XX:+UseG1GC", "-Xss100m", "-Xms1g", "-Xmx1g", "-disableassertions" }) /* set fork to 0 if you want to debug */
+    public static class Base {
+        @Param({ "10", "100", "1000", "10000" })
+        public int CONTAINER_SIZE;
+
+        public Integer[] ELEMENTS;
+
+        @Setup
+        public void setup() {
+            final Random random = new Random(0);
+
+            ELEMENTS = new Integer[CONTAINER_SIZE];
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                final int value = random.nextInt(CONTAINER_SIZE) - (CONTAINER_SIZE / 2);
+                ELEMENTS[i] = value;
+            }
+        }
+
+        protected static <T> void assertEquals(T a, T b) {
+            if (!Objects.equals(a, b)) {
+                throw new IllegalStateException(a + " != " + b);
+            }
+        }
+    }
+
+    public static class AddAll extends Base {
+        @Benchmark
+        @SuppressWarnings("ManualArrayToCollectionCopy")
+        public void java_mutable() {
+            final java.util.ArrayList<Integer> values = new java.util.ArrayList<>(ELEMENTS.length);
+            for (Integer element : ELEMENTS) {
+                values.add(element);
+            }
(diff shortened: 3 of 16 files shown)


paplorinc:
After some possibly controversial optimizations (@danieldietrich, please don't get mad at me), the `add` and `iterate` speeds are surprisingly close to the mutable one:
```java
Ratios for: [10, 100, 1000, 10000]
Group 'AddAll':
slang_persistent/fjava_persistent       : [2.00, 2.20, 1.91, 1.95]
slang_persistent/java_mutable           : [1.50, 1.28, 1.14, 1.21]
slang_persistent/pcollections_persistent: [1.18, 1.17, 0.91, 1.45]
slang_persistent/scala_mutable          : [1.68, 1.56, 1.15, 1.22]
slang_persistent/java_mutable_linked: [1.56, 1.60, 1.22, 1.35]
slang_persistent/scala_persistent       : [1.21, 1.44, 1.08, 1.33]

Ratios for: [10, 100, 1000, 10000]
Group 'Iterate':
slang_persistent/fjava_persistent       : [1.43, 1.48, 1.54, 1.04]
slang_persistent/java_mutable           : [0.91, 0.56, 0.46, 0.43]
slang_persistent/pcollections_persistent: [1.01, 1.00, 1.34, 1.01]
slang_persistent/scala_mutable          : [1.38, 1.08, 0.96, 0.93]
slang_persistent/java_mutable_linked: [1.00, 0.86, 0.89, 1.29]
slang_persistent/scala_persistent       : [1.43, 1.44, 1.21, 0.96]
```
i.e. slang's `List` is now almost always faster than `pcollections`,`fjava`, and persistent and mutable `scala`.
It's slightly slower (but very close) to `ArrayList` and `LinkedList`.

Please take a look at the commits and tell me what I should keep, split in separate commits or just revert.

paplorinc:
The `Array` optimizations for `add` and `iteration` are more evident, i.e.

Before:
```Groovy
Group 'AddAll':
java_mutable/slang_persistent: [5.04, 10.79, 76.37]
Group 'Iterate':
java_mutable/slang_persistent: [4.25, 24.84, 427.45]
```
after
```Groovy
Group 'AddAll':
java_mutable/slang_persistent: [10.35, 22.46, 109.96]
Group 'Iterate':
java_mutable/slang_persistent: [2.08, 5.09, 9.85]
```
While add became slightly slower (it's still in the tens of millions per second), iteration via `tail` became a lot faster.

javaslang-benchmark/pom.xml
```diff
@@ -33,6 +33,21 @@
...
+            <artifactId>pcollections</artifactId>
+            <version>2.1.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.functionaljava</groupId>
+            <artifactId>functionaljava</artifactId>
+            <version>4.5</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
```
paplorinc: for the `@RegExp` annotation

javaslang-benchmark/src/test/java/javaslang/benchmark/ListBenchmark.java
```diff
@@ -0,0 +1,268 @@
...
+                .build();
+
+        final Collection<RunResult> results = new Runner(opts).run();
+        displayRatios(results, "^slang.+$");
+    }
+
+    @State(Scope.Benchmark)
+    @BenchmarkMode(Mode.Throughput)
+    @OutputTimeUnit(TimeUnit.SECONDS)
+    @Warmup(iterations = 20, time = 1, timeUnit = TimeUnit.SECONDS)
```
paplorinc: `List`s need more foreplay

javaslang-benchmark/src/test/java/javaslang/benchmark/ListBenchmark.java
```diff
@@ -0,0 +1,268 @@
...
+                scalaMutable.clear();
+            }
+        }
+
+        @Benchmark
+        @SuppressWarnings("ForLoopReplaceableByForEach")
+        public void java_mutable(Initialized state) {
+            int aggregate = 0;
+            for (final Iterator<Integer> iterator = state.javaMutable.iterator(); iterator.hasNext(); ) {
+                aggregate ^= iterator.next();
```
paplorinc: xor is faster than adding to a list

javaslang-benchmark/src/test/java/javaslang/benchmark/ArrayBenchmark.java
```diff
@@ -0,0 +1,129 @@
...
+                javaMutable = null;
+            }
+        }
+
+        @Benchmark
+        @SuppressWarnings("ForLoopReplaceableByForEach")
+        public void java_mutable(Initialized state) {
+            int aggregate = 0;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                aggregate ^= state.javaMutable[i];
```
paplorinc: xor is faster than adding to a list, so it doesn't distort the ratios and provides a very basic validation also

javaslang/src/main/java/javaslang/$.java
```diff
@@ -33,8 +33,7 @@
     // -- javaslang.collection
 
     // List
-    @Unapply static <T> Tuple2<T, List<T>> List(List.Cons<T> cons) { return Tuple.of(cons.head(), cons.tail()); }
-    @Unapply static <T> Tuple0 List(List.Nil<T> nil) { return Tuple.empty(); }
+    @Unapply static <T> Tuple2<T, List<T>> List(List<T> cons) { return Tuple.of(cons.head(), cons.tail()); }
```
paplorinc: @danieldietrich, this wasn't tested so I presumed it wasn't working before either :p

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -22,31 +22,43 @@
  * @since 2.0.0
  */
 public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serializable {
-
     private static final long serialVersionUID = 1L;
 
     private static final Array<?> EMPTY = new Array<>(new Object[0]);
 
-    private final Object[] back;
+    private final int startIndex, length;
```
paplorinc: this way we have `O(1)`: `head`, `tail`, `init`, `drop`, `dropRight`

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -807,23 +801,25 @@ public int length() {
...
-            final Array<T> tail = tail();
-            if (tail.isEmpty()) {
-                return of(this);
-            } else {
-                final Array<Array<T>> zero = empty();
-                return distinct().foldLeft(zero, (xs, x) -> {
-                    final Function<Array<T>, Array<T>> prepend = l -> l.prepend(x);
-                    return xs.appendAll(remove(x).permutations().map(prepend));
-                });
+            Array<Array<T>> results = empty();
```
paplorinc: eliminated recursion inside a foldLeft to make it debuggable and avoid problems with the stack

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -807,23 +801,25 @@ public int length() {
...
+                }
             }
+            return results;
         }
     }
 
     @Override
     public Array<T> prepend(T element) {
-        return insert(0, element);
+        final Object[] copy = new Object[length() + 1];
```
paplorinc: it's apparently faster this way :/

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -1075,34 +1075,31 @@ public int length() {
...
-        System.arraycopy(back, beginIndex, arr, 0, arr.length);
-        return wrap(arr);
     }
 
     @Override
     public Array<T> tail() {
         if (isEmpty()) {
             throw new UnsupportedOperationException("tail() on empty Array");
+        } else {
+            return with(startIndex + 1, length);
```
paplorinc: this line was the main motivation of this change :)

javaslang/src/test/java/javaslang/collection/VectorTest.java
```diff
@@ -202,46 +202,4 @@ public void shouldStringifyNil() {
     public void shouldStringifyNonNil() {
         assertThat(of(null, 1, 2, 3).toString()).isEqualTo("Vector(null, 1, 2, 3)");
     }
-
```
paplorinc: I think this was a wrong copy-paste here, it was testing `List` also :/

javaslang/src/main/java/javaslang/$.java
danieldietrich: Hi @paplorinc, the removed lines should work. These are the patterns that are generated at compile time:

```java
// @Unapply static <T> Tuple2<T, List<T>> List(List.Cons<T> cons)
public static <T, _1 extends T, _2 extends List<T>> Pattern2<List.Cons<T>, _1, _2> List(Pattern<_1, ?> p1, Pattern<_2, ?> p2) {
    return Pattern2.of(List.Cons.class, p1, p2, $::List);
}

// @Unapply static <T> Tuple0 List(List.Nil<T> nil)
public static <T> Pattern0<List.Nil<T>> List() {
    return Pattern0.of(List.Nil.class);
}
```

Some tests are in MatchTest.java. But I will add these (which work with lines deleted above):

```java
@Test
public void shouldDecomposeEmptyList() {
    final List<Integer> list = List.empty();
    final boolean isEmpty = Match(list).of(
            Case(List($(), $()), (x, xs) -> false),
            Case(List(), true)
    );
    assertThat(isEmpty).isTrue();
}

@Test
public void shouldDecomposeNonEmptyList() {
    final List<Integer> list = List.of(1);
    final boolean isNotEmpty = Match(list).of(
            Case(List($(), $()), (x, xs) -> true),
            Case(List(), false)
    );
    assertThat(isNotEmpty).isTrue();
}
```

Could you please explain your observation?

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -546,24 +558,22 @@ public T get(int index) {
     public Array<T> drop(long n) {
         if (n <= 0) {
             return this;
-        }
-        if (n >= length()) {
+        } else if (n >= length()) {
             return empty();
+        } else {
+            return with(startIndex + (int) n, length - (int) n);
```
danieldietrich: the cast does not work for big `n`.
It is confusing that we use both `length()` and `length`. We should use always `length` because is an accessible instance variable.

javaslang/src/main/java/javaslang/collection/Array.java
danieldietrich: Wow - same result? Nice!

javaslang/src/test/java/javaslang/collection/VectorTest.java
danieldietrich: Nice finding, thx!!

javaslang/src/test/java/javaslang/collection/ListTest.java
```diff
@@ -292,47 +292,6 @@ public void shouldStringifyNonNil() {
...
-            /*
-             * Hack the serialized data and fake zero elements.
-             */
-            listWithOneElement[index] = 0;
-            Serializables.deserialize(listWithOneElement);
-        } catch (IllegalStateException x) {
-            throw (x.getCause() != null) ? x.getCause() : x;
-        }
-    }
-
```
danieldietrich: You entirely deleted the method. Do we have a replacement for the test?

javaslang/src/main/java/javaslang/collection/Array.java
danieldietrich: Maybe we should move this code to the `insert` method and add an `if-else` for the case that the index is `0`. This could also boost the insert operation a little. What do you think?

paplorinc: That was done previously, but when I know the index is `0` I can spare the index checks, probably those 3-4 ifs make it measurably quicker, as it's inherently a very fast operation.

javaslang/src/main/java/javaslang/collection/Array.java
paplorinc: Yes, I thought I already inlined them :).
The cast is needed as you can't index arrays with `long`s.

javaslang/src/test/java/javaslang/collection/ListTest.java
paplorinc: You can serialize the entire object now, I think this was testing that the interface is non-serializable, just its implementations. Please correct me.

javaslang/src/main/java/javaslang/$.java
paplorinc: If I'm mistaken correctly, only a few classes are currently decomposable to `head` and `tail`, so I took the liberty of not making a separate `nil` object for `List` - though we will probably have to update pattern logic to accommodate.

javaslang/src/main/java/javaslang/collection/Array.java
danieldietrich: ok, I just thought `insert(0, ...)` could also benefit - but I think that is a minor optimization and we are trading it for additional if's. It is ok as it is!

javaslang/src/main/java/javaslang/collection/Array.java
danieldietrich: Ok, I think it is the best we can do. Segmenting using a two-dimensional array would be too much...

javaslang/src/test/java/javaslang/collection/ListTest.java
danieldietrich: Ruslan added these tests to increase the code coverage towards 100%. They cover the internals of `readObject` and `writeObject`. We test that `InvalidObjectException` is thrown if the bytes representing the serialized object are broken. It is not-so-good that the test makes assumptions about implementation details, like the length of the list is stored at index 0 and so on...

javaslang/src/main/java/javaslang/$.java
danieldietrich: We need them in order to decompose algebraic data types (ADTs) (see my [previous comment](https://github.com/javaslang/javaslang/pull/1323/files/0756c6164f8e6c186578cee0cbecf2ee101e43a8#r62576805)).

Your version does throw `NoSuchElementException` on empty List (see `cons.head()`):

```java
@Unapply static <T> Tuple2<T, List<T>> List(List<T> cons) { return Tuple.of(cons.head(), cons.tail()); }
```

`Option` is a good example for an ADT, it can be `Some` and `None`. `List` is another example for an ADT, it can be `Cons` and `Nil`. Because `Stream` is similar to `List`, it can be `Cons` and `Empty`, I decided to choose different names than in Scala. In Javaslang it is just `List` and `Stream`.

Scala:

```scala
list match {
    case head :: tail => ...
    case Nil => ...
}
```

Javaslang:

```java
Match(list).of(
    Case(List($(), $()), (head, tail) => ...),
    Case(List(), ...)
);
```

We can't generalize the pattern matching framework to decompose arbitrary types into head/tail because we decide it _by type_  and not by _isEmpty_ check.

javaslang/src/main/java/javaslang/collection/List.java
```diff
@@ -85,9 +81,80 @@
  * @author Daniel Dietrich
  * @since 1.1.0
  */
-public interface List<T> extends Kind1<List<?>, T>, LinearSeq<T>, Stack<T> {
+public final class List<T> implements Stack<T>, LinearSeq<T>, Kind1<List<?>, T>, Serializable {
```
danieldietrich: @paplorinc Ohhh my gosh - I just saw that you made List a class O_O. You can't do such big changes in a PR named "Added ... Benchmarks". List is an algebraic data type, like Option (Some/None). We need to distinguish the states by type. It is essential for pattern matching. Also we follow best practices from the book "Purely Functional Data Structures". Scala and Haskell do the same for example.

Please revert the List changes and open a discussion/issue if you think there is potential to enhance the implementation.

danieldietrich: Never mind, I was a little bit shocked because it was the first collection I implemented in Javaslang 1.0. There was already a discussion why it is an interface and not a class but it is intentionally designed that way. It should be `sealed` in order to prevent others to implement it but in Java it is not possible to express it.

danieldietrich:
I think the PR's get too big. We should separate our changes. E.g. first PR: benchmarks, then additional PRs: improved List, improved Array, improved ...

javaslang/src/main/java/javaslang/collection/List.java
paplorinc: This change started as a benchmark experiment, pushed it and said that it's "controversial" and that I will gladly split it into many, more cohesive commits, if needed; I pushed it to gather comments. Though I expected something else than "dude, you're not supposed to touch that, because that's how Haskell and Scala did it, so it's the way we should do it also - even if it might not be the best solution in Java".
Will split it into more commits.

danieldietrich: > Though I expected something else than "dude, you're not supposed to touch that, because that's how Haskell and Scala did it, so it's the way we should do it also - even if it might not be the best solution in Java".

Sorry! I was not prepared for such a change in this PR - it was a shocking moment and I was emotional. Changing the List from an ADT to a plain class has so much implications for the whole code base that many of my design decisions would be thrown over board.

There are different types of changes to the code base of Javaslang:

* reformatting - this shouldn't be done by committers in general (but solving inconsistencies is ok)
* adding features - this is the 'normal' case for PRs
* changing/optimizing implementations - also a 'normal' PR case
* performing design changes - this should be discussed in depth because it may have deep implications.

Design changes are very dangerous. Even if they are right they most probably will break backward compatibility. We need to be careful here and slow down a bit.

We should separate benchmarks and optimizations into multiple PRs instead of multiple commits. Is that possible without too much effort?

PR #1300, opened 2016-04-26
Added persistent `PriorityQueue` implementation

paplorinc:
Fixes https://github.com/javaslang/javaslang/issues/1264

TODO:
* Understand what I just implemented :p
 * http://www.brics.dk/RS/96/37/BRICS-RS-96-37.pdf
 * https://en.wikipedia.org/wiki/Binomial_heap
 * http://scottlobdell.me/2016/04/purely-functional-data-structures-python
* make the rest of the methods `O(1)` by providing a global forest root and via bootstrapping, described above 
* Add proper documentation

PR #1323, opened 2016-05-08
Added List decomposition tests

danieldietrich:


javaslang/src/test/java/javaslang/MatchTest.java
```diff
@@ -219,6 +219,26 @@ public void shouldDecomposeSomeSomeTuple() {
 
     // -- List
 
+    @Test
+    public void shouldDecomposeEmptyList() {
```
paplorinc: can we decompose e.g. `(Priority)Queue` head and tails also?

danieldietrich: Decomposition, more specifically pattern matching, does only work for algebraic data types. Wikipedia has a good explanation (https://en.wikipedia.org/wiki/Algebraic_data_type), pattern matching is mentioned in the 4th paragraph.

An empty PriorityQueue has no head. Therefore we can't decompose it into a `Tuple2(head, tail)`.

paplorinc: > An empty PriorityQueue has no head

An empty `List` doesn't have a `head` either. If we can decompose a `List`, we should be able to decompose any `Traversable` - it should be possible to base it on `isEmpty` or to provide separate `empty` classes for (FYI: https://github.com/javaslang/javaslang/pull/1300#discussion_r61717723).

danieldietrich: The match pattern definitions (see $.java) have to return a Tuple. Therefore using isEmpty is not possible (the method returns either `Tuple0` or `Tuple2` but it cannot return both).

Providing separate classes is the only possibility I see. Then my dream would come true and all collections could be interfaces with two nested classes `NonEmpty` and `Empty`. I like that idea. Those collections without an ordering/Comparator could have a singleton Empty class.

(Having interfaces is nice because we are able to proxy them. I make use of that in `Lazy.val` for example. But the nested generic `T` in `A<T>` is nasty.)

Issue #1271, opened 2016-04-14
Create a persistent `BitSet` data structure


paplorinc:
`BitSet` is extremely useful when you want to test the presence of some enumerable entity, i.e. a compact `Set` of `Booleans`.
We might want a pure one in Javaslang also.

It could be based on [Scala](https://github.com/scala/scala/blob/2.12.x/src/library/scala/collection/immutable/BitSet.scala)'s or  [Haskell](http://hackage.haskell.org/package/bitset)'s implementations.

danieldietrich:
Nice idea. Basically we align to Scala (type hierarchy and naming). Looking at Haskell is better in terms of impl details because it is more clean/pure. However, when it comes to recursion we use iteration with mutable local method vars instead.

ruslansennov:
@danieldietrich @paplorinc please take a look on this skeleton: b63602eeec6d4f70853b5e8691425cd6b78876c1 (copied from Scala implementation)

ruslansennov:
#1334

Issue #1332, opened 2016-05-14
Add Java-like Value.collect() methods to increase interoperability


danieldietrich:
`java.util.stream.Stream` has:

* [`collect(Collector<? super T,A,R> collector)`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#collect-java.util.stream.Collector-)
* [`collect(Supplier<R> supplier, BiConsumer<R,? super T> accumulator, BiConsumer<R,R> combiner)`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#collect-java.util.function.Supplier-java.util.function.BiConsumer-java.util.function.BiConsumer-)

When we add these as `default` methods to `javaslang.Value`, Javaslang types will be _✨automagically✨_ convertible to **all libraries** that are built on top of `java.util.stream.Stream` (e.g. [jooq/jool](https://github.com/jOOQ/jOOL), [amaembo/streamex](https://github.com/amaembo/streamex), ...), but more important this allows us to re-use user-defined aggregation functions based on Collectors.

Because this is a conversion method, we implement it in `Value` (or `Convertible`) rather than in `Traversable`. `Iterator` will benefit from it because in a future Javaslang version `Iterator` will not extend `Traversable` any more (see [#1295](https://github.com/javaslang/javaslang/issues/1295#issuecomment-213835541)).

danieldietrich:
Since this is a conversion method, we may implement it in `Value` (or `Convertible`) rather than in `Traversable`. `Iterator` would benefit from it because in a future Javaslang version `Iterator` will not extend `Traversable` any more. Instead `Value` is split into `Value` and `Convertible` and `Iterator` will extend `Convertible`. (see [#1295](https://github.com/javaslang/javaslang/issues/1295#issuecomment-213835541))

mvh77:
Is this really necessary? You can just do .toJavaStream() and call collect(...) on that.

danieldietrich:
It is more concise. Also it works for all Values. It might be confusing to stream an Option for example because `java.util.stream.Stream` is associated with collections.

mvh77:
I'm not really against it and I thought about it myself some time ago, but it's a bit bloaty in my opinion since you can do without and it's something you need very rarely. And btw, Option is a kind of collection! :wink:

danieldietrich:
I will think about it... we should only pull in changes that make sense.

danieldietrich:
[...]

Issue #1295, opened 2016-04-23
[type hierarchy] Change Iterator not to inherit from Iterable


mvh77:
As  javaslang Iterator is iterable, it has an iterator() method, but currently it returns this, which breaks the contract of Iterable. It should always return a new iterator that can be traversed. This prints both foo and bar twice in scala but not with javaslang:

``` scala
def printTwice[T](it: Iterable[T]) = {
  it.iterator.foreach(println)
  it.iterator.foreach(println)
}

printTwice(Vector("foo"))
printTwice(Iterable("bar"))
```

This works differently with and without the call to .iterator() which IMO is not ok:

``` java
List<Integer> transform = List.of(1, 2, 3)
  .iterator()
  .transform(i -> List.ofAll(i).appendAll(i));
System.out.println(transform);
```

ruslansennov:
Your javaslang code not consistent with your Scala code.
It should be something like this:

```java
public class It {

    private static <T> void printTwice(Iterable<T> it) {
        it.iterator().forEachRemaining(System.out::println);
        it.iterator().forEachRemaining(System.out::println);
    }

    public static void main(String[] args) {
        printTwice(Vector.of("foo"));   // print twice, ok
        printTwice(Iterator.of("bar")); // print once, ok
    }
}
```
I believe it is impossible to return a new iterator in <code>Iterator#iterator()</code> method because Iterator is mutable object.

mvh77:
It's not the same code, but that's not the point. The point is that since Iterator is an Iterable, you can pass it to any method that expects an Iterable, and an Iterable is supposed to give you iterators that you can traverse the collection with. An Iterator can be iterated once, but an Iterable must provide new Iterators on all calls to iterator(). The method receiving the Iterable has no way to know that the Iterable provides the same Iterator every time.

Iterator needs to keep a reference to the sequence it was created with, that's how it can create new iterators on every call.

danieldietrich:
Yep, I think Ruslan is right, we need to live with that caveat. There is a reason for that:

Scala's type hierarchy is different, `Iterable` extends `Traversable` and `Iterator` extends `Traversable`. Especially `Iterator` is not an `Interable`.

In Javaslang `Traversable` extends `Iterable` and `Iterator` extends `Traversable`. Especially `Iterator` is an `Iterable`.

Because of this difference and the fact that Iterator is stateful, we cannot create a new Iterator based on an Iterator.

danieldietrich:
[...]

PR #1334, opened 2016-05-15
Bitset #1271

ruslansennov:
Internal offset is not implemented yet, and I'm not sure it needed. Should be discussed in own issue

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/BitSet.java
+++ b/javaslang/src/main/java/javaslang/collection/BitSet.java
@@ -31,9 +31,7 @@ public interface BitSet<T> extends SortedSet<T> {
 
     long serialVersionUID = 1L;
 
-    class Builder<T> implements Serializable {
-
-        private static final long serialVersionUID = 1L;
+    class Builder<T> {
 
         final static Builder<Integer> DEFAULT = new Builder<>(i -> i, i -> i);
 
@@ -54,15 +52,15 @@ public interface BitSet<T> extends SortedSet<T> {
         }
 
         public BitSet<T> empty() {
-            return new BitSetModule.BitSet1<>(this, 0L);
+            return new BitSetModule.BitSet1<>(fromInt, toInt, 0L);
         }
 
         public BitSet<T> of(T t) {
             final int value = toInt.apply(t);
             if(value < BitSetModule.BITS_PER_WORD) {
-                return new BitSetModule.BitSet1<>(this, 1L << value);
+                return new BitSetModule.BitSet1<>(fromInt, toInt, 1L << value);
             } else if(value < 2 * BitSetModule.BITS_PER_WORD) {
-                return new BitSetModule.BitSet2<>(this, 0L, 1L << value);
+                return new BitSetModule.BitSet2<>(fromInt, toInt, 0L, 1L << value);
             } else {
                 return empty().add(t);
             }
@@ -71,13 +69,7 @@ public interface BitSet<T> extends SortedSet<T> {
         @SuppressWarnings("varargs")
         @SafeVarargs
         public final BitSet<T> of(T... values) {
-            if(values.length == 0) {
-                return empty();
-            } else if(values.length == 1) {
-                return of(values[0]);
-            } else {
-                return empty().addAll(Array.wrap(values));
-            }
+            return empty().addAll(Array.wrap(values));
         }
 
         public BitSet<T> ofAll(Iterable<? extends T> values) {
@@ -373,8 +365,7 @@ public interface BitSet<T> extends SortedSet<T> {
 
     @Override
     default <U> SortedSet<U> flatMap(Function<? super T, ? extends Iterable<? extends U>> mapper) {
-        Objects.requireNonNull(mapper, "mapper is null");
-        return TreeSet.ofAll(naturalComparator(), iterator().flatMap(mapper));
+        return flatMap(naturalComparator(), mapper);
     }
 
     @Override
@@ -591,10 +582,12 @@ interface BitSetModule {
 
         private static final long serialVersionUID = 1L;
 
-        Builder<T> builder;
+        final Function1<Integer, T> fromInt;
+        final Function1<T, Integer> toInt;
 
-        AbstractBitSet(Builder<T> builder) {
-            this.builder = builder;
+        AbstractBitSet(Function1<Integer, T> fromInt, Function1<T, Integer> toInt) {
+            this.fromInt = fromInt;
+            this.toInt = toInt;
         }
 
         abstract int getWordsNum();
@@ -603,18 +596,26 @@ interface BitSetModule {
 
         abstract long getWord(int index);
 
+        BitSet<T> createEmpty() {
+            return new BitSet1<>(fromInt, toInt, 0L);
+        }
+
+        BitSet<T> createFromAll(Iterable<? extends T> values) {
+            return createEmpty().addAll(values);
+        }
+
         BitSet<T> fromBitMaskNoCopy(long[] elements) {
             final int len = elements.length;
             if (len == 0) {
-                return builder.empty();
+                return createEmpty();
             }
             if (len == 1) {
-                return new BitSet1<>(builder, elements[0]);
+                return new BitSet1<>(fromInt, toInt, elements[0]);
             }
             if (len == 2) {
-                return new BitSet2<>(builder, elements[0], elements[1]);
+                return new BitSet2<>(fromInt, toInt, elements[0], elements[1]);
             }
-            return new BitSetN<>(builder, elements);
+            return new BitSetN<>(fromInt, toInt, elements);
         }
 
         private void setElement(long[] words, int element) {
@@ -642,7 +643,7 @@ interface BitSetModule {
             if (contains(t)) {
                 return this;
             } else {
-                final int element = builder.toInt.apply(t);
+                final int element = toInt.apply(t);
                 final long[] copy = copyExpand(1 + (element >> ADDRESS_BITS_PER_WORD));
                 setElement(copy, element);
                 return fromBitMaskNoCopy(copy);
@@ -652,13 +653,13 @@ interface BitSetModule {
         @Override
         public BitSet<T> distinctBy(Comparator<? super T> comparator) {
             Objects.requireNonNull(comparator, "comparator is null");
-            return builder.ofAll(iterator().distinctBy(comparator));
+            return createFromAll(iterator().distinctBy(comparator));
         }
 
         @Override
         public <U> BitSet<T> distinctBy(Function<? super T, ? extends U> keyExtractor) {
             Objects.requireNonNull(keyExtractor, "keyExtractor is null");
-            return builder.ofAll(iterator().distinctBy(keyExtractor));
+            return createFromAll(iterator().distinctBy(keyExtractor));
         }
 
         @Override
@@ -666,9 +667,9 @@ interface BitSetModule {
             if (n <= 0) {
                 return this;
             } else if (n >= length()) {
-                return builder.empty();
+                return createEmpty();
             } else {
-                return builder.ofAll(iterator().drop(n));
+                return createFromAll(iterator().drop(n));
             }
         }
 
@@ -677,16 +678,16 @@ interface BitSetModule {
             if (n <= 0) {
                 return this;
             } else if (n >= length()) {
-                return builder.empty();
+                return createEmpty();
             } else {
-                return builder.ofAll(iterator().dropRight(n));
+                return createFromAll(iterator().dropRight(n));
             }
         }
 
         @Override
         public BitSet<T> dropWhile(Predicate<? super T> predicate) {
             Objects.requireNonNull(predicate, "predicate is null");
-            final BitSet<T> bitSet = builder.ofAll(iterator().dropWhile(predicate));
+            final BitSet<T> bitSet = createFromAll(iterator().dropWhile(predicate));
             return (bitSet.length() == length()) ? this : bitSet;
         }
 
@@ -694,13 +695,13 @@ interface BitSetModule {
         public BitSet<T> intersect(Set<? extends T> elements) {
             Objects.requireNonNull(elements, "elements is null");
             if (isEmpty() || elements.isEmpty()) {
-                return builder.empty();
+                return createEmpty();
             } else {
                 int size = size();
                 if (size <= elements.size()) {
                     return retainAll(elements);
                 } else {
-                    BitSet<T> results = builder.ofAll(elements).retainAll(this);
+                    BitSet<T> results = createFromAll(elements).retainAll(this);
                     return (size == results.size()) ? this : results;
                 }
             }
@@ -708,13 +709,13 @@ interface BitSetModule {
 
         @Override
         public Iterator<BitSet<T>> sliding(long size, long step) {
-            return iterator().sliding(size, step).map(builder::ofAll);
+            return iterator().sliding(size, step).map(this::createFromAll);
         }
 
         @Override
         public Tuple2<BitSet<T>, BitSet<T>> span(Predicate<? super T> predicate) {
             Objects.requireNonNull(predicate, "predicate is null");
-            return iterator().span(predicate).map(builder::ofAll, builder::ofAll);
+            return iterator().span(predicate).map(this::createFromAll, this::createFromAll);
         }
 
         @Override
@@ -723,44 +724,44 @@ interface BitSetModule {
             return Collections.scanLeft(this, zero, operation, new java.util.ArrayList<T>(), (arr, t) -> {
                 arr.add(t);
                 return arr;
-            }, builder::ofAll);
+            }, this::createFromAll);
         }
 
         @Override
         public Tuple2<BitSet<T>, BitSet<T>> partition(Predicate<? super T> predicate) {
             Objects.requireNonNull(predicate, "predicate is null");
-            return iterator().partition(predicate).map(builder::ofAll, builder::ofAll);
+            return iterator().partition(predicate).map(this::createFromAll, this::createFromAll);
         }
 
         @Override
         public BitSet<T> filter(Predicate<? super T> predicate) {
             Objects.requireNonNull(predicate, "predicate is null");
-            final BitSet<T> bitSet = builder.ofAll(iterator().filter(predicate));
+            final BitSet<T> bitSet = createFromAll(iterator().filter(predicate));
             return (bitSet.length() == length()) ? this : bitSet;
         }
 
         @Override
         public  <C> Map<C, BitSet<T>> groupBy(Function<? super T, ? extends C> classifier) {
             Objects.requireNonNull(classifier, "classifier is null");
-            return iterator().groupBy(classifier).map((key, iterator) -> Tuple.of(key, builder.ofAll(iterator)));
+            return iterator().groupBy(classifier).map((key, iterator) -> Tuple.of(key, createFromAll(iterator)));
         }
 
         @Override
         public Comparator<T> comparator() {
-            return (t1, t2) -> Integer.compare(builder.toInt.apply(t1), builder.toInt.apply(t2));
+            return (t1, t2) -> Integer.compare(toInt.apply(t1), toInt.apply(t2));
         }
 
         @Override
         public BitSet<T> takeWhile(Predicate<? super T> predicate) {
             Objects.requireNonNull(predicate, "predicate is null");
-            final BitSet<T> result = builder.ofAll(iterator().takeWhile(predicate));
+            final BitSet<T> result = createFromAll(iterator().takeWhile(predicate));
             return (result.length() == length()) ? this : result;
         }
 
         @Override
         @SuppressWarnings("unchecked")
         public BitSet<T> addAll(Iterable<? extends T> elements) {
-            final Stream<Integer> source = Stream.ofAll(elements).map(builder.toInt);
+            final Stream<Integer> source = Stream.ofAll(elements).map(toInt);
             final long[] copy = copyExpand(1 + (source.max().getOrElse(0) >> ADDRESS_BITS_PER_WORD));
             source.forEach(element -> {
                 if (element < 0) {
@@ -773,7 +774,7 @@ interface BitSetModule {
 
         @Override
         public boolean contains(T t) {
-            final int element = builder.toInt.apply(t);
+            final int element = toInt.apply(t);
             if (element < 0) {
                 throw new IllegalArgumentException("bitset element must be >= 0");
             }
@@ -788,7 +789,7 @@ interface BitSetModule {
             } else {
                 final long last = getWord(getWordsNum() - 1);
                 final int element = BITS_PER_WORD * (getWordsNum() - 1) + BITS_PER_WORD - Long.numberOfLeadingZeros(last) - 1;
-                return remove(builder.fromInt.apply(element));
+                return remove(fromInt.apply(element));
             }
         }
 
@@ -796,7 +797,7 @@ interface BitSetModule {
         public Iterator<T> iterator() {
             return Stream.range(0, getWordsNum() << ADDRESS_BITS_PER_WORD)
                     .filter(i -> (getWord(i >> ADDRESS_BITS_PER_WORD) & (1L << i)) != 0)
-                    .map(builder.fromInt)
+                    .map(fromInt)
                     .iterator();
         }
 
@@ -812,29 +813,29 @@ interface BitSetModule {
         @Override
         public BitSet<T> take(long n) {
             if (n <= 0) {
-                return builder.empty();
+                return createEmpty();
             } else if (n >= length()) {
                 return this;
             } else {
-                return builder.ofAll(iterator().take(n));
+                return createFromAll(iterator().take(n));
             }
         }
 
         @Override
         public BitSet<T> takeRight(long n) {
             if (n <= 0) {
-                return builder.empty();
+                return createEmpty();
             } else if (n >= length()) {
                 return this;
             } else {
-                return builder.ofAll(iterator().takeRight(n));
+                return createFromAll(iterator().takeRight(n));
             }
         }
 
         @Override
         public BitSet<T> remove(T t) {
             if (contains(t)) {
-                final int element = builder.toInt.apply(t);
+                final int element = toInt.apply(t);
                 final long[] copy = copyExpand(getWordsNum());
                 unsetElement(copy, element);
                 return fromBitMaskNoCopy(shrink(copy));
@@ -846,7 +847,7 @@ interface BitSetModule {
         @Override
         @SuppressWarnings("unchecked")
         public BitSet<T> removeAll(Iterable<? extends T> elements) {
-            final Stream<Integer> source = Stream.ofAll(elements).map(builder.toInt);
+            final Stream<Integer> source = Stream.ofAll(elements).map(toInt);
             final long[] copy = copyExpand(getWordsNum());
             source.forEach(element -> {
                 unsetElement(copy, element);
@@ -883,8 +884,8 @@ interface BitSetModule {
 
         private final long elements;
 
-        BitSet1(Builder<T> builder, long elements) {
-            super(builder);
+        BitSet1(Function1<Integer, T> fromInt, Function1<T, Integer> toInt, long elements) {
+            super(fromInt, toInt);
             this.elements = elements;
         }
 
@@ -915,7 +916,7 @@ interface BitSetModule {
         @Override
         public T head() {
             if (elements != 0) {
-                return builder.fromInt.apply(Long.numberOfTrailingZeros(elements));
+                return fromInt.apply(Long.numberOfTrailingZeros(elements));
             }
             throw new NoSuchElementException("head of empty BitSet");
         }
@@ -927,8 +928,8 @@ interface BitSetModule {
 
         private final long elements1, elements2;
 
-        BitSet2(Builder<T> builder, long elements1, long elements2) {
-            super(builder);
+        BitSet2(Function1<Integer, T> fromInt, Function1<T, Integer> toInt, long elements1, long elements2) {
+            super(fromInt, toInt);
             this.elements1 = elements1;
             this.elements2 = elements2;
         }
@@ -965,9 +966,9 @@ interface BitSetModule {
         @Override
         public T head() {
             if (elements1 != 0) {
-                return builder.fromInt.apply(Long.numberOfTrailingZeros(elements1));
+                return fromInt.apply(Long.numberOfTrailingZeros(elements1));
             } else if (elements2 != 0) {
-                return builder.fromInt.apply(BITS_PER_WORD + Long.numberOfTrailingZeros(elements2));
+                return fromInt.apply(BITS_PER_WORD + Long.numberOfTrailingZeros(elements2));
             }
             throw new NoSuchElementException("head of empty BitSet");
         }
@@ -979,8 +980,8 @@ interface BitSetModule {
 
         private final long[] elements;
 
-        BitSetN(Builder<T> builder, long[] elements) {
-            super(builder);
+        BitSetN(Function1<Integer, T> fromInt, Function1<T, Integer> toInt, long[] elements) {
+            super(fromInt, toInt);
             this.elements = elements;
         }
 
@@ -1013,7 +1014,7 @@ interface BitSetModule {
             int offset = 0;
             for (int i = 0; i < getWordsNum(); i++) {
                 if(elements[i] != 0) {
-                    return builder.fromInt.apply(offset + Long.numberOfTrailingZeros(elements[i]));
+                    return fromInt.apply(offset + Long.numberOfTrailingZeros(elements[i]));
                 }
                 offset += BITS_PER_WORD;
             }
--- a/javaslang/src/main/java/javaslang/collection/BitSet.java
+++ b/javaslang/src/main/java/javaslang/collection/BitSet.java
@@ -11,6 +11,7 @@ import javaslang.Tuple2;
 import javaslang.Tuple3;
 import javaslang.control.Option;
 
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.NoSuchElementException;
@@ -30,14 +31,16 @@ public abstract class BitSet<T> implements SortedSet<T> {
 
     private static final long serialVersionUID = 1L;
 
-    public static class Builder<T> {
+    public static class Builder<T> implements Serializable {
+
+        private static final long serialVersionUID = 1L;
 
         final static Builder<Integer> DEFAULT = new Builder<>(i -> i, i -> i);
 
         final Function<Integer, T> fromInt;
         final Function<T, Integer> toInt;
 
-        Builder(Function<Integer, T> fromInt, Function<T, Integer> toInt) {
+        Builder(Function1<Integer, T> fromInt, Function1<T, Integer> toInt) {
             this.fromInt = fromInt;
             this.toInt = toInt;
         }
@@ -57,9 +60,9 @@ public abstract class BitSet<T> implements SortedSet<T> {
         public BitSet<T> of(T t) {
             final int value = toInt.apply(t);
             if(value < BitSetModule.BITS_PER_WORD) {
-                return new BitSetModule.BitSet1<>(this, value);
+                return new BitSetModule.BitSet1<>(this, 1L << value);
             } else if(value < 2 * BitSetModule.BITS_PER_WORD) {
-                return new BitSetModule.BitSet2<>(this, 0L, value);
+                return new BitSetModule.BitSet2<>(this, 0L, 1L << value);
             } else {
                 return empty().add(t);
             }
@@ -93,13 +96,13 @@ public abstract class BitSet<T> implements SortedSet<T> {
         }
     }
 
-    public static <T> Builder<T> withRelations(Function<Integer, T> fromInt, Function<T, Integer> toInt) {
+    public static <T> Builder<T> withRelations(Function1<Integer, T> fromInt, Function1<T, Integer> toInt) {
         return new Builder<>(fromInt, toInt);
     }
 
     public static <T extends Enum<T>> Builder<T> withEnum(Class<T> clz) {
-        final Function<Integer, T> fromInt = i -> clz.getEnumConstants()[i];
-        final Function<T, Integer> toInt = Enum<T>::ordinal;
+        final Function1<Integer, T> fromInt = i -> clz.getEnumConstants()[i];
+        final Function1<T, Integer> toInt = Enum<T>::ordinal;
         return new Builder<>(fromInt, toInt);
     }
 
@@ -382,6 +385,11 @@ public abstract class BitSet<T> implements SortedSet<T> {
     }
 
     @Override
+    public String toString() {
+        return mkString(stringPrefix() + "(", ", ", ")");
+    }
+
+    @Override
     public Comparator<T> comparator() {
         return (t1, t2) -> Integer.compare(builder.toInt.apply(t1), builder.toInt.apply(t2));
     }
@@ -466,7 +474,11 @@ public abstract class BitSet<T> implements SortedSet<T> {
 
     @Override
     public BitSet<T> tail() {
-        return remove(head());
+        if (isEmpty()) {
+            throw new UnsupportedOperationException("tail of empty BitSet");
+        } else {
+            return remove(head());
+        }
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+            Objects.requireNonNull(s, "s is null");
+            return empty().addAll(Collections.fill(n, s));
+        }
+    }
+
+    static <T> Builder<T> withRelations(Function1<Integer, T> fromInt, Function1<T, Integer> toInt) {
+        return new Builder<>(fromInt, toInt);
+    }
+
+    static <T extends Enum<T>> Builder<T> withEnum(Class<T> clz) {
```
paplorinc: minor suggestion: instead of using abbreviations and misspellings of words (cla*zz*), could we please rename to reflect the intention, i.e. `Class<T> enumClass`?

paplorinc: :+1: for enums :)

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+
+    static Builder<Short> withShorts() {
+        return new Builder<>(Integer::shortValue, Short::intValue);
+    }
+
+    /**
+     * Returns a {@link java.util.stream.Collector} which may be used in conjunction with
+     * {@link java.util.stream.Stream#collect(java.util.stream.Collector)} to obtain a {@link javaslang.collection.BitSet}.
+     *
+     * @return A javaslang.collection.List Collector.
```
paplorinc:
```java
@return A  {@link javaslang.collection.BitSet} Collector
```

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+    }
+
+    /**
+     * Returns a BitSet containing {@code n} values of a given Function {@code f}
+     * over a range of integer values from 0 to {@code n - 1}.
+     *
+     * @param n The number of elements in the BitSet
+     * @param f The Function computing element values
+     * @return A BitSet consisting of elements {@code f(0),f(1), ..., f(n - 1)}
+     * @throws NullPointerException if {@code f} are null
```
paplorinc: `f` ~~are~~ is null?

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+        return Builder.DEFAULT.tabulate(n, f);
+    }
+
+    /**
+     * Returns a BitSet containing {@code n} values supplied by a given Supplier {@code s}.
+     *
+     * @param n The number of elements in the BitSet
+     * @param s The Supplier computing element values
+     * @return A BitSet of size {@code n}, where each element contains the result supplied by {@code s}.
+     * @throws NullPointerException if {@code s} are null
```
paplorinc: same

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+
+    static BitSet<Integer> ofAll(Iterable<Integer> values) {
+        return Builder.DEFAULT.ofAll(values);
+    }
+
+    /**
+     * Creates a BitSet based on the elements of a boolean array.
+     *
+     * @param array a boolean array
+     * @return A new BitSet of Boolean values
```
paplorinc: ~~Boolean~~ boolean values

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+    static BitSet<Byte> ofAll(byte[] array) {
+        Objects.requireNonNull(array, "array is null");
+        return BitSet.withBytes().ofAll(Iterator.ofAll(array));
+    }
+
+    /**
+     * Creates a BitSet based on the elements of a char array.
+     *
+     * @param array a char array
+     * @return A new BitSet of Character values
```
paplorinc: I think all of these should be primitives in the doc also

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+
+    @Override
+    Tuple2<BitSet<T>, BitSet<T>> span(Predicate<? super T> predicate);
+
+    @Override
+    default BitSet<T> tail() {
+        if (isEmpty()) {
+            throw new UnsupportedOperationException("tail of empty BitSet");
+        } else {
+            return remove(head());
```
paplorinc: `drop(1)` ?

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+    BitSet<T> take(long n);
+
+    @Override
+    BitSet<T> takeRight(long n);
+
+    @Override
+    default BitSet<T> takeUntil(Predicate<? super T> predicate) {
+        Objects.requireNonNull(predicate, "predicate is null");
+        final BitSet<T> result = takeWhile(predicate.negate());
+        return (result.length() == length()) ? this : result;
```
paplorinc: shouldn't `takeWhile` take care of this already?

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+    }
+
+    @Override
+    BitSet<T> takeWhile(Predicate<? super T> predicate);
+
+    @Override
+    default java.util.SortedSet<T> toJavaSet() {
+        return toJavaSet(() -> new java.util.TreeSet<>(comparator()));
```
paplorinc: hmmm, maybe we should introduce a `toBitSet` to `Value` also

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+    // TODO
+    @Override
+    default <T1, T2> Tuple2<TreeSet<T1>, TreeSet<T2>> unzip(
+            Function<? super T, Tuple2<? extends T1, ? extends T2>> unzipper) {
+        Objects.requireNonNull(unzipper, "unzipper is null");
+        return iterator().unzip(unzipper).map(i1 -> TreeSet.ofAll(naturalComparator(), i1),
+                i2 -> TreeSet.ofAll(naturalComparator(), i2));
+    }
+
+    // TODO
```
paplorinc: could you please explain what these `TODO`s mean?

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+        BitSet<T> createFromAll(Iterable<? extends T> values) {
+            return createEmpty().addAll(values);
+        }
+
+        BitSet<T> fromBitMaskNoCopy(long[] elements) {
+            final int len = elements.length;
+            if (len == 0) {
+                return createEmpty();
+            }
+            if (len == 1) {
```
paplorinc: minor: could you please `else if` mutually exclusive branches

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+                return new BitSet1<>(fromInt, toInt, elements[0]);
+            }
+            if (len == 2) {
+                return new BitSet2<>(fromInt, toInt, elements[0], elements[1]);
+            }
+            return new BitSetN<>(fromInt, toInt, elements);
+        }
+
+        private void setElement(long[] words, int element) {
+            final int index = element >> ADDRESS_BITS_PER_WORD;
```
paplorinc: are we using the sign bit also for storage?
Do we need a `>>` or `>>>`?
Would this change if later we decide that negative indices mean indexing from the end somehow (not sure how it could be done in this case, though)?

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+            words[index] |= (1L << element);
+        }
+
+        private void unsetElement(long[] words, int element) {
```
paplorinc: * `unset` is `clear` in `Java`, is it different in `Scala`?
* could we have a toggle also (via `xor`)?

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+                final long[] copy = copyExpand(1 + (element >> ADDRESS_BITS_PER_WORD));
+                setElement(copy, element);
+                return fromBitMaskNoCopy(copy);
+            }
+        }
+
+        @Override
+        public BitSet<T> distinctBy(Comparator<? super T> comparator) {
+            Objects.requireNonNull(comparator, "comparator is null");
+            return createFromAll(iterator().distinctBy(comparator));
```
paplorinc: could we return the original in case of equality? Or it's unreasonable in this case?

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+        @Override
+        public BitSet<T> takeWhile(Predicate<? super T> predicate) {
+            Objects.requireNonNull(predicate, "predicate is null");
+            final BitSet<T> result = createFromAll(iterator().takeWhile(predicate));
+            return (result.length() == length()) ? this : result;
+        }
+
+        @Override
+        @SuppressWarnings("unchecked")
+        public BitSet<T> addAll(Iterable<? extends T> elements) {
```
paplorinc: nice, it's linear in `elements`, right :)?

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+        }
+
+        @Override
+        public BitSet<T> dropRight(long n) {
+            if (n <= 0) {
+                return this;
+            } else if (n >= length()) {
+                return createEmpty();
+            } else {
+                return createFromAll(iterator().dropRight(n));
```
paplorinc: this may be slow, as I think a `Queue` is backing the `iterator()`

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+        }
+
+        @Override
+        public BitSet<T> intersect(Set<? extends T> elements) {
+            Objects.requireNonNull(elements, "elements is null");
+            if (isEmpty() || elements.isEmpty()) {
+                return createEmpty();
+            } else {
+                int size = size();
+                if (size <= elements.size()) {
```
paplorinc: nice :)

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+                setElement(copy, element);
+            });
+            return fromBitMaskNoCopy(copy);
+        }
+
+        @Override
+        public boolean contains(T t) {
+            final int element = toInt.apply(t);
+            if (element < 0) {
+                throw new IllegalArgumentException("bitset element must be >= 0");
```
paplorinc: if we implement that offset we mentioned, we could even map negative elements here with a negative offset :)

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+            }
+            final int index = element >> ADDRESS_BITS_PER_WORD;
+            return (getWord(index) & (1L << element)) != 0;
```
paplorinc: could we extract all the `1L << element` to something like `private int indexOf(element)`?

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+        }
+
+        @Override
+        public BitSet<T> init() {
+            if (isEmpty()) {
+                throw new UnsupportedOperationException("init of empty TreeSet");
+            } else {
+                final long last = getWord(getWordsNum() - 1);
```
paplorinc: :+1: for not using `dropRight` here

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+                final int element = BITS_PER_WORD * (getWordsNum() - 1) + BITS_PER_WORD - Long.numberOfLeadingZeros(last) - 1;
+                return remove(fromInt.apply(element));
+            }
+        }
+
+        @Override
+        public Iterator<T> iterator() {
+            return Stream.range(0, getWordsNum() << ADDRESS_BITS_PER_WORD)
```
paplorinc: hmmm, this seams linear to the number of possible elements, and not to the existing ones.

Since this is an essential method, could we rather return a custom `Iterator` that uses `Integer.highestOneBit` to find the next bit in constant time for a block and jump through unset blocks the same way via `values[i] != 0` or something? It would still be linear, but should be a lot faster :/

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+                    .filter(i -> (getWord(i >> ADDRESS_BITS_PER_WORD) & (1L << i)) != 0)
+                    .map(fromInt)
+                    .iterator();
+        }
+
+        @Override
+        public int length() {
+            int len = 0;
```
paplorinc: I think this should rather be stored, it's the way we do it elsewhere and should be possible to do it in constant time :)

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+                return this;
+            }
+        }
+
+        @Override
+        @SuppressWarnings("unchecked")
+        public BitSet<T> removeAll(Iterable<? extends T> elements) {
+            final Stream<Integer> source = Stream.ofAll(elements).map(toInt);
+            final long[] copy = copyExpand(getWordsNum());
+            source.forEach(element -> {
```
paplorinc: minor: I think enhanced for loops are a bit cheaper (no lambda) and are just as readable

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+        }
+
+        @Override
+        public String toString() {
+            return mkString(stringPrefix() + "(", ", ", ")");
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (o == this) {
```
paplorinc: minor: can be a one-liner via:

```Java
return o == this || o instanceof BitSet && Collections.equals(this, (Iterable) o);
```

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+        }
+
+        @Override
+        int getWordsNum() {
+            return 1;
+        }
+
+        @Override
+        long[] copyExpand(int wordsNum) {
+            if(wordsNum < 1) {
```
paplorinc: Not sure if it's more readable or not, just an alternative one-liner instead of 899-902:
```java
long[] arr = new long[Math.max(1, wordsNum)];
```

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+            arr[0] = elements;
+            return arr;
+        }
+
+        @Override
+        long getWord(int index) {
+            if (index == 0) {
+                return elements;
+            } else {
+                return 0L;
```
paplorinc: what does this branch mean exactly, seems weird for me :/

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+            }
+        }
+
+        @Override
+        public T head() {
+            if (elements != 0) {
```
paplorinc: could you please invert the if, usually the guard is in the first branch throughout the codebase :)

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+        int getWordsNum() {
+            return 2;
+        }
+
+        @Override
+        long[] copyExpand(int wordsNum) {
+            if(wordsNum < 2) {
+                wordsNum = 2;
+            }
+            long[] arr = new long[wordsNum];
```
paplorinc: same, 
```java
long[] arr = new long[Math.max(1, wordsNum)];
```

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+            arr[0] = elements1;
+            arr[1] = elements2;
+            return arr;
+        }
+
+        @Override
+        long getWord(int index) {
+            if (index == 0) {
+                return elements1;
+            } else {
```
paplorinc: merge `else if`?

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+        int getWordsNum() {
+            return elements.length;
+        }
+
+        @Override
+        long[] copyExpand(int wordsNum) {
+            if(wordsNum < elements.length) {
+                wordsNum = elements.length;
+            }
+            long[] arr = new long[wordsNum];
```
paplorinc: same :)

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+            System.arraycopy(elements, 0, arr, 0, elements.length);
+            return arr;
+        }
+
+        @Override
+        long getWord(int index) {
+            if (index < elements.length) {
+                return elements[index];
+            } else {
+                return 0L;
```
paplorinc: same

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+        }
+
+        @Override
+        public T head() {
+            int offset = 0;
+            for (int i = 0; i < getWordsNum(); i++) {
+                if(elements[i] != 0) {
+                    return fromInt.apply(offset + Long.numberOfTrailingZeros(elements[i]));
+                }
+                offset += BITS_PER_WORD;
```
paplorinc: please else if this line to make it obvious that we have multiple returns

paplorinc: please else if this line to make it obvious that we have multiple returns

javaslang/src/test/java/javaslang/collection/BitSetTest.java
```diff
@@ -0,0 +1,389 @@
...
+                return fromIntMap.get(i);
+            } else {
+                throw new RuntimeException();
+            }
+        }
+
+        synchronized Integer toInt(T value) {
+            Integer i = toIntMap.get(value);
+            if (i == null) {
+                i = nextValue++;
```
paplorinc: could we use an `AtomicLong` instead of a synchronized test class?

javaslang/src/test/java/javaslang/collection/BitSetTest.java
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    @Override
+    protected <T> BitSet<T> empty() {
+        return this.<T>bsBuilder().empty();
+    }
+
+    @Override
+    protected boolean emptyShouldBeSingleton() {
+        return false;
```
paplorinc: `true`?

javaslang/src/test/java/javaslang/collection/BitSetTest.java
```diff
@@ -0,0 +1,389 @@
...
+            return this.<Long>bsBuilder().ofAll(Iterator.rangeClosedBy(from, toInclusive, step));
+        } else {
+            return BitSet.rangeClosedBy(from, toInclusive, step);
+        }
+    }
+
+    // BitSet specific
+
+    @Test
+    public void testBitSet1() {
```
paplorinc: :+1:

javaslang/src/test/java/javaslang/collection/BitSetTest.java
```diff
@@ -0,0 +1,389 @@
...
+        assertThat(bs.head()).isEqualTo(700);
+        assertThat(bs.contains(2)).isFalse();
+
+        bs = bs.remove(700);
+        assertThat(bs.isEmpty()).isTrue();
+    }
+
+    @Test
+    public void testFactories() {
+        assertThat(BitSet.of(7).contains(7)).isTrue();     // BitSet1
```
paplorinc: * < 64
* < 2*64
* >= 2*64

? we may include it in the comments or the values to make it obvious why the numbers were chosen

javaslang/src/test/java/javaslang/collection/BitSetTest.java
```diff
@@ -0,0 +1,389 @@
...
+        assertThat(BitSet.of(77).contains(77)).isTrue();   // BitSet2
+        assertThat(BitSet.of(777).contains(777)).isTrue(); // BitSetN
+    }
+
+    @Test
+    public void testEnums() {
+        BitSet<E> bs = BitSet.withEnum(E.class).empty();
+        bs = bs.add(E.V2);
```
paplorinc: niiice :dancer:

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+            }
+            throw new NoSuchElementException("head of empty BitSet");
+        }
+    }
+
+    class BitSetN<T> extends AbstractBitSet<T> {
+
+        private static final long serialVersionUID = 1L;
+
+        private final long[] elements;
```
paplorinc: It may be late for me to suggest it, but could we use `Array<T>` here instead? It would take care of all the copying and it even has an offset and end now, many operations would be constant time for free :/

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+public interface BitSet<T> extends SortedSet<T> {
+
+    long serialVersionUID = 1L;
+
+    class Builder<T> {
+
+        final static Builder<Integer> DEFAULT = new Builder<>(i -> i, i -> i);
+
+        final Function1<Integer, T> fromInt;
+        final Function1<T, Integer> toInt;
```
danieldietrich: If there is no special reason to use `Function1` then please use `Function` instead. In general it is a good practice because it widens the applicability (`Function1` is a `Function`).

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+
+        Builder(Function1<Integer, T> fromInt, Function1<T, Integer> toInt) {
```
danieldietrich: dito

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+            return empty().addAll(Collections.tabulate(n, f));
+        }
+
+        public BitSet<T> fill(int n, Supplier<? extends T> s) {
+            Objects.requireNonNull(s, "s is null");
+            return empty().addAll(Collections.fill(n, s));
+        }
+    }
+
+    static <T> Builder<T> withRelations(Function1<Integer, T> fromInt, Function1<T, Integer> toInt) {
```
danieldietrich: dito ... (and all other occurrences)

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+            this.fromInt = fromInt;
+            this.toInt = toInt;
+        }
+
+        public Collector<T, ArrayList<T>, BitSet<T>> collector() {
+            final BinaryOperator<ArrayList<T>> combiner = (left, right) -> {
+                left.addAll(right);
+                return left;
+            };
+            return Collector.of(ArrayList::new, ArrayList::add, combiner, this::ofAll);
```
danieldietrich: Maybe we can omit the `ArrayList` because `this::ofAll` is the same as `empty().addAll(...)`:

```java
return Collector.of(this::empty, BitSet::add, ?BitSet::addAll?, this::ofAll)
```

For other collections we have special concat operations that can be used instead of `addAll` I think. E.g. `merge`, `union`, etc. Can we do s.th. similar here for the combiner, too (see above)?

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+     * @param from        the first number
+     * @param toInclusive the last number
+     * @param step        the step
+     * @return a range of int values as specified or the empty range if<br>
+     * {@code from > toInclusive} and {@code step > 0} or<br>
+     * {@code from < toInclusive} and {@code step < 0}
+     * @throws IllegalArgumentException if {@code step} is zero
+     */
+    static BitSet<Integer> rangeClosedBy(int from, int toInclusive, int step) {
+        return BitSet.ofAll(Iterator.rangeClosedBy(from, toInclusive, step));
```
danieldietrich: With #1295 we will change `Iterator`, it will not be `Iterable` any more, so we cannot use `BitSet.ofAll(Iterator)` in a future version.

With #1332 we will still be able to convert an `Iterator` to an arbitrary type, especially `BitSet`. But @paplorinc also suggested to add the conversion method `Value.toBitSet()`.

Currently I'm not clear how a conversion method `toBitSet` deals with the currently Builder (`withXxx`). Because of type erasure we are not aware of the component type `T`. I guess we have to provide the type via `toBitSet(Class<T>)`.

javaslang/src/main/java/javaslang/collection/BitSet.java
```diff
@@ -0,0 +1,1024 @@
...
+    default <T1, T2> Tuple2<TreeSet<T1>, TreeSet<T2>> unzip(
+            Function<? super T, Tuple2<? extends T1, ? extends T2>> unzipper) {
+        Objects.requireNonNull(unzipper, "unzipper is null");
+        return iterator().unzip(unzipper).map(i1 -> TreeSet.ofAll(naturalComparator(), i1),
+                i2 -> TreeSet.ofAll(naturalComparator(), i2));
+    }
+
+    // TODO
+    @Override
+    default <T1, T2, T3> Tuple3<TreeSet<T1>, TreeSet<T2>, TreeSet<T3>> unzip3(
```
danieldietrich: I've seen the TODOs. We need to return `Tuple3<BitSet<T1>, BitSet<T2>, BitSet<T3>>`, same applies for other methods.

danieldietrich:
Nice work so far!

javaslang/src/main/java/javaslang/collection/BitSet.java
ruslansennov: you're completely right, thank you

javaslang/src/main/java/javaslang/collection/BitSet.java
ruslansennov: I forget to remove it and I forget to make this comment:
We know nothing about types &lt;T1, T2, T3&gt; so we can't return tuple of BitSets because we need two conversion functions of every unknown types.

javaslang/src/main/java/javaslang/collection/BitSet.java
ruslansennov: see comment [above](https://github.com/javaslang/javaslang/pull/1334/files#r63306886)

ruslansennov:
Hi guys, thanks for comments, I need one or two days to answer/reject/fix them :smile:

javaslang/src/main/java/javaslang/collection/BitSet.java
ruslansennov: Both <code>fromInt</code> and <code>toInt</code> fields of <code>AbstractBitSet</code> should be serializable. For example, you can see that <code>BitSetTest.Mapper</code> is serializable class. If we change signatures from <code>Function1</code> to <code>Function</code> these tests (I'll add them to BitSetTest) will fail
```java
    @Test
    public void shouldSerializeDeserializeNativeBitSet() {
        final Object actual = deserialize(serialize(BitSet.of(1, 2, 3)));
        final Object expected = BitSet.of(1, 2, 3);
    }

    @Test
    public void shouldSerializeDeserializeEnumBitSet() {
        final Object actual = deserialize(serialize(BitSet.withEnum(E.class).of(E.V1, E.V2)));
        final Object expected = BitSet.withEnum(E.class).of(E.V1, E.V2);
    }
```
Or we can introduce new <code>SerializableFunction</code> interface but I'm not sure this is good way

javaslang/src/test/java/javaslang/collection/BitSetTest.java
```diff
@@ -0,0 +1,389 @@
...
+
+    private enum E {
+        V1, V2, V3
+    }
+
+    private static class Mapper<T> implements Serializable {
+
+        private static final long serialVersionUID = 1L;
+
+        private final java.util.Map<Integer, T> fromIntMap = new java.util.HashMap<>();
```
paplorinc: could we use our own datastructures here instead of the mutable ones?

javaslang/src/test/java/javaslang/collection/BitSetTest.java
```diff
@@ -0,0 +1,389 @@
...
+                    final Tuple2<?, ?> t1 = (Tuple2<?, ?>) actual;
+                    final Tuple2<?, ?> t2 = (Tuple2<?, ?>) expected;
+                    assertThat((Iterable<?>) t1._1).isEqualTo(t2._1);
+                    assertThat((Iterable<?>) t1._2).isEqualTo(t2._2);
+                    return this;
+                } else if (actual instanceof Tuple3) {
+                    final Tuple3<?, ?, ?> t1 = (Tuple3<?, ?, ?>) actual;
+                    final Tuple3<?, ?, ?> t2 = (Tuple3<?, ?, ?>) expected;
+                    assertThat((Iterable<?>) t1._1).isEqualTo(t2._1);
+                    assertThat((Iterable<?>) t1._2).isEqualTo(t2._2);
```
paplorinc: are you deliberately ignoring the 3rd parameter?

javaslang/src/test/java/javaslang/collection/BitSetTest.java
```diff
@@ -0,0 +1,389 @@
...
+    @Override
+    protected <T> ObjectAssert<T> assertThat(T actual) {
+        return new ObjectAssert<T>(actual) {
+            @Override
+            public ObjectAssert<T> isEqualTo(Object expected) {
+                if (actual instanceof Tuple2) {
+                    final Tuple2<?, ?> t1 = (Tuple2<?, ?>) actual;
+                    final Tuple2<?, ?> t2 = (Tuple2<?, ?>) expected;
+                    assertThat((Iterable<?>) t1._1).isEqualTo(t2._1);
+                    assertThat((Iterable<?>) t1._2).isEqualTo(t2._2);
```
paplorinc: isn't this simply `actual.equals(expected)`?

```java
    @Override
        if (o == this) {
            return true;
        } else if (!(o instanceof Tuple2)) {
            return false;
        } else {
            final Tuple2<?, ?> that = (Tuple2<?, ?>) o;
            return Objects.equals(this._1, that._1)
                  && Objects.equals(this._2, that._2);
        }
    }
```

javaslang/src/test/java/javaslang/collection/BitSetTest.java
```diff
@@ -0,0 +1,389 @@
...
+    protected BitSet<Character> rangeBy(char from, char toExclusive, int step) {
+        return BitSet.rangeBy(from, toExclusive, step);
+    }
+
+    @Override
+    protected BitSet<Double> rangeBy(double from, double toExclusive, double step) {
+        return this.<Double>bsBuilder().ofAll(Iterator.rangeBy(from, toExclusive, step));
+    }
+
+    private static boolean badRange(int a, int b) {
```
paplorinc: could you please prefix it with `is`, as I thought quickly that this method will generate illegal ranges :)

javaslang/src/test/java/javaslang/collection/BitSetTest.java
```diff
@@ -0,0 +1,389 @@
...
+        return a < 0 || b < 0 || a > MAX_BIT || b > MAX_BIT;
+    }
+
+    private static boolean badRange(long a, long b) {
+        return a < 0 || b < 0 || a > MAX_BIT || b > MAX_BIT;
```
paplorinc: alternative:
```java
return Math.min(a, b) < 0 || Math.max(a, b) > MAX_BIT;
```

javaslang/src/test/java/javaslang/collection/BitSetTest.java
ruslansennov: No, it's usual behavior, but we need to to check that Tuple2&lt;TreeSet,TreeSet&gt; **corresponds** to Tuple2&lt;BitSet,BitSet&gt; i.e. has same elements set. It's used when we check result of methods that returns something else than <code>BitSet</code>, for example <code>unzip()</code>

paplorinc: You mean that you want to see exactly whether it was `_1` or `_2` that failed and `equals` would only show the end result without details?

ruslansennov: No, I mean that <code>equals</code> fails if objects are instances of different classes, but we should accept equality of BitSet and TreeSet. So we should use overridden <code>IterableAssert&lt;T&gt; assertThat(Iterable&lt;T&gt; actual)</code> (see above)

paplorinc:
In order to increase confidence in all those state changes, could we add a property based test that validates it against Java's BitSet after each step? Something similar to https://github.com/javaslang/javaslang/blob/master/javaslang/src/test/java/javaslang/collection/PriorityQueueTest.java#L226 (or better)

Also, if you think it is appropriate, please consider adding benchmarks for your important methods (depends on https://github.com/javaslang/javaslang/pull/1335)

javaslang/src/main/java/javaslang/collection/BitSet.java
ruslansennov: No, I would leave this as is. <code>Collector::accumulator()</code> method should be fast, because it called with each element of external collection. <code>ArrayList::add</code> is fast enough, but <code>BitSet::add</code> can be very slow for big length, its complexity is O(n).

javaslang/src/main/java/javaslang/collection/BitSet.java
ruslansennov: > With #1295 we will change Iterator, it will not be Iterable any more, so we cannot use BitSet.ofAll(Iterator) in a future version.

I see. But let's do it in future (milestone 3.0.0 I guess) and in own separate commit, m? :smile:

javaslang/src/main/java/javaslang/collection/BitSet.java
ruslansennov: > Would this change if later we decide that negative indices mean indexing from the end

I don't know, this is not the near future :smile:

javaslang/src/main/java/javaslang/collection/BitSet.java
ruslansennov: yes, thank you for pointing this in BitSet's skeleton :smile:

javaslang/src/main/java/javaslang/collection/BitSet.java
ruslansennov: Sorry, but I don't see any reason to wrap <code>long[]</code> to <code>Array&lt;Long&gt;</code> for now. Maybe in future, when (and if) <code>BitSet.offset</code> will be implemented... But now there is exactly one case when <code>elements</code> can be changed: the method <code>copyExpand()</code> which called just before changing BitSet (with the obligatory copying)

javaslang/src/test/java/javaslang/collection/BitSetTest.java
ruslansennov: thanx :+1:

javaslang/src/test/java/javaslang/collection/BitSetTest.java
ruslansennov: No. BitSet's empty instance depends from conversion functions

javaslang/src/main/java/javaslang/collection/BitSet.java
paplorinc: ... and the `shrink` method, but that still might not be enough reason. You know the details, I just have some ideas :).
The advantage of `Array` would be that you could `head`, `tail`, `drop`, `dropRight`, `take` etc. in constant time, instead of linear, as it is currently.

javaslang/src/test/java/javaslang/collection/BitSetTest.java
paplorinc: because of the type, i.e `empty character bitset` vs `empty enum bitset`?

ruslansennov:
> In order to increase confidence in all those state changes, could we add a property based test that validates it against Java's BitSet after each step?

Yes, I'll do it in separate PR. Also benchmarks and Performance table

javaslang/src/test/java/javaslang/collection/BitSetTest.java
ruslansennov: yep

javaslang/src/main/java/javaslang/collection/BitSet.java
ruslansennov: > The advantage of Array would be that you could head, tail, drop, dropRight, take etc. in constant time, instead of linear, as it is currently.

So sad, but no, <code>Array&lt;Long&gt</code> can't help us. For example, <code>BitSet.of(1, 3, 80, 81, ...).tail()</code> requires full-copy of it's content anyway. The only good case if we have exactly one 1-bit in first non-empty word...

paplorinc: ... or we could express the offset in bits, instead of bytes :)
But that could of course be done in a separate commit, as we need a good feature coverage to pull that of without errors :)

javaslang/src/main/java/javaslang/collection/BitSet.java
danieldietrich: Yep, you're right. I'm not sure if TreeSet is the right type. Why not allow duplicates?

Maybe we should return `Tuple3<Traversable<T1>, Traversable<T2>, Traversable<T3>>` because it is the most liberal solution. Under the hood we could use `List`. I expect a `BitSet` to contain not many elements in general...

danieldietrich:
It would be nice if we have benchmarks for all methods of the performance table - just a thought. (midterm)

javaslang/src/main/java/javaslang/collection/BitSet.java
danieldietrich: An additional `SerializableFunction` would be too much. It is possible - please cast the method reference in the `withXxx`methods like this:

```java
SAM s1 = (SAM & Serializable) Test::bar;
```

Then we can use `Function`.

**Example (TreeSet):**

```java
public static <T extends Comparable<? super T>> Collector<T, ArrayList<T>, TreeSet<T>> collector() {
    return collector((Comparator<? super T> & Serializable) T::compareTo);
}
```

javaslang/src/main/java/javaslang/collection/BitSet.java
danieldietrich: ok, thx

javaslang/src/main/java/javaslang/collection/BitSet.java
danieldietrich: Yep

javaslang/src/main/java/javaslang/collection/BitSet.java
ruslansennov: This method should override corresponding method in SortedSet:
```java
    @Override
    <T1, T2, T3> Tuple3<? extends SortedSet<T1>, ? extends SortedSet<T2>, ? extends SortedSet<T3>> 
    unzip3(Function<? super T, Tuple3<? extends T1, ? extends T2, ? extends T3>> unzipper);
```
We can't use <code>BitSet</code> here, so I decided it will be TreeSet

javaslang/src/main/java/javaslang/collection/BitSet.java
ruslansennov: done

javaslang/src/main/java/javaslang/collection/BitSet.java
danieldietrich: Yes, you are right. I tested a bit using the Scala REPL. The Scala language architects decided for each Set impl to return a tuple of Sets. We should do the same. Because it breaks backward compatibility, I deferred the change to 3.0.0. See #1342.

paplorinc:
It's progressing really nicely, @danieldietrich, @ruslansennov, what's the state of this PR? :)

ruslansennov:
LGTM :smile:

danieldietrich:
LveryGTM :-)

PR #1331, opened 2016-05-14
Benchmark

eduardmanas:
I have created a `BenchmarkPerformanceReporter` that removes sample outliers.

For now I have only made changes to `ListBenchmark` until we decide whether we want to go ahead with it.

These are the main changes:
* I have configured the report to remove 30% of bottom outliers and 10% of top outliers. The idea is that we want to remove as many samples impacted by GC as possible.
* Changed `time=500, timeUnit=TimeUnit.MILLISECONDS`. The rationale is that we want to increase the number of samples to improve the confidence interval of the T-distribution.  Also, the more samples, the more effective the outlier filtering.
* Changed `WarmUp=10 iterations` (5 secs), and `Measurement=40 iterations` (20 secs). Thanks to the outlier removal, we could reduce number of measurement iterations to 20 during development without impacting `Error %` much (`Error %` goes up approx from '0.8%' to '2.0%')
* Set `shouldDoGC(true)`, as we want to force a GC before the iteration to reduce chances of GC kicking in during the tests

**Feature:** https://github.com/javaslang/javaslang/issues/1328

PR #1335, opened 2016-05-15
Remove BenchmarkResultAggregator

eduardmanas:
Tidy up after https://github.com/javaslang/javaslang/pull/1331

I have created a class `JmhRunner` to centrilise test set-up and ensure all tests run with optional parameters.  

I have added two methods to `JmhRunner`:
* `devRunAndReport` - quick execution (warm up for 2.5 secs, measure for 5 secs)
* `runAndReport` - much longer but more reliable execution (warm up for 5 secs, measure for 20 secs)

The rationale is that we could use `devRunAndReport` for quick feedback during development, but we should also have `runAndReport` when we want more reliable results.

Also, I have moved the existing `Benchmark`s to a `collections` package.


eduardmanas:
I also removed `BenchmarkResultAggregator` as it has been superseded by `BenchmarkPerformanceReporter`

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
```diff
@@ -0,0 +1,66 @@
...
+                .shouldDoGC(true)
+                .shouldFailOnError(true)
+                .mode(Mode.Throughput)
+                .timeUnit(TimeUnit.SECONDS)
+                .warmupTime(TimeValue.milliseconds(500))
+                .warmupIterations(warmupIterations)
+                .measurementTime(TimeValue.milliseconds(500))
+                .measurementIterations(measurementIterations)
+                .forks(1)
+                .jvmArgsAppend("-XX:+UseG1GC", "-Xss100m", "-Xms1g", "-Xmx1g", assertions.vmArg)
```
paplorinc: Please consider something like: `@Fork(value = 1, jvmArgsAppend = { "-server", "-XX:+UseG1GC", "-Xss100m", "-Xms4g", "-Xmx4g", "-XX:+PrintGC", "-XX:MaxGCPauseMillis=1000", "-XX:+UnlockExperimentalVMOptions", "-XX:G1NewSizePercent=100", "-XX:G1MaxNewSizePercent=100", "-disableassertions" })`

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
```diff
@@ -0,0 +1,66 @@
...
+            return new Runner(opts).run();
+        }
+        catch (Exception e) {
+            throw new RuntimeException(e.getMessage(), e);
+        }
+    }
+
+    private static enum Assertions {
+        Enable("-enableassertions"),
+        Disable("-disableassertions")
```
paplorinc: I strongly think that assertions should always be disabled, we certainly don't want to test the speed of them in production code accidentally.

javaslang-benchmark/src/test/java/javaslang/benchmark/collection/ArrayBenchmark.java
```diff
@@ -1,33 +1,24 @@
...
-                .include(ArrayBenchmark.class.getSimpleName())
-                .shouldDoGC(false)
-                .shouldFailOnError(true)
-                .build();
-
-        final Collection<RunResult> results = new Runner(opts).run();
-        displayRatios(results, "^.*slang.*$");
+
+    public static void main(String... args) { /* main is more reliable than a test */
+        JmhRunner.runAndReport(ArrayBenchmark.class);
```
paplorinc: Nice abstraction, good job! :)

javaslang-benchmark/src/test/java/javaslang/benchmark/collection/ListBenchmark.java
```diff
@@ -1,50 +1,28 @@
...
 
     @State(Scope.Benchmark)
-    @BenchmarkMode(Mode.Throughput)
-    @OutputTimeUnit(TimeUnit.SECONDS)
-    @Warmup(iterations = 10, time = 500, timeUnit = TimeUnit.MILLISECONDS)
-    @Measurement(iterations = 40, time = 500, timeUnit = TimeUnit.MILLISECONDS)
-    @Fork(value = 1, jvmArgsAppend = { "-XX:+UseG1GC", "-Xss100m", "-Xms1g", "-Xmx1g", "-disableassertions" }) /* set fork to 0 if you want to debug */
     public static class Base {
-        @Param({ "10", "100", "1000", "10000" })
+        @Param({ "10", "100", })
```
paplorinc: was this done on purpose?

eduardmanas: good catch!

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
eduardmanas: The idea here was to have two types of benchmark executions: the more reliable version (`runAndReport`), and the less reliable but quicker version to use during development (`devRunAndReport`).

I thought that during development we would want to run with assertions on, and given we are not that bothered about reliability, it wouldn't be a major issue.

If you prefer, we could create a third method to make it explicit: `devRunAndReportWithAssertions`. Or if you prefer, we could also shorten the names of all three methods by removing the `AndReport`.

eduardmanas: And by the way, is there any reason why we use `assertEquals` instead of assertions in `ListBenchmark`?

Also, I think that asserting on `size()` might be very expensive.

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
eduardmanas: I understand the rationale of changing NewGen to 100% of the heap.  I must admit, however, it makes me a bit uneasy, as it is an artificial setup.  However, I agree that in theory by removing all GCs we should get stable results.

I did a quick test. I run the `devRunAndReport` on `ListBenchmark` with container size = 10000 and a 4Gb heap. The first run with default NewGen size, and the second with NewGen size=100%.  No GC was triggered in the second run. 

However, to my surprise, there were still considerable swings from iteration to iteration, so not sure what to make of these results. 

First run, normal NewGen:
```
Target         Operation   Impl                         Params  Count            Score  ±     Error  
ListBenchmark  Iterate     slang_persistent              10000      6       32,704.013  ±     4.23%  
ListBenchmark  Iterate     scala_persistent              10000      6       35,181.237  ±     1.87%  
ListBenchmark  Iterate     scala_mutable                 10000      6       41,459.476  ±     2.74%  
ListBenchmark  Iterate     pcollections_persistent       10000      6       30,591.533  ±     7.57%  
ListBenchmark  Iterate     java_mutable_linked           10000      6       28,613.690  ±    12.88%  
ListBenchmark  Iterate     java_mutable                  10000      6       90,556.997  ±     6.98%  
ListBenchmark  Iterate     fjava_persistent              10000      6       35,779.948  ±     5.46%  
ListBenchmark  AddAll      slang_persistent              10000      6       26,124.064  ±    20.22%  
ListBenchmark  AddAll      scala_persistent              10000      6       20,810.554  ±     9.16%  
ListBenchmark  AddAll      scala_mutable                 10000      6       20,356.731  ±     1.28%  
ListBenchmark  AddAll      pcollections_persistent       10000      6       22,217.533  ±    36.81%  
ListBenchmark  AddAll      java_mutable_linked           10000      6       21,513.086  ±    20.40%  
ListBenchmark  AddAll      java_mutable                  10000      6       23,199.166  ±    52.08%  
ListBenchmark  AddAll      fjava_persistent              10000      6       14,222.033  ±     3.23%  
```

Second run, NewGen = 100%
```
Target         Operation   Impl                         Params  Count            Score  ±     Error  
ListBenchmark  Iterate     slang_persistent              10000      6       25,462.352  ±     1.42%  
ListBenchmark  Iterate     scala_persistent              10000      6       25,151.763  ±     2.27%  
ListBenchmark  Iterate     scala_mutable                 10000      6       44,893.597  ±     4.30%  
ListBenchmark  Iterate     pcollections_persistent       10000      6       25,520.437  ±     4.26%  
ListBenchmark  Iterate     java_mutable_linked           10000      6       30,922.490  ±     4.99%  
ListBenchmark  Iterate     java_mutable                  10000      6      112,805.879  ±     4.02%  
ListBenchmark  Iterate     fjava_persistent              10000      6       24,831.025  ±     9.02%  
ListBenchmark  AddAll      slang_persistent              10000      6       26,400.629  ±    17.79%  
ListBenchmark  AddAll      scala_persistent              10000      6       21,334.084  ±     4.55%  
ListBenchmark  AddAll      scala_mutable                 10000      6       19,364.354  ±     9.11%  
ListBenchmark  AddAll      pcollections_persistent       10000      6       23,310.641  ±    45.25%  
ListBenchmark  AddAll      java_mutable_linked           10000      6       22,698.641  ±    31.58%  
ListBenchmark  AddAll      java_mutable                  10000      6       23,148.883  ±    50.05%  
ListBenchmark  AddAll      fjava_persistent              10000      6       14,240.217  ±     2.97%  
```

paplorinc: The iteration swings can be attributed to numerous other factors, context switching, cpu power saving, false sharing etc.

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
paplorinc: Will try to explain more clearly :).
There are expensive asserts in the production code also, e.g.: https://github.com/javaslang/javaslang/blob/master/javaslang/src/main/java/javaslang/collection/PriorityQueue.java#L535. We want to include these when we're testing correctness, but ignore them when we're testing speed :).
Just run the PriorityQueue benchmark WITH assertions, and you'll understand the difference :)

eduardmanas: Do we need asserts in the production code? 

I don't see the need if we have proper junit coverage.

eduardmanas: But I agree, there are several expensive looking assertions in the `PriorityQueueBenchmark`, such as this one:

```
assertEquals(values.values().map(Traversable::size).sum().intValue(), CONTAINER_SIZE)
```

I don't mind assertions in the Benchmark classes, as we don't have junits for them. 

I think, however, we should always use `assert` instead of `assertEquals`, as the latter always gets executed.

eduardmanas:
I have updated the pull request with @paplorinc comments.

I also changed the `JmhRunner` methods to make them shorter (i.e. removed `AndReport` part as it is redundant). I have also added a `JmhRunner.devRunWithAssertions` method to make it clear when assertions are turned on.

Regarding the JVM memory setup, I am still having doubts whether we should configure the JVM with  `NewGen=100%`. Yes, I agree this configuration generates no GC in all of our current benchmarks, which is a good thing! (one less variable to worry about). 

However, I have this little fella in my head that keeps telling me it doesn't "feel natural" (ok, kinda weird reason, I admit). Also, we might run into massive problems if one of our benchmarks uses more than 4Gb memory on each iteration. At the moment the `ListBenchmark` is using up over 2.5Gb for each 500ms iteration on my 3 year old PC. A faster PC might just push through the 4Gb threshold.  Also, I was expecting that by removing the GC we would reduce the swings among iterations considerably. This however doesn't seem to be the case, which is disappointing.

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
paplorinc: Asserts are for internal state validation (white box), while tests are for black box.
Also, asserts can document the code or fail early at the client site, if used illegally, which we should know asap.

paplorinc:
> I have this little fella in my head that keeps telling me it doesn't "feel natural"

de1027627c008e3d5e7fa9e4d523d3b0 1

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
```diff
@@ -0,0 +1,73 @@
...
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.TimeValue;
+
+import java.util.Collection;
+import java.util.concurrent.TimeUnit;
+
+public class JmhRunner {
+    private static final int WARMUP_ITERATIONS = 10;
+    private static final int MEASUREMENT_ITERATIONS = 40;
```
paplorinc: do we need this many, now that we don't have `GC`s during?

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
```diff
@@ -0,0 +1,73 @@
...
+                .timeUnit(TimeUnit.SECONDS)
+                .warmupTime(TimeValue.milliseconds(500))
+                .warmupIterations(warmupIterations)
+                .measurementTime(TimeValue.milliseconds(500))
+                .measurementIterations(measurementIterations)
+                .forks(1)
+                // We are using 4Gb and setting NewGen to 100% to avoid GC during testing.
+                // Any GC during testing will destroy the iteration, which should get ignored as an outlier
+                .jvmArgsAppend("-XX:+UseG1GC", "-Xss100m", "-Xms4g", "-Xmx4g", "-XX:+PrintGC", "-XX:MaxGCPauseMillis=1000", "-XX:+UnlockExperimentalVMOptions", "-XX:G1NewSizePercent=100", "-XX:G1MaxNewSizePercent=100", assertions.vmArg)
+//                .jvmArgsAppend("-XX:+UseG1GC", "-Xss100m", "-Xms4g", "-Xmx4g", "-XX:+PrintGC", assertions.vmArg)
```
paplorinc: is there any reason for leaving the commented out code here?

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.concurrent.TimeUnit;
+
+public class JmhRunner {
+    private static final int WARMUP_ITERATIONS = 10;
+    private static final int MEASUREMENT_ITERATIONS = 40;
+
+    private static final int QUICK_WARMUP_ITERATIONS = 5;
+    private static final int QUICK_MEASUREMENT_ITERATIONS = 10;
+
+    public static void run(Class clazz) {
```
paplorinc: `Class<?> benchmarkClass`?

javaslang-benchmark/src/test/java/javaslang/benchmark/collection/ListBenchmark.java
```diff
@@ -1,50 +1,28 @@
...
 
     @State(Scope.Benchmark)
-    @BenchmarkMode(Mode.Throughput)
-    @OutputTimeUnit(TimeUnit.SECONDS)
-    @Warmup(iterations = 10, time = 500, timeUnit = TimeUnit.MILLISECONDS)
-    @Measurement(iterations = 40, time = 500, timeUnit = TimeUnit.MILLISECONDS)
-    @Fork(value = 1, jvmArgsAppend = { "-XX:+UseG1GC", "-Xss100m", "-Xms1g", "-Xmx1g", "-disableassertions" }) /* set fork to 0 if you want to debug */
     public static class Base {
-        @Param({ "10", "100", "1000", "10000" })
+        @Param({ "10", "100", "1000"})
```
paplorinc: do you think `10000` is unrealistic?

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
eduardmanas: If you want to assert for internal state consistency, I don't think you should rely on `assert`, as it can be disabled thus leaving the internal state compromised.

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
eduardmanas: We need it for proper statistical analysis imo. Even without GC, the swings are still considerable and therefore when running with 10 iterations the margin of error is still high.  Also remember each iteration is just 500ms.

I think it is ok to use `JmhRunner.devRun` during development for quick feedback (10 iterations). 

However, if we are gathering reports for publication or to want to make certain decisions, I think 40 is a good number.

javaslang-benchmark/src/test/java/javaslang/benchmark/collection/ListBenchmark.java
eduardmanas: Ah, good point. I actually wanted to talk about this one. No, 10000 is not unrealistic. I just think we should be consistent across all Benchmarks. 

I would personally use 10, 100, 1000, 10000 and 100000.  But then benchmarks will take forever (as we know). So the question is how many we want to use across the board? and which ones?

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
danieldietrich: @paplorinc, @eduardmanas

> Asserts are for internal state validation (white box), while tests are for black box.

I tend to remove assertions from our production code. I was tempted to do so when developing the RedBlackTree class. In the end I wrote property tests using the javaslang property test framework. The properties I tested checked the RB-Tree invariants, like no subsequent red nodes, max length of a path to ensure the tree is balanced after insert/delete, etc.
Those tests disappeared when I refactored the Javaslang module structure because of cyclic dependencies. However, such tests could move to the `javaslang-test` module or to a separate module that contains property tests for the javaslang classes.

paplorinc: Internal asserts are meant for redundancy checks, i.e. that the stored `size` is the same as the calculated one.
Tests are not supposed to validate internals.

e.g. how could you validate in a test that a tree is always balanced, complete or doesn't contain duplicates: via asserts that would get evaluated all the time, except when speed matters -- in production.

I happened to find a bug in `PQ` that no test caught (the output was the same, but stored in a sub-optimal way), once I introduced invariant assertions.

danieldietrich:
I question the use of assertions in benchmark code at all. We should not debug at all in benchmark code. If we think there is something wrong we should write a unit test or property test for that purpose. When we debug, the information gets lost. When we write a test, Javaslang gets hardened.

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
danieldietrich: > Internal asserts are meant for redundancy checks, i.e. that the stored size is the same as the calculated one.

I understand. That's ok.

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
```diff
@@ -17,39 +17,38 @@
...
         final Options opts = new OptionsBuilder()
-                .include(clazz.getSimpleName())
+                .include(benchmarkClass.getSimpleName())
                 .shouldDoGC(true)
                 .shouldFailOnError(true)
                 .mode(Mode.Throughput)
                 .timeUnit(TimeUnit.SECONDS)
                 .warmupTime(TimeValue.milliseconds(500))
-                .warmupIterations(warmupIterations)
+                .warmupIterations(warmUpIterations)
```
paplorinc: the method name is actually with lowercase, but I'm ok either way :)

danieldietrich: +1 for keeping names consistent

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
```diff
@@ -29,20 +29,20 @@ public static void devRunWithAssertions(Class benchmarkClass) {
         runAndReport(benchmarkClass, QUICK_WARMUP_ITERATIONS, QUICK_MEASUREMENT_ITERATIONS, Assertions.Enable);
     }
 
-    private static void runAndReport(Class benchmarkClass, int warmUpIterations, int measurementIterations, Assertions assertions) {
-        Collection<RunResult> results = run(benchmarkClass, warmUpIterations, measurementIterations, assertions);
+    private static void runAndReport(Class benchmarkClass, int warmupIterations, int measurementIterations, Assertions assertions) {
```
paplorinc: Quick question: do you deliberately ignore the type of the `Class`, i.e. why no `Class<?>`?

eduardmanas: I have added `Class<?>` now. 

I left it out because it is not needed and doesn't give you anything. A different story would have been if we were using `Class<? extends Benchmark>`. However, I concede `Class<?>` is more accurate as `Class<>` is a parameterized class.

javaslang-benchmark/src/test/java/javaslang/benchmark/JmhRunner.java
```diff
@@ -17,24 +17,24 @@
     private static final int QUICK_WARMUP_ITERATIONS = 5;
     private static final int QUICK_MEASUREMENT_ITERATIONS = 10;
 
-    public static void run(Class benchmarkClass) {
+    public static void run(Class<?> benchmarkClass) {
```
paplorinc: :+1:

danieldietrich:
@eduardmanas @paplorinc Are there any open points left? If not I will merge it. Thx!

paplorinc:
I don't have to agree for it to be accepted, though I still keep my original concern about enabling assertions for performance tests.

eduardmanas:
@paplorinc we have three `JmhRunner` methods :
* `run` - 5 secs warmup and 20 secs measurement (default)
* `devRun` - 2.5 secs warmup and 5 secs measurement
* `devRunWithAssertions` - same as `devRun` but running with assertions

Have you seen the latest? Are you still concerned about this?

eduardmanas:
`run` and `devRun` run without assertions

eduardmanas:
@paplorinc Can you answer the above? What is you concern?
@danieldietrich Can we please merge?

danieldietrich:
@eduardmanas yep, we will merge.

Issue #1333, opened 2016-05-15
Add PriorityQueue to Performance Characteristics table


ruslansennov:
[Here](https://github.com/javaslang/javaslang/blob/master/javaslang/src/main/java/javaslang/collection/package-info.java) is a table

ruslansennov:
I'm not sure that this is necessary, because a lot of methods are not implemented in PQ

paplorinc:
Yeah, `Queue` is in the first list (sequential), while `TreeSet` is in the second.
Added it in the first, even though it doesn't fit perfectly.

PR #1336, opened 2016-05-16
Added `PriorityQueue` complexities to package info

paplorinc:
Fixes https://github.com/javaslang/javaslang/issues/1333


danieldietrich:
Great, thanks!
Yes, we have different categories of collections: Seq(uential), Map/Set ... and Tree and PriorityQueue.
For now it is ok to have PriorityQueue in the Seq section but I think we should create a *Traversable* or *Other* section midterm.

PR #1340, opened 2016-05-17
Added Stream.ofAll(java.util.stream.Stream)

### LINKED ISSUE #1338: Lazily convert a java.util.stream.Stream to a javaslang.collection.Stream
Check, if this is possible. Currently we convert it by calling `j.u.s.Stream.collect(Stream.collector())`. The `collector()` method first eagerly builds an `ArrayList` and then a `Stream`. Goal is to access the elements of a `j.u.s.Stream` lazily.

danieldietrich:
Fixes #1338

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Stream.java
+++ b/javaslang/src/main/java/javaslang/collection/Stream.java
@@ -6,7 +6,8 @@
 package javaslang.collection;
 
 import javaslang.*;
-import javaslang.collection.Stream.*;
+import javaslang.collection.Stream.Cons;
+import javaslang.collection.Stream.Empty;
 import javaslang.collection.StreamModule.*;
 import javaslang.control.Option;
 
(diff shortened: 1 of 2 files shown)


javaslang/src/main/java/javaslang/collection/Stream.java
```diff
@@ -6,7 +6,8 @@
 package javaslang.collection;
 
 import javaslang.*;
-import javaslang.collection.Stream.*;
+import javaslang.collection.Stream.Cons;
+import javaslang.collection.Stream.Empty;
```
danieldietrich: (Mhh, we seem to use different formatters - we should unify them. I like the wildcards)

javaslang/src/main/java/javaslang/collection/Stream.java
```diff
@@ -305,7 +306,7 @@ public T next() {
      *
      * @param <T>      Component type of the Stream.
      * @param elements An Iterable of elements.
-     * @return A list containing the given elements in the same order.
+     * @return A Stream containing the given elements in the same order.
```
danieldietrich: fixed javadoc

javaslang/src/main/java/javaslang/collection/Stream.java
```diff
@@ -802,7 +815,7 @@ public T next() {
         }
         final Stream<T> finalStream = stream;
         return stream.isEmpty() ? Stream.empty()
-                                : cons(stream.head(), () -> finalStream.tail().filter(predicate));
+                : cons(stream.head(), () -> finalStream.tail().filter(predicate));
```
danieldietrich: also different formatting rule!?

PR #1352, opened 2016-05-24
implementation for #1332

earlzero:
Implementation collect methods for Value interface.

### DIFF
--- a/javaslang/src/main/java/javaslang/Value.java
+++ b/javaslang/src/main/java/javaslang/Value.java
@@ -431,7 +431,31 @@ public interface Value<T> extends Iterable<T> {
      */
     @Override
     Iterator<T> iterator();
+    
+    /**
+     * Collects the underlying value(s) (if present) using the provided {@code collector}.
+     *
+     * @param collector Collector performing reduction
+     * @return R reduction result
+     */
+    default <R, A> R collect(Collector<? super T,A,R> collector) {
+        return StreamSupport.stream(spliterator(), false).collect(collector);
+    }
 
+    /**
+     * Collects the underlying value(s) (if present) using the given {@code supplier}, {@code accumulator} and
+     * {@code combiner}.
+     *
+     * @param supplier    provide unit value for reduction
+     * @param accumulator perform reduction with unit value
+     * @param combiner    function for combining two values, which must be
+     *                    compatible with the accumulator.
+     * @return R reduction result
+     */
+    default <R> R collect(Supplier<R> supplier, BiConsumer<R,? super T> accumulator, BiConsumer<R,R> combiner) {
+        return StreamSupport.stream(spliterator(), false).collect(supplier, accumulator, combiner);
+    }
+    
     // -- conversion methods
 
     /**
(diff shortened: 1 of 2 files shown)


javaslang/src/main/java/javaslang/Value.java
```diff
@@ -818,6 +819,39 @@ default CharSeq toCharSeq() {
...
+    default <R, A> R collect(Collector<? super T,A,R> collector) {
+		return collector.finisher()
+				.apply(this.collect(collector.supplier(), 
+						collector.accumulator(), null));
+    }
+
+    /**
+     * Collect value using supplier, accumulator and combiner
+     * <p>
+     * Since it is provided for a single value, it do not use combiner
```
ruslansennov: <code>Value</code> is not necessarily a single value, see <code>Value::isSingleValued()</code>

danieldietrich: @earlzero yes, that's right, we need the `combiner`, too. Could you please add it?

danieldietrich:
Hi @earlzero, thank you for your contribution!
I will add some comments regarding formatting to match the Javaslang code base style.

javaslang/src/main/java/javaslang/Value.java
```diff
@@ -818,6 +819,39 @@ default CharSeq toCharSeq() {
...
+     * <p>
+     * Use {@link #collect(Supplier, BiConsumer, BiConsumer)} internally 
+     * 
+     * @param collector Collector performing reduction
+     * @return R result of "reduction" with collector
+     */
+    default <R, A> R collect(Collector<? super T,A,R> collector) {
+		return collector.finisher()
+				.apply(this.collect(collector.supplier(), 
+						collector.accumulator(), null));
```
danieldietrich: please use spaces instead of tabs

javaslang/src/main/java/javaslang/Value.java
```diff
@@ -818,6 +819,39 @@ default CharSeq toCharSeq() {
...
+     * Since it is provided for a single value, it do not use combiner
+     *
+     * @param supplier provide unit value for reduction
+     * @param accumulator perform reduction with unit value
+     * @param combiner is not used 
+     * @return R result of "reduction" with collector
+     */
+
+    default <R> R collect(Supplier<R> supplier, BiConsumer<R,? super T> accumulator, BiConsumer<R,R> combiner) {
+    	R container = supplier.get();
```
danieldietrich: make all local vars `final` if possible

javaslang/src/test/java/javaslang/ValueTest.java
```diff
@@ -18,4 +21,19 @@ public void shouldNarrowValue() {
         final Value<Number> numbers = Value.narrow(doubles);
         assertThat(numbers.get()).isEqualTo(1.0d);
     }
+    
+    @Test
+    public void collectWorkAsExpected() {
+        final Value<Double> doubles = List.of(1.0d);
+        Double result = doubles.collect(Collectors.toList()).get(0);
```
danieldietrich: can be also final

javaslang/src/test/java/javaslang/ValueTest.java
```diff
@@ -18,4 +21,19 @@ public void shouldNarrowValue() {
...
+        assertThat(result).isEqualTo(1.0d);
+    }
+    
+    @Test
+    public void collectWorkAsExpected2() {
+        final Value<Double> doubles = List.of(1.0d);
+        Double result = doubles.collect(() -> new ArrayList<Double>(),
```
danieldietrich: dito

earlzero:
@ruslansennov thanks, I didn't realise that it may hold multple values initially. I've implemented multi-value case.
@danieldietrich Don't mention :). I've fixed issues you mentioned (spaces, final, etc).

danieldietrich:
I've modified your code a bit and used `StreamSupport`, it does only depend on the underlying `splitterator/iterator`. For multi-valued types (like Traversable) we iterate the elements only once (instead of twice in the previous solution).

```java
    /**
     * Collect value using provided collector
     * <p>
     * Use {@link #collect(Supplier, BiConsumer, BiConsumer)} internally
     *
     * @param collector Collector performing reduction
     * @return R result of "reduction" with collector
     */
    default <R, A> R collect(Collector<? super T, A, R> collector) {
        return StreamSupport.stream(spliterator(), false).collect(collector);
    }

    /**
     * Collect value using supplier, accumulator and combiner
     * <p>
     * Since it is provided for a single value, it do not use combiner
     *
     * @param supplier    provide unit value for reduction
     * @param accumulator perform reduction with unit value
     * @param combiner    function for combining two values, which must be
     *                    compatible with the accumulator.
     * @return R result of "reduction"
     */
    default <R> R collect(Supplier<R> supplier, BiConsumer<R, ? super T> accumulator, BiConsumer<R, R> combiner) {
        return StreamSupport.stream(spliterator(), false).collect(supplier, accumulator, combiner);
    }
```

Please use the formatting above, your formatter seems to differ from our IntelliJ IDEA settings ([provided in this repo](https://github.com/javaslang/javaslang/blob/master/.ide/idea-settings.jar)).

danieldietrich:
Please insert the code right behind the line

```java
    // -- conversion methods
```

earlzero:
@danieldietrich Thanks for modification. I do not use IDEA, so I can't import this settings. I've checked for tabs/spaces again and moved everything according to your second comment.

danieldietrich:
Thanks for your effort, I really appreciate it!

Issue #79, opened 2015-01-06
More Stream.in() and Traversable.out() methods


danieldietrich:
```java
Stream<String> Stream.in() // same as Stream.in(System::in, Charset.default()).lines()
Stream<Byte> Stream.in(InputStream in)
Stream<Char> Stream.in(InputStream in, Charset charset)

Stream<String> stream.lines() // needs to match current stream component type and convert elements

Traversable.out(OutputStream) // writes bytes
Traversable.out(OutputStream, Charset) // writes chars
```

**Update:** `Traversable.toInputStream()` and `Traversable.toOutputStream()` could be also useful.

lukaseder:
Ooooh, beautiful ideas! I suspect that passing an `InputStream` will produce a `Stream<Byte>` in your case?

mperry:
@danieldietrich  and @lukaseder 

Should this return the stream within the IO monad to indicate that it is not pure?  Alternately, return a lazy stream, e.g. in FunctionalJava this is a P1<Stream<A>>.  You can then concatenate this stream with another stream without the effects causing problems. The stream you create here won't compose if it is eager on the head with a lazy tail.

mperry:
Another potential problem is what happens if the stream is not fully evaluated.  Does the resource ever close if it is lazy?

danieldietrich:
@lukaseder Thank you :-) Yes, Stream<Byte> is the idea...

lukaseder:
> @mperry: Does the resource ever close if it is lazy?

`java.util.stream.Stream` extends `AutoCloseable`, so the `close()` and resource semantics should already be covered...

danieldietrich:
@mperry:
> The stream you create here won't compose if it is eager on the head with a lazy tail.
I think I had such a problem when iterating locking head calculation from the standard input stream. There is a bugfix for such a situation in the Stream.iterator() method but I currently don't fully understand the core problem you mentioned and will investigate it this evening...

> Another potential problem is what happens if the stream is not fully evaluated. Does the resource ever close if it is lazy?
An infinite input stream will not close, currently... I will look at Runar's post this evening!

mperry:
@lukaseder /cc @danieldietrich  
[...]

PR #1317, opened 2016-05-05
Added support for navigable fields in SortedMap/TreeMap

eduardmanas:
1) Added support for navigable fields in `SortedMap`/`TreeMap`\- Added `descendingIterator()`/`descendingKeySet()` methods
- Added `descendingMap()` method to create a map with reversed order
- Added `subMap()`/`headMap()`/`tailMap()` methods to create a subMap view of this map
- Added `floor`/`ceiling`/`higher`/`lower` methods to navigate the sorted map

2) Refactored `RedBlackTree` to:
- support ascending/descending sub-trees
- support descending iterator
- Changed hierarchy structure:

```
    RedBlackTree  <----- Node  <--- ValueNode
                   \            \-- Empty
                    \
                     \-- AbstractSubTreeView  <----  AscendingSubTreeView
                                               \---  DescendingSubTreeView

```

3) Added `TreeMapBenchmark`
- Measure `TreeMap` and new navigable methods performance versus Java/Scala implementation
- Created `BenchmarkPerformanceReporter` to show nice looking performance comparison reports.

PR #1361, opened 2016-05-30
Stream.in(InputStream) & Stream.in(Reader) #79

ruslansennov:


danieldietrich:
Maybe this is the solution for _dirty_ IO. I wrapped my head around adding `in`-methods while ensuring that resources are closed properly when _combining_ streams.

`java.util.stream.Stream` extends `AutoClosable`. Maybe Javaslang's Stream also needs to implement it when baking IO into it. But then composed Streams (like zipped Streams) need to close resources properly.

If one sub-Stream fails, all other sub-Streams need to close resources, too.

I'm not sure that this impl is the correct way to do it. Will think about it...

javaslang/src/main/java/javaslang/collection/Stream.java
```diff
@@ -418,6 +418,76 @@ public T next() {
...
+     * {@link IOException}'s thrown be the <code>InputStream</code> are wrapped
+     * by {@link UncheckedIOException}'s.
+     */
+    static Stream<Byte> in(InputStream in) {
+        return StreamFactory.create(new AbstractIterator<Byte>() {
+            Byte next = null;
+
+            @Override
+            protected Byte getNext() {
+                Byte result = next;
```
paplorinc: `final`?

javaslang/src/main/java/javaslang/collection/Stream.java
```diff
@@ -418,6 +418,76 @@ public T next() {
...
+            @Override
+            public boolean hasNext() {
+                if (next == null) {
+                    try {
+                        int i = in.read();
+                        if (i >= 0) {
+                            next = (byte) i;
+                        }
+                    } catch (IOException e) {
+                        throw new UncheckedIOException(e);
```
paplorinc: I'm not comfortable with `hasNext` throwing, instead of returning false in this case, since that's why I'm calling it :/

javaslang/src/main/java/javaslang/collection/Stream.java
```diff
@@ -418,6 +418,76 @@ public T next() {
         return Stream.ofAll(Iterator.ofAll(array));
     }
 
+    /**
+     * Wrap a <code>InputStream</code> into a <code>Stream</code> of bytes.
+     * <p>
+     * Client code must close the <code>InputStream</code>. All
+     * {@link IOException}'s thrown be the <code>InputStream</code> are wrapped
```
paplorinc: > All `IOException`s thrown be the `InputStream` are wrapped

not sure what this means

javaslang/src/main/java/javaslang/collection/Stream.java
```diff
@@ -418,6 +418,76 @@ public T next() {
...
+
+            @Override
+            protected Character getNext() {
+                Character result = next;
+                next = null;
+                return result;
+            }
+
+            @Override
+            public boolean hasNext() {
```
paplorinc: In my opinion it would be nicer if the return would simply be in in the `try` and the `throw` in the `catch`, so that it's obvious that there are multiple exits from the method, i.e.:
```java
@Override
public boolean hasNext() {
    try {
        if (next == null) {
            int i = reader.read();
            if (i >= 0) {
                next = (char) i;
            }
            return next != null;
        } catch(IOException e) {
            throw new UncheckedIOException(e);
        }
    }
}
```

danieldietrich:
> danieldietrich added the in progress label 2 days ago

I think this may have to do with waffle.io and the new story board. Not sure. I did not added 'in progress' explicitly.

ruslansennov:
> I think this may have to do with waffle.io and the new story board. Not sure. I did not added 'in progress' explicitly.

Oh! :)

javaslang/src/main/java/javaslang/collection/Stream.java
ruslansennov: @paplorinc thank for your comments, but firstly we should decide something about `AutoClosable` (see Daniel's comment above). Maybe this PR will be closed ;)

danieldietrich: Yes. I've to come up with an idea - it will take some time because I only have some hours in the evening.

danieldietrich:
Couldn't effort the time to look into it, yet. Have to finish #1317 before. But I did not forget it...

ruslansennov:
I'm playing with GWT now. There are problems with match mechanism (isAssignable method). I'll make resume soon.

danieldietrich:
ok, let me know if I can do sth

danieldietrich:
We track the big picture currently in milestone [2.2.0](https://github.com/javaslang/javaslang/milestone/17). However, I don't think there is a simple solution and I'm currently not sure if this feature should be baked into Javaslang.

I will close this PR for now.

Issue #1348, opened 2016-05-23
Optimize `Vector`


paplorinc:
We should optimize `Vector`, e.g. it's `slice`, `subSequence`, `drop`, `dropRight`, `dropWhile`, `dropUntil`, `init`, `tail`, `take`, `takeRight`, `takeUntil`, `takeWhile`, `remove`, `removeAt` should all be made `O(1)` time (some are close already) with greater structural sharing, possibly with an `offset` and `length`, as done in `Array` also.

Notes:
* `reverse` of a single element should also return itself
* `init` can be `dropRight(1)`, `tail` can be `drop(1)`
* `intersperce` should probably use `iterator()`
* equals can be: `return o == this || o instanceof Vector && Collections.equals(this, (Vector<?>) o)`

On a related note, we should create a `JMH` performance benchmark for all important interfaces (`Seq`, `Traversable`, `Map`, `Set`...) with their important methods.
We should be able to tell that way which implementations could be made faster and whether certain optimizations actually result in faster code.

paplorinc:
@danieldietrich, I think this is mislabeled

mvh77:
Is anybody working on [this](https://github.com/javaslang/javaslang/issues/568)?

paplorinc:
@mvh77, I have 2 open PRs for `Vector` currently, one for benchmarks, the other for some optimizations.

PR #1364, opened 2016-05-31
Optimized front and back operations for `Vector`

paplorinc:
Fixes: https://github.com/javaslang/javaslang/issues/1348

> Vector

```java
Ratios slang / <alternative_impl>
Target           Operation   Ratio                                                    10        100       1000 
VectorBenchmark  Head        slang_persistent/fjava_persistent                     0.27x      0.25x      0.24x 
VectorBenchmark  Head        slang_persistent/java_mutable                         0.24x      0.22x      0.21x 
VectorBenchmark  Head        slang_persistent/pcollections_persistent              0.62x      1.02x      1.48x 
VectorBenchmark  Head        slang_persistent/scala_persistent                     0.28x      0.34x      0.32x 

VectorBenchmark  Tail        slang_persistent/fjava_persistent                    61.34x    114.41x    128.79x 
VectorBenchmark  Tail        slang_persistent/java_mutable                         7.01x      7.24x     15.55x 
VectorBenchmark  Tail        slang_persistent/pcollections_persistent              8.33x     14.09x     21.29x 
VectorBenchmark  Tail        slang_persistent/scala_persistent                     5.08x      8.90x     12.61x 

VectorBenchmark  Get         slang_persistent/fjava_persistent                     9.46x      7.46x     10.37x 
VectorBenchmark  Get         slang_persistent/java_mutable                         0.14x      0.03x      0.02x 
VectorBenchmark  Get         slang_persistent/pcollections_persistent              1.40x      0.95x      1.51x 
VectorBenchmark  Get         slang_persistent/scala_persistent                     0.27x      0.14x      0.09x 

VectorBenchmark  Update      slang_persistent/fjava_persistent                    47.86x     88.49x    103.10x 
VectorBenchmark  Update      slang_persistent/java_mutable                         0.11x      0.05x      0.03x 
VectorBenchmark  Update      slang_persistent/pcollections_persistent              1.30x      1.72x      2.09x 
VectorBenchmark  Update      slang_persistent/scala_persistent                     0.94x      0.60x      0.40x 

VectorBenchmark  Prepend     slang_persistent/fjava_persistent                     1.87x      2.46x      2.20x 
VectorBenchmark  Prepend     slang_persistent/java_mutable                         0.48x      0.47x      0.84x 
VectorBenchmark  Prepend     slang_persistent/pcollections_persistent              0.87x      1.75x      2.32x 
VectorBenchmark  Prepend     slang_persistent/scala_persistent                     0.44x      0.39x      0.29x 

VectorBenchmark  Append      slang_persistent/fjava_persistent                     2.53x      2.94x      2.63x 
VectorBenchmark  Append      slang_persistent/java_mutable                         0.13x      0.09x      0.08x 
VectorBenchmark  Append      slang_persistent/pcollections_persistent              1.13x      2.30x      3.53x 
VectorBenchmark  Append      slang_persistent/scala_persistent                     0.53x      0.42x      0.37x 

VectorBenchmark  GroupBy     slang_persistent/java_mutable                         0.21x      0.18x      0.25x 
VectorBenchmark  GroupBy     slang_persistent/scala_persistent                     0.85x      0.43x      0.30x 

VectorBenchmark  Iterate     slang_persistent/fjava_persistent                    30.15x     51.62x     37.36x 
VectorBenchmark  Iterate     slang_persistent/java_mutable                         0.18x      0.07x      0.04x 
VectorBenchmark  Iterate     slang_persistent/pcollections_persistent              2.11x      1.72x      1.15x 
VectorBenchmark  Iterate     slang_persistent/scala_persistent                     0.34x      0.19x      0.12x 

VectorBenchmark  Slice       slang_persistent/java_mutable                         0.90x      0.91x      0.79x
VectorBenchmark  Slice       slang_persistent/scala_persistent                     6.24x     14.79x     15.39x
```

### DIFF
--- a/javaslang-benchmark/src/test/java/javaslang/benchmark/collection/VectorBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/benchmark/collection/VectorBenchmark.java
@@ -1,0 +1,446 @@
+package javaslang.benchmark.collection;
+
+import javaslang.benchmark.JmhRunner;
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.infra.Blackhole;
+import scala.compat.java8.JFunction;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static javaslang.benchmark.JmhRunner.*;
+
+public class VectorBenchmark {
+    public static void main(String... args) {
+        JmhRunner.runDev(VectorBenchmark.class);
+    }
+
+    @State(Scope.Benchmark)
+    public static class Base {
+        @Param({ "10", "100", "1000" })
+        public int CONTAINER_SIZE;
+
+        int expectedAggregate = 0;
+        public Integer[] ELEMENTS;
+
+        /* Only use these for non-mutating operations */
+        final java.util.ArrayList<Integer> javaMutable = new java.util.ArrayList<>();
+
+        fj.data.Seq<Integer> fjavaPersistent = fj.data.Seq.empty();
+        org.pcollections.PVector<Integer> pcollectionsPersistent = org.pcollections.TreePVector.empty();
+        scala.collection.immutable.Vector<Integer> scalaPersistent = scala.collection.immutable.Vector$.MODULE$.empty();
+        javaslang.collection.Vector<Integer> slangPersistent = javaslang.collection.Vector.empty();
+
+        @Setup
+        public void setup() {
+            ELEMENTS = getRandomValues(CONTAINER_SIZE, 0);
+
+            assertEquals(javaMutable.size(), 0);
+            assertEquals(fjavaPersistent.length(), 0);
+            assertEquals(pcollectionsPersistent.size(), 0);
+            assertEquals(scalaPersistent.size(), 0);
+            assertEquals(slangPersistent.size(), 0);
+            for (Integer element : ELEMENTS) {
+                expectedAggregate ^= element;
+
+                javaMutable.add(element);
+
+                fjavaPersistent = fjavaPersistent.snoc(element);
+                pcollectionsPersistent = pcollectionsPersistent.plus(element);
+                scalaPersistent = scalaPersistent.appendBack(element);
+                slangPersistent = slangPersistent.append(element);
+            }
+            assertEquals(javaMutable.size(), CONTAINER_SIZE);
+            assertEquals(fjavaPersistent.length(), CONTAINER_SIZE);
+            assertEquals(pcollectionsPersistent.size(), CONTAINER_SIZE);
+            assertEquals(scalaPersistent.size(), CONTAINER_SIZE);
+            assertEquals(slangPersistent.size(), CONTAINER_SIZE);
+        }
+    }
+
+    public static class Head extends Base {
+        @Benchmark
+        public Object java_mutable() { return javaMutable.get(0); }
+
+        @Benchmark
+        public Object scala_persistent() { return scalaPersistent.head(); }
+
+        @Benchmark
+        public Object fjava_persistent() { return fjavaPersistent.head(); }
+
+        @Benchmark
+        public Object pcollections_persistent() { return pcollectionsPersistent.get(0); }
+
+        @Benchmark
+        public Object slang_persistent() { return slangPersistent.head(); }
+    }
+
+    public static class Tail extends Base {
+        @State(Scope.Thread)
+        public static class Initialized {
+            final java.util.ArrayList<Integer> javaMutable = new java.util.ArrayList<>();
+
+            @Setup(Level.Invocation)
+            public void initializeMutable(Base state) {
+                assertEquals(javaMutable.size(), 0);
+                Collections.addAll(javaMutable, state.ELEMENTS);
+                assertEquals(javaMutable.size(), state.CONTAINER_SIZE);
+            }
+
+            @TearDown(Level.Invocation)
+            public void tearDown() {
+                javaMutable.clear();
+            }
+        }
+
+        @Benchmark
+        public void java_mutable(Initialized state) {
+            final java.util.ArrayList<Integer> values = state.javaMutable;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values.remove(0);
+            }
+            assertEquals(values, new java.util.ArrayList<>());
+        }
+
+        @Benchmark
+        public void scala_persistent() {
+            scala.collection.immutable.Vector<Integer> values = scalaPersistent;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values = values.tail();
+            }
+            assertEquals(values, scala.collection.immutable.Vector.empty());
+        }
+
+        @Benchmark
+        public void fjava_persistent() {
+            fj.data.Seq<Integer> values = fjavaPersistent;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values = values.tail();
+            }
+            assertEquals(values, fj.data.Seq.empty());
+        }
+
+        @Benchmark
+        public void pcollections_persistent() {
+            org.pcollections.PVector<Integer> values = pcollectionsPersistent;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values = values.minus(0);
+            }
+            assertEquals(values, org.pcollections.TreePVector.empty());
+        }
+
+        @Benchmark
+        public void slang_persistent() {
+            javaslang.collection.Vector<Integer> values = slangPersistent;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values = values.tail();
+            }
+            assertEquals(values, javaslang.collection.Vector.empty());
+        }
+    }
+
+    public static class Get extends Base {
+        @Benchmark
+        public void java_mutable() {
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                assertEquals(javaMutable.get(i), ELEMENTS[i]);
+            }
+        }
+
+        @Benchmark
+        public void scala_persistent() {
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                assertEquals(scalaPersistent.apply(i), ELEMENTS[i]);
+            }
+        }
+
+        @Benchmark
+        public void fjava_persistent() {
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                assertEquals(fjavaPersistent.index(i), ELEMENTS[i]);
+            }
+        }
+
+        @Benchmark
+        public void pcollections_persistent() {
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                assertEquals(pcollectionsPersistent.get(i), ELEMENTS[i]);
+            }
+        }
+
+        @Benchmark
+        public void slang_persistent() {
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                assertEquals(slangPersistent.get(i), ELEMENTS[i]);
+            }
+        }
+    }
+
+    public static class Update extends Base {
+        @State(Scope.Thread)
+        public static class Initialized {
+            final java.util.ArrayList<Integer> javaMutable = new java.util.ArrayList<>();
+
+            @Setup(Level.Invocation)
+            public void initializeMutable(Base state) {
+                assertEquals(javaMutable.size(), 0);
+                Collections.addAll(javaMutable, state.ELEMENTS);
+                assertEquals(javaMutable.size(), state.CONTAINER_SIZE);
+            }
+
+            @TearDown(Level.Invocation)
+            public void tearDown() {
+                javaMutable.clear();
+            }
+        }
+
+        @Benchmark
+        public Object java_mutable(Initialized state) {
+            final java.util.ArrayList<Integer> values = state.javaMutable;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values.set(i, 0);
+            }
+            return values;
+        }
+
+        @Benchmark
+        public Object scala_persistent() {
+            scala.collection.immutable.Vector<Integer> values = scalaPersistent;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values = values.updateAt(i, 0);
+            }
+            return values;
+        }
+
+        @Benchmark
+        public Object fjava_persistent() {
+            fj.data.Seq<Integer> values = fjavaPersistent;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values = values.update(i, 0);
+            }
+            return values;
+        }
+
+        @Benchmark
+        public Object pcollections_persistent() {
+            org.pcollections.PVector<Integer> values = pcollectionsPersistent;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values = values.with(i, 0);
+            }
+            return values;
+        }
+
+        @Benchmark
+        public Object slang_persistent() {
+            javaslang.collection.Vector<Integer> values = slangPersistent;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values = values.update(i, 0);
+            }
+            return values;
+        }
+    }
+
+    public static class Prepend extends Base {
+        @Benchmark
+        @SuppressWarnings("ManualArrayToCollectionCopy")
+        public void java_mutable() {
+            final java.util.ArrayList<Integer> values = new java.util.ArrayList<>(CONTAINER_SIZE);
+            for (Integer element : ELEMENTS) {
+                values.add(0, element);
+            }
+            assertEquals(values.size(), CONTAINER_SIZE);
+        }
+
+        @Benchmark
+        public void scala_persistent() {
+            scala.collection.immutable.Vector<Integer> values = scala.collection.immutable.Vector$.MODULE$.empty();
+            for (Integer element : ELEMENTS) {
+                values = values.appendFront(element);
+            }
+            assertEquals(values.size(), CONTAINER_SIZE);
+        }
+
+        @Benchmark
+        public void fjava_persistent() {
+            fj.data.Seq<Integer> values = fj.data.Seq.empty();
+            for (Integer element : ELEMENTS) {
+                values = values.cons(element);
+            }
+            assertEquals(values.length(), CONTAINER_SIZE);
+        }
+
+        @Benchmark
+        public void pcollections_persistent() {
+            org.pcollections.PVector<Integer> values = org.pcollections.TreePVector.empty();
+            for (Integer element : ELEMENTS) {
+                values = values.plus(element);
+            }
+            assertEquals(values.size(), CONTAINER_SIZE);
+        }
+
+        @Benchmark
+        public void slang_persistent() {
+            javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.empty();
+            for (Integer element : ELEMENTS) {
+                values = values.prepend(element);
+            }
+            assertEquals(values.size(), CONTAINER_SIZE);
+        }
+    }
+
+    public static class Append extends Base {
+        @Benchmark
+        @SuppressWarnings("ManualArrayToCollectionCopy")
+        public void java_mutable() {
+            final java.util.ArrayList<Integer> values = new java.util.ArrayList<>(CONTAINER_SIZE);
+            for (Integer element : ELEMENTS) {
+                values.add(element);
+            }
+            assertEquals(values.size(), CONTAINER_SIZE);
+        }
+
+        @Benchmark
+        public void scala_persistent() {
+            scala.collection.immutable.Vector<Integer> values = scala.collection.immutable.Vector$.MODULE$.empty();
+            for (Integer element : ELEMENTS) {
+                values = values.appendBack(element);
+            }
+            assertEquals(values.size(), CONTAINER_SIZE);
+        }
+
+        @Benchmark
+        public void fjava_persistent() {
+            fj.data.Seq<Integer> values = fj.data.Seq.empty();
+            for (Integer element : ELEMENTS) {
+                values = values.snoc(element);
+            }
+            assertEquals(values.length(), CONTAINER_SIZE);
+        }
+
+        @Benchmark
+        public void pcollections_persistent() {
+            org.pcollections.PVector<Integer> values = org.pcollections.TreePVector.empty();
+            for (Integer element : ELEMENTS) {
+                values = values.plus(values.size(), element);
+            }
+            assertEquals(values.size(), CONTAINER_SIZE);
+        }
+
+        @Benchmark
+        public void slang_persistent() {
+            javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.empty();
+            for (Integer element : ELEMENTS) {
+                values = values.append(element);
+            }
+            assertEquals(values.size(), CONTAINER_SIZE);
+        }
+    }
+
+    public static class GroupBy extends Base {
+        @Benchmark
+        public Object java_mutable() { return javaMutable.stream().collect(Collectors.groupingBy(Integer::bitCount)); }
+
+        @Benchmark
+        public Object scala_persistent() { return scalaPersistent.groupBy(JFunction.func(Integer::bitCount)); }
+
+        @Benchmark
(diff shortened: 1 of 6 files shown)


javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -929,7 +906,7 @@ public int length() {
             if (found) {
                 arr[i] = get(i);
             } else {
-                if (currentElement.equals(value)) {
+                if (Objects.equals(currentElement, value)) {
```
danieldietrich: good!

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -26,16 +26,16 @@
 
     private static final Array<?> EMPTY = new Array<>(new Object[0]);
 
-    private final int startIndex, length;
+    private final int offset, length;
```
paplorinc: Aligned `Array` to `Vector`

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -946,7 +923,7 @@ public int length() {
         boolean changed = false;
         for (int i = 0; i < length(); i++) {
             final T value = get(i);
-            if (currentElement.equals(value)) {
+            if (Objects.equals(currentElement, value)) {
```
paplorinc: it's safer this way

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -526,30 +547,22 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
     public Vector<T> drop(long n) {
         if (n <= 0) {
             return this;
-        }
-        if (n >= length()) {
+        } else if (n >= length()) {
```
danieldietrich: I've not understood yet why we call a method `length()` when the instance variable `length` is in scope. Isn't it additional overhead to perform internal abstractions?

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -23,21 +23,34 @@
...
  */
 public final class Vector<T> implements Kind1<Vector<?>, T>, IndexedSeq<T>, Serializable {
-
     private static final long serialVersionUID = 1L;
 
     private static final Vector<?> EMPTY = new Vector<>(HashArrayMappedTrie.empty());
 
-    private final HashArrayMappedTrie<Integer, T> trie;
-    private final int indexShift;
+    private final int offset, length;
```
paplorinc: Until now the `indexShift` was only used as an offset for indices, i.e. if you deleted the first element, the first index became `1`, so we had to subtract `1` for every index. Now the underlying `HashArrayMappedTrie` remains the same (i.e. structural sharing), and the internal indexing can be negative also (see prepend)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -476,16 +489,24 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
-        return new Vector<>(indexShift, trie.put(length() + indexShift, element));
+        return appendAll(List.of(element));
     }
 
     @Override
     public Vector<T> appendAll(Iterable<? extends T> elements) {
-        HashArrayMappedTrie<Integer, T> result = trie;
+        HashArrayMappedTrie<Integer, T> result = delegate;
+
+        for (int i = offset + length(); i < delegate.size(); i++) {
```
paplorinc: in case of append, we delete elements from the end, add the elements back with the correct index and update the length

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: Just because the length is constant at the moment, doesn't mean it won't change (i.e. it's an implementation detail that may change, and we don't want to update it later everywhere, just in the `length` method)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -526,30 +547,22 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
     public Vector<T> drop(long n) {
         if (n <= 0) {
             return this;
-        }
-        if (n >= length()) {
+        } else if (n >= length()) {
             return empty();
+        } else {
+            return with(offset + (int) n, length() - (int) n);
```
paplorinc: `drop` is `O(1)` now, we don't need to reindex every remaining element :)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -526,30 +547,22 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
     @Override
     public Vector<T> dropRight(long n) {
         if (n <= 0) {
             return this;
-        }
-        if (n >= length()) {
+        } else if (n >= length()) {
             return empty();
+        } else {
+            return with(offset, length() - (int) n);
```
paplorinc: same, `O(1)`

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1090,40 +1050,30 @@ public int length() {
...
-        return wrap(trie);
     }
 
     @Override
     public Vector<T> tail() {
         if (isEmpty()) {
-            throw new UnsupportedOperationException("tail of empty vector");
-        }
-        if (length() == 1) {
-            return empty();
```
danieldietrich: `take(n)` returns `empty()`, why `tail()` not anymore?

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -662,61 +677,41 @@ public int indexOf(T element, int from) {
...
-        for (int i = 0; i <= length(); i++) {
-            if (i == index) {
-                for (T element : elements) {
-                    trie = trie.put(trie.size(), element);
+        if (index < 0 || index > length()) {
+            throw new IndexOutOfBoundsException("insertAll(" + index + ", e) on Vector of length " + length());
+        } else if (index == 0) {
+            return prependAll(elements);
+        } else if (index == length()) {
+            return appendAll(elements);
```
paplorinc: optimized insert in front or back

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -727,17 +722,16 @@ public boolean isTraversableAgain() {
...
             @Override
             public boolean hasNext() {
-                return index < size;
+                return index < Vector.this.length();
             }
 
             @Override
             public T getNext() {
-                return trie.get(index++).get();
+                return Vector.this.get(index++);
```
paplorinc: unified access to the `delegate` for simplicity

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -820,76 +815,61 @@ public int length() {
...
     public Vector<T> prependAll(Iterable<? extends T> elements) {
-        List<T> list = List.ofAll(elements);
-        final int newIndexShift = indexShift - list.length();
-        HashArrayMappedTrie<Integer, T> newTrie = trie;
-        for (int i = newIndexShift; !list.isEmpty(); i++) {
-            newTrie = newTrie.put(i, list.head());
-            list = list.tail();
+        HashArrayMappedTrie<Integer, T> result = delegate;
+
+        for (int i = 0; i < offset; i++) {
```
paplorinc: delete all leftover elements from the beginning to be able to prepend the new ones in `O(elements.size)` time

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -820,76 +815,61 @@ public int length() {
...
-            if (found) {
-                trie = trie.put(trie.size(), value);
-            } else {
-                if (element.equals(value)) {
-                    found = true;
-                } else {
-                    trie = trie.put(trie.size(), value);
-                }
+            if (Objects.equals(get(i), element)) {
+                return removeAt(i);
```
paplorinc: let's reuse `removeAt`

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -820,76 +815,61 @@ public int length() {
...
-            if (found) {
-                trie = trie.put(trie.size(), value);
-            } else {
-                if (predicate.test(value)) {
-                    found = true;
-                } else {
-                    trie = trie.put(trie.size(), value);
-                }
+            if (predicate.test(get(i))) {
+                return removeAt(i);
```
paplorinc: same

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -905,19 +885,21 @@ public int length() {
...
             throw new IndexOutOfBoundsException("removeAt(" + index + ")");
-        }
-        HashArrayMappedTrie<Integer, T> trie = HashArrayMappedTrie.empty();
-        for (int i = 0; i < length(); i++) {
-            if (i != index) {
-                trie = trie.put(trie.size(), get(i));
+        } else if (index == 0) {
+            return drop(1);
+        } else if (index == length() - 1) {
+            return dropRight(1);
```
paplorinc: optimized front and back removal to `O(1)`

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -973,11 +940,7 @@ public int length() {
 
     @Override
     public Vector<T> reverse() {
-        HashArrayMappedTrie<Integer, T> trie = HashArrayMappedTrie.empty();
-        for (int i = 0; i < length(); i++) {
-            trie = trie.put(i, get(length() - 1 - i));
-        }
-        return trie.isEmpty() ? empty() : new Vector<>(trie);
+        return (length() <= 1) ? this : foldLeft(empty(), Vector::prepend);
```
paplorinc: reverse of a single or empty `Vector` should return itself

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1090,40 +1050,30 @@ public int length() {
...
-            return Vector.empty();
-        }
-        HashArrayMappedTrie<Integer, T> trie = HashArrayMappedTrie.empty();
-        for (int i = beginIndex; i < endIndex; i++) {
-            trie = trie.put(trie.size(), get(i));
+            throw new IndexOutOfBoundsException("subSequence(" + beginIndex + ", " + endIndex + ") on Vector of length " + length());
+        } else if (beginIndex == endIndex) {
+            return empty();
+        } else {
+            return slice(beginIndex, endIndex);
```
paplorinc: let's reuse `slice`

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1136,30 +1086,22 @@ public int length() {
     public Vector<T> take(long n) {
         if (n >= length()) {
             return this;
-        }
-        if (n <= 0) {
+        } else if (n <= 0) {
             return empty();
+        } else {
+            return with(offset, (int) n);
```
paplorinc: now it's `O(1)`, it was linear

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1136,30 +1086,22 @@ public int length() {
...
     @Override
     public Vector<T> takeRight(long n) {
         if (n >= length()) {
             return this;
-        }
-        if (n <= 0) {
+        } else if (n <= 0) {
             return empty();
+        } else {
+            return with(length() + offset - (int) n, (int) n);
```
paplorinc: same

danieldietrich:
Please also update the performance table in `javaslang/collection/package-info.java` if necessary.

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: `with` will return empty if it's empty:

```java
if (length <= 0) {
    return empty();
}
```

paplorinc:
Everything was `constᵉᶠᶠ` before in the table, even though I think they were logarithmic in reality (i.e. `Map` has `log` for `remove`)

paplorinc:
@danieldietrich, @ruslansennov, @eduardmanas, @zsolt-donca, please review and try to find counter-examples that should still be tested :)

ruslansennov:
I'm trying to run this test
```java
    @Test
    public void tt() {
        Vector<Integer> vect = Stream.range(4, 1000)
                .foldLeft(Vector.of(1, 2, 3), (v, i) -> v.append(i).tail());
        System.out.println(vect); // expected: Vector(997, 998, 999)
    }
```
But your implementation fails (current master works fine).

----
Well, my goal is to check size of HAMT in resulting `Vector`. Current implementation shows `size = 3`. I guess your implementation will show `size = 1000`. 
Let's talk about this but not about bug that I shown above. OK, my test is a little bit synthetic and `Vector` can be replaced by `Queue` for example, but you can see the problem with memory consumption.

paplorinc:
Thank you @ruslansennov, very useful comments, I force pushed the fixed to the bug you pointed out :) (it was an index mismatch in `append`).
Abut storing the underlying map, people can simply normalize it via `ofAll(iterator())` any time, I consider it a separate memory optimization issue.

javaslang/src/test/java/javaslang/collection/VectorTest.java
```diff
@@ -208,46 +205,4 @@ public void shouldStringifyNil() {
     public void shouldStringifyNonNil() {
         assertThat(of(null, 1, 2, 3).toString()).isEqualTo("Vector(null, 1, 2, 3)");
     }
-
```
paplorinc: these were testing `List`, not applicable to `Vector`, since its `Empty` is not a separate class

javaslang/src/test/java/javaslang/collection/AbstractSeqTest.java
```diff
@@ -138,7 +138,30 @@ public void shouldAppendElementToNonNil() {
...
     @Test
     public void shouldMixAppendAndPrepend() {
-        assertThat(of(1).append(2).prepend(0).prepend(-1).append(3).append(4)).isEqualTo(of(-1, 0, 1, 2, 3, 4));
+        final Seq<Integer> append = of(1).append(2).prepend(0).prepend(-1).append(3).append(4);
+        assertThat(append).isEqualTo(of(-1, 0, 1, 2, 3, 4));
+    }
+
+    @Test
+    public void shouldAppend() {
+        final Seq<Integer> actual = Stream.range(4, 100).foldLeft(of(1, 2, 3),
```
paplorinc: thanks @ruslansennov

ruslansennov:
> About storing the underlying map, people can simply normalize it via ofAll(iterator()) any time, I consider it a separate memory optimization issue.

I'm not sure that we can force the user to do the normalization. This non-obvious requirement, which looks like a limitation. My snippet code with range to Long.MAX_VALUE can runs along years and it is expected behavior, but your fails with out of memory.

And another question. The complexity of all operations depends from size of `HAMT,` but not from size of Vector. Let's imagine code something like this:
```java
Vector<Integer> vect = Vector.range(0, 1_000_000).take(10)
```

Size of `Vector` is 10, but size of delegated `HAMT` is one million. So, all operations with vector (set, get, etc.) will be strange slow.

Sorry, my opinion is that this PR should be rejected.

danieldietrich:
> Sorry, my opinion is that this PR should be rejected.

Yes, I also think that this is a major issue.

paplorinc:
@danieldietrich, @ruslansennov, this is the same as `Array`, i.e. the delegate might be bigger than it appears.
If this seems like a valid problem, it could be resolved similarly to how `ArrayList` solves it (which has usually more allocated space than necessary also), i.e. if the masked out region is bigger than the visible one, the `with` method won't change the offsets, but copy the actual content instead.

If I'm mistaken correctly, this way the complexity of e.g. consuming `n` elements with `drop(1)` would be around `O(2n)`, while before it was `O(n²)`.

Opinions?

ruslansennov:
> this is the same as Array, i.e. the delegate might be bigger than it appears.

My failt, I have not seen this
Should be reverted ASAP

ruslansennov:
> :-1: 

this is just my opinion :)

paplorinc:
Could you please react to my response to your concern also?

ruslansennov:
> Could you please react to my response to your concern also?

IMO previous implementation of Array was simple and fast enough for its main goals: get element and iterate along elements. All others operation should used very carefully. If someone try to get tail of Array -  OK, but it must know what he doing and must understand all about bad performance.

paplorinc:
> but it must know what he doing

I think this logic is applicable to my version also!

javaslang-benchmark/src/test/java/javaslang/benchmark/collection/VectorBenchmark.java
```diff
@@ -0,0 +1,429 @@
...
+import javaslang.benchmark.JmhRunner;
+import org.openjdk.jmh.annotations.*;
+import scala.compat.java8.JFunction;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static javaslang.benchmark.JmhRunner.*;
+
+public class VectorBenchmark {
```
paplorinc: same as in https://github.com/javaslang/javaslang/pull/1372

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -45,12 +45,15 @@ private Array(Object[] delegate, int startIndex, int length) {
...
 
-    Array<T> with(int startIndex, int length) {
-        length = Math.min(delegate.length - startIndex, length);
+    private Array<T> with(int offset, int length) {
+        length = Math.min(delegate.length - offset, length);
         if (length <= 0) {
             return empty();
         } else {
-            return new Array<T>(delegate, startIndex, length);
+            /* In order to avoid excessive memory usage, we only store the offsets and length if the kept area's size is bigger than the leftover size, otherwise we trim it (similarly to how ArrayList operates) */
```
paplorinc: This is important @ruslansennov, @danieldietrich

javaslang/src/test/java/javaslang/collection/ArrayTest.java
```diff
@@ -188,11 +188,23 @@ public void shouldNarrowArray() {
         assertThat(actual).isEqualTo(3);
     }
 
+    // -- delegate
+
+    @Test
+    public void shouldPeriodicallyTrimExcess() {
```
paplorinc: test internals @ruslansennov

danieldietrich:
This PR is open for a while now. I share Ruslan's opinion that we should take memory consumption seriously.

Memory sharing makes only sense if there are actual users of different parts of a collection that form a whole. But if we keep a whole HAMT but only a little part is used at all then this could be considered as memory leak. However, we still need one reference to prevent the garbage from being collected.

We should not try to provide hacks/optimizations to an existing algorithm. Over time the algorithm will be hard to maintain because its simplicity got lost. Instead we should consider to change the algorithm - by taking a look at how other (like Scala) solved the problem. @paplorinc provided a [link](https://github.com/andrewoma/dexx/blob/master/collection/src/main/java/com/github/andrewoma/dexx/collection/Vector.java) to another Vector impl.

We should revert these change/close the PR. We should rewrite the whole Vector implementation without remorse if, and only if, we get an overall better performance. It does not make sense to rewrite it if we only get partial better results (for some methods) and others get worse.

The same applies to the recent Array changes.

---

**Note: In this PR not only Vector was changed but also Array. The Array changes should have been done in another PR.**

**Note 2: During my office hours I'm not able to answer question (even not at gitter.im). I will come back once a day / in the evening (CET).**

paplorinc:
@danieldietrich, I think I addressed @ruslansennov's concerns about memory consumption and I think these changes could be applied even if the Vector impl is changed!

Just take a look at how much faster the `tail` and `slice` operations are compared to anything else (e.g. 10x faster than Scala), while consuming the same amount of amortized memory (i.e. `O(2n)` worst case while providing `O(1)` operations for many operations).

I don't think the implementation got a lot more complicated, since I reused lots of parts internally.

danieldietrich:
I see the effort and the improvements that you've done with this PR and I really appreciate it. That's the reason I will pull it in. Please resolve the conflicts (see above) - then I will merge it (in the evening hours (CET)).

We always have to move forward, not backward. If there are other caveats with the current impl we have to solve them, right? So keep on the great work (@all).

paplorinc:
The memory consumption of my implementation has the same amortized complexity (!), while being a lot faster for some methods (e.g. 1500x faster than Scala for 1000 elements). These optimizations should be done even if the HAMT is changed: it's unrelated.
Not sure why @ruslansennov didn't explain the "whole picture" here, so that we can talk about it, but it's your library, you decide.

Feel free to use the "unrelated" parts from this commit, I'm not planning on doing it.
Keep up the good work with Javaslang!

danieldietrich:
But isn't Ruslan's example (currently) valid?

```java
// backing HAMT stays big
Vector<Integer> vect = Vector.range(0, 1_000_000).take(10)
```

paplorinc:
Apparently `result.toVector()` on line `55` in `Vector.java` isn't trimming correctly.
If we change it to `result.map(Function.identity())` instead, the above expression will have exactly `10` backing elements:
```java
@Test
public void test() {
    assertThat(Vector.range(0, 1_000_000).take(10).delegate.size()).isLessThanOrEqualTo(2 * 10);
}
```

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -23,21 +23,37 @@
...
+
+    private Vector<T> with(int offset, int length) {
+        length = Math.min(delegate.size() - offset, length);
+        if (length <= 0) {
+            return empty();
+        } else {
+            /* In order to avoid excessive memory usage, we only store the offsets and length if the kept area's size is bigger than the leftover size, otherwise we trim it (similarly to how ArrayList operates) */
+            final int maskedAreaLength = delegate.size() - length;
+            final Vector<T> result = new Vector<>(delegate, offset, length);
+            return (length >= maskedAreaLength) ? result : result.toVector();
```
paplorinc: This should have been:
```java
return (length >= maskedAreaLength) ? result : result.map(Function.identity())
```

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -45,12 +45,15 @@ private Array(Object[] delegate, int startIndex, int length) {
...
+            final int maskedAreaLength = delegate.length - length;
+            final Array<T> result = new Array<>(delegate, offset, length);
+            return (length >= maskedAreaLength) ? result : result.toArray();
```
paplorinc: This should have been:
```java
return (length >= maskedAreaLength) ? result : result.map(Function.identity())
```

ruslansennov:
What about `Vector.range(0, 1_000_000).take(500_001)`?

paplorinc:
It would still consume `O(n)` memory, just as
```java
List.range(0, 1_000_000).take(550_000).toJavaList()
```
will have `800_000` elements internally, instead of `500_000`, it's the same concept, just a different multiplier (1.5 instead of 2).

javaslang/src/test/java/javaslang/collection/VectorTest.java
```diff
@@ -189,11 +186,23 @@ public void shouldNarrowVector() {
         assertThat(actual).isEqualTo(3);
     }
 
+    // -- delegate
+
+    @Test
+    public void shouldPeriodicallyTrimExcess() {
```
paplorinc: This should have been:

```java
@Test
public void shouldPeriodicallyTrimExcess() {
    Set<Integer> sizes = TreeSet.empty();
    for (Vector<Integer> values = range(0, 100); !values.isEmpty(); values = values.tail()) {
        final int length = values.length(), delegateLength = values.delegate.size();
        if (length != delegateLength) {
            sizes = sizes.add(delegateLength);
        }
    }
    assertThat(sizes).isEqualTo(TreeSet.of(100, 49, 24, 11, 5, 2));
}
```

ruslansennov:
> List.range(0, 1_000_000).take(550_000).toJavaList()
will have 800_000 elements internally, instead of 500_000

Good catch, thank you.
We should use something like `new ArrayList(size())` instead of `new ArrayList()`

PR #1370, opened 2016-06-02
add the unfold function

emmanueltouzery:
Add the unfold function.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Array.java
+++ b/javaslang/src/main/java/javaslang/collection/Array.java
@@ -473,6 +473,60 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
         return ofAll(Iterator.rangeClosedBy(from, toInclusive, step));
     }
 
+    /**
+     * Creates a Array from a seed value and a function.
+     * The function takes the seed at first.
+     * The function should return {@code None} when it's
+     * done generating the Array, otherwise {@code Some} {@code Tuple}
+     * of the element for the next call and the value to add to the
+     * resulting Array.
+     * <p>
+     * Example:
+     * <pre>
+     * <code>
+     * Array.unfoldRight(10, x -&gt; x == 0
+     *             ? Option.none()
+     *             : Option.of(new Tuple2&lt;gt;(x, x-1)));
+     * // Array(10, 9, 8, 7, 6, 5, 4, 3, 2, 1))
+     * </code>
+     * </pre>
+     *
+     * @param seed the start value for the iteration
+     * @param f    the function to get the next step of the iteration
+     * @return an Array with the values built up by the iteration
+     * @throws IllegalArgumentException if {@code f} is null
+     */
+    static <T,U> Array<U> unfoldRight(T seed, Function1<T,Option<Tuple2<U,T>>> f) {
+        return Iterator.unfoldRight(seed, f).toArray();
+    }
+
+    /**
+     * Creates an Array from a seed value and a function.
+     * The function takes the seed at first.
+     * The function should return {@code None} when it's
+     * done generating the list, otherwise {@code Some} {@code Tuple}
+     * of the value to add to the resulting list and
+     * the element for the next call.
+     * <p>
+     * Example:
+     * <pre>
+     * <code>
+     * Array.unfoldLeft(10, x -&gt; x == 0
+     *             ? Option.none()
+     *             : Option.of(new Tuple2&lt;gt;(x-1, x)));
+     * // Array(1, 2, 3, 4, 5, 6, 7, 8, 9, 10))
+     * </code>
+     * </pre>
+     *
+     * @param seed the start value for the iteration
+     * @param f    the function to get the next step of the iteration
+     * @return an Array with the values built up by the iteration
+     * @throws IllegalArgumentException if {@code f} is null
+     */
+    static <T,U> Array<U> unfoldLeft(T seed, Function<T,Option<Tuple2<T,U>>> f) {
+        return Iterator.unfoldLeft(seed, f).toArray();
+    }
+
     @Override
     public Array<T> append(T element) {
         final Object[] copy = new Object[length() + 1];
--- a/javaslang/src/main/java/javaslang/collection/Array.java
+++ b/javaslang/src/main/java/javaslang/collection/Array.java
@@ -527,6 +527,33 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
         return Iterator.unfoldLeft(seed, f).toArray();
     }
 
+    /**
+     * Creates an Array from a seed value and a function.
+     * The function takes the seed at first.
+     * The function should return {@code None} when it's
+     * done generating the list, otherwise {@code Some} {@code Tuple}
+     * of the value to add to the resulting list and
+     * the element for the next call.
+     * <p>
+     * Example:
+     * <pre>
+     * <code>
+     * Array.unfold(10, x -&gt; x == 0
+     *             ? Option.none()
+     *             : Option.of(new Tuple2&lt;gt;(x-1, x)));
+     * // Array(1, 2, 3, 4, 5, 6, 7, 8, 9, 10))
+     * </code>
+     * </pre>
+     *
+     * @param seed the start value for the iteration
+     * @param f    the function to get the next step of the iteration
+     * @return an Array with the values built up by the iteration
+     * @throws IllegalArgumentException if {@code f} is null
+     */
+    static <T> Array<T> unfold(T seed, Function<? super T,Option<Tuple2<? extends T, ? extends T>>> f) {
+        return Iterator.unfold(seed, f).toArray();
+    }
+
     @Override
     public Array<T> append(T element) {
         final Object[] copy = new Object[length() + 1];
--- a/javaslang/src/main/java/javaslang/collection/Array.java
+++ b/javaslang/src/main/java/javaslang/collection/Array.java
@@ -496,7 +496,7 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
      * @return an Array with the values built up by the iteration
      * @throws IllegalArgumentException if {@code f} is null
      */
-    static <T,U> Array<U> unfoldRight(T seed, Function1<T,Option<Tuple2<U,T>>> f) {
+    static <T,U> Array<U> unfoldRight(T seed, Function<? super T,Option<Tuple2<? extends U, ? extends T>>> f) {
         return Iterator.unfoldRight(seed, f).toArray();
     }
 
@@ -523,7 +523,7 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
      * @return an Array with the values built up by the iteration
      * @throws IllegalArgumentException if {@code f} is null
      */
-    static <T,U> Array<U> unfoldLeft(T seed, Function<T,Option<Tuple2<T,U>>> f) {
+    static <T,U> Array<U> unfoldLeft(T seed, Function<? super T,Option<Tuple2<? extends T, ? extends U>>> f) {
         return Iterator.unfoldLeft(seed, f).toArray();
     }
 
--- a/javaslang/src/main/java/javaslang/collection/CharSeq.java
+++ b/javaslang/src/main/java/javaslang/collection/CharSeq.java
@@ -239,7 +239,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @return a CharSeq with the values built up by the iteration
      * @throws IllegalArgumentException if {@code f} is null
      */
-    static <T> CharSeq unfoldRight(T seed, Function<T,Option<Tuple2<Character, T>>> f) {
+    static <T> CharSeq unfoldRight(T seed, Function<? super T,Option<Tuple2<? extends Character, ? extends T>>> f) {
         return CharSeq.ofAll(Iterator.unfoldRight(seed, f));
     }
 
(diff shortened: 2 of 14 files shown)


javaslang/src/main/java/javaslang/collection/List.java
```diff
@@ -555,6 +555,40 @@
...
+     * // List(10, 9, 8, 7, 6, 5, 4, 3, 2, 1))
+     * </code>
+     * </pre>
+     *
+     * @param seed the start value for the iteration
+     * @param f    the function to get the next step of the iteration
+     * @return a list with the values built up by the iteration
+     * @throws IllegalArgumentException if {@code f} is null
+     */
+    static <T,U> List<U> unfold(T seed, Function1<T,Option<Tuple2<U,T>>> f) {
```
paplorinc: This seems to be the same as
```java
List.ofAll(Iterator.iterate(10, x -> x - 1).takeWhile(x -> x > 0));
```

emmanueltouzery:
regarding the analogy to `iterate`, the difference is that for iterate the input to the next iteration is necessarily the result of the current iteration, while with `unfold` you are free to specify any input (plus the fact that the termination is builtin, but that's more a detail I think).

For instance the situation I needed it in was with `Scanner`:

        Scanner scanner = new Scanner(input);
        return unfold(
            scanner,
            s -> Option.of(s.findInLine(regex))
            .map(f -> new Tuple2<>(s.match(), s)));

So here I keep passing the scanner as input for the next iteration, while in the return list I put the matches. It's actually unnecessary in this case, and `iterate` would be enough for this case, since `Scanner` is side-effect based and mutable.. If Scanner was immutable and would return a new Scanner when you modify it, then `unfold` would be needed. As it is, you can just refer to the outside Scanner from `iterate`, but I think there will be situations where the full unfold will come in handy.

Otherwise I lifted the API from [the haskell stdlib](https://hackage.haskell.org/package/base-4.8.1.0/docs/Data-List.html#g:9)

danieldietrich:
`unfold` should be provided for all `Traversable`s(?), maybe as part of the `Foldable` interface. Currently you provided only an impl for `List`, right?

emmanueltouzery:
Yes. I thought about that, I didn't do it or mention it because I wanted to find out whether you're interested at all, and also I'm not sure whether it really makes sense for `Set`? It could generate two elements which are equal and they would appear only once in the set, that would be a bit strange wouldn't it?
But yes, for `Seq` and `List` at least it makes sense.

emmanueltouzery:
So waiting for the answer on whether it makes sense also for `Set` before I resubmit. If it shouldn't be done for `Set`, we could implement it on the `Seq` interface?

danieldietrich:
Yes, I see the use cases, it makes perfectly sense to have it. The example @paplorinc provided is simpler in the sense of intermediate object creation but it is only a special case.

Haskell provides dual methods `unfold`, `unfoldLeft` and `unfoldRight` for the methods `fold`, `foldLeft`, `foldRight`.

We should provide static methods `unfold`, `unfoldLeft` and `unfoldRight` for all `Seq` impls (i.e. not for `Seq`, `LinearSeq` and `IndexSeq`), namely `CharSeq`, `Array`, `Vector`, `Queue`, `List` and `Stream`.

**My suggestion is to provide methods `Iterator.unfold*`. The `Seq` impls then just call `Iterator.unfoldXxx(...).toList()` for example.**

emmanueltouzery:
Just two things:

1. The example @paplorinc provided is from the apidoc in the PR, which i lifted from the haskell apidoc. It suggests maybe we need a better example in the doc but i can't think of a short standalone one.

2. Afaik haskell provides only unfoldr? See also http://stackoverflow.com/questions/23779494/why-the-r-in-unfoldr - but yeah why stop there.

Ok i'll probably work on it through the week-end. Implementation on Iterator, all `Seq` implementations get it.

emmanueltouzery:
Yeah the reason haskell only provides a right unfold is because it only provides list unfold and a list left unfold would have disastrous performance.

danieldietrich:
We should only implement what makes sense, e.g. start with just `unfold`. You can decide it.

> list left unfold would have disastrous performance.

Interesting. I understand left/right unfold as dual to left/right fold. They should have same performance characteristics as their dual operations...

emmanueltouzery:
Hmm maybe I had the wrong understanding of what would be a left unfold. I thought it would build the list from the left, so doing appends in a loop, instead of prepends in a list for a right unfold. But maybe I wrong in this understanding. I'll try to google for left unfold implementations somewhere to make sure.

danieldietrich:
Ok. I will take also a look this weekend! Maybe I can provide some useful input...

emmanueltouzery:
So I looked into building the unfold implementation on `Iterator`. Turns out it's a poor fit for `unfoldRight`, but a very elegant one for `unfoldLeft`. The left unfold appends each new value that is obtained to the result sequence. That works great with the `Iterator` API.
The right unfold however *prepends* each new value that is obtained to the result. That doesn't work at all with the `Iterator` API... We would have to expand the whole iterator in memory then revert it.

I don't know enough the javaslang collections hierarchy... One idea that I have from the top of my head right now is to have the `unfoldLeft` implementation in `Iterator`, and the `unfoldRight` in each specifc type: `List`, `Seq`..?

Btw just curious about `fold` being a left fold but with a single type instead of two. Is it meant to make the job simpler for the type checker?

PS: for reference, [haskell's Data.Sequence implements both unfoldr & unfoldl](http://hackage.haskell.org/package/containers-0.5.7.1/docs/Data-Sequence.html#v:unfoldr)

emmanueltouzery:
I guess one question is, do we want to offer `unfoldRight` on `Iterator` for coherence/symetry although it's widely inefficient?
In any case `List` will keep its custom more efficient version.

emmanueltouzery:
hmm well also `List.unfoldRight` must reverse the list in the end... Maybe then it can also call `Iterator.unfoldRight` and performance is the same in the end...

emmanueltouzery:
I pushed a commit which is closer to what we want, I think. I wrote that we still need the normal `fold` call, plus a more tolerant type for the `Function<>` (with proper use of `super` & `extends` and so on). However I'm not sure about the `fold`?
If the type will be <T,T> instead of <T,U> then I'm not sure it brings much over `iterate`?

danieldietrich:
Thank you, I will take an in-depth look this evening.

I think the best place for the implementation might be the package-private `javaslang.collection.Collections` helper class instead of `Iterator`.

emmanueltouzery:
hmm but otherwise we still want to export `Iterator.unfold*` right? If that's case why does it help to move the implementation somewhere else?

danieldietrich:
You are right. Location `Iterator` for impls is ok...

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -473,6 +473,60 @@ private Array(Object[] delegate, int startIndex, int length) {
...
+     * // Array(10, 9, 8, 7, 6, 5, 4, 3, 2, 1))
+     * </code>
+     * </pre>
+     *
+     * @param seed the start value for the iteration
+     * @param f    the function to get the next step of the iteration
+     * @return an Array with the values built up by the iteration
+     * @throws IllegalArgumentException if {@code f} is null
+     */
+    static <T,U> Array<U> unfoldRight(T seed, Function1<T,Option<Tuple2<U,T>>> f) {
```
danieldietrich: We should use `Function` instead of `Function1` for method argument types (because `Function1 extends Function`).

Also I think it could make sense to widen the bounds:

```java
static <T,U> Array<U> unfoldRight(T seed, Function<? super T, Option<Tuple2<? extends U, ? extends T>>> f)
```

I think we do not need `? extends Option` in that signature because the Option API always returns Option.

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -473,6 +473,60 @@ private Array(Object[] delegate, int startIndex, int length) {
...
+     * // Array(1, 2, 3, 4, 5, 6, 7, 8, 9, 10))
+     * </code>
+     * </pre>
+     *
+     * @param seed the start value for the iteration
+     * @param f    the function to get the next step of the iteration
+     * @return an Array with the values built up by the iteration
+     * @throws IllegalArgumentException if {@code f} is null
+     */
+    static <T,U> Array<U> unfoldLeft(T seed, Function<T,Option<Tuple2<T,U>>> f) {
```
danieldietrich: also `Function<? super T, Option<Tuple2<? extends U, ? extends T>>>`

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -216,6 +216,60 @@ public static CharSeq rangeClosedBy(char from, char toInclusive, int step) {
...
+     * // CharSeq.of("jihgfedcb"))
+     * </code>
+     * </pre>
+     *
+     * @param seed the start value for the iteration
+     * @param f    the function to get the next step of the iteration
+     * @return a CharSeq with the values built up by the iteration
+     * @throws IllegalArgumentException if {@code f} is null
+     */
+    static <T> CharSeq unfoldRight(T seed, Function<T,Option<Tuple2<Character, T>>> f) {
```
danieldietrich: dito

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -216,6 +216,60 @@ public static CharSeq rangeClosedBy(char from, char toInclusive, int step) {
...
+     * // CharSeq.of("bcdefghij"))
+     * </code>
+     * </pre>
+     *
+     * @param seed the start value for the iteration
+     * @param f    the function to get the next step of the iteration
+     * @return a CharSeq with the values built up by the iteration
+     * @throws IllegalArgumentException if {@code f} is null
+     */
+    static <T> CharSeq unfoldLeft(T seed, Function<T,Option<Tuple2<T,Character>>> f) {
```
danieldietrich: Because Character is final we do:

```java
Function<? super T, Option<Tuple2<? extends T, Character>>>
```

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1156,6 +1156,77 @@ public boolean hasNext() {
...
+     * // List(1, 2, 3, 4, 5, 6, 7, 8, 9, 10))
+     * </code>
+     * </pre>
+     *
+     * @param seed the start value for the iteration
+     * @param f    the function to get the next step of the iteration
+     * @return a list with the values built up by the iteration
+     * @throws IllegalArgumentException if {@code f} is null
+     */
+    static <T,U> Iterator<U> unfoldLeft(T seed, Function<T,Option<Tuple2<T,U>>> f) {
```
danieldietrich: dito

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1156,6 +1156,77 @@ public boolean hasNext() {
...
+     * // List(10, 9, 8, 7, 6, 5, 4, 3, 2, 1))
+     * </code>
+     * </pre>
+     *
+     * @param seed the start value for the iteration
+     * @param f    the function to get the next step of the iteration
+     * @return a list with the values built up by the iteration
+     * @throws IllegalArgumentException if {@code f} is null
+     */
+    static <T,U> Iterator<U> unfoldRight(T seed, Function<T,Option<Tuple2<U,T>>> f) {
```
danieldietrich: dito

danieldietrich:
The code looks great. I've added comments about widening the function arguments. Please give me a hint if you think that could work.

emmanueltouzery:
Thanks for spelling it out about the widening! I'll make the changes tomorrow. There was another open question: do we implement an `unfold` besides `unfoldLeft` and right?

I wrote: 

> However I'm not sure about the fold?
If the type will be <T,T> instead of <T,U> then I'm not sure it brings much over iterate?

emmanueltouzery:
Ok I widened the type signatures. Note that I tried simply `Character` since `Character` is final, but that does not typecheck. `? extends Character` is needed for it to type-check.

So as I said I think the last question is do we add an `unfold` besides `unfoldLeft` & `unfoldRight`. See my previous questions on that topic: https://github.com/javaslang/javaslang/pull/1370#issuecomment-225448815

I also asked:

> Btw just curious about fold being a left fold but with a single type instead of two. Is it meant to make the job simpler for the type checker?

danieldietrich:
Thank you!

> `? extends Character is needed for it to type-check.`

ok, LGTM

> the last question is do we add an `unfold` besides `unfoldLeft` & `unfoldRight`

Yes. `fold*` is _dual_ to `unfold*`, so we need unfold.

`fold` is a simple version of `foldLeft` that _operates_  on values of same type. We find such methods everywhere, e.g.

Java: `Stream.reduce(BinaryOperator<T> accumulator)`

Scala: `GenTraversableOnce`

```scala
  def reduce[A1 >: A](op : scala.Function2[A1, A1, A1]) : A1
  def reduceOption[A1 >: A](op : scala.Function2[A1, A1, A1]) : scala.Option[A1]
  def fold[A1 >: A](z : A1)(op : scala.Function2[A1, A1, A1]) : A1
  def /:[B](z : B)(op : scala.Function2[B, A, B]) : B
  def :\[B](z : B)(op : scala.Function2[A, B, B]) : B
  def foldLeft[B](z : B)(op : scala.Function2[B, A, B]) : B
  def foldRight[B](z : B)(op : scala.Function2[A, B, B]) : B
```

---

Coming back to the original question - the dual counterpart of `fold` looks like this:

```java
T fold(T zero, BiFunction<? super T, ? super T, ? extends T> combine)

static <T> ReturnType<T> unfoldLeft(T seed, Function<? super T, Option<Tuple2<? extends T, ? extends T>>> f);
```

emmanueltouzery:
ok added the less-polymorphic unfold(). Note that for CharSeq it's not polymorphic at all anymore => it's fully hardcoded to `Character`.

danieldietrich:
Well done, thank you!!

PR #1325, opened 2016-05-11
Reduced many operations in `Array` from linear to constant time and space

paplorinc:
> ArrayBenchmark

Before:
```java
Group 'AddAll'                 :    10 |    100 |   1000 |
java_mutable / slang_persistent: 6.30x | 13.17x | 97.02x |

Group 'Iterate'                :    10 |    100 |    1000 |
java_mutable / slang_persistent: 4.19x | 27.39x | 448.73x |
```

After:
```java
Group 'AddAll'                 :     10 |    100 |    1000 | 
java_mutable / slang_persistent: 10.15x | 23.06x | 111.37x | 

Group 'Iterate'                :    10 |   100 |   1000 | 
java_mutable / slang_persistent: 1.99x | 5.14x | 10.31x | 
```

PR #1373, opened 2016-06-08
Array: reverts Object[] sharing #1325

ruslansennov:


danieldietrich:
Hard decision for me as there are two general opposed opinions by committers who do awesome work.

I understand _memory sharing_ in terms of _sharing what is needed_ and not _keeping what is not needed any more_. Therefore I will pull it.

Thank you!

PR #1376, opened 2016-06-10
add the Map.putOrUpdate function.

emmanueltouzery:
Adding the "putOrUpdate" function.
java8 has an equivalent function [Map.merge()](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#merge-K-V-java.util.function.BiFunction-). Haskell calls it [insertWith](https://hackage.haskell.org/package/containers-0.5.7.1/docs/Data-Map-Strict.html#v:insertWith).
`Map.merge()` is already taken in javaslang, so we need another name.

I think it makes sense that the name starts with `put` so that you spot it immediately next to the normal `put`. Then it could be `purOrMerge`, `putWith` or something else I guess.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/AbstractMap.java
+++ b/javaslang/src/main/java/javaslang/collection/AbstractMap.java
@@ -48,6 +48,7 @@ abstract class AbstractMap<K, V, M extends AbstractMap<K, V, M>> implements Map<
     @SuppressWarnings("unchecked")
     @Override
     public M putWith(K key, V value, BiFunction<? super V, ? super V, ? extends V> merge) {
+        Objects.requireNonNull(merge, "the merge function is null");
         Option<V> currentValue = get(key);
         if (currentValue.isEmpty()) {
             return (M) put(key, value);
@@ -59,6 +60,7 @@ abstract class AbstractMap<K, V, M extends AbstractMap<K, V, M>> implements Map<
     @Override
     public M putWith(Tuple2<? extends K, ? extends V> entry,
                      BiFunction<? super V, ? super V, ? extends V> merge) {
+        Objects.requireNonNull(merge, "the merge function is null");
         Option<V> currentValue = get(entry._1);
         if (currentValue.isEmpty()) {
             return put(entry);
--- a/javaslang/src/main/java/javaslang/collection/AbstractMap.java
+++ b/javaslang/src/main/java/javaslang/collection/AbstractMap.java
@@ -5,7 +5,6 @@
  */
 package javaslang.collection;
 
-import javaslang.Function2;
 import javaslang.Tuple;
 import javaslang.Tuple2;
 import javaslang.control.Option;
@@ -48,7 +47,7 @@ abstract class AbstractMap<K, V, M extends AbstractMap<K, V, M>> implements Map<
 
     @SuppressWarnings("unchecked")
     @Override
-    public M putWith(K key, V value, Function2<? super V, ? super V, ? extends V> merge) {
+    public M putWith(K key, V value, BiFunction<? super V, ? super V, ? extends V> merge) {
         Option<V> currentValue = get(key);
         if (currentValue.isEmpty()) {
             return (M) put(key, value);
@@ -59,7 +58,7 @@ abstract class AbstractMap<K, V, M extends AbstractMap<K, V, M>> implements Map<
 
     @Override
     public M putWith(Tuple2<? extends K, ? extends V> entry,
-                         Function2<? super V, ? super V, ? extends V> merge) {
+                     BiFunction<? super V, ? super V, ? extends V> merge) {
         Option<V> currentValue = get(entry._1);
         if (currentValue.isEmpty()) {
             return put(entry);
--- a/javaslang/src/main/java/javaslang/collection/AbstractMap.java
+++ b/javaslang/src/main/java/javaslang/collection/AbstractMap.java
@@ -5,6 +5,7 @@
  */
 package javaslang.collection;
 
+import javaslang.Function2;
 import javaslang.Tuple;
 import javaslang.Tuple2;
 import javaslang.control.Option;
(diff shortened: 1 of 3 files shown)


ruslansennov:
Maybe we should also add `Map::putOrUpdate(Tuple2)` (see `Map::put(Tuple2)`)

emmanueltouzery:
sure, I can modify the PR, that's a simple one. I'll make a second commit shortly adding `putOrUpdate(Tuple2, Function2<V, V, V>)`

emmanueltouzery:
I made a second commit to add the Tuple support. I reimplemented the function to call the lower-level Tuple version instead of calling directly putOrUpdate(x._1, x_2, merge), assuming that also `put` was reimplemented for tuples because it can be specially optimized.

danieldietrich:
Great!

Let's call it `putWith`, it is shorter and indicates that it is _put with a merge function_.

Please move the implementation to `AbstractMap`. The complete `Map` type hierarchy needs methods with adjusted return types. Take a look at `put` to see an example.

emmanueltouzery:
thanks, will do.
Also looking at the type of the Function2. Java8 map.merge is:

    Function2<? super V, ? super V, ? extends V>

(well they use `BiFunction` obviously) I've read a bit on `super` vs `extends` now and... I'm hoping you're familiar with this, because I won't dig too deep in this right now ;)

 So do I go also with this super/extends setup?

emmanueltouzery:
pushed with the suggested changes plus the super & extends in the `Function` type.

javaslang/src/main/java/javaslang/collection/AbstractMap.java
```diff
@@ -47,6 +48,29 @@ public M put(Tuple2<? extends K, ? extends V> entry) {
 
     @SuppressWarnings("unchecked")
     @Override
+    public M putWith(K key, V value, Function2<? super V, ? super V, ? extends V> merge) {
```
ruslansennov: `Function2` extends `BiFunction`, so use it please (like in Java8)

javaslang/src/main/java/javaslang/collection/AbstractMap.java
```diff
@@ -47,6 +48,29 @@ public M put(Tuple2<? extends K, ? extends V> entry) {
...
+        if (currentValue.isEmpty()) {
+            return (M) put(key, value);
+        } else {
+            return (M) put(key, merge.apply(currentValue.get(), value));
+        }
+    }
+
+    @Override
+    public M putWith(Tuple2<? extends K, ? extends V> entry,
+                         Function2<? super V, ? super V, ? extends V> merge) {
```
ruslansennov: also here

emmanueltouzery:
...in case someone would pass an actual BiFunction & not a lambda with type inference. I get it. Pushed.

emmanueltouzery:
I had forgotten the defensive null check for the merge function. Pushed now. Do squash the commit if you pull it ;-)

javaslang/src/main/java/javaslang/collection/AbstractMap.java
```diff
@@ -47,6 +47,31 @@ public M put(Tuple2<? extends K, ? extends V> entry) {
...
+        if (currentValue.isEmpty()) {
+            return (M) put(key, value);
+        } else {
+            return (M) put(key, merge.apply(currentValue.get(), value));
+        }
+    }
+
+    @Override
+    public M putWith(Tuple2<? extends K, ? extends V> entry,
+                     BiFunction<? super V, ? super V, ? extends V> merge) {
```
danieldietrich: I've compared the signature with `Map.merge`:

```java
<U extends V> Map<K, V> merge(Map<? extends K, U> that,                  BiFunction<? super V, ? super U, ? extends V> collisionResolution);
              M         putWith(Tuple2<? extends K, ? extends V> entry,  BiFunction<? super V, ? super V, ? extends V> merge);
```

They are very similar. We should adjust `putWith` like this:

```java
<U extends V> M         putWith(Tuple2<? extends K, ? extends U> entry,  BiFunction<? super V, ? super U, ? extends V> merge);
```

That opens the type space of the second covariant argument of the `merge` function.

/cc @ruslansennov right?

ruslansennov: I believe `Tuple2<? extends K, U>` is enough

danieldietrich: Yep, edit-error by me - will correct it in the example above.

javaslang/src/main/java/javaslang/collection/AbstractMap.java
```diff
@@ -47,6 +47,31 @@ public M put(Tuple2<? extends K, ? extends V> entry) {
 
     @SuppressWarnings("unchecked")
     @Override
+    public M putWith(K key, V value, BiFunction<? super V, ? super V, ? extends V> merge) {
+        Objects.requireNonNull(merge, "the merge function is null");
+        Option<V> currentValue = get(key);
```
danieldietrich: One hint (I know it is nasty): Please use the `final` keyword on _local vars_ if possible. That aligns the code style to the  Javaslang code base.

Once I decided to do it from the ground up while building Javaslang because Scala has `var` and `val`. One of the most evil enemies of functional programming is Mr. Mutation. Having final vars hardens our code.

The only thing I don't like about `final` is its verbosity. In an utopic Java language it should have been the default. But with [JEP #286](http://openjdk.java.net/jeps/286) comes the rescue (I pray every night for it). Then all of our local vars will be either `var` or `val`, of course most of them `val`.

emmanueltouzery:
trust me, on the `final` issue [you're  preaching to the converted](https://twitter.com/deech/status/728046838109110272) :)

I'm still picking up the discipline to actually type `final` in java, but I support it and want it 200%.

I'll make another commit with the requested changes sometime until the evening, thanks for the feedback! This PR is definitely interesting for me from the point of view of API design & inheritance-related type constraints =D

danieldietrich:
> trust me, on the final issue you're preaching to the converted :)

Haha, nice link :-))))

emmanueltouzery:
done, I think!

emmanueltouzery:
actually it crossed my mind that with overloading in java would have even just an overload to `put` with the three parameters. No need for `putWith`. Maybe `putWith` is clearer though. https://github.com/javaslang/javaslang/issues/1379#issuecomment-226170608 may be relevant where we could have `mapKeys` or `mapKeysWith` potentially.

danieldietrich:
In Javaslang we use `Map.merge` for both cases. We have no `Map.mergeWith`. To be consistent we should use `put` for both cases and not `putWith`. Same applies to `mapKeys`.

emmanueltouzery:
renamed it to `put` as discussed.

danieldietrich:
Well done! This looks really nice now. Great addition.

PR #1385, opened 2016-06-16
Implement new method: CheckedFunction_#recover

### LINKED ISSUE #1381: Add CheckedFunction_#recover(Function_)
Sometimes I want to use trusted checked functions in work, and leave behind all exception handling.
Something like
```java
public class DigestTest {
	public static String digest(String algorithm, String value) throws NoSuchAlgorithmException {
		MessageDigest md = MessageDigest.getInstance(algorithm);
		md.update(value.getBytes(StandardCharsets.UTF_8));
		return toHexString(md.digest());
	}

	public static CheckedFunction2<String, String, String> digest = DigestTest::digest;
	
	public static Function<String, String> md5 = digest.apply("MD5").rethrow();

	@Test
	public void testMd5() throws Exception {
		assertThat(md5.apply("123")).isNotNull().isEqualToIgnoringCase("202cb962ac59075b964b07152d234b70");
	}
}
```

In some cases we know that function can not throw Exception, or we can not recover if it will be really throwed.

But this case can have more general purpose: recover from error.
This can be implemented somethis like this:
```java
    default Function1<T1, R> recover(Function2<? super T1, ? super Throwable, ? extends R> recover) {
        return (t1) -> {
            try {
                return this.apply(t1);
            } catch (Throwable throwable) {
                return recover.apply(t1, throwable);
            }
        };
    }

	default <X extends RuntimeException> Function1<T1, R> rethrow(Function<Throwable, X> exceptionMapper) {
		return recover(((t1, throwable) -> {
			throw exceptionMapper.apply(throwable);
		}));
	}

	default Function1<T1, R> trusted() {
		return rethrow(IllegalStateException::new);
	}

```

Compare API 

1. `CheckedFunction0<T1, R> func`:
  * `recover`
    * NEW: `func.recover(() -> new R())`
    * OLD: `CheckedFunction0.lift(func).andThen(o -> o.getOrElse(() -> new R()))`
  * `trusted`
    * NEW: `func.rethrow(ex -> new IllegalArgumentException("Message", ex))`
    * OLD: exception lost after `lift`
1. `CheckedFunction1<T1, R> func`:
  * `recover`
    * NEW: `func.recover(t1 -> new R(t1))`
    * OLD: arguments lost after `lift`
  * `trusted`
    * NEW: `func.rethrow(ex -> new IllegalArgumentException("Message", ex))`
    * OLD: exception lost after `lift`
1. `CheckedFunction2<T1, T2, R> func`:
  * `recover`
    * NEW: `func.recover((t1, t2) -> new R(t1, t2))`
    * OLD: arguments lost after `lift`
  * `trusted`
    * NEW: `func.rethrow(ex -> new IllegalArgumentException("Message", ex))`
    * OLD: exception lost after `lift`

valery1707:
fix #1381

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -1525,37 +1525,73 @@ def generateTestClasses(): Unit = {
                   $assertThat(memo.isMemoized()).isTrue();
               }
 
-              ${(checked && i == 0).gen(xs"""
-                @$test
-                public void shouldRecover() {
-                    final $AtomicInteger integer = new $AtomicInteger();
-                    $name$i<MessageDigest> digest = () -> ${im.getType("java.security.MessageDigest")}.getInstance(integer.get() == 0 ? "MD5" : "Unknown");
-                    Function$i<MessageDigest> recover = digest.recover((tuple, throwable) -> null);
-                    MessageDigest md5 = recover.apply();
-                    assertThat(md5).isNotNull();
-                    assertThat(md5.getAlgorithm()).isEqualToIgnoringCase("MD5");
-                    assertThat(md5.getDigestLength()).isEqualTo(16);
-                    integer.incrementAndGet();
-                    assertThat(recover.apply()).isNull();
-                }
-              """)}
-              ${(checked && i > 0).gen(xs"""
-                @$test
-                public void shouldRecover() {
-                    ${
-                      val types = s"<${(1 to i).gen(j => "String")(", ")}, MessageDigest>"
-                      def toArgList (s: String) = s.split("", i).mkString("\"", "\", \"", "\"") + (s.length + 2 to i).gen(j => ", \"\"")
-                      xs"""
-                        $name$i$types digest = (${(1 to i).gen(j => s"s$j")(", ")}) -> ${im.getType("java.security.MessageDigest")}.getInstance(${(1 to i).gen(j => s"s$j")(" + ")});
-                        Function$i<${(1 to i).gen(j => "String")(", ")}, MessageDigest> recover = digest.recover((tuple, throwable) -> null);
-                        MessageDigest md5 = recover.apply(${toArgList("MD5")});
-                        assertThat(md5).isNotNull();
-                        assertThat(md5.getAlgorithm()).isEqualToIgnoringCase("MD5");
-                        assertThat(md5.getDigestLength()).isEqualTo(16);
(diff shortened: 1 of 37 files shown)


valery1707:
Tests for `Function_#liftTry()` will be created soon

danieldietrich:
That was fast! The implementation looks great.

But I think we need a second iteration. While looking at our current idea to use Tuples (because of the Function arity bound of 8) I think currying fits better. We can make the following simplifications:

**Simplify method _argument_ types**

(Function0..2 have super interface in java.util.function)

* Instead of `Function0<R>` we write `Supplier<R>`
* Instead of `Function1<T1, R>` we write `Function<T1, R>`
* Instead of `Function2<T1, T2, R>` we write `BiFunction<T1, T2, R>`
* For CheckedFunctions we cannot make these 'optimizations'

**Currying instead of Tuples:**

(Curried type signatures look ugly in Java but are nice to write: `x -> (t1, ..., tn) -> r`)

* Instead of `CheckedFunction2<Tuple0, Throwable, R>` we should use `CheckedFunction1<Throwable, CheckedFunction0<R>>`
* Instead of `CheckedFunction2<Tuple1<T1>, Throwable, R>` we should use `CheckedFunction1 <Throwable, CheckedFunction1<T1, R>>`
* Instead of `CheckedFunction2<Tuple2<T1, T2>, Throwable, R>` we should use `CheckedFunction1<Throwable, CheckedFunction2<T1, T2, R>>`
* ...
* Instead of `Function2<Tuple0, Throwable, R>` we should use `Function<Throwable, Supplier<R>>`
* Instead of `Function2<Tuple1<T1>, Throwable, R>` we should use `Function<Throwable, Function<T1, R>>`
* Instead of `Function2<Tuple2<T1, T2>, Throwable, R>` we should use `Function<Throwable, BiFunction<T1, T2, R>>`
* Instead of `Function3<Tuple3<T1, T2, T3>, Throwable, R>` we should use `Function<Throwable, Function3<T1, T2, T3, R>>`
* ...

I hope that it is not too much effort for you to apply the changes. The good thing is that we use a generator. There are already examples of switching functions (Supplier, Function, BiFunction).

Thank you very much!

valery1707:
@danieldietrich, fixed:
* `currying instead of Tuples` applied to `recover` and `unchecked`
* `simplify method argument types` applied to `recover`, `unchecked` and `liftTry`

danieldietrich:
This, Sir, looks awesome! Many thanks.

PR #1387, opened 2016-06-17
new static function Function*.constant

emmanueltouzery:
taking as a model [the haskell const function](http://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html#v:const), this adds `Function*.constant`. I regret the longer name, but `const` is a keyword in java, and right now I can't think of a shorter, clear name. Maybe `Function*.value()` but probably not (?).
I think if java allowed `_` as a variable name we wouldn't really need this, you could just write `_ -> 7` for instance. But since we can't have that, it does feel clunky to write `x -> 7` where x stands for nothing and is useless.

But then `x -> 7` is considerably shorter than `Function1.constant(7)` (and comparable to `constant(7)` if you static import)... So I think there's really valid ground to reject this PR. I'll let you judge...

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -751,6 +751,17 @@ def generateMainClasses(): Unit = {
                   return $i;
               }
 
+              /**
+               * Returns a function that always returns the constant
+               * value that you give in parameter.
+               *
+               * @param value the value to be returned
+               * @return a function always returning the given value
+               */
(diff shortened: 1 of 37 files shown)


danieldietrich:
Great, thank you. I find it also useful. Java has `Function.identity()`, which can be also written `t -> t`, so adding a function `constant` is pretty legitimate.

Issue #1386, opened 2016-06-16
Check generic bounds and widen argument types of (Checked)Function.*


danieldietrich:
Similar to #1385 we **Simplify method _argument_ types**

(Function0..2 have super interface in java.util.function)

* Instead of `Function0<R>` we write `Supplier<R>`
* Instead of `Function1<T1, R>` we write `Function<T1, R>`
* Instead of `Function2<T1, T2, R>` we write `BiFunction<T1, T2, R>`
* For CheckedFunctions we cannot make these 'optimizations'

danieldietrich:
Also some generic types may need to be (co)variant (see also `recover` and `unchecked`).

danieldietrich:
Fixed with #1395

PR #1395, opened 2016-06-20
Wide generic argument in function

valery1707:
For issue #1386

### DIFF
--- a/javaslang/src-gen/main/java/javaslang/CheckedFunction0.java
+++ b/javaslang/src-gen/main/java/javaslang/CheckedFunction0.java
@@ -63,8 +63,8 @@ public interface CheckedFunction0<R> extends λ<R> {
      * @param <R> return type
      * @return a {@code CheckedFunction0}
      */
-    static <R> CheckedFunction0<R> of(CheckedFunction0<? extends R> methodReference) {
-        return methodReference::apply;
+    static <R> CheckedFunction0<R> of(CheckedFunction0<R> methodReference) {
+        return methodReference;
     }
 
     /**
@@ -75,8 +75,9 @@ public interface CheckedFunction0<R> extends λ<R> {
      * @return a function that applies arguments to the given {@code partialFunction} and returns {@code Some(result)}
      *         if the function is defined for the given arguments, and {@code None} otherwise.
      */
+    @SuppressWarnings("RedundantTypeArguments")
     static <R> Function0<Option<R>> lift(CheckedFunction0<? extends R> partialFunction) {
-        return () -> Try.of(of(partialFunction)::apply).getOption();
+        return () -> Try.<R>of(partialFunction::apply).getOption();
     }
 
     /**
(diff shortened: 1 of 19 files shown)


javaslang/src-gen/main/java/javaslang/CheckedFunction1.java
```diff
@@ -79,7 +79,7 @@
      * @return a function that applies arguments to the given {@code partialFunction} and returns {@code Some(result)}
      *         if the function is defined for the given arguments, and {@code None} otherwise.
      */
-    static <T1, R> Function1<T1, Option<R>> lift(CheckedFunction1<T1, R> partialFunction) {
+    static <T1, R> Function1<T1, Option<R>> lift(CheckedFunction1<? super T1, R> partialFunction) {
```
danieldietrich: I think we will also need to widen the result type R: `? extends R` for al occurrences (lift, liftTry, recover).

Could you also widen the parameters (`? super`for args and `? extends` for return type) for the method `of(...)` of CheckedFunction and Function?

valery1707:
@danieldietrich, fixed

danieldietrich:
Thank you for applying my suggestions. Looking at the changes of `of`, `lift` and `liftTry` I'm double-minded if the bounds are really needed. I need to sleep one night about the topic. Maybe widening the bounds does not make sense for one or more of the methods I mentioned above.

Sorry for being undecided here.

danieldietrich:
Okay, we are nearly finished.

Please revert the changes for the `of` Method. The function argument has to be the same type like the result function. The `of` method does not use generic type bounds.

Example:

```java
static <T1, T2, R> Function2<T1, T2, R> of(Function2<T1, T2, R> methodReference) {
    return methodReference;
}
```

Explanation:

The purpose of the `of` method is to turn a lambda expression into a Javaslang function. The arg is returned as-is. It does only work well with lambda expressions. The types of method references cannot be inferred by Java.

javaslang/src-gen/main/java/javaslang/CheckedFunction0.java
```diff
@@ -75,8 +75,8 @@
      * @return a function that applies arguments to the given {@code partialFunction} and returns {@code Some(result)}
      *         if the function is defined for the given arguments, and {@code None} otherwise.
      */
-    static <R> Function0<Option<R>> lift(CheckedFunction0<R> partialFunction) {
-        return () -> Try.of(partialFunction::apply).getOption();
+    static <R> Function0<Option<R>> lift(CheckedFunction0<? extends R> partialFunction) {
+        return () -> Try.of(of(partialFunction)::apply).getOption();
```
danieldietrich: We can replace the method call `of(partialFunction)` by providing a type hint `<R>`:

```java
return () -> Try.<R> of(partialFunction::apply).getOption();
```

valery1707:
@danieldietrich, fixed

danieldietrich:
Great, thank you!

PR #1396, opened 2016-06-20
Deprecate Tuple*.transform and rename them to Tuple*.apply

### LINKED ISSUE #1314: Deprecate Tuple*.transform and rename them to Tuple*.apply
We apply a function to a tuple by applying the tuple components to the function. The name _transform_ is confusing.

**Example:** Tuple2.apply(BiFunction)

```java
t(f) := f(t._1, t._2)
```

We add `apply` and deprecate `transform` starting with 2.1.0. In 3.0.0 `transform` will be removed.

valery1707:
Fix #1314

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -469,10 +469,10 @@ def generateMainClasses(): Unit = {
                                       return $OptionType.none();
                                   } else {
                                       ${if (i == 1) xs"""
-                                        return unapply.apply(obj).transform(u1 -> ((Pattern<U1, ?>) p1).apply(u1).map(_1 -> (T1) u1));
+                                        return unapply.apply(obj).apply(u1 -> ((Pattern<U1, ?>) p1).apply(u1).map(_1 -> (T1) u1));
                                       """ else xs"""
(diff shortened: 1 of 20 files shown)


javaslang/generator/Generator.scala
```diff
@@ -1136,6 +1136,28 @@ def generateMainClasses(): Unit = {
...
+             * @deprecated Use {@link #apply($functionType)} instead, will be removed in 3.0.0
+             * @param f Transformation which creates a new object of type U based on this tuple's contents.
+             * @param <U> type of the transformation result
+             * @return An object of type U
+             * @throws NullPointerException if {@code f} is null
+             */
+            @Deprecated(/* Use apply instead, will be removed in 3.0.0 */)
+            ${if (i == 0) xs"""
               public <U> U transform($functionType<? extends U> f) {
                   $Objects.requireNonNull(f, "f is null");
```
danieldietrich: Great, thank you!

I think it does not really matter (because it is generated code) but could you delegate to `apply`?

```scala
public <U> U transform($functionType<? extends U> f) {
    return apply(f);
}
```

danieldietrich:
Thx!

PR #1398, opened 2016-06-22
Benchmark updates

paplorinc:
* Fixed parameter and benchmark order sorting.
* Added `Clojure` comparisons to `List` and `Vector`.
* Added tests for each benchmark in quick mode, with internal assertions (which are now disabled if run from `main`) and enabled the tests to run with every build (~3 additional minutes).
* Added precision-mode run which runs all benchmarks in slow mode.
* Running only the specified groups in a given order is now possible from `main` (needed production code modification also, see the comments).
* Added asserts and returns to every benchmark.

I recommend the review [commit-by-commit](https://github.com/javaslang/javaslang/pull/1398/commits), as heavy refactorings were applied.

Any comments are welcome @danieldietrich, @eduardmanas.

```
Ratios slang / <alternative_impl>
Target                  Operation  Ratio                                           10       32      100     1000     1024    32768 
ArrayBenchmark          Append     slang_persistent/fjava_mutable               1.93x             1.62x    1.12x                  
ArrayBenchmark          Append     slang_persistent/java_mutable                0.26x             0.15x    0.02x                  

ArrayBenchmark          Create     slang_persistent/fjava_persistent           20.55x            43.75x   49.13x                  
ArrayBenchmark          Create     slang_persistent/java_mutable                1.19x             1.04x    1.02x                  

ArrayBenchmark          Get        slang_persistent/fjava_mutable               0.92x             0.99x    1.00x                  
ArrayBenchmark          Get        slang_persistent/java_mutable                1.00x             0.99x    1.00x                  

ArrayBenchmark          Head       slang_persistent/fjava_mutable               0.99x             0.99x    0.99x                  
ArrayBenchmark          Head       slang_persistent/java_mutable                1.01x             1.00x    1.00x                  

ArrayBenchmark          Iterate    slang_persistent/fjava_mutable               1.21x             1.07x    1.41x                  
ArrayBenchmark          Iterate    slang_persistent/java_mutable                0.96x             1.03x    0.99x                  

ArrayBenchmark          Prepend    slang_persistent/fjava_mutable               1.11x             1.05x    1.03x                  
ArrayBenchmark          Prepend    slang_persistent/java_mutable                0.70x             0.57x    0.28x                  

ArrayBenchmark          Tail       slang_persistent/java_mutable                3.91x             1.53x    0.36x                  

ArrayBenchmark          Update     slang_persistent/fjava_mutable               0.03x             0.00x    0.00x                  
[...]

PR #1402, opened 2016-06-25
Changed and unified all `groupBy` methods to keep insertion order

paplorinc:
Separated it from https://github.com/javaslang/javaslang/pull/1398 upon the [request](https://github.com/javaslang/javaslang/pull/1398#discussion_r68176461) of @ruslansennov.

This way the output of the benchmarks will have has the same order as its definition, not random.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/AbstractMap.java
+++ b/javaslang/src/main/java/javaslang/collection/AbstractMap.java
@@ -5,16 +5,11 @@
  */
 package javaslang.collection;
 
-import javaslang.Tuple;
-import javaslang.Tuple2;
+import javaslang.*;
 import javaslang.control.Option;
 
-import java.util.Comparator;
-import java.util.Objects;
-import java.util.function.BiFunction;
-import java.util.function.Consumer;
-import java.util.function.Function;
-import java.util.function.Predicate;
+import java.util.*;
+import java.util.function.*;
 
 /**
  * An abstract {@link Map} implementation (not intended to be public).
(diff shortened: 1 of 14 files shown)


javaslang/src/main/java/javaslang/collection/AbstractMap.java
```diff
@@ -134,14 +129,7 @@ public M filter(Predicate<? super Tuple2<K, V>> predicate) {
...
     public <C> Map<C, M> groupBy(Function<? super Tuple2<K, V>, ? extends C> classifier) {
-        Objects.requireNonNull(classifier, "classifier is null");
-        return foldLeft(HashMap.empty(), (map, entry) -> {
-            final C key = classifier.apply(entry);
-            final Map<K, V> values = map.get(key)
-                    .map(entries -> entries.put(entry._1, entry._2))
-                    .getOrElse(createFromEntries(Iterator.of(entry)));
-            return map.put(key, (M) values);
-        });
+        return Collections.groupBy(this, classifier, this::createFromEntries);
```
paplorinc: unified all `groupBy` methods to use the same template

javaslang/src/main/java/javaslang/collection/Collections.java
```diff
@@ -53,13 +48,13 @@
         Objects.requireNonNull(classifier, "classifier is null");
         Objects.requireNonNull(mapper, "mapper is null");
 
-        final java.util.Map<C, Collection<T>> mutableResults = new java.util.HashMap<>();
+        final java.util.Map<C, Collection<T>> mutableResults = new java.util.LinkedHashMap<>();
```
paplorinc: @danieldietrich, @ruslansennov, this is where the magic happens, i.e. all `groupBy`s using a `LinkedHashMap` instead of a `HashMap`

ruslansennov: Thank you! I can't review this PR now (out of my console), but it seems it significantly shorter than prev

danieldietrich:
Great, thank you! The changes look good to me.

I think the formatting might change again (whitespace) because we use different formatters. But that is of minor importance.

PR #1415, opened 2016-07-04
List Tail benchmark java_linked_mutable should use LinkedList

shelajev:
List Tail benchmark java_linked_mutable should use LinkedList rather than ArrayList as the collection under test.


danieldietrich:
Hi Oleg, thank you - good point :)

PR #1419, opened 2016-07-06
Iterator#iterate should not call supplier function until value is rea…

jorander:
…lly needed.

Calling the supplier function to generate the next value "one value ahead" is a really counter intuitive behaviour.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Iterator.java
+++ b/javaslang/src/main/java/javaslang/collection/Iterator.java
@@ -968,8 +968,11 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
     static <T> Iterator<T> iterate(T seed, Function<? super T, ? extends T> f) {
         Objects.requireNonNull(f, "f is null");
         return new AbstractIterator<T>() {
-            Function<? super T, ? extends T> nextFunc = (s) -> seed;
-            T previous = null;
+            Function<? super T, ? extends T> nextFunc = s -> {
+                nextFunc = f;
+                return seed;
+            };
+            T current = null;
 
             @Override
             public boolean hasNext() {
@@ -978,9 +981,8 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
 
             @Override
             public T getNext() {
-                previous = nextFunc.apply(previous);
-                nextFunc = f;
-                return previous;
+                current = nextFunc.apply(current);
+                return current;
             }
         };
     }
--- a/javaslang/src/main/java/javaslang/collection/Iterator.java
+++ b/javaslang/src/main/java/javaslang/collection/Iterator.java
@@ -968,7 +968,8 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
     static <T> Iterator<T> iterate(T seed, Function<? super T, ? extends T> f) {
         Objects.requireNonNull(f, "f is null");
         return new AbstractIterator<T>() {
-            T next = seed;
+            Function<? super T, ? extends T> nextFunc = (s) -> seed;
+            T previous = null;
 
             @Override
             public boolean hasNext() {
@@ -977,9 +978,9 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
 
             @Override
             public T getNext() {
-                final T result = next;
-                next = f.apply(next);
-                return result;
+                previous = nextFunc.apply(previous);
+                nextFunc = f;
+                return previous;
             }
         };
     }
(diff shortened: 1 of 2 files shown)


jorander:
Hi,

Been away from Javaslang for a while and coming back I find it more impressive than ever. Great job done!

I was going over and bringing my own Euler solutions up-to-date with Javaslang 2.1.0 and since I use another approach to generating prime numbers (problem 7) than you have in the current solution I ran into this issue. The problem I got was because I use the already generated primes in the function generating the next. Due to the early call to calculate "next" I got an infinite loop and hence a StackOverflowException.

Best Regards,
Jörgen

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -968,7 +968,8 @@ public T getNext() {
     static <T> Iterator<T> iterate(T seed, Function<? super T, ? extends T> f) {
         Objects.requireNonNull(f, "f is null");
         return new AbstractIterator<T>() {
-            T next = seed;
+            Function<? super T, ? extends T> nextFunc = (s) -> seed;
```
paplorinc: I agree that iterator should be totally lazy, thanks for providing a PR.
However, it may be simpler if we implemented it without a supplier, e.g.:
```java
    T current;

    @Override
    public boolean hasNext() {
        return true;
    }

    @Override
    public T getNext() {
        current = (current == null) ? seed : f.apply(current);
        return current;
    }
};
```

javaslang/src/test/java/javaslang/collection/IteratorTest.java
```diff
@@ -360,6 +360,11 @@ public void shouldGenerateInfiniteStreamBasedOnSupplierWithAccessToPreviousValue
         assertThat(iterate(2, (i) -> i + 2).take(3).reduce((i, j) -> i + j)).isEqualTo(12);
     }
 
+    @Test
+    public void shouldNotCallSupplierUntilNecessary() {
+        assertThat(iterate(2, (i) -> {throw new RuntimeException("The supplier function should not be called since the next value is not needed in order to get head.");}).head()).isEqualTo(2);
```
paplorinc: my preference would be to ommit the text in the Exception, the name of the test already explains the intent :)

javaslang/src/main/java/javaslang/collection/Iterator.java
jorander: Yes, simpler, but perhaps too simple. What if null happens to be a valid value somewhere in the stream? Then we would return the seed next time getNext() is called.

javaslang/src/test/java/javaslang/collection/IteratorTest.java
jorander: Yeah, I agree. The name might be enough.

javaslang/src/main/java/javaslang/collection/Iterator.java
paplorinc: @danieldietrich will decide anyway, and he doesn't hate `null`s as much as I do :).
I personally would simply document in this case, not to use `null` irresponsibly :)

ruslansennov: Javaslang accepts null elements as ordinary elements

danieldietrich:
Hi @jorander, good to see you again :)

Thanks! @ruslansennov is right, we allow `null`'s in Javaslang. /cc @paplorinc 

I will review it this evening...

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: Yep, we should embrace `null` because it is Java we use. Java will have `null`s forever and so we should support this billion dollar mistake (or use another language which does not have `null`).

Please use this slightly optimized version:

```java
    static <T> Iterator<T> iterate(T seed, Function<? super T, ? extends T> f) {
            Function<? super T, ? extends T> nextFunc = s -> {
                nextFunc = f;
                return seed;
            };
            T next = null;

            @Override
            public boolean hasNext() {
                return true;
            }

            @Override
            public T getNext() {
                return next = nextFunc.apply(next);
            }
        };
    }
```

You may write

```java
            public T getNext() {
                next = nextFunc.apply(next);
                return next;
            }
```

if you are not comfortable with having a return statement and an assignment in one line. At least @paplorinc will not be comfortable with it ;-p

---
one side-note about style: we write `s -> ...` instead of `(s) -> ...` in case of single arg lambdas..

danieldietrich: I changed the name `prev` to `next` because we recently started to implement navigable collections having a reverse Iterator. The name `prev` is therefore too confusing.

paplorinc: isn't this rather the `current` element?
It's confusing that we're storing a `next`, that's actually the `previous` :/

danieldietrich: You're right. `current` or `curr` is the best name

jorander:
Ok, thanks for all the input. I'll try to be back with an updated version later tonight.

/Jörgen

danieldietrich:
Thank you Jörgen, nice finding!! This one is really deep down the rabbit hole of functional laziness :-)

Issue #1418, opened 2016-07-06
TreeMap#keyComparator fails if TreeMap has been created with TreeMap#ofAll


goerge:
Following code fails

```
TreeMap.ofAll(Collections.singletonMap("key", "value")).keyComparator();
```

with

```
java.lang.ClassCastException: javaslang.collection.RedBlackTree$$Lambda$1/761960786 cannot be cast to javaslang.collection.TreeMap$EntryComparator

	at javaslang.collection.TreeMap.keyComparator(TreeMap.java:455)
```

Same error occurs when `keySet()` is called because `keyComparator()` is called inside.

danieldietrich:
Thank you!

PR #1420, opened 2016-07-06
Fix TreeMap creation from Java Map

paplorinc:
Fixes https://github.com/javaslang/javaslang/issues/1418, @ruslansennov, @danieldietrich and @goerge, please review.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/TreeMap.java
+++ b/javaslang/src/main/java/javaslang/collection/TreeMap.java
@@ -167,7 +167,7 @@ public final class TreeMap<K, V> extends AbstractMap<K, V, TreeMap<K, V>> implem
      */
     public static <K extends Comparable<? super K>, V> TreeMap<K, V> ofAll(java.util.Map<? extends K, ? extends V> map) {
         Objects.requireNonNull(map, "map is null");
-        RedBlackTree<Tuple2<K, V>> result = RedBlackTree.empty();
+        RedBlackTree<Tuple2<K, V>> result = RedBlackTree.empty(new EntryComparator<>((Comparator<? super K> & Serializable) K::compareTo));
         for (java.util.Map.Entry<? extends K, ? extends V> entry : map.entrySet()) {
             result = result.insert(Tuple.of(entry.getKey(), entry.getValue()));
         }
--- a/javaslang/src/main/java/javaslang/collection/TreeMap.java
+++ b/javaslang/src/main/java/javaslang/collection/TreeMap.java
@@ -5,11 +5,16 @@
  */
 package javaslang.collection;
 
-import javaslang.*;
+import javaslang.Kind2;
+import javaslang.Tuple;
+import javaslang.Tuple2;
 import javaslang.control.Option;
 
 import java.io.Serializable;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.function.*;
 import java.util.stream.Collector;
 
@@ -35,29 +40,29 @@ public final class TreeMap<K, V> extends AbstractMap<K, V, TreeMap<K, V>> implem
     }
 
     /**
-     * Returns a {@link java.util.stream.Collector} which may be used in conjunction with
-     * {@link java.util.stream.Stream#collect(java.util.stream.Collector)} to obtain a
-     * {@link javaslang.collection.TreeMap}.
+     * Returns a {@link Collector} which may be used in conjunction with
+     * {@link java.util.stream.Stream#collect(Collector)} to obtain a
+     * {@link TreeMap}.
      * <p>
      * The natural comparator is used to compare TreeMap keys.
      *
      * @param <K> The key type
      * @param <V> The value type
-     * @return A {@link javaslang.collection.TreeMap} Collector.
+     * @return A {@link TreeMap} Collector.
      */
     public static <K extends Comparable<? super K>, V> Collector<Tuple2<K, V>, ArrayList<Tuple2<K, V>>, TreeMap<K, V>> collector() {
         return collector((Comparator<? super K> & Serializable) K::compareTo);
     }
 
     /**
-     * Returns a {@link java.util.stream.Collector} which may be used in conjunction with
-     * {@link java.util.stream.Stream#collect(java.util.stream.Collector)} to obtain a
-     * {@link javaslang.collection.TreeMap}.
+     * Returns a {@link Collector} which may be used in conjunction with
+     * {@link java.util.stream.Stream#collect(Collector)} to obtain a
+     * {@link TreeMap}.
      *
      * @param <K>           The key type
      * @param <V>           The value type
      * @param keyComparator A key comparator
-     * @return A {@link javaslang.collection.TreeMap} Collector.
+     * @return A {@link TreeMap} Collector.
      */
     public static <K, V> Collector<Tuple2<K, V>, ArrayList<Tuple2<K, V>>, TreeMap<K, V>> collector(Comparator<? super K> keyComparator) {
         Objects.requireNonNull(keyComparator, "keyComparator is null");
@@ -67,7 +72,7 @@ public final class TreeMap<K, V> extends AbstractMap<K, V, TreeMap<K, V>> implem
             left.addAll(right);
             return left;
         };
-        final Function<ArrayList<Tuple2<K, V>>, TreeMap<K, V>> finisher = list -> TreeMap.ofEntries(keyComparator, list);
+        final Function<ArrayList<Tuple2<K, V>>, TreeMap<K, V>> finisher = list -> ofEntries(keyComparator, list);
         return Collector.of(supplier, accumulator, combiner, finisher);
     }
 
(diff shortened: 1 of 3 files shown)


javaslang/src/main/java/javaslang/collection/TreeMap.java
```diff
@@ -162,7 +167,7 @@ private TreeMap(RedBlackTree<Tuple2<K, V>> entries) {
      */
     public static <K extends Comparable<? super K>, V> TreeMap<K, V> ofAll(java.util.Map<? extends K, ? extends V> map) {
         Objects.requireNonNull(map, "map is null");
-        RedBlackTree<Tuple2<K, V>> result = RedBlackTree.empty();
+        RedBlackTree<Tuple2<K, V>> result = RedBlackTree.empty(new EntryComparator<>(Comparator.naturalOrder()));
```
paplorinc: `K` isn't comparable, so we should hope the natural order works

javaslang/src/test/java/javaslang/collection/TreeMapTest.java
```diff
@@ -126,6 +125,12 @@ public void shouldWrapMap() {
         assertThat(TreeMap.ofAll(source)).isEqualTo(emptyIntInt().put(1, 2).put(3, 4));
     }
 
+    @Test
+    public void shouldCreateKeyComparatorForJavaUtilMap() {
+        final TreeMap<String, String> map = TreeMap.ofAll(Collections.singletonMap("key", "value"));
+        assertThat(map.keyComparator()).isNotNull();
```
paplorinc: should I assert more than this?

javaslang/src/main/java/javaslang/collection/TreeMap.java
ruslansennov: `K` **is** comparable because `K extends Comparable<? super K>`. We can't use natural comparator here. I believe this is correct fix:

```java
final Comparator<Tuple2<K, V>> comparator = new EntryComparator<>((Comparator<? super K> & Serializable) K::compareTo);
RedBlackTree<Tuple2<K, V>> result = RedBlackTree.empty(comparator);
```

ruslansennov: Hmmm...
I looked inside `Comparator.naturalOrder()`, it seems that your solution also works :)

danieldietrich:
Thank you, will check that this evening...

javaslang/src/test/java/javaslang/collection/TreeMapTest.java
danieldietrich: We could ensure that the natural Tuple2 comparator is used, namely the component-wise comparator.

An additional test for this would create `TreeMap.of(Tuple.of("b", "b"), Tuple.of("a", "b"), Tuple.of("b", "a")))`.

Don't know what's the best way to check the validity. Maybe `assertThat(treeMap.toList()).isEqualTo(List.of(Tuple.of("a", "b"), Tuple.of("b", "a"), Tuple.of("b", "b")))`?

javaslang/src/main/java/javaslang/collection/TreeMap.java
danieldietrich: @paplorinc, @ruslansennov We use the `naturalComparator()` only in the case we have no Comparable (generic) in scope. If we have a Comparable, like K here, we are able to make it Serializable like Ruslan showed. We use that solution in RedBlackTree for example.

So I would prefer to use

```java
final Comparator<Tuple2<K, V>> comparator = new EntryComparator<>((Comparator<? super K> & Serializable) K::compareTo);
RedBlackTree<Tuple2<K, V>> result = RedBlackTree.empty(comparator);
```

because the `naturalComparator()` is not Serializable.

We should write a simple Serialize/Deserialize test for the empty case here. There is a helper for that: `Serializables`, e.g.

```java
final CharSequence s = deserialize(serialize("test"));
assertThat(s).isEqualTo("test");
```

javaslang/src/main/java/javaslang/collection/TreeMap.java
```diff
@@ -5,11 +5,16 @@
  */
 package javaslang.collection;
 
-import javaslang.*;
+import javaslang.Kind2;
```
paplorinc: The modified files had too many warnings, reduced and extracted them to simplify the review of next commit

javaslang/src/main/java/javaslang/collection/TreeMap.java
```diff
@@ -167,7 +167,7 @@ private TreeMap(RedBlackTree<Tuple2<K, V>> entries) {
      */
     public static <K extends Comparable<? super K>, V> TreeMap<K, V> ofAll(java.util.Map<? extends K, ? extends V> map) {
         Objects.requireNonNull(map, "map is null");
-        RedBlackTree<Tuple2<K, V>> result = RedBlackTree.empty();
+        RedBlackTree<Tuple2<K, V>> result = RedBlackTree.empty(new EntryComparator<>((Comparator<? super K> & Serializable) K::compareTo));
```
paplorinc: inlined it in the end, as that's how it's done in other similar cases also

javaslang/src/test/java/javaslang/collection/AbstractMapTest.java
```diff
@@ -751,6 +751,13 @@ public void shouldPreserveSingletonInstanceOnDeserialization() {
         assertThat(actual).isTrue();
     }
 
+    @Test
+    public void shouldSerializeDeserializeNonEmptyMap() {
+        final Object expected = mapOfTuples(Tuple.of(1, "a"), Tuple.of(2, "b"), Tuple.of(3, "c"));
```
paplorinc: this passes without the fix also ... is this what you wanted, @danieldietrich?

javaslang/src/test/java/javaslang/collection/TreeMapTest.java
```diff
@@ -122,6 +126,22 @@ public void shouldWrapMap() {
...
+    @Test
+    public void shouldCreateKeyComparatorForJavaUtilMap() {
+        final TreeMap<String, Integer> actual = TreeMap.ofAll(mapOfTuples(Tuple.of("c", 0), Tuple.of("a", 0), Tuple.of("b", 0)).toJavaMap());
+        final List<String> expected = List.of("a", "b", "c");
+        assertThat(actual.keySet().toList()).isEqualTo(expected);
+    }
+
+    @Test
+    public void shouldSerializeDeserializeNonEmptyMap() {
+        final Object expected = TreeMap.ofAll(Collections.singletonMap(0, 1));
```
paplorinc: same here

javaslang/src/test/java/javaslang/collection/TreeMapTest.java
```diff
@@ -122,6 +126,22 @@ public void shouldWrapMap() {
         assertThat(TreeMap.ofAll(source)).isEqualTo(emptyIntInt().put(1, 2).put(3, 4));
     }
 
+    // -- ofAll
+
+    @Test
+    public void shouldCreateKeyComparatorForJavaUtilMap() {
+        final TreeMap<String, Integer> actual = TreeMap.ofAll(mapOfTuples(Tuple.of("c", 0), Tuple.of("a", 0), Tuple.of("b", 0)).toJavaMap());
+        final List<String> expected = List.of("a", "b", "c");
+        assertThat(actual.keySet().toList()).isEqualTo(expected);
```
paplorinc: `toList()` doesn't call the `keyComparator`, only `keySet()` does.
Is this ok, @danieldietrich?

javaslang/src/main/java/javaslang/collection/TreeMap.java
danieldietrich: ok

javaslang/src/test/java/javaslang/collection/TreeMapTest.java
danieldietrich: Yes, great!

danieldietrich: @paplorinc But wait - what is the result of `actual.toList()`? Because `TreeMap<K, V>` is and ordered Traversable<Tuple2<K,V>>` I would expect that `toList()` returns the contained entries in the underlying order. If that isn't the case it is a bug.

paplorinc: That's why I asked. Could you please investigate if for me?

danieldietrich: Yes, will do!

danieldietrich: @paplorinc This works on my current RedBlackTree branch:

```java
    @Test
    public void shouldCreateKeyComparatorForJavaUtilMap() {
        final TreeMap<String, Integer> actual = TreeMap.ofAll(List.of(Tuple.of("c", 0), Tuple.of("a", 0), Tuple.of("b", 0)).toJavaMap(t -> t));
        final List<Tuple2<String, Integer>> expected = List.of(Tuple.of("a", 0), Tuple.of("b", 0), Tuple.of("c", 0));
        assertThat(actual.toList()).isEqualTo(expected);
    }
```

Could you replace your test with it, please. No bug here as far as I can see...

paplorinc: Does it fail before?

danieldietrich: All right. It passed before. Just removed `mapOfTuples` because it creates a TreeMap to create a TreeMap. But here it does not really matter.

danieldietrich:
Thank you!! Will pull it now.

PR #1422, opened 2016-07-07
Added solution for Euler number 12.

jorander:
I'm doing this to learn FP and Javaslang, so all kind of comments are highly appreciated.

BR,
Jörgen


ashrko619:
Hi @jorander ,
Your code works fine :+1: But there is a much faster algorithm based on the fact that n, n+1 are co-primes. See this http://codereview.stackexchange.com/a/83680  :)

jorander:
Hi @ashrko619 ,
Thanks for pointing that out. If I understand the answer you linked correctly it is the calculation of number of divisors that could be smarter. That would mean calculating the number of divisors directly, taking advantage of some more involved mathematical idiom, rather than first creating a stream of factors and then count them.

I'm sure it will be faster, but I'm not sure that is the right path to go for the Euler problems as tests for Javaslang. I think the main objective here is to validate and show case Javaslang's capabilities as an API for functional programming in Java. In doing so I would like to keep the logic as simple as possible and the solutions as close as possible to the problem statement so that any reader should be able to focus on the Javaslang constructs rather than pondering the mathematical logic we use to arrive at the right solution. From that perspective I think the more naïve solution I used is preferable.

Other point-of-views? //cc @danieldietrich 

Regards,
Jörgen

javaslang/src/test/java/javaslang/collection/euler/Euler12Test.java
```diff
@@ -0,0 +1,61 @@
...
+     * What is the value of the first triangle number to have over five hundred
+     * divisors?
+     */
+    @Test
+    public void shouldSolveProblem12() {
+        assertThat(valueOfFirstTriangleNumberWithMoreDivisorsThan(5)).isEqualTo(28);
+        assertThat(valueOfFirstTriangleNumberWithMoreDivisorsThan(500)).isEqualTo(76_576_500);
+    }
+
+    private static long valueOfFirstTriangleNumberWithMoreDivisorsThan(long nofDivisors) {
```
paplorinc: `nof`?
I think `divisorCount` would make more sense?

javaslang/src/test/java/javaslang/collection/euler/Euler12Test.java
```diff
@@ -0,0 +1,61 @@
...
+     * We can see that 28 is the first triangle number to have over five
+     * divisors.
+     * <p>
+     * What is the value of the first triangle number to have over five hundred
+     * divisors?
+     */
+    @Test
+    public void shouldSolveProblem12() {
+        assertThat(valueOfFirstTriangleNumberWithMoreDivisorsThan(5)).isEqualTo(28);
+        assertThat(valueOfFirstTriangleNumberWithMoreDivisorsThan(500)).isEqualTo(76_576_500);
```
paplorinc: how did you come up with the end result :)?

javaslang/src/test/java/javaslang/collection/euler/Utils.java
```diff
@@ -27,6 +27,13 @@ public static BigInteger factorial(int n) {
         return Stream.rangeClosed(1, n).map(BigInteger::valueOf).fold(BigInteger.ONE, BigInteger::multiply);
     }
 
+    public static Stream<Long> factors(long l) {
```
paplorinc: please rename `l`, it looks like `1` :)
```java
factors(long number)
```

javaslang/src/test/java/javaslang/collection/euler/Euler12Test.java
```diff
@@ -0,0 +1,61 @@
...
+        return triangleNumbers()
+                .find(t -> nofDivisors(t) > nofDivisors)
+                .get();
+    }
+
+    private static Stream<Long> triangleNumbers() {
+        return Stream.from(1L).scanLeft(0L, (a, l) -> a + l);
+    }
+
+    private static long nofDivisors(long l) {
```
paplorinc: same

javaslang/src/test/java/javaslang/collection/euler/Utils.java
ruslansennov: please stop this flood

javaslang/src/test/java/javaslang/collection/euler/Euler12Test.java
jorander: You mean that the first triangle number with more than 500 divisors is 76 576 500? I calculated it using my program and then went to https://projecteuler.net/problem=12 to enter my answer and get it evaluated. It turned out to be right.

danieldietrich:
> From that perspective I think the more naïve solution I used is preferable.
> Other point-of-views? //cc @danieldietrich

I also prefer the more naïve solution. My intention of having Eule tests is to implement real world examples with the help of the Javaslang API. The more concise the code is, the better. Often optimization means low level, plain Java code. That's not the goal here.

> Is there an equivalent for distributing the workload for data in a Javaslang Stream over more cores?

I had it in mind but not further investigated it yet. The _work_ has to be splitted into parts and the result have to be joined again (in the right order, if there is an order). Most of the Javaslang types are Iterable, i.e. they have a method `splitterator()`. Javaslang's Stream is a _linear_ sequence, i.e. hard/impossible to split. Java's Stream works in a different way. It is not a Stream as we know it from Scala. In Javaslang maybe we could leverage the Future implementation to move parts of a computation to different threads but I have to think about an API here...

danieldietrich:
Thank you very much!

PR #1424, opened 2016-07-08
Added solution for Euler number 17.

jorander:
I think this problem is a good show-case for the Pattern Matching API.


danieldietrich:
Awesome, thank you! This looks really great :-)

PR #1426, opened 2016-07-09
Some minor enhancements to solution for Euler number 17 to showcase P…

jorander:
…attern Matching even more and make the code cleaner.


danieldietrich:
Great, thank you!

Issue #1313, opened 2016-05-01
Add method zipWith(Iterable, BiFunction) as shortcut for zip(Iterable).map(Function)


danieldietrich:
We should add `zipWith` to all types that already have `zip`.

Here is an example of an implementation in Stream. Because Stream is lazy evaluated, other types that are eager evaluated may have another implementation.

``` java
default <U, R> Stream<R> zipWith(Iterable<? extends U> iterable, BiFunction<? super T, ? super U, ? extends R> combiner) {
    return zip(iterable).map(t -> combiner.apply(t._1, t._2));
}
```

danieldietrich:
**Note:** It would be more _functional_ to switch the arguments if we had curried argument lists in Java (like in Scala):

```java
// zip = stream1.zipWith(f); zipped = zip.apply(stream2);
default <U, R> Stream<R> zipWith(BiFunction<? super T, ? super U, ? extends R> combiner)(Iterable<? extends U> iterable) {
    ...
}
```

or have a static method to obtain a curried form:

```java
// zip = zipWith(f); zipped = zip.apply(stream1, stream2);
static <T, U, R> BiFunction<Stream<T>, Stream<U>, Stream<R>> zipWith(BiFunction<? super T, ? super U, ? extends R> combiner) {
    ...
}
```

But since Java is an object-oriented language, this seems more natural:

```java
// = stream1.zip(stream2).map(t -> f.apply(t._1, t._2))
stream1.zipWith(stream2, f);
```

valery1707:
This issue was implemented by #1427 isn't it?

danieldietrich:
Yes, thx, will close it.

PR #1427, opened 2016-07-09
add zipWith & zipWithIndex with BiFunction

ronanM:
Added on Traversable:
- ```<U, R> Traversable<R> zipWith(Iterable<? extends U> that, BiFunction<? super T, ? super U, ? extends R> mapper);```
- ```<U> Traversable<U> zipWithIndex(BiFunction<? super T, ? super Long, ? extends U> mapper);```

See #1313

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/LinkedHashSet.java
+++ b/javaslang/src/main/java/javaslang/collection/LinkedHashSet.java
@@ -834,8 +834,14 @@ public final class LinkedHashSet<T> implements Kind1<LinkedHashSet<?>, T>, Set<T
 
     @Override
     public <U> LinkedHashSet<Tuple2<T, U>> zip(Iterable<? extends U> that) {
+        return zipWith(that, Tuple::of);
+    }
+
+    @Override
+    public <U, R> LinkedHashSet<R> zipWith(Iterable<? extends U> that, BiFunction<? super T, ? super U, ? extends R> mapper) {
         Objects.requireNonNull(that, "that is null");
-        return LinkedHashSet.ofAll(iterator().zip(that));
+        Objects.requireNonNull(mapper, "mapper is null");
+        return LinkedHashSet.ofAll(iterator().zipWith(that, mapper));
     }
 
     @Override
@@ -846,7 +852,13 @@ public final class LinkedHashSet<T> implements Kind1<LinkedHashSet<?>, T>, Set<T
 
     @Override
     public LinkedHashSet<Tuple2<T, Long>> zipWithIndex() {
-        return LinkedHashSet.ofAll(iterator().zipWithIndex());
+        return zipWithIndex(Tuple::of);
+    }
+
+    @Override
+    public <U> LinkedHashSet<U> zipWithIndex(BiFunction<? super T, ? super Long, ? extends U> mapper) {
+        Objects.requireNonNull(mapper, "mapper is null");
+        return LinkedHashSet.ofAll(iterator().zipWithIndex(mapper));
     }
 
     // -- Object
--- a/javaslang/src/main/java/javaslang/collection/PriorityQueue.java
+++ b/javaslang/src/main/java/javaslang/collection/PriorityQueue.java
@@ -498,8 +498,15 @@ public final class PriorityQueue<T> extends AbstractsQueue<T, PriorityQueue<T>> 
 
     @Override
     public <U> PriorityQueue<Tuple2<T, U>> zip(Iterable<? extends U> that) {
+        return zipWith(that, Tuple::of);
+    }
+
+
+    @Override
+    public <U, R> PriorityQueue<R> zipWith(Iterable<? extends U> that, BiFunction<? super T, ? super U, ? extends R> mapper) {
         Objects.requireNonNull(that, "that is null");
-        return ofAll(iterator().zip(that));
+        Objects.requireNonNull(mapper, "mapper is null");
+        return ofAll(naturalComparator(), iterator().zipWith(that, mapper));
     }
 
     @Override
@@ -510,7 +517,13 @@ public final class PriorityQueue<T> extends AbstractsQueue<T, PriorityQueue<T>> 
 
     @Override
     public PriorityQueue<Tuple2<T, Long>> zipWithIndex() {
-        return ofAll(iterator().zipWithIndex());
+        return zipWithIndex(Tuple::of);
+    }
+
+    @Override
+    public <U> PriorityQueue<U> zipWithIndex(BiFunction<? super T, ? super Long, ? extends U> mapper) {
+        Objects.requireNonNull(mapper, "mapper is null");
+        return ofAll(naturalComparator(), iterator().zipWithIndex(mapper));
     }
 
     @Override
@@ -558,8 +571,8 @@ public final class PriorityQueue<T> extends AbstractsQueue<T, PriorityQueue<T>> 
                 assert rank == tree.rank;
 
                 return comparator.isLessOrEqual(this.root, tree.root)
-                       ? of(this.root, this.rank + 1, tree.appendTo(this.children))
-                       : of(tree.root, tree.rank + 1, this.appendTo(tree.children));
+                        ? of(this.root, this.rank + 1, tree.appendTo(this.children))
+                        : of(tree.root, tree.rank + 1, this.appendTo(tree.children));
             }
 
             /**
--- a/javaslang/src/main/java/javaslang/collection/Queue.java
+++ b/javaslang/src/main/java/javaslang/collection/Queue.java
@@ -1135,11 +1135,17 @@ public final class Queue<T> extends AbstractsQueue<T, Queue<T>> implements Linea
         return ofAll(toList().update(index, element));
     }
 
-    @SuppressWarnings("unchecked")
     @Override
     public <U> Queue<Tuple2<T, U>> zip(Iterable<? extends U> that) {
+        return zipWith(that, Tuple::of);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public <U, R> Queue<R> zipWith(Iterable<? extends U> that, BiFunction<? super T, ? super U, ? extends R> mapper) {
         Objects.requireNonNull(that, "that is null");
-        return ofAll(toList().zip((Iterable<U>) that));
+        Objects.requireNonNull(mapper, "mapper is null");
+        return ofAll(toList().zipWith((Iterable<U>) that, mapper));
     }
 
     @Override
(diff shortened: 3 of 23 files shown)


javaslang/src/test/java/javaslang/collection/IntMap.java
```diff
@@ -23,7 +23,7 @@
     @SuppressWarnings("unchecked")
     public static <T> IntMap<T> of(Map<Integer, T> original) {
         return original.isEmpty() ? (IntMap<T>) EMPTY
-                                  : new IntMap<>(original);
+                : new IntMap<>(original);
```
paplorinc: Minor: please leave the original indentation in these cases.
In Idea you can configure it in:
```
File | Settings | Editor | Code Style | Java
Wrapping and Braces | Ternary operation | Align when multiline | ☑
```

javaslang/src/test/java/javaslang/collection/IntMultimap.java
```diff
@@ -313,8 +313,14 @@ public int characteristics() {
 
     @Override
     public <U> Seq<Tuple2<T, U>> zip(Iterable<? extends U> that) {
+        return zipWith(that, Tuple::of);
+    }
+
+    @Override
+    public <U, R> Seq<R> zipWith(Iterable<? extends U> that, BiFunction<? super T, ? super U, ? extends R> mapper) {
         Objects.requireNonNull(that, "that is null");
```
paplorinc: `that is null` doesn't sound very useful to me.
Could we rename `that` to `target` or something less context dependent :)?

javaslang/src/main/java/javaslang/collection/Traversable.java
```diff
@@ -1203,4 +1219,19 @@ default Number sum() {
      * @return A new traversable containing all elements of this traversable paired with their index, starting with 0.
      */
     Traversable<Tuple2<T, Long>> zipWithIndex();
+
+    /**
+     * Returns a traversable formed from this traversable and another Iterable collection by mapping elements.
+     * If one of the two iterables is longer than the other, its remaining elements are ignored.
```
paplorinc: @danieldietrich is it ok that we're ignoring the leftovers?
Shouldn't they have the same length?

javaslang/src/main/java/javaslang/collection/Traversable.java
```diff
@@ -1203,4 +1219,19 @@ default Number sum() {
...
+     * <p>
+     * The length of the returned traversable is the minimum of the lengths of this traversable and {@code that}
+     * iterable.
+     *
+     * @param <U>    The type of the mapped elements.
+     * @param mapper a mapper.
```
paplorinc: please provide a more useful description :)

javaslang/src/main/java/javaslang/collection/PriorityQueue.java
```diff
@@ -558,8 +571,8 @@ private Node(T root, int rank, Seq<Node<T>> children) {
                 assert rank == tree.rank;
 
                 return comparator.isLessOrEqual(this.root, tree.root)
-                       ? of(this.root, this.rank + 1, tree.appendTo(this.children))
-                       : of(tree.root, tree.rank + 1, this.appendTo(tree.children));
+                        ? of(this.root, this.rank + 1, tree.appendTo(this.children))
```
paplorinc: Minor: unnecessary formatting (the previous was better aligned)

javaslang/src/test/java/javaslang/collection/IntMap.java
danieldietrich: We provide a formatter for IntelliJ, see `.ide/idea-settings.jar`. Do you use it?

javaslang/src/main/java/javaslang/collection/LinkedHashSet.java
```diff
@@ -846,7 +852,13 @@ public String mkString(CharSequence prefix, CharSequence delimiter, CharSequence
 
     @Override
     public LinkedHashSet<Tuple2<T, Long>> zipWithIndex() {
-        return LinkedHashSet.ofAll(iterator().zipWithIndex());
+        return zipWithIndex(Tuple::of);
+    }
+
+    @Override
+    public <U> LinkedHashSet<U> zipWithIndex(BiFunction<? super T, ? super Long, ? extends U> mapper) {
```
paplorinc: Can we weaken the return type?

danieldietrich:
Very nice! I will check the changes now...

danieldietrich:
Very nice! I will check the changes now...

javaslang/src/test/java/javaslang/collection/IntMap.java
ronanM: Yes, I've imported settings.

paplorinc: Than please format only the `VCS changed` lines :) (or manually revert the ones that were better formatted before)

javaslang/src/test/java/javaslang/collection/IntMultimap.java
ronanM: All other `zip` function have `that` parameter named.

paplorinc: I know.

javaslang/src/main/java/javaslang/collection/Traversable.java
ronanM: I follow `map` and `flatMap`.

javaslang/src/main/java/javaslang/collection/Queue.java
```diff
@@ -1135,11 +1135,17 @@ public boolean startsWith(Iterable<? extends T> that, int offset) {
...
+        return zipWith(that, Tuple::of);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public <U, R> Queue<R> zipWith(Iterable<? extends U> that, BiFunction<? super T, ? super U, ? extends R> mapper) {
         Objects.requireNonNull(that, "that is null");
-        return ofAll(toList().zip((Iterable<U>) that));
+        Objects.requireNonNull(mapper, "mapper is null");
+        return ofAll(toList().zipWith((Iterable<U>) that, mapper));
```
danieldietrich: Does a type-hint work instead of casting? `return ofAll(toList().<U> zipWith(that, mapper));`

javaslang/src/test/java/javaslang/collection/IntMultimap.java
danieldietrich: Let's leave it as is. We have `that` in many places and is correct to say `this.zipWith(that)`.

By convention be write `<param-name> is null`. @ronanM has done it correct.

javaslang/src/main/java/javaslang/collection/Queue.java
ronanM: With type-hints it doesn't compile. Strange !?!

javaslang/src/main/java/javaslang/collection/Traversable.java
danieldietrich: We could fix the descriptions of map, flatMap, zipWithIndex et.al. in a separate PR

javaslang/src/main/java/javaslang/collection/Traversable.java
danieldietrich: `zipWithIndex` has no other Iterable, so the description is not correct. After zipWithIndex we have the same length.

javaslang/src/main/java/javaslang/collection/Queue.java
paplorinc: You can try extract to variable (ctrl alt v) make it compile and inline it back (ctrl alt v).
This usually works for me :)

javaslang/src/main/java/javaslang/collection/PriorityQueue.java
danieldietrich: Looks like number of spaces % 4 is 0. It is likely the IDE changed indentation. For me it is ok...

javaslang/src/main/java/javaslang/collection/LinkedHashSet.java
danieldietrich: Why? It is correct in my opinion.

javaslang/src/test/java/javaslang/collection/IntMap.java
danieldietrich: It is my fault. We should provide a formatter that aligns rows correct when applied to a file. There most by some formatting option to do that. We have to fix the formatter and check it in (in a different PR). We can then also reformat the whole code base (modulo Generator.scala and generated code).

@ronanM has done all right. You may leave it as is!

javaslang/src/main/java/javaslang/collection/Queue.java
danieldietrich: It was just a question. I think you are right and it does not compile then. Just leave it as is.

javaslang/src/main/java/javaslang/collection/PriorityQueue.java
paplorinc: It's the same ternary alignment I mentioned below.
It's an unrelated modification anyway (that doesn't make the code better), it's reason I recommended a revert.

javaslang/src/main/java/javaslang/collection/LinkedHashSet.java
paplorinc: It's like returning an `LinkedList`, the user of the code shouldn't care whether we change it to `ArrayList`, they should only see `List` in the public interface.
The same applies here, we should probably return `Set`, so that we can freely modify the underlying structure later.

javaslang/src/main/java/javaslang/collection/Queue.java
ronanM: I just push the deletion of this cast.

javaslang/src/main/java/javaslang/collection/LinkedHashSet.java
danieldietrich: I've thought about it - and took a look at the Scala API reference. For example Scala's [TreeSet.zipWithIndex](http://www.scala-lang.org/api/2.11.8/index.html#scala.collection.immutable.TreeSet@zipWithIndex:scala.collection.immutable.TreeSet[(A,Int)]) returns `TreeSet`.
For now I would keep it as is. I see currently no reason to change it.

danieldietrich:
The PR LGTM. I will pull it now. If I should have forgotten s.th. @paplorinc remarked (other than formatting which we will cover in another PR) please give me a hint.

danieldietrich:
Btw - many thanks, great work!

PR #1433, opened 2016-07-10
mkCharSeq(...) : CharSeq version of mkString(...)

ronanM:


danieldietrich:
Great, thank you! It makes sense to close the gap between other Traversables and our rich string `CharSeq`.

PR #1436, opened 2016-07-11
Added solution for Euler number 23.

jorander:


javaslang/src/test/java/javaslang/collection/euler/Euler23Test.java
```diff
@@ -0,0 +1,76 @@
...
+    private static long sumOfAllPositiveIntegersThatCannotBeWrittenAsTheSumOfTwoAbundantNumbers() {
+        return Stream.rangeClosed(1, LOWER_LIMIT_FOUND_BY_MATHEMATICAL_ANALYSIS_FOR_NUMBERS_THAT_CAN_BE_WRITTEN_AS_THE_SUM_OF_TO_ABUNDANT_NUMBERS)
+                .filter(l -> !canBeWrittenAsTheSumOfTwoAbundantNumbers(l))
+                .sum().longValue();
+    }
+
+    private static boolean canBeWrittenAsTheSumOfTwoAbundantNumbers(long l) {
+        return API.Match(l).of(
+                Case($(n -> n < SMALLEST_NUMBER_WRITTEN_AS_THE_SUM_OF_TO_ABUNDANT_NUMBERS), false),
+                Case($(SMALLEST_NUMBER_WRITTEN_AS_THE_SUM_OF_TO_ABUNDANT_NUMBERS), true),
```
danieldietrich: ... `Case(value, ...)` is a shortcut for `Case($(value), ...)`

but it is up to you which to use, just sayin'

danieldietrich:
The solution looks great. I will pull it in!

javaslang/src/test/java/javaslang/collection/euler/Euler23Test.java
jorander: I'll keep that in mind. But for now I think I prefer the $(*)-version
because it helps give a clear distinction between the  matching
predicate/value and the resulting function/value. In addition it is in-line
with the default case $() that tends to often show up at the end.

PR #1440, opened 2016-07-13
Added solution for Euler number 24

jorander:
Made two solutions. One that is very straight-forward and uses only the List-API but is not performant enough for the final task (takes about 40 seconds on my laptop) and one that is much more performant, but with a more involved algorithm.


danieldietrich:
Thank you!

PR #1442, opened 2016-07-13
Added solution for Euler number 25. Had to upgrade the return type of…

jorander:
… Utils#fibonacci to allow for a long enough list of Fibonaccis.


danieldietrich:
Thx!

PR #1444, opened 2016-07-14
Implement Value converters: to[Linked|Sorted][Map|Set]

### LINKED ISSUE #1434: Simple methods fo convert Value into LinkedHasMap/TreeMap/LinkedHashSet/TreeSet
Now I can convert any `Value` into `HashMap` and `HashSet` only.
But sometimes I want save key order while convert `Iterable` into `Map`.

Current code:
```java
LinkedHashMap.ofEntries(stream.filter(...).map(...).sortBy(...).map(o -> Tuple.of(o.getKey(), o)))
```

Supposed code:
```java
stream.filter(...).map(...).sortBy(...).toLinkedMap(o -> Tuple.of(o.getKey(), o))
```

Supposed methods in `Value`:
* `toLinkedMap(Function<...>)`
* `toTreeMap(Function<...>)`
* `toTreeMap(Comparator<...>, Function<...>)`
* `toLinkedSet(Function<...>)`
* `toTreeSet(Function<...>)`
* `toTreeSet(Comparator<...>, Function<...>)`

valery1707:
Fix #1434

### DIFF
--- a/javaslang/src/main/java/javaslang/Value.java
+++ b/javaslang/src/main/java/javaslang/Value.java
@@ -13,6 +13,7 @@ import javaslang.collection.HashSet;
 import javaslang.collection.Iterator;
 import javaslang.collection.List;
 import javaslang.collection.Map;
+import javaslang.collection.SortedMap;
 import javaslang.collection.PriorityQueue;
 import javaslang.collection.Queue;
 import javaslang.collection.Set;
@@ -100,7 +101,7 @@ import java.util.stream.StreamSupport;
  * <li>{@link #toList()}</li>
  * <li>{@link #toMap(Function)}</li>
  * <li>{@link #toLinkedMap(Function)}</li>
- * <li>{@link #toTreeMap(Function)}</li>
+ * <li>{@link #toSortedMap(Function)}</li>
  * <li>{@link #toOption()}</li>
  * <li>{@link #toQueue()}</li>
  * <li>{@link #toRight(Object)}</li>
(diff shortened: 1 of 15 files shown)


javaslang/src/main/java/javaslang/Value.java
```diff
@@ -660,13 +668,46 @@ default CharSeq toCharSeq() {
...
     default <K, V> Map<K, V> toMap(Function<? super T, ? extends Tuple2<? extends K, ? extends V>> f) {
         Objects.requireNonNull(f, "f is null");
-        if (isEmpty()) {
-            return HashMap.empty();
-        } else if (isSingleValued()) {
-            return HashMap.of(f.apply(get()));
-        } else {
-            return HashMap.ofEntries(Iterator.ofAll(this).map(f));
-        }
+		return ValueModule.toMap(this, HashMap.empty(), HashMap::of, HashMap::ofEntries, f);
```
danieldietrich: Very nice, thank you!

I think we should additionally check in all toMap and toSet methods that `this` Value is not already of type Map or Set, e.g.

```java
default <K, V> Map<K, V> toMap(Function<? super T, ? extends Tuple2<? extends K, ? extends V>> f) {
    if (Map.class.isAssignableFrom(getClass())) {
        return (Map<K, V>) this;
    } else {
        return ValueModule.toMap(this, HashMap.empty(), HashMap::of, HashMap::ofEntries, f);
    }
}
```

Maybe we can move the check into ValueModule.toMap / toSet by adding a mapType / setType parameter of type class? But generics in `Class<...>` could make problems. It could be just `Class<?>` and be cast to `M`...

valery1707: @danieldietrich I implement such check for `javaslang.ValueModule#toTraversable` because this method convert `Value<T>` into `Traversable<T>`, and in case `Set<T>#toSet()` we can use same object.
But I think this is not needed for `javaslang.ValueModule#toMap`, because we alway have mapping function and in case of converting `HashMap<K, V>` to map we can not return same object because we must consider to mapping function. And only in case than mapping function for new Map return same Tuple we can return original Map. But how to detect this?

danieldietrich: You are right! Thank you.

PR #1449, opened 2016-07-17
Implemented `Vector` as bit-mapped trie, with primitive support

paplorinc:
EDIT: this PR was separated in multiple smaller ones:
* https://github.com/javaslang/javaslang/pull/1503
* https://github.com/javaslang/javaslang/pull/1504
* https://github.com/javaslang/javaslang/pull/1522

---

Vector is the default Seq implementation that provides effectively constant time access to any element.
Many other operations (e.g. `tail`, `drop`, `slice`) are also effectively constant.

It's implemented using a `bit-mapped vector trie`, i.e. a tree where each node has `32` children.
For integers this means that the tree's depth will be `≤ 6`.

For symmetrically fast beginning and end operations (e.g. `drop` vs `dropRight`) two small arrays are used as depthless staging areas.
The leading and trailing arrays have dynamic lengths (their max size equals the branching factor),
but the middle tree's leafs only contain full blocks - though the first one may be padded by an offset.
The leading is only empty if everything is empty; if the trailing is empty, so is the middle.

For smaller memory footprint, primitive inputs are not boxed, resulting in a `4x` smaller size than `ArrayList`, with `~2x` faster iteration and `~5x` faster creation.
Some operations may box the primitive values, but it's possible to retrieve the values without boxing also (by returning the corresponding array).

Already boxed primitives are not unboxed (i.e. are stored as wrappers).

---

It's not based on [Clojure](https://github.com/paplorinc/clojure/blob/master/src/jvm/clojure/lang/PersistentVector.java) or [Scala](https://github.com/paplorinc/scala/blob/2.12.x/src/library/scala/collection/immutable/Vector.scala), but it has the same underlying idea: logarithmic access using a very wide and shallow tree, with a high branching factor.

This implementation is optimized algorithmically - though small wins are probably still lurking around :).

Compared to the old implementation, it's usually a lot faster and uses a *lot* less memory:
Memory - almost `70x` less memory overhead for large enough datasets (with `13x` less total memory for primitives):
```
for 32768 elements
 `javaslang.collection.OldVector` uses `1.9 MB` (`1.3 MB` overhead, `41.4` bytes overhead per element)
 `javaslang.collection.Vector` uses `659 KB` (`20.7 KB` overhead, `0.6` bytes overhead per element)

for 1024 elements
 `javaslang.collection.OldVector` uses `58.8 KB` (`40.1 KB` overhead, `40.1` bytes overhead per element)
 `javaslang.collection.Vector` uses `19.3 KB` (`672 bytes` overhead, `0.7` bytes overhead per element)

for 32 elements
 `javaslang.collection.OldVector` uses `1.6 KB` (`1.1 KB` overhead, `35.2` bytes overhead per element)
 `javaslang.collection.Vector` uses `544 bytes` (`40 bytes` overhead, `1.2` bytes overhead per element)
```

Speed - the new impl is `~4-5x` faster on average:
```groovy
Operation Ratio                                    32     1024      32768
Create   slang_persistent/old_slang_persistent 74.21×   44.92×     60.12× 
Head     slang_persistent/old_slang_persistent  3.82×    4.28×      5.52× 
Tail     slang_persistent/old_slang_persistent  4.09×    3.91×      6.55× 
Get      slang_persistent/old_slang_persistent  3.53×    5.94×      4.40× 
Update   slang_persistent/old_slang_persistent  1.40×    1.60×      2.33× 
Prepend  slang_persistent/old_slang_persistent  2.51×    3.96×      5.53× 
Append   slang_persistent/old_slang_persistent  1.55×    2.20×      3.08× 
GroupBy  slang_persistent/old_slang_persistent  1.23×    2.76×      3.94× 
Slice    slang_persistent/old_slang_persistent 44.33× 2014.66× 124956.66× 
Iterate  slang_persistent/old_slang_persistent  9.26×    7.78×      8.56× 
```

Compared to `Java`'s mutable `ArrayList`, and `Scala`'s, `Clojure`'s, `Functional Java`'s and `Pcollections`'s persistent ones, these are the results (e.g. `2x` means that our implementation is `2` times faster - can be executed twice while the other only once - for the given number of elements):

```groovy
Operation  Ratio                                               32     1024      32768 
Create     slang_persistent/java_mutable                    1.13×    0.20×      0.24×
Create     slang_persistent/fjava_persistent              315.71×   89.95×     75.30×
Create     slang_persistent/pcollections_persistent       160.86×   99.40×    152.24×
Create     slang_persistent/clojure_persistent              1.26×    5.88×      5.29×
Create     slang_persistent/scala_persistent               19.82×    4.01×      2.79×
                                                                             
Head       slang_persistent/java_mutable                    0.85×    0.81×      0.85×
Head       slang_persistent/fjava_persistent                1.00×    0.95×      1.00×
Head       slang_persistent/pcollections_persistent         2.68×    5.76×      9.26×
Head       slang_persistent/clojure_persistent              0.92×    1.32×      1.91×
Head       slang_persistent/scala_persistent                1.02×    0.96×      1.02×
                                                                             
Tail       slang_persistent/java_mutable                    0.53×    0.36×      0.29×
Tail       slang_persistent/fjava_persistent               42.17×   29.95×     35.59×
Tail       slang_persistent/pcollections_persistent         4.98×    5.71×      9.24×
Tail       slang_persistent/scala_persistent                2.52×    3.15×      4.48×
                                                                             
Get        slang_persistent/java_mutable                    0.86×    0.30×      0.23×
Get        slang_persistent/fjava_persistent               60.01×   71.98×     50.88×
Get        slang_persistent/pcollections_persistent         8.57×   10.06×     11.66×
Get        slang_persistent/clojure_persistent              0.99×    0.96×      0.91×
Get        slang_persistent/scala_persistent                1.26×    0.76×      0.62×
                                                                             
Update     slang_persistent/java_mutable                    0.11×    0.05×      0.02×
Update     slang_persistent/fjava_persistent              135.14×  227.17×    226.08×
Update     slang_persistent/pcollections_persistent         2.51×    3.63×      5.27×
Update     slang_persistent/clojure_persistent              0.92×    1.08×      1.10×
Update     slang_persistent/scala_persistent                1.07×    0.92×      1.18×
                                                                             
Prepend    slang_persistent/java_mutable                    1.61×    3.67×     83.34×
Prepend    slang_persistent/fjava_persistent                7.22×    8.73×      7.64×
Prepend    slang_persistent/pcollections_persistent         4.89×   13.11×     20.31×
Prepend    slang_persistent/clojure_persistent             29.43×  674.17×  22795.68×
Prepend    slang_persistent/scala_persistent                1.13×    1.18×      1.10×
                                                                             
Append     slang_persistent/java_mutable                    0.27×    0.20×      0.13×
Append     slang_persistent/fjava_persistent                5.17×    5.07×      5.06×
Append     slang_persistent/pcollections_persistent         2.72×    6.08×     10.26×
Append     slang_persistent/clojure_persistent              1.00×    0.78×      0.85×
Append     slang_persistent/scala_persistent                0.82×    0.74×      0.74×
                                                                             
Slice      slang_persistent/java_mutable                    0.24×    0.30×      0.32×
Slice      slang_persistent/clojure_persistent              0.21×    0.26×      0.29×
Slice      slang_persistent/scala_persistent                0.97×    3.45×      5.95×
                                                                             
Iterate    slang_persistent/java_mutable                    0.87×    0.30×      0.37×
Iterate    slang_persistent/fjava_persistent              388.66×  281.62×    275.59×
Iterate    slang_persistent/pcollections_persistent        16.37×    9.52×      7.21×
Iterate    slang_persistent/clojure_persistent              0.89×    0.73×      1.00×
Iterate    slang_persistent/scala_persistent                1.97×    1.06×      0.89×
```                                                                          
                                                                             
with significant wins in create and iterate for primitives:
```java
Operation  Ratio                                               32     1024    32768 
Create     slang_persistent_int/java_mutable                3.88×    0.48×    0.43×
Create     slang_persistent_int/fjava_persistent         1079.05×  214.89×  136.56×
Create     slang_persistent_int/pcollections_persistent   549.81×  237.47×  276.11×
Create     slang_persistent_int/clojure_persistent          4.31×   14.06×    9.60×
Create     slang_persistent_int/scala_persistent           67.74×    9.58×    5.06×

Iterate    slang_persistent_int/java_mutable                2.42×    1.72×    2.09×
Iterate    slang_persistent_int/fjava_persistent         1078.59× 1600.60× 1542.54×
Iterate    slang_persistent_int/pcollections_persistent    45.43×   54.11×   40.35×
Iterate    slang_persistent_int/clojure_persistent          2.48×    4.14×    5.62×
Iterate    slang_persistent_int/scala_persistent            5.46×    6.01×    4.98×
```

and memory usages:
```groovy
for 32768 elements
 `java.util.ArrayList` uses `638.4 KB` (`0 bytes` overhead, `0.0` bytes overhead per element)
 `scala.collection.immutable.Vector` uses `659 KB` (`20.7 KB` overhead, `0.6` bytes overhead per element)
 `clojure.lang.PersistentVector` uses `683.8 KB` (`45.4 KB` overhead, `1.4` bytes overhead per element)
 `org.pcollections.TreePVector` uses `1.7 MB` (`1.1 MB` overhead, `36.0` bytes overhead per element)
 `fj.data.Seq` uses `3.3 MB` (`2.6 MB` overhead, `84.1` bytes overhead per element)
 `javaslang.collection.Vector` uses `659 KB` (`20.7 KB` overhead, `0.6` bytes overhead per element)
 `javaslang.collection.Vector` uses `148.7 KB` (`-501400 bytes` overhead, `-15.3` bytes overhead per element) (with primitives)

for 1024 elements
 `java.util.ArrayList` uses `18.6 KB` (`0 bytes` overhead, `0.0` bytes overhead per element)
 `scala.collection.immutable.Vector` uses `19.3 KB` (`672 bytes` overhead, `0.7` bytes overhead per element)
 `clojure.lang.PersistentVector` uses `20 KB` (`1.4 KB` overhead, `1.4` bytes overhead per element)
 `org.pcollections.TreePVector` uses `54.7 KB` (`36.1 KB` overhead, `36.1` bytes overhead per element)
 `fj.data.Seq` uses `102.3 KB` (`83.7 KB` overhead, `83.7` bytes overhead per element)
 `javaslang.collection.Vector` uses `19.3 KB` (`672 bytes` overhead, `0.7` bytes overhead per element)
 `javaslang.collection.Vector` uses `4.7 KB` (`-14248 bytes` overhead, `-13.9` bytes overhead per element) (with primitives)

for 32 elements
 `java.util.ArrayList` uses `504 bytes` (`0 bytes` overhead, `0.0` bytes overhead per element)
 `scala.collection.immutable.Vector` uses `536 bytes` (`32 bytes` overhead, `1.0` bytes overhead per element)
 `clojure.lang.PersistentVector` uses `704 bytes` (`200 bytes` overhead, `6.2` bytes overhead per element)
 `org.pcollections.TreePVector` uses `1.7 KB` (`1.2 KB` overhead, `37.8` bytes overhead per element)
 `fj.data.Seq` uses `3 KB` (`2.5 KB` overhead, `80.0` bytes overhead per element)
 `javaslang.collection.Vector` uses `544 bytes` (`40 bytes` overhead, `1.2` bytes overhead per element)
 `javaslang.collection.Vector` uses `232 bytes` (`-272 bytes` overhead, `-8.5` bytes overhead per element) (with primitives)
```

The implementation is quite complex, please review it from your IDE instead (preferably commit-by-commit) :).

Primitive storage also enabled the creation of very big `Vectors`, i.e.
```java
Create      slang_persistent_int     1048576     330.374  ops/s                     
Create      slang_persistent_int    33554432       8.309 ops/s    
```

Every comment is very welcome! :)

@danieldietrich, @zsolt-donca, @szili88, @ovidiudeac, @djspiewak, please review.

### DIFF
--- a/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
@@ -21,6 +21,7 @@ import static scala.collection.JavaConversions.asScalaBuffer;
 @SuppressWarnings({"ALL", "unchecked"})
 public class VectorBenchmark {
     static final Array<Class<?>> CLASSES = Array.of(
+            Access.class,
             Create.class,
             Head.class,
             Tail.class,
@@ -50,6 +51,7 @@ public class VectorBenchmark {
 
         int EXPECTED_AGGREGATE;
         Integer[] ELEMENTS;
+        int[] INT_ELEMENTS;
         int[] RANDOMIZED_INDICES;
 
         /* Only use this for non-mutating operations */
@@ -60,11 +62,13 @@ public class VectorBenchmark {
         fj.data.Seq<Integer> fjavaPersistent;
         org.pcollections.PVector<Integer> pcollectionsPersistent;
         javaslang.collection.Vector<Integer> slangPersistent;
+        javaslang.collection.Vector<Integer> slangPersistentInt;
 
         @Setup
         public void setup() {
             final Random random = new Random(0);
             ELEMENTS = getRandomValues(CONTAINER_SIZE, false, random);
+            INT_ELEMENTS = Arrays2.toPrimitiveArray(int.class, ELEMENTS);
             RANDOMIZED_INDICES = Arrays2.shuffle(Array.range(0, CONTAINER_SIZE).toJavaStream().mapToInt(Integer::intValue).toArray(), random);
 
             EXPECTED_AGGREGATE = Array.of(ELEMENTS).reduce(JmhRunner::aggregate);
@@ -75,6 +79,36 @@ public class VectorBenchmark {
             pcollectionsPersistent = create(org.pcollections.TreePVector::from, javaMutable, v -> areEqual(v, javaMutable));
             fjavaPersistent = create(fj.data.Seq::fromJavaList, javaMutable, v -> areEqual(v, javaMutable));
             slangPersistent = create(javaslang.collection.Vector::ofAll, javaMutable, v -> areEqual(v, javaMutable));
+            slangPersistentInt = create(v -> javaslang.collection.Vector.ofAll(INT_ELEMENTS), javaMutable, v -> areEqual(v, javaMutable) && (v.type == int.class));
+        }
+    }
+
+    /** Bulk creation from array based, boxed source */
+    public static class Access extends Base {
+        @Benchmark
+        public Object create() {
+            return Arrays2.arrayCopy(Integer.class, CONTAINER_SIZE, ELEMENTS, 0, 0, CONTAINER_SIZE);
+        }
+
+        @Benchmark
+        public Object set() {
+            final Object[] values = new Object[CONTAINER_SIZE];
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                Arrays2.setAt(values, i, ELEMENTS[i]);
+            }
+            return values;
+        }
+
+        @Benchmark
+        public void get(Blackhole bh) {
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                bh.consume(Arrays2.<Object> getAt(ELEMENTS, i));
+            }
+        }
+
+        @Benchmark
+        public int length() {
+            return Arrays2.getLength(ELEMENTS);
         }
     }
 
@@ -83,6 +117,16 @@ public class VectorBenchmark {
         @Benchmark
         public Object java_mutable() {
             final ArrayList<Integer> values = new ArrayList<>(javaMutable);
+            assert areEqual(values, javaMutable);
+            return values;
+        }
+
+        @Benchmark
+        public Object java_mutable_boxed() {
+            final ArrayList<Integer> values = new ArrayList<>();
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values.add(INT_ELEMENTS[i]);
+            }
             assert areEqual(values, javaMutable);
             return values;
         }
@@ -119,7 +163,14 @@ public class VectorBenchmark {
         public Object slang_persistent() {
             final javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.ofAll(javaMutable);
             assert areEqual(values, javaMutable);
-            return values.head();
+            return values;
+        }
+
+        @Benchmark
+        public Object slang_persistent_int() {
+            final javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.ofAll(INT_ELEMENTS);
+            assert (values.type == int.class) && areEqual(values, javaMutable);
+            return values;
         }
     }
 
@@ -163,6 +214,13 @@ public class VectorBenchmark {
         public Object slang_persistent() {
             final Object head = slangPersistent.head();
             assert Objects.equals(head, javaMutable.get(0));
+            return head;
+        }
+
+        @Benchmark
+        public int slang_persistent_int() {
+            final int head = ((int[]) slangPersistentInt.leafUnsafe(0))[0];
+            assert Objects.equals(head, javaMutable.get(slangPersistentInt.leafIndex(0)));
             return head;
         }
     }
@@ -233,6 +291,16 @@ public class VectorBenchmark {
             assert values.isEmpty();
             return values;
         }
+
+        @Benchmark
+        public Object slang_persistent_int() {
+            javaslang.collection.Vector<Integer> values = slangPersistentInt;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values = values.tail();
+            }
+            assert values.isEmpty();
+            return values;
+        }
     }
 
     /** Aggregated, randomized access to every element */
@@ -292,6 +360,17 @@ public class VectorBenchmark {
             int aggregate = 0;
             for (int i : RANDOMIZED_INDICES) {
                 aggregate ^= slangPersistent.get(i);
+            }
+            assert aggregate == EXPECTED_AGGREGATE;
+            return aggregate;
+        }
+
+        @Benchmark
+        public int slang_persistent_int() {
+            int aggregate = 0;
+            for (int i : RANDOMIZED_INDICES) {
+                final int[] leaf = (int[]) slangPersistentInt.leafUnsafe(i);
+                aggregate ^= leaf[slangPersistentInt.leafIndex(i)];
             }
             assert aggregate == EXPECTED_AGGREGATE;
             return aggregate;
@@ -375,6 +454,16 @@ public class VectorBenchmark {
             assert values.forAll(e -> e == 0);
             return values;
         }
+
+        @Benchmark
+        public Object slang_persistent_int() {
+            javaslang.collection.Vector<Integer> values = slangPersistentInt;
+            for (int i : RANDOMIZED_INDICES) {
+                values = values.update(i, 0);
+            }
+            assert (values.type == int.class) && values.forAll(e -> e == 0);
+            return values;
+        }
     }
 
     public static class Prepend extends Base {
@@ -439,6 +528,16 @@ public class VectorBenchmark {
                 values = values.prepend(element);
             }
             assert areEqual(values.reverse(), javaMutable);
+            return values;
+        }
+
+        @Benchmark
+        public Object slang_persistent_int() {
+            javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.ofAll(new int[] {ELEMENTS[0]});
+            for (int i = 1; i < ELEMENTS.length; i++) {
+                values = values.prepend(ELEMENTS[i]);
+            }
+            assert (values.type == int.class) && areEqual(values.reverse(), javaMutable);
             return values;
         }
     }
@@ -513,6 +612,18 @@ public class VectorBenchmark {
             assert areEqual(values, javaMutable);
             return values;
         }
+
+        @Benchmark
+        public Object slang_persistent_int() {
+            javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.ofAll(new int[] {ELEMENTS[0]});
+            for (int i = 1; i < ELEMENTS.length; i++) {
+                final Vector<Integer> newValues = values.append(ELEMENTS[i]);
+                assert values != newValues;
+                values = newValues;
+            }
+            assert (values.type == int.class) && areEqual(values, javaMutable);
+            return values;
+        }
     }
 
     public static class GroupBy extends Base {
@@ -567,6 +678,16 @@ public class VectorBenchmark {
         @Benchmark
         public void slang_persistent(Blackhole bh) {
             javaslang.collection.Vector<Integer> values = this.slangPersistent;
+            while (!values.isEmpty()) {
+                values = values.slice(1, values.size());
+                values = values.slice(0, values.size() - 1);
+                bh.consume(values);
+            }
+        }
+
+        @Benchmark
+        public void slang_persistent_int(Blackhole bh) {
+            javaslang.collection.Vector<Integer> values = this.slangPersistentInt;
             while (!values.isEmpty()) {
                 values = values.slice(1, values.size());
                 values = values.slice(0, values.size() - 1);
@@ -648,6 +769,21 @@ public class VectorBenchmark {
             int aggregate = 0;
             for (final Iterator<Integer> iterator = slangPersistent.iterator(); iterator.hasNext(); ) {
                 aggregate ^= iterator.next();
+            }
+            assert aggregate == EXPECTED_AGGREGATE;
+            return aggregate;
+        }
+
+        @Benchmark
+        public int slang_persistent_int() {
+            int aggregate = 0;
+            for (int i = 0; i < slangPersistentInt.length(); ) {
+                final int[] leaf = (int[]) slangPersistentInt.leafUnsafe(i);
+                final int offset = slangPersistentInt.leafIndex(i);
+                for (int j = offset; j < leaf.length; j++) {
+                    aggregate ^= leaf[j];
+                }
+                i += leaf.length - offset;
             }
             assert aggregate == EXPECTED_AGGREGATE;
             return aggregate;
--- a/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
@@ -6,14 +6,19 @@ import org.openjdk.jmh.annotations.*;
 import org.openjdk.jmh.infra.Blackhole;
 import scala.compat.java8.JFunction;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Objects;
+import java.util.Random;
 import java.util.stream.Collectors;
 
 import static java.util.Arrays.asList;
-import static javaslang.JmhRunner.*;
+import static javaslang.JmhRunner.create;
+import static javaslang.JmhRunner.getRandomValues;
 import static javaslang.collection.Collections.areEqual;
-import static scala.collection.JavaConversions.*;
+import static scala.collection.JavaConversions.asJavaCollection;
+import static scala.collection.JavaConversions.asScalaBuffer;
 
+@SuppressWarnings({"ALL", "unchecked"})
 public class VectorBenchmark {
     static final Array<Class<?>> CLASSES = Array.of(
             Create.class,
@@ -34,30 +39,34 @@ public class VectorBenchmark {
     }
 
     public static void main(String... args) {
+        JmhRunner.runDebugWithAsserts(CLASSES);
         JmhRunner.runNormalNoAsserts(CLASSES);
     }
 
     @State(Scope.Benchmark)
     public static class Base {
-        @Param({ "32", "1024", "32768"/*, "1048576"*/ }) // i.e. depth 1,2,3(,4) for a branching factor of 32
+        @Param({"32", "1024", "32768" /*, "1048576", "33554432", "1073741824" */}) /* i.e. depth 1,2,3(,4,5,6) for a branching factor of 32 */
         public int CONTAINER_SIZE;
 
         int EXPECTED_AGGREGATE;
         Integer[] ELEMENTS;
+        int[] RANDOMIZED_INDICES;
 
         /* Only use this for non-mutating operations */
         java.util.ArrayList<Integer> javaMutable;
 
-        fj.data.Seq<Integer> fjavaPersistent = fj.data.Seq.empty();
-        org.pcollections.PVector<Integer> pcollectionsPersistent = org.pcollections.TreePVector.empty();
         scala.collection.immutable.Vector<Integer> scalaPersistent;
         clojure.lang.PersistentVector clojurePersistent;
+        fj.data.Seq<Integer> fjavaPersistent;
+        org.pcollections.PVector<Integer> pcollectionsPersistent;
         javaslang.collection.Vector<Integer> slangPersistent;
 
         @Setup
-        @SuppressWarnings("unchecked")
         public void setup() {
-            ELEMENTS = getRandomValues(CONTAINER_SIZE, 0);
+            final Random random = new Random(0);
+            ELEMENTS = getRandomValues(CONTAINER_SIZE, false, random);
+            RANDOMIZED_INDICES = Arrays2.shuffle(Array.range(0, CONTAINER_SIZE).toJavaStream().mapToInt(Integer::intValue).toArray(), random);
+
             EXPECTED_AGGREGATE = Array.of(ELEMENTS).reduce(JmhRunner::aggregate);
 
             javaMutable = create(java.util.ArrayList::new, asList(ELEMENTS), v -> areEqual(v, asList(ELEMENTS)));
@@ -69,6 +78,7 @@ public class VectorBenchmark {
         }
     }
 
+    /** Bulk creation from array based, boxed source */
     public static class Create extends Base {
         @Benchmark
         public Object java_mutable() {
@@ -157,7 +167,6 @@ public class VectorBenchmark {
         }
     }
 
-    @SuppressWarnings("Convert2MethodRef")
     public static class Tail extends Base {
         @State(Scope.Thread)
         public static class Initialized {
@@ -177,9 +186,9 @@ public class VectorBenchmark {
 
         @Benchmark
         public Object java_mutable(Initialized state) {
-            final java.util.ArrayList<Integer> values = state.javaMutable;
+            java.util.List<Integer> values = state.javaMutable;
             for (int i = 0; i < CONTAINER_SIZE; i++) {
-                values.remove(0);
+                values = values.subList(1, values.size()); // remove(0) would copy everything, but this will slow access down because of nesting
             }
             assert values.isEmpty();
             return values;
@@ -190,16 +199,6 @@ public class VectorBenchmark {
             scala.collection.immutable.Vector<Integer> values = scalaPersistent;
             for (int i = 0; i < CONTAINER_SIZE; i++) {
                 values = values.tail();
-            }
-            assert values.isEmpty();
-            return values;
-        }
-
-        @Benchmark
-        public Object clojure_persistent() {
-            clojure.lang.PersistentVector values = clojurePersistent;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
-                values = values.pop();
             }
             assert values.isEmpty();
             return values;
@@ -236,11 +235,12 @@ public class VectorBenchmark {
         }
     }
 
+    /** Aggregated, randomized access to every element */
     public static class Get extends Base {
         @Benchmark
         public int java_mutable() {
             int aggregate = 0;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
                 aggregate ^= javaMutable.get(i);
             }
             assert aggregate == EXPECTED_AGGREGATE;
@@ -250,7 +250,7 @@ public class VectorBenchmark {
         @Benchmark
         public int scala_persistent() {
             int aggregate = 0;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
                 aggregate ^= scalaPersistent.apply(i);
             }
             assert aggregate == EXPECTED_AGGREGATE;
@@ -260,7 +260,7 @@ public class VectorBenchmark {
         @Benchmark
         public int clojure_persistent() {
             int aggregate = 0;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
                 aggregate ^= (int) clojurePersistent.get(i);
             }
             assert aggregate == EXPECTED_AGGREGATE;
@@ -270,7 +270,7 @@ public class VectorBenchmark {
         @Benchmark
         public int fjava_persistent() {
             int aggregate = 0;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
                 aggregate ^= fjavaPersistent.index(i);
             }
             assert aggregate == EXPECTED_AGGREGATE;
@@ -280,7 +280,7 @@ public class VectorBenchmark {
         @Benchmark
         public int pcollections_persistent() {
             int aggregate = 0;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
                 aggregate ^= pcollectionsPersistent.get(i);
             }
             assert aggregate == EXPECTED_AGGREGATE;
@@ -290,7 +290,7 @@ public class VectorBenchmark {
         @Benchmark
         public int slang_persistent() {
             int aggregate = 0;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
                 aggregate ^= slangPersistent.get(i);
             }
             assert aggregate == EXPECTED_AGGREGATE;
@@ -298,6 +298,7 @@ public class VectorBenchmark {
         }
     }
 
+    /** Randomized update of every element */
     public static class Update extends Base {
         @State(Scope.Thread)
         public static class Initialized {
@@ -318,7 +319,7 @@ public class VectorBenchmark {
         @Benchmark
         public Object java_mutable(Initialized state) {
             final java.util.ArrayList<Integer> values = state.javaMutable;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
                 values.set(i, 0);
             }
             assert Array.ofAll(values).forAll(e -> e == 0);
@@ -328,7 +329,7 @@ public class VectorBenchmark {
         @Benchmark
         public Object scala_persistent() {
             scala.collection.immutable.Vector<Integer> values = scalaPersistent;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
                 values = values.updateAt(i, 0);
             }
             assert Array.ofAll(asJavaCollection(values)).forAll(e -> e == 0);
@@ -338,7 +339,7 @@ public class VectorBenchmark {
         @Benchmark
         public Object clojure_persistent() {
             clojure.lang.PersistentVector values = clojurePersistent;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
                 values = values.assocN(i, 0);
             }
             assert Array.of(values.toArray()).forAll(e -> Objects.equals(e, 0));
@@ -348,7 +349,7 @@ public class VectorBenchmark {
         @Benchmark
         public Object fjava_persistent() {
             fj.data.Seq<Integer> values = fjavaPersistent;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
                 values = values.update(i, 0);
             }
             assert Array.ofAll(values).forAll(e -> e == 0);
@@ -358,7 +359,7 @@ public class VectorBenchmark {
         @Benchmark
         public Object pcollections_persistent() {
             org.pcollections.PVector<Integer> values = pcollectionsPersistent;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
                 values = values.with(i, 0);
             }
             assert Array.ofAll(values).forAll(e -> e == 0);
@@ -368,7 +369,7 @@ public class VectorBenchmark {
         @Benchmark
         public Object slang_persistent() {
             javaslang.collection.Vector<Integer> values = slangPersistent;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
                 values = values.update(i, 0);
             }
             assert values.forAll(e -> e == 0);
@@ -376,11 +377,10 @@ public class VectorBenchmark {
         }
     }
 
-    @SuppressWarnings("ManualArrayToCollectionCopy")
     public static class Prepend extends Base {
         @Benchmark
         public Object java_mutable() {
-            final java.util.ArrayList<Integer> values = new java.util.ArrayList<>(CONTAINER_SIZE);
+            final java.util.ArrayList<Integer> values = new java.util.ArrayList<>(); // no initial value, as we're simulating dynamic usage
             for (Integer element : ELEMENTS) {
                 values.add(0, element);
             }
@@ -395,6 +395,20 @@ public class VectorBenchmark {
                 values = values.appendFront(element);
             }
             assert areEqual(Array.ofAll(asJavaCollection(values)).reverse(), javaMutable);
+            return values;
+        }
+
+        @Benchmark
+        public Object clojure_persistent() {
+            clojure.lang.PersistentVector values = clojure.lang.PersistentVector.EMPTY;
+            for (int i = 0; i < ELEMENTS.length; i++) {
+                clojure.lang.PersistentVector prepended = clojure.lang.PersistentVector.create(ELEMENTS[i]);
+                for (Object value : values) {
+                    prepended = prepended.cons(value); // rebuild everything via append
+                }
+                values = prepended;
+            }
+            assert areEqual(Array.ofAll(values).reverse(), javaMutable);
             return values;
         }
 
@@ -429,11 +443,10 @@ public class VectorBenchmark {
         }
     }
 
-    @SuppressWarnings("ManualArrayToCollectionCopy")
     public static class Append extends Base {
         @Benchmark
         public Object java_mutable() {
-            final java.util.ArrayList<Integer> values = new java.util.ArrayList<>(CONTAINER_SIZE);
+            final java.util.ArrayList<Integer> values = new java.util.ArrayList<>(); // no initial value as we're simulating dynamic usage
             for (Integer element : ELEMENTS) {
                 values.add(element);
             }
@@ -445,7 +458,9 @@ public class VectorBenchmark {
         public Object scala_persistent() {
             scala.collection.immutable.Vector<Integer> values = scala.collection.immutable.Vector$.MODULE$.empty();
             for (Integer element : ELEMENTS) {
-                values = values.appendBack(element);
+                final scala.collection.immutable.Vector<Integer> newValues = values.appendBack(element);
+                assert values != newValues;
+                values = newValues;
             }
             assert areEqual(asJavaCollection(values), javaMutable);
             return values;
@@ -455,7 +470,9 @@ public class VectorBenchmark {
         public Object clojure_persistent() {
             clojure.lang.PersistentVector values = clojure.lang.PersistentVector.EMPTY;
             for (Integer element : ELEMENTS) {
-                values = values.cons(element);
+                final clojure.lang.PersistentVector newValues = values.cons(element);
+                assert values != newValues;
+                values = newValues;
             }
             assert areEqual(values, javaMutable);
             return values;
@@ -465,7 +482,9 @@ public class VectorBenchmark {
         public Object fjava_persistent() {
             fj.data.Seq<Integer> values = fj.data.Seq.empty();
             for (Integer element : ELEMENTS) {
-                values = values.snoc(element);
+                final fj.data.Seq<Integer> newValues = values.snoc(element);
+                assert values != newValues;
+                values = newValues;
             }
             assert areEqual(values, javaMutable);
             return values;
@@ -475,7 +494,9 @@ public class VectorBenchmark {
         public Object pcollections_persistent() {
             org.pcollections.PVector<Integer> values = org.pcollections.TreePVector.empty();
             for (Integer element : ELEMENTS) {
-                values = values.plus(element);
+                final org.pcollections.PVector<Integer> newValues = values.plus(element);
+                assert values != newValues;
+                values = newValues;
             }
             assert areEqual(values, javaMutable);
             return values;
@@ -485,7 +506,9 @@ public class VectorBenchmark {
         public Object slang_persistent() {
             javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.empty();
             for (Integer element : ELEMENTS) {
-                values = values.append(element);
+                final Vector<Integer> newValues = values.append(element);
+                assert values != newValues;
+                values = newValues;
             }
             assert areEqual(values, javaMutable);
             return values;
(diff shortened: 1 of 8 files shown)


javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -34,30 +39,34 @@ public void testAsserts() {
...
 
     public static void main(String... args) {
+        JmhRunner.runDebugWithAsserts(CLASSES);
         JmhRunner.runNormalNoAsserts(CLASSES);
     }
 
     @State(Scope.Benchmark)
     public static class Base {
-        @Param({ "32", "1024", "32768"/*, "1048576"*/ }) // i.e. depth 1,2,3(,4) for a branching factor of 32
+        @Param({"32", "1024", "32768" /*, "100000" , "1048576", "33554432", "1073741824" */}) /* i.e. depth 1,2,3(,4,5,6) for a branching factor of 32 */
```
paplorinc: Once this is finished, I would be very interested in `>10m` - I can't run that on my laptop anymore :p

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -196,16 +205,6 @@ public Object scala_persistent() {
         }
 
         @Benchmark
-        public Object clojure_persistent() {
-            clojure.lang.PersistentVector values = clojurePersistent;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
-                values = values.pop();
```
paplorinc: `pop` is the same as `init`, not as `tail` :/

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -236,11 +235,12 @@ public Object slang_persistent() {
         }
     }
 
+    /** Aggregated, randomized access to every element */
     public static class Get extends Base {
         @Benchmark
         public int java_mutable() {
             int aggregate = 0;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
```
paplorinc: random access is what we're after, not linear (Scala optimizes for linear, by storing the previous element. I don't think we want that.)

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -368,19 +369,18 @@ public Object pcollections_persistent() {
...
             return values;
         }
     }
 
-    @SuppressWarnings("ManualArrayToCollectionCopy")
     public static class Prepend extends Base {
         @Benchmark
         public Object java_mutable() {
-            final java.util.ArrayList<Integer> values = new java.util.ArrayList<>(CONTAINER_SIZE);
+            final java.util.ArrayList<Integer> values = new java.util.ArrayList<>(); // no initial value as we're simulating dynamic usage
```
paplorinc: it's more similar to the rest this way

javaslang/src/test/java/javaslang/collection/VectorTest.java
```diff
@@ -263,40 +263,6 @@ public void shouldNotSerializeEnclosingClass() throws Throwable {
...
     }
 
-    @Test(expected = InvalidObjectException.class)
-    public void shouldNotDeserializeListWithSizeLessThanOne() throws Throwable {
-        try {
-            /*
-             * This implementation is stable regarding jvm impl changes of object serialization. The index of the number
-             * of List elements is gathered dynamically.
-             */
-            final byte[] listWithOneElement = Serializables.serialize(List.of(0));
```
paplorinc: was testing `List`, not `Vector`

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -9,38 +9,53 @@
...
+ * @author Pap Lőrinc
+ * @since 3.0.0
  */
-
 @SuppressWarnings({"SuspiciousArrayCast", "unchecked"})
 public class Vector<T> implements Kind1<Vector<?>, T>, IndexedSeq<T> {
     private static final long serialVersionUID = 1L;
 
-    private static final Vector<?> EMPTY = new Vector<>(HashArrayMappedTrie.empty());
+    static int BRANCHING_BASE = 5;
```
paplorinc: it's changeable from code, is used from tests to validate multiple levels.
In Scala they're not testing the upper levels, as it would take too much time
(it wasn't even functional in Scala for 6 years, when someone finally gave it a try for billions of elements)
```scala
display4 = display5((newIndex >> 20) & 31).asInstanceOf[Array[AnyRef]] // 20 -> should be 25
if (display4 == null) display4 = new Array(32)
display3 = display4((newIndex >> 20) & 31).asInstanceOf[Array[AnyRef]]
if (display3 == null) display3 = new Array(32)
display2 = display3((newIndex >> 15) & 31).asInstanceOf[Array[AnyRef]]
```

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -9,38 +9,53 @@
...
+
+    public Vector(T[] firstLeaf, VectorTree<T> middle, T[] lastLeaf) {
+        this.firstLeaf = firstLeaf;
+        this.middle = middle;
+        this.lastLeaf = lastLeaf;
 
-    private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
-        this.trie = trie;
-        this.indexShift = indexShift;
+        assert (this.firstLeaf.length <= branchingFactor()) && (this.lastLeaf.length <= branchingFactor());
```
paplorinc: I'm using lots of asserts at this stage, they helped me a lot, as there are very many stages to go to/from, I want to fail early

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -153,12 +169,39 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
+        }
+    }
+
+    public static <T> Vector<T> ofAll(Collection<? extends T> elements) {
+        Objects.requireNonNull(elements, "elements is null");
+        return elements.isEmpty() ? empty()
+                                  : ofAll(elements, elements.size());
+    }
+
+    public static <T> Vector<T> ofAll(Iterable<? extends T> iterable, int size) {
```
paplorinc: This way of creating everything from bottom up and proved to be `4-10`x faster than both `clojure` and `scala`

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -0,0 +1,40 @@
...
+ * /___/\_/  \_/\____/\_/  \_/\__\/__/\__\_/  \_//  \__/\_____/   Licensed under the Apache License, Version 2.0
+ */
+package javaslang.collection;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+// TODO should rewrite javaslang-test, once it's accessible from here
```
paplorinc: @danieldietrich

javaslang/src/main/java/javaslang/collection/Arrays2.java
```diff
@@ -55,6 +55,18 @@
         return array;
     }
 
+    static <T> Iterator<T> iterator(final T[] array, int offset) {
+        return new Iterator<T>() { /* it's faster than Iterator.of(root) */
```
paplorinc: `AbstractIterator` is too heavy here

paplorinc: `AbstractIterator` is too heavy here

javaslang/src/main/java/javaslang/collection/Arrays2.java
```diff
@@ -74,6 +74,22 @@ public static Object get(Object[] array, int i) {
         };
     }
 
+    static <T> T[] inject(Object arrayObject, int index, T element) {
```
paplorinc: `prepend`, `set` and `append`

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -814,7 +814,34 @@ public boolean isTraversableAgain() {
...
+            final List<Iterator<T>> iterators = List.of(
+                    Arrays2.iterator(firstLeaf, 0),
+                    middle.<T> iterator(),
+                    Arrays2.iterator(lastLeaf, 0)
+            ).filter(Iterator::nonEmpty);
+
+            if (iterators.size() == 1) {
+                return iterators.head();
+            } else {
+                return new Iterator<T>() { // a lot faster than Iterator.concat // TODO it's still too slow :(
```
paplorinc: ideas are welcome on how to iterate quickly using 3 iterators :/

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1581,6 +1618,34 @@ public T next() {
             };
         }
 
+        // TODO we should speed this up by minimizing one-by-one copying and replacing it with copying chunks with System.arraycopy
```
paplorinc: this is the bottleneck for `prepend` and `append`.

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -130,6 +133,98 @@ public void shouldDropRight() {
...
+                for (int j = expected.length(); j >= i; j -= 11) {
+                    expected = expected.slice(i, j);
+                    actual = assertAreEqual(actual, Tuple.of(i, j), (a, p) -> a.slice(p._1, p._2), expected);
+                }
+            }
+        }
+    }
+
+    @Test
+    public void shouldBehaveLikeArray() {
```
paplorinc: It's amazing how many bugs were revealed by this :trollface:

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -0,0 +1,240 @@
...
+                    assertThat(actual.tail().toJavaList()).isEqualTo(expected.tail().toJavaList());
+                }
+            }
+
+            {
+                if (!expected.isEmpty()) {
+                    final int index = random.nextInt(expected.size());
+                    assertThat(actual.get(index)).isEqualTo(expected.get(index));
+                }
+            }
```
stancalau: Are these sections needed if there is no new declaration wihtin?

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -0,0 +1,240 @@
...
+
+            if (random.nextInt(100) < 30) {
+                final int n = random.nextInt(expected.size() + 1);
+                expected = expected.take(n);
+                actual = assertAreEqual(actual, n, Vector::take, expected);
+            }
+
+            {
+                if (!expected.isEmpty()) {
```
stancalau: Are these sections needed if there is no new declaration wihtin?

paplorinc: yes, it's an `if` statement, not an empty block

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -0,0 +1,240 @@
...
+                    assertThat(actual.head()).isEqualTo(expected.head());
+                    assertThat(actual.tail().toJavaList()).isEqualTo(expected.tail().toJavaList());
+                }
+            }
+
+            {
```
paplorinc: these aren't necessarily needed :)

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
stancalau: Wrong place to put my comment, I was referring to the section above the if.

paplorinc: Thanks, fixed it in a next commit already :)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -11,33 +11,82 @@
...
+ * Vector is the default Seq implementation that provides effectively constant time access to any element.
+ * Many other operations (e.g. `tail`, `drop`, `slice`) are also effectively constant.
+ *
+ * It's implemented using a `bit-mapped vector trie`, i.e. a tree where each node has 32 children.
+ * For integers this means that the tree's depth will be ≤ 6.
+ *
+ * For symmetrically fast beginning and end operations (e.g. `drop` vs `dropRight`) two small arrays are used as depthless staging areas.
+ * The first and last arrays have dynamic lengths (are shifted to the correct index on drop), but the middle one has internal offset to avoid shifting.
+ * The first one is only empty if everything is empty; if the last is empty, so is the middle.
+ * The middle should only contain full blocks - padded via the offset.
```
paplorinc: Not yet the case ...

zsolt-donca:
Hi, Lőrinc,

It's great to see development in the area of persistent collections. You did a great job, especially considering that you didn't just copy-paste an existing implementation (such as Scala's), but you decided to implement it yourself.

It's great that you measured the performance of your new implementation, together with the performance of existing (reference) implementations. You could further improve the performance details by:
- explain what each operation is in the context of each implementation; this would be particularly helpful in the context of *mutable* collection types, where the comparison inherently feels like comparing apples to oranges;
- adding a table with the big-o complexities (time, memory); surely, this would be obviously be mostly filled with "effectively constant", but even then, its an important thing to have and to visualise;

Also, I noticed that this is a "work in process' pull request. It's not clear to me what is actually left to do here before merging. You mention that some operations are not optimized enough; do you think that these should be improved before actually merging?

javaslang/src/main/java/javaslang/collection/Arrays2.java
```diff
@@ -0,0 +1,131 @@
...
+
+            @Override
+            public boolean hasNext() { return index < array.length; }
+
+            @Override
+            public T next() { return array[index++]; }
+        };
+    }
+
+    static <T> T[] inject(Object arrayObject, int index, T element) {
```
zsolt-donca: It's somewhat counter-intuitive to me that this method covers three different use cases:
1. injecting to an empty or null array produces a new single-element array, ignoring `index`;
2. injecting with a positive index "inserts" the elem to the `index` position;
3. injecting with a negative index shifts the whole array and inserts the to the beginning.

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+import java.util.function.Function;
+
+import static javaslang.Function2.constant;
+import static org.assertj.core.api.Assertions.assertThat;
+
+// TODO should rewrite javaslang-test, once it's accessible from here
```
zsolt-donca: +1 to rewriting the below test with a property-based testing framework. I don't know about _javaslang-test_, but I can recommend [scalacheck](https://www.scalacheck.org/).

javaslang/src/main/java/javaslang/collection/Vector.java
danieldietrich: > it wasn't even functional in Scala for 6 years, when someone finally gave it a try for billions of elements

wow

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -565,6 +608,7 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
         } else if (isEmpty()) {
             return ofAll(elements);
         } else {
+            // TODO should be done via merging of two Vectors, there are too many unnecessary array copies done this way
```
danieldietrich: please file an issue

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -611,15 +655,10 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
-        if (n >= length()) {
+        } else if (n >= length()) {
             return empty();
         }
-        HashArrayMappedTrie<Integer, T> trie = HashArrayMappedTrie.empty();
-        for (int i = n; i < length(); i++) {
-            trie = trie.put(i - n, get(i));
-        }
-        return wrap(trie);
+        throw new UnsupportedOperationException();
```
danieldietrich: I think we should leave this PR open until Vector is finished. The master branch should contain production ready code.

javaslang/src/main/java/javaslang/collection/Arrays2.java
```diff
@@ -13,10 +13,49 @@
  * @author Pap Lőrinc
  * @since 3.0.0
  */
+@SuppressWarnings({"unchecked", "SuspiciousArrayCast"})
 final class Arrays2 { // TODO reuse these in `Array` also
```
danieldietrich: Why do we need Arrays**2**?

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -11,33 +11,82 @@
...
 
-    private final HashArrayMappedTrie<Integer, T> trie;
-    private final int indexShift;
+    static final Vector<?> EMPTY = new Vector<>(emptyArray(), new Vector1<>(emptyArray(), 0), emptyArray());
 
-    private Vector(HashArrayMappedTrie<Integer, T> trie) {
-        this(0, trie);
-    }
+    final VectorTree<T> middle;
+    final T[] firstLeaf, lastLeaf; // TODO we could avoid dropping these by providing a start and end offset (would be a constant memory surplus)
```
danieldietrich: Are these the first/last leaves of all branches?

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
danieldietrich: @zsolt-donca javaslang-test should work similar to scalacheck, both are inspired by Haskell's Quickcheck

javaslang/src/main/java/javaslang/collection/Arrays2.java
danieldietrich: @zsolt-donca any suggestions how to do it better? splitting it into three different methods?

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: It is production ready, this is an intermediary commit :)
But we can leave it open, I don't mind, I'm planning on optimizing it further when I will have time again :)

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: If it's merged, I will, but I consider it WIP until I optimize all the usecases I TODOed :)

javaslang/src/main/java/javaslang/collection/Arrays2.java
paplorinc: Valid question, because we use `java.util.Arrays` from the same classes and I didn't want to fully qualify it.
I think the internal `Collections` should also be renamed to `Collections2`  ([Guava](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/Collections2.html) style), or `SlangCollections`, since we definitely need both from the same namespace.

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: Yes, the ones I'm using as staging areas.
> The first and last arrays have dynamic lengths (are shifted to the correct index on drop), but the middle one has internal offset to avoid shifting.
The first one is only empty if everything is empty; if the last is empty, so is the middle.
The first and last have a max size equal to the branching factor.
// TODO The middle should only contain full blocks - padded via the offset.

javaslang/src/main/java/javaslang/collection/Arrays2.java
paplorinc: It's a valid argument, but unfortunately I need it as-is for the moment, as it should work for `prepend`, `update` and `append` also.
I might be able to refactor it to something more coherent when I'm done with the rest of the optimizations :)

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
paplorinc: @danieldietrich, could you help me in making `javaslang-test` accessible from the project, and rewriting a few of the tests from here, `javaslang-test` style?

javaslang/src/main/java/javaslang/collection/Arrays2.java
danieldietrich: Our approach is to name them the same and fully qualify the `java.*` stuff. It will not leak to the outside because these helper classes are package-private / internal.

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
danieldietrich: I can't handle it before my vacation. I will file an issue and assign me...

I think it is not possible to have javaslang-test as Maven test-dependency within javaslang (core) because it will lead to a cyclic dependency. Moving javaslang-test into javaslang is also no option. Maybe we can create a separate Maven module that contains only tests - we will see...

paplorinc: ... or move the build to Gradle, which probably recognizes that it's still a DAG ...

javaslang/src/main/java/javaslang/collection/Arrays2.java
paplorinc: Thanks @zsolt-donca, fixed this in the latest commit :)

javaslang/src/main/java/javaslang/collection/Arrays2.java
paplorinc: Done

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1013,44 +1089,27 @@ public int length() {
     }
 
     @Override
-    public IndexedSeq<T> removeAll(Predicate<? super T> predicate) {
+    public Vector<T> removeAll(Predicate<? super T> predicate) {
```
paplorinc: Why don't the test catch this?

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1387,23 +1414,376 @@ public String toString() {
...
+
+    static final class Vector1<T> implements VectorTree<T> {
+        private static final long serialVersionUID = 1L;
+
+        private final T[] array;
+
+        Vector1(T[] array) {
+            this.array = array;
+
+            assert length() <= (fullTreeSize(1));
```
paplorinc: lots of expensive asserts for now, to make sure the invariants are kept.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1387,23 +1414,376 @@ public String toString() {
...
+        }
+
+        @Override
+        public VectorTree<T> update(int index, T element) {
+            final Object[] root = recursiveSet(array, offset + index, shiftValue(depth), element, 0);
+            return build(root, offset, length(), depth);
+        }
+
+        /** Since the depth of the tree is always small, a non-tailrecursive call is actually faster than pre-calculating the path (is probably unrolled) */
+        static <T> Object[] recursiveSet(Object arrayObject, int index, int shift, T element, int endShift) {
```
paplorinc: this was a surprising optimization for me :)

paplorinc:
@danieldietrich, @zsolt-donca, @eduardmanas, @ovidiudeac, @djspiewak, please review :)

javaslang/src/main/java/javaslang/collection/Arrays2.java
```diff
@@ -81,7 +81,7 @@
     static <T> T[] append(Object arrayObject, T element) {
         final T[] array = ((T[]) arrayObject);
         final T[] copy = (T[]) new Object[array.length + 1];
-        System.arraycopy(array, 0, copy, 0, array.length);
+        System.arraycopy(array, 0, copy, 0, array.length); // why is this a LOT slower than prepend???
```
paplorinc: https://github.com/javaslang/javaslang/pull/1449/commits/ebad20ce474b04764d3e988ffb04ec59b53b88fb#diff-c6c1876105d111251f38e0916d4d99a4R65 WHY?

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1387,23 +1414,376 @@ public String toString() {
...
+
+        int offset()             { return array.length - length(); }
+        byte length()            { return length; }
+
+        T start(int index) {
+            final int offset = offset();
+            assert (offset >= 0) && (offset < array.length);
+            return end(offset + index);
+        }
+        T end(int index) {
```
zsolt-donca: What do `start`, `end`, `start(int)`, `end(int)` do? It's obvious that they have something to do with indexing relative to the _beginning_ and _end_ of the array, but I can't figure out much more than that. Could you give them better names?

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1387,23 +1414,376 @@ public String toString() {
...
+        private final byte length;
 
-        static <T> Vector<Vector<T>> apply(Vector<T> elements, int k) {
-            if (k == 0) {
-                return Vector.of(Vector.empty());
+        static <T> Leaf<T> empty()                         { return (Leaf<T>) EMPTY; }
+        public static <T> Leaf<T> of(T[] leaf)             { return of(leaf, leaf.length); }
+        public static <T> Leaf<T> of(T[] leaf, int length) { return new Leaf<T>(leaf, (byte) length); }
+
+        private Leaf(T[] array, byte length) {
```
zsolt-donca: Why is the array's length different than the leaf's length ?

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: these apply the offsets to the leading and tailing arrays, truncating them to the stored length.
Will try to think of a better name for them :)

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: This enables fast `drop` for front and back, i.e. in case of the leading and trailing arrays (which are not part of the tree structure), we can keep the full array and store how much we've dropped.
This is useful in case of multiple drop calls (e.g. `tail`) and has a constant memory surplus (in worst case, `2*(branchingFactor-1)`)

javaslang/src/main/java/javaslang/collection/Vector.java
zsolt-donca: > Also, when does it make sense to use start() instead of end()? When otherwise? 

I figured that one out myself: `start` and `start(int)` are only used on `Vector.firstLeaf`, and `end` and `end(int)` are only used on `Vector.lastLeaf`. It seems to me that the two Vector fields (`firstLeaf` and `lastLeaf`) should have different types, something that only exposes the right element collection (`start` and `end`) and indexing (`start(int)`, `end(int)`) methods.

zsolt-donca: Also, I find it counter-intuitive the naming of the `Leaf` class. The only actual tree structure here is `VectorN`, and `Leaf` is not actually used by it. `Leaf` is just the data type for the two buffer-like data structures at the beginning and at the end of the vector (`firstLeaf` and `lastLeaf`); correct me if I'm wrong.

paplorinc: You're perfectly right.
I will try to unify it (maybe even get rid of the class as it uses more memory and is slightly slower because of that additional indirection) :/

paplorinc: I removed the additional indirection without making the code worse - I think it's actually cleaner now.
The speed is also considerably better, almost always faster than anything else, while usually using less or equal memory :).

Thanks @zsolt-donca, anything else? @danieldietrich? @eduardmanas?

javaslang/src/main/java/javaslang/collection/Arrays2.java
```diff
@@ -71,6 +71,13 @@
         return copy;
     }
 
+    static <T> T[] pushBack(T[] array, T element) {
```
paplorinc: basically `append`, but renamed in order to be able to statically import it

javaslang/src/main/java/javaslang/collection/Arrays2.java
```diff
@@ -71,6 +71,13 @@
...
+        final T[] copy = (T[]) new Object[array.length + 1];
+        System.arraycopy(array, 0, copy, 0, array.length); // why is this a *LOT* slower than prepend???
```
paplorinc: Any help would be appreciated here :/

why is 
```java
System.arraycopy(array, 0, copy, 0, array.length);
```
almost twice as fast as
```java
System.arraycopy(array, 0, copy, 0, array.length)
```
?

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -51,12 +53,48 @@
     final T[] leading, trailing;
 
     private Vector(T[] leading, VectorTree<T> middle, T[] trailing) {
+        /* Normalize the 3 containers */
+        if (isNullOrEmpty(leading)) {
```
paplorinc: these assure the invariants presented in the description, i.e.
* that the `leading` is only empty if the whole vector is empty;
* the `middle` tree's leafs have size equal to the `branchingFactor()`;
* the `trailing` is only empty, if the `midle` tree is also.

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -146,12 +148,117 @@ public void shouldDropRight() {
...
+                    if (!expected.isEmpty()) {
+                        final int from = random.nextInt(expected.size());
+                        final int to = random.nextInt(expected.size());
+                        expected = expected.slice(from, to);
+                        actual = assertAreEqual(actual, null, (a, p) -> a.slice(from, to), expected);
+                        history = history.append(Tuple.of(expected, actual));
+                    }
+                }
+
+                history.forEach(t -> assertAreEqual(t._1, t._2)); // test that the modifications are persistent
```
paplorinc: I'm even testing that the previous versions are all intact - just to be sure I didn't accidentally modify an existing array;

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -399,6 +399,20 @@ public Object scala_persistent() {
         }
 
         @Benchmark
+        public Object clojure_persistent() {
```
paplorinc: `Clojure` has only an linear prepend :/

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -44,8 +44,10 @@
     private static final long serialVersionUID = 1L;
 
     static int BRANCHING_BASE = 5;
-    static int branchingMask()   { return (-1 >>> -BRANCHING_BASE); }
-    static int branchingFactor() { return (1 << BRANCHING_BASE); }
+    static int branchingFactor()                   { return (1 << BRANCHING_BASE); }
+    static int firstDigit(int num, int depthShift) { return num >> depthShift; }
```
paplorinc: we don't want to see bit shift operations spread around the code ...

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -56,8 +56,22 @@
...
+        this.trailingLength = trailingLength;
+
+        assert (leading.length <= branchingFactor()) && (trailing.length <= branchingFactor());
+        assert (leadingLength() == 0) || ((leadingOffset < leading.length) && (length() > 0));
+        assert (trailingLength == 0) || (leading.length > this.leadingOffset);
+        assert (middle.length() == 0) || (trailingLength > 0);
+        assert List.range(0, middle.length()).forAll(i -> middle.getLeaf(i).length == branchingFactor());
+    }
+
+    private static <T> Vector<T> normalized(T[] leading, byte leadingOffset, VectorTree<T> middle, T[] trailing, byte trailingLength) {
```
paplorinc: strangely the size of the constructor made the algorithms a lot slower, even if all the below `if` statements were false :|

Separated it to a simple constructor and normalization method, used mainly by `drop` and `take`.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -241,13 +248,16 @@ private Vector(T[] leading, byte leadingOffset, VectorTree<T> middle, T[] traili
         }
 
         final T[] leading = (T[]) copyOfRange(array, 0, Math.min(size, branchingFactor()));
-        final T[] trailing = (T[]) copyOfRange(array, Math.max(leading.length, array.length - branchingFactor()), array.length);
-        if (trailing.length < branchingFactor()) {
+        final int remaining = size - leading.length;
+        int trailingSize = lastDigit(remaining);
+        if (trailingSize == 0) { trailingSize += branchingFactor(); }
+        final T[] trailing = (T[]) copyOfRange(array, array.length - trailingSize, array.length);
+        final int middleSize = remaining - trailingSize;
```
paplorinc: we can calculate the middle size to be full always, and put the rest in `trailing`

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -652,7 +662,11 @@ private Vector(T[] leading, byte leadingOffset, VectorTree<T> middle, T[] traili
         if (isEmpty()) {
             return of(element);
         } else {
-            if (trailingLength < branchingFactor()) {
+            if ((trailingLength == 0) && (leadingLength() < branchingFactor())) {
+                assert (middle.length() == 0) && (trailingLength == 0);
```
paplorinc: if we don't have a `trailing` yet (just a `leading`), we should append there instead

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -812,7 +826,7 @@ public T head() {
         if (isEmpty()) {
             throw new NoSuchElementException("head of empty Vector");
         } else {
-            return get(0);
+            return leading[leadingOffset];
```
paplorinc: `head` will always be in the `leading`, no need to run the checks present in `get` (it's a lot faster this way)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1136,7 +1157,7 @@ public int length() {
 
     @Override
     public Vector<T> reverse() {
-        return (length() <= 1) ? this : foldLeft(empty(), Vector::prepend);
+        return (length() <= 1) ? this : ofAll(reverseIterator());
```
paplorinc: unrelated, but useful :)

javaslang/src/main/java/javaslang/collection/Arrays2.java
```diff
@@ -63,22 +60,22 @@
         return array;
     }
 
-    static <T> T[] pushFront(T[] array, T element) {
+    static <T> T[] copyPrepend(T[] array, T element) {
```
paplorinc: will try to rebase all these modifications to the correct commit, so that I don't modify the same line twice ...

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1454,31 +1454,33 @@ public String toString() {
         return mkString(stringPrefix() + "(", ", ", ")");
     }
 
-    enum Helper {
-        ;
-        static <T> VectorTree<T> build(Object[] array, int offset, int length, byte depthShift) {
-            assert (array != null) && (offset >= 0) && (length >= 0);
+    static final class VectorTree<T> implements Serializable {
```
paplorinc: yup, removed `Helper` and `Vector1` :)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -49,13 +48,13 @@
     static int digit(int num, int depthShift)      { return lastDigit(num >> depthShift); }
     static int lastDigit(int num)                  { return (num & (-1 >>> -BRANCHING_BASE)); }
 
-    static final Vector<?> EMPTY = new Vector<>(emptyArray(), (byte) 0, new Vector1<>(emptyArray()), emptyArray(), (byte) 0);
+    static final Vector<?> EMPTY = new Vector<>(emptyArray(), 0, emptyTree(), emptyArray(), 0);
 
     final VectorTree<T> middle;
     final T[] leading, trailing;
-    final byte leadingOffset, trailingLength;
+    final int leadingOffset, trailingLength;
```
paplorinc: `byte`s weren't as useful as I hoped they would :/

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -75,6 +78,7 @@ public void setup() {
             pcollectionsPersistent = create(org.pcollections.TreePVector::from, javaMutable, v -> areEqual(v, javaMutable));
             fjavaPersistent = create(fj.data.Seq::fromJavaList, javaMutable, v -> areEqual(v, javaMutable));
             slangPersistent = create(javaslang.collection.Vector::ofAll, javaMutable, v -> areEqual(v, javaMutable));
+            slangPersistentInt = create(v -> javaslang.collection.Vector.ofAll(INT_ELEMENTS), javaMutable, v -> areEqual(v, javaMutable) && (v.type == int.class));
```
paplorinc: if you create the `Vector` from primitives, it will store it as such internally

danieldietrich:
Hi Lorinc,

please let me say that your implementation of Vector is a great piece of work and maybe currently the best immutable Vector implementation available inside and outside of the JVM. I'm deeply impressed!

leonardo-da-vinci-vitruvianische-mensch

I've invested thousands of hours within the last three years (beside my daylight job and a family including my wife and 3 (in Nov/Dec 4) kids) to create a community driven platform. I was deeply convinced that I'm able to reify my vision to enhance Java and that the energy will not get lost. I'm proud that so many great people like you bring life to this place.

Please don't feel uncomfortable or upset if I steer Javaslang in a direction you don't fully understand (#1480) and keep on the extraordinary work!

Today I will start to review your Vector impl. Please provide me with all links to resources or papers you've used (if there are some).

Daniel

danieldietrich:
While starting the review of the implementation I came to the following result:

I question the approach of leading and trailing objects in Vector. They seem to make sense only the first BRANCHING_FACTOR (= 32) times using `drop` and `tail`. Beyond that the code appears to be more clunky and significantly less maintainable.

I know that you have a different opinion - and that is good! We need diversity, otherwise progress is stopped. I asked Ruslan this morning to implement a more minimal Bit Mapped Trie for comparison. This is the only way I see to make the difference visible.

My wish, as always, is not an extremely optimized version of a new feature. In Javaslang we need a working version with *main focus on simplicity*. Further improvements are discussed by the team/community. Please do not go that deep with your optimizations with your first PR of a new feature!

In the past we two often had these discussions. This costs unnecessary energy. My experience is that you follow your own goals but there is a clear Javaslang vision. Simplicity is as important as user experience. If in doubt, simplicity and maintainability win.

I see that you have the main vision to get the very most out of the performance. Performance is very important. But we follow the rule that there are several collections, each one with its own strengths and drawbacks. The Javaslang way (along with other collections/languages) is to choose the right collection for the right job.

However, this does not mean that we will not get the most out of the Vector operations. Optimization is a difficult task with many facets. The new Vector will come. And your effort will be honored.

paplorinc:
@danieldietrich, would it help if I explained the whole commit via a hangouts session?

I think the code is significantly simpler than my first few attempts and not over-optimized with questionable gains, like I consider the https://github.com/paplorinc/scala/blob/2.12.x/src/library/scala/collection/immutable/Vector.scala#L273 implementation (I urge you to check it out).
I removed very many premature optimizations, kept only those that have a serious performance or memory impact.

> They seem to make sense only the first BRANCHING_FACTOR (= 32) times using drop and tail.

The `trailing` and `leading` play a very important role in `append` and `prepend` also (they don't depend on the depth of the tree this way, making a lot fewer copies), they make the `head` actually constant, not just effectively, they save `drop` and `dropRight` (therefore `slice`, `tail`, `init` etc.) from making unnecessary copies, saving memory etc.
If you have doubts about their usefulness, I really urge you to remove it temporarily and rerun the benchmarks - I would be the happiest if we can have comparable speeds with simpler codebase!

> In Javaslang we need a working version with main focus on simplicity

It may surprise you, but I fully agree. I think my implementation is among the simplest `Vector` implementations (though compared to Clojure it supports for `prepend` and `slice` also). Feel free to compare it against a `Java` impl: https://github.com/andrewoma/dexx/blob/master/collection/src/main/java/com/github/andrewoma/dexx/collection/Vector.java#L481
If you have concrete suggestions in the code (where a lot simpler code has comparable performance), please let me know!
Would it help if I extracted parts in separate methods/classes?

> I asked Ruslan this morning to implement a more minimal Bit Mapped Trie for comparison.

I worked a lot on achieving this balance between simplicity and performance,  I would rather urge @ruslansennov to review this codebase instead and suggest simplifications.

> I see that you have the main vision to get the very most out of the performance. 

Again, I understand that it may seem that way, but I accepted many drawback (e.g. `get` is sometimes a lot slower than `Scala`), for the sake of simplicity and readable code. If you think I failed in that regard, please let me know how I can fix it.
Please also keep in mind that this is a very complicated algorithm (not unlike [`PriorityQueue`](https://github.com/javaslang/javaslang/blob/master/javaslang/src/main/java/javaslang/collection/PriorityQueue.java#L587) or [`HashArrayMappedTrie`](https://github.com/javaslang/javaslang/blob/master/javaslang/src/main/java/javaslang/collection/HashArrayMappedTrie.java#L317)).

ruslansennov:
> > I asked Ruslan this morning to implement a more minimal Bit Mapped Trie for comparison.

> I worked a lot on achieving this balance between simplicity and performance, I would rather urge @ruslansennov to review this codebase instead and suggest simplifications.

I started investigate your Vector implementation. But it is **EPIC** change, we should review as much variants as possible. I prefer spend some time to see that your proposal is the best (hope it is!)

zsolt-donca:
I'd like to speak a couple of words about this pull request. In the last couple of months @paplorinc  and I have talked, in person, a lot about his new Vector implementation; he explained to me the algorithm, he'd shown me various intermediate versions of code (many of which were significantly more complicated than the current one), we've done code review pairing sessions together, and I've seen how the code gradually matured into what is in the current PR.

This data structure (bit-mapped trie) is inherently complex; and the fact that we want a persistent one further adds complexity. Besides this complexity, there are certain optimizations put in place to reduce the "constant" factor (execution time or memory) for certain operations.

I'd summarize the code in the following way:

1. the base algorithm of the _bit-mapped trie_ data structure is represented by the internal class **`VectorTree`**. The main aspect of the implementation is that it can be the _middle_ of an n-ary tree: that is, a tree that can be incomplete on both its _left_ and _right_ side at the same time (but fully complete in the middle), while keeping the indexing start from `0`, no matter how incomplete the tree is on the left. This allows `Vector` to have effectively-constant `prepend`, `append`, `drop` and `dropRight` operations. How many elements are missing on the left side is represented by the field `VectorTree.offset`, and how many on the right is represented by `VectorTree.length`. The structure itself and the operations on it are a piece of art; @paplorinc  could give us more details on this; maybe in the form of a blog, an article, similarly to how he explained to me, in person, all these.

2. the `leading` and `trailing` fields of `Vector` are an optimization for the operations that operate with the beginning or with the end of the Vector (`head`, `tail`, `prepend`, `append`, `drop`, `dropRight` etc.) to reduce the constant factor in their time complexity. Having these fields indeed adds some complexity to the code; it would be an interesting experiment to have, side by side, the two code base (with and without this optimization), along with the performance results. That way, we could investigate what are the exact benefits (performance) and drawbacks (added code complexity). Maybe @paplorinc could provide us a new branch along with the performance results.

3. the latest version of the code stores primitive types (currently only `int` and `char`) in primitive arrays. This optimization is happening behind the scenes (the user is now aware of this), and has the added benefit of using significantly less memory (4x less in case of `int`s). This optimization has the drawback that the code now needs to delegate all data-array-related operations to helper methods, but the main "outline" of the code remains the same. Also, care must be taken in certain corner-cases (such as appending an object to primitive-backed Vector). I believe that this optimization is not fully functional (certain cases may be erroneous now). As above, it would be nice to have a different branch with just this optimization, along with the performance and memory measurements.

I believe that the above two optimizations (_2._ and _3._ above) should be treated separately, in different pull requests, and have in this pull request for just the main algorithm (described in _1._ above). This way, it's easier to do review and merge: it's easier to see what's part of the core algorithm and what's an optimization, and it is easier to weigh the costs/benefits of any particular optimization.

Overall, I believe that this PR represents tremendous value, and the code base is _relatively_ simple, even with the above two optimizations, especially when comparing to other Vector implementations (Scala's being the worst in terms of code quality, in my opinion).

A small correction:

> I question the approach of leading and trailing objects in Vector. They seem to make sense only the first BRANCHING_FACTOR (= 32) times using drop and tail. Beyond that the code appears to be more clunky and significantly less maintainable.

The right thing to say is that they make sense all time, but every BRANCHING_FACTOR (=32) times the code falls back to accessing the underlying bit-mapped trie. This effectively means that the "constant factor" of the related operations is divided by BRANCHING_FACTOR * tree depth. This can actually lead to pretty significant performance improvement, but, again, we'll see exactly how much it is once we have measurements with and without this optimization.

paplorinc:
Thanks @zsolt-donca, I would gladly split the PR in many smaller PRs (currently did the same thing via commits, as I can't wait for PRs to be merged first).
@danieldietrich, @ruslansennov, please let me know how it would make your work simpler.

PR #1450, opened 2016-07-17
Value::toTraversable refactored

ruslansennov:
I believe this check should be moved to certain collection factory.
Also `Class::isAssignableFrom` is gwt-incompatible.
/cc @valery1707


ruslansennov:
something wrong with travis again, but all tests are passed on my computer

danieldietrich:
Thx!

PR #1451, opened 2016-07-18
Implement converting to map with two functions (key and value mappers)

### LINKED ISSUE #1445: Converter Value to Map Java8-style
Current Value have `toMap(Function<? super T, ? extends Tuple2<? extends K, ? extends V>>)`, but sometimes I want to use method reference in method `toMap(Function<? super T, ? extends K> keyMapper, Function<? super T, ? extends V> valueMapper)`.
Before:
```java
Stream.of(items).toMap(i -> Tuple.of(i.getCode(), i.getWeight()))
```
After:
```java
Stream.of(items).toMap(Item::getCode, Item::getWeight)
```

I suggest add this `alias` with two functions to all methods converting `Value` to `Map`.

 If this methods is applicable to project rules, then I can implement it.

valery1707:
Fix #1445

### DIFF
--- a/javaslang/src/main/java/javaslang/Value.java
+++ b/javaslang/src/main/java/javaslang/Value.java
@@ -559,6 +559,23 @@ public interface Value<T> extends Iterable<T> {
     /**
      * Converts this to a specific {@link java.util.Map}.
      *
+     * @param factory     A {@code java.util.Map} factory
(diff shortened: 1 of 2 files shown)


valery1707:
TravisCI stuck again

danieldietrich:
That looks very nice, thank you!

PR #1453, opened 2016-07-19
Added solution for Euler number 26

jorander:
The solution uses two rather long chains of manipulation of CharSeq/Stream and I thought it got the code hard to read. In order to make the main methods easier to read I experimented with using #transform() as a way to encapsulate manipulations with a common goal without breaking the laziness of Stream. I think I got an pretty good result. What do you think about such an approach?

### DIFF
--- a/javaslang/src/test/java/javaslang/collection/euler/Euler26Test.java
+++ b/javaslang/src/test/java/javaslang/collection/euler/Euler26Test.java
@@ -100,7 +100,7 @@ public class Euler26Test {
     private static Function1<CharSeq, CharSeq> removeTrailingZeroes() {
         return seq -> seq
                 .reverse()
-                .dropWhile(c -> Character.valueOf('0').equals(c)) //Remove any trailing zeroes
+                .dropWhile(c -> c == '0') //Remove any trailing zeroes
                 .reverse();
     }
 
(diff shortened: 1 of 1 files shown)


javaslang/src/test/java/javaslang/collection/euler/Euler26Test.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private static Function1<CharSeq, CharSeq> removeRoundingDigit() {
+        return seq -> seq.dropRight(1);
+    }
+
+    private static Function1<CharSeq, CharSeq> removeTrailingZeroes() {
+        return seq -> seq
+                .reverse()
+                .dropWhile(c -> Character.valueOf('0').equals(c)) //Remove any trailing zeroes
```
danieldietrich: Does unboxing work: `.dropWhile(c -> c == '0')`?

danieldietrich:
Is that Java? ;-) That looks really nice.

javaslang/src/test/java/javaslang/collection/euler/Euler26Test.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private static Function1<CharSeq, CharSeq> removeRoundingDigit() {
+        return seq -> seq.dropRight(1);
+    }
+
+    private static Function1<CharSeq, CharSeq> removeTrailingZeroes() {
+        return seq -> seq
+                .reverse()
+                .dropWhile(c -> c == '0') //Remove any trailing zeroes
+                .reverse();
```
jorander: This rather clunky reversing and re-reversing is because there is no CharSeq#dropRightWhile. Should we add that (and dropRightUntil), or does adding that to CharSeq mean it should be added to other collections as well where it might not be a good method to use?

danieldietrich:
Thank you!

javaslang/src/test/java/javaslang/collection/euler/Euler26Test.java
danieldietrich: Mmhh, if we add it, we should add it to all collections, i.e. to Traversable.

PR #1456, opened 2016-07-21
Added benchmarks for common java idioms

mping:
Hopefully will have the following benchmarks:
- For comprehension
- Tuple vs Pair class?
- Try
- Pattern Matching

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ See [User Guide](http://docs.javaslang.io)
 
 ### Benchmarks
 
-Currently only basic microbenchmarks are available. To run
+Currently only basic microbenchmarks are available. To run:
 
 ```bash
 mvn clean test -P benchmark -pl javaslang-benchmark
--- a/javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
@@ -1,0 +1,76 @@
+package javaslang.idiom;
+
+import javaslang.JmhRunner;
+import javaslang.collection.Array;
+import javaslang.collection.Iterator;
+import org.junit.Test;
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+
+import static javaslang.API.For;
+import static javaslang.JmhRunner.getRandomValues;
+
+/**
+ * Benchmark for nested loops vs javaslang's For().yield comprehensions.
+ *
+ * @see javaslang.API.For2
+ */
+public class ForBenchmark {
+    static final Array<Class<?>> CLASSES = Array.of(
+            ForComprehensionNestedFor.class
+    );
+
+    @Test
+    public void testAsserts() {
+        JmhRunner.runDebugWithAsserts(CLASSES);
+    }
+
+    public static void main(String... args) {
+        JmhRunner.runNormalNoAsserts(CLASSES);
+    }
+
+    @State(Scope.Benchmark)
+    public static class Base {
+        @Param({ "10", "20", "30" })
+        public int CONTAINER_SIZE;
+
+        List<Integer> ELEMENTS;
+
+        @Setup
+        public void setup(Blackhole bh) {
+            Integer[] array = getRandomValues(CONTAINER_SIZE, 0, true);
+            ELEMENTS = Arrays.asList(array);
+        }
+    }
+
+    public static class ForComprehensionNestedFor extends Base {
+
+        @Benchmark
+        public void slang_for(Blackhole bh) {
+            BiFunction<Integer,Integer,Integer> bf = (i,j) -> {bh.consume(i);bh.consume(j); return 0;};
+            List<Integer> res = For(ELEMENTS, ELEMENTS).yield(bf).collect(Collectors.toList());
+
+            assert Iterator.ofAll(res).forAll(i -> i == 0);
+        }
+
+        @Benchmark
+        public void java_for(Blackhole bh) {
+            BiFunction<Integer,Integer,Integer> bf = (i,j) -> {bh.consume(i);bh.consume(j); return 0;};
+            List<Integer> result = new ArrayList<>(CONTAINER_SIZE * CONTAINER_SIZE);
+            for (Integer i : ELEMENTS) {
+                for (Integer j : ELEMENTS) {
+                    result.add(bf.apply(i,j));
+                }
+            }
+
+            assert Iterator.ofAll(result).forAll(i -> i == 0);
+        }
+    }
+
+}
--- a/javaslang-benchmark/src/test/java/javaslang/idiom/PatternMatchingBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/idiom/PatternMatchingBenchmark.java
@@ -1,0 +1,87 @@
+package javaslang.idiom;
+
+import javaslang.JmhRunner;
+import javaslang.collection.Array;
+import org.junit.Test;
+import org.openjdk.jmh.annotations.*;
+
+import java.util.Random;
+
+import static javaslang.API.*;
+
+/**
+ * Benchmark for nested loops vs javaslang's For().yield comprehensions.
+ *
+ * @see javaslang.API.For2
+ */
+public class PatternMatchingBenchmark {
+    static final Array<Class<?>> CLASSES = Array.of(
+            PatternMatchingMatches.class
+    );
+
+    @Test
+    public void testAsserts() {
+        JmhRunner.runDebugWithAsserts(CLASSES);
+    }
+
+    public static void main(String... args) {
+        JmhRunner.runNormalNoAsserts(CLASSES);
+    }
+
+
+    @State(Scope.Benchmark)
+    public static class Base {
+        @Param({ "2", "3", "4" })
+        public int MATCH_CASES;
+
+        int random;
+
+        @Setup
+        public void setup() {
+            random = new Random(0).nextInt(MATCH_CASES);
+        }
+    }
+
+    public static class PatternMatchingMatches extends Base {
+
+        @Benchmark
+        public Object slang_match_ok() {
+            String res = Match(random).of(
+                    Case($(0), "0"),
+                    Case($(1), "1"),
+                    Case($(2), "2"),
+                    Case($(3), "3"),
+                    Case($(), "error")
+            );
+
+            assert String.valueOf(random).equals(res);
+            return res;
+        }
+
+        @Benchmark
+        public Object java_switch() {
+            String res;
+
+            switch (random) {
+                case 0:
+                    res = "0";
+                    break;
+                case 1:
+                    res = "1";
+                    break;
+                case 2:
+                    res = "2";
+                    break;
+                case 3:
+                    res = "3";
+                    break;
+                default:
+                    res = "error";
+            }
+
+            assert String.valueOf(random).equals(res);
+            return res;
+        }
+    }
+
+}
--- a/javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
@@ -1,0 +1,92 @@
+package javaslang.idiom;
+
+import javaslang.JmhRunner;
+import javaslang.collection.Array;
+import javaslang.control.Try;
+import org.junit.Test;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.util.Random;
+
+/**
+ * Benchmark for Try vs try/catch.
+ */
+public class TryBenchmark {
+    static final Array<Class<?>> CLASSES = Array.of(
+            TryThrowsCatchBenchmark.class,
+            TryNoThrowsCatchBenchmark.class
+    );
+
+    @Test
+    public void testAsserts() {
+        JmhRunner.runDebugWithAsserts(CLASSES);
+    }
+
+    public static void main(String... args) {
+        JmhRunner.runNormalNoAsserts(CLASSES);
+    }
+
+
+    @State(Scope.Benchmark)
+    public static class Base {
+
+        Integer throwsException() {
+            throw new RuntimeException();
+        }
+
+        Integer doesntThrow() {
+            return 0;
+        }
+    }
+
+    public static class TryThrowsCatchBenchmark extends Base {
+
+        @Benchmark
+        public void slang_try_actual_exception(Blackhole bh) {
+            Integer res = Try.of(this::throwsException)
+                    .recover(RuntimeException.class, 1)
+                    .getOrElse(0);
+
+            assert res == 1;
+            bh.consume(res);
+        }
+
+        @Benchmark
+        public void java_try_actual_exception(Blackhole bh) {
+            try {
+                this.throwsException();
+                assert false;
+            } catch (RuntimeException e) {
+                bh.consume(e);
+            }
+        }
+    }
+
+    public static class TryNoThrowsCatchBenchmark extends Base {
+
+        @Benchmark
+        public void slang_try_no_exception(Blackhole bh) {
+            Integer res = Try.of(this::doesntThrow)
+                    .recover(RuntimeException.class, 1)
+                    .getOrElse(0);
+
+            assert res == 0;
+            bh.consume(res);
+        }
+
+        @Benchmark
+        public void java_try_no_exception(Blackhole bh) {
+            try {
+                Integer res = this.doesntThrow();
+                bh.consume(res);
+            } catch (RuntimeException e) {
+                assert false;
+            }
+        }
+    }
+
+
+}
--- a/javaslang-benchmark/src/test/java/javaslang/idiom/TupleBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/idiom/TupleBenchmark.java
@@ -1,0 +1,72 @@
+package javaslang.idiom;
+
+import javaslang.JmhRunner;
+import javaslang.Tuple;
+import javaslang.Tuple2;
+import javaslang.Tuple3;
+import javaslang.collection.Array;
+import javaslang.collection.Iterator;
+import org.junit.Test;
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.util.*;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+
+import static javaslang.API.For;
+import static javaslang.JmhRunner.getRandomValues;
+
+/**
+ * Benchmark for Tuple2, Tuple3 vs an array.
+ */
+public class TupleBenchmark {
+    static final Array<Class<?>> CLASSES = Array.of(
+            Tuple2Benchmark.class,
+            Tuple3Benchmark.class
+    );
+
+    @Test
+    public void testAsserts() {
+        JmhRunner.runDebugWithAsserts(CLASSES);
+    }
+
+    public static void main(String... args) {
+        JmhRunner.runNormalNoAsserts(CLASSES);
+    }
+
+    @State(Scope.Benchmark)
+    public static class Base {
+    }
+
+    public static class Tuple2Benchmark extends Base {
+
+        @Benchmark
+        public Object slang_tuple2_creation() {
+            Tuple2<String, String> pair = Tuple.of("some", "string");
+            return pair;
+        }
+
+        @Benchmark
+        public Object java_tuple2_creation() {
+            String[] pair = new String[] { "some", "string" };
+            return pair;
+        }
+    }
+
+    public static class Tuple3Benchmark extends Base {
+
+        @Benchmark
+        public Object slang_tuple2_creation() {
+            Tuple3<String, String, String> triplet = Tuple.of("some", "string", "value");
+            return triplet;
+        }
+
+        @Benchmark
+        public Object java_tuple2_creation() {
+            String[] triplet = new String[] { "some", "string", "value" };
+            return triplet;
+        }
+    }
+
+}
--- a/javaslang-benchmark/src/test/java/javaslang/collection/ListBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/collection/ListBenchmark.java
@@ -172,7 +172,7 @@ public class ListBenchmark {
                 java.util.Collections.addAll(javaMutable, state.ELEMENTS);
                 javaMutableLinked.addAll(javaMutable);
                 assert areEqual(javaMutable, asList(state.ELEMENTS))
-                       && areEqual(javaMutableLinked, javaMutable);
+                        && areEqual(javaMutableLinked, javaMutable);
             }
 
             @TearDown(Level.Invocation)
@@ -332,8 +332,8 @@ public class ListBenchmark {
                 }
 
                 assert areEqual(javaMutable, asList(state.ELEMENTS))
-                       && areEqual(javaMutableLinked, javaMutable)
-                       && areEqual(asJavaCollection(scalaMutable), javaMutable);
+                        && areEqual(javaMutableLinked, javaMutable)
+                        && areEqual(asJavaCollection(scalaMutable), javaMutable);
             }
 
             @TearDown(Level.Invocation)


mping:
@danieldietrich let me know how is it looking in #1456

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,66 @@
...
+        @Setup
+        public void setup() {
+            ELEMENTS = getRandomValues(CONTAINER_SIZE, 0);
+            ELEMENTS_ARRAY = Arrays.asList(ELEMENTS);
+            EXPECTED_AGGREGATE = Iterator.of(ELEMENTS).reduce(JmhRunner::aggregate);
+        }
+
+        @Benchmark
+        public Object javaslang_for() {
+            Integer result = For(ELEMENTS_ARRAY).yield(i -> i).reduce((i, j) -> i + j);
```
paplorinc: Thanks for providing alternative benchmarks :).
It would be nice if the build would be passing before asking for code reviews.
Also, please use the same style as the other benchmarks do (e.g. final local variables, method naming, encapsulation in separate classes with a base class)

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,66 @@
...
+
+        @Benchmark
+        public Object java_for() {
+            Integer result = 0;
+            ELEMENTS_ARRAY
+                    .stream()
+                    .map(i -> i)
+                    .reduce((i, j) -> i + j);
+
+            assert result.intValue() == EXPECTED_AGGREGATE;
```
paplorinc: isn't result `0` here???

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,66 @@
...
+        public Object javaslang_for() {
+            Integer result = For(ELEMENTS_ARRAY).yield(i -> i).reduce((i, j) -> i + j);
+            assert result.intValue() == EXPECTED_AGGREGATE;
+            return result;
+        }
+
+        @Benchmark
+        public Object java_for() {
+            Integer result = 0;
+            ELEMENTS_ARRAY
```
paplorinc: you are testing `reduce` here also, not just the `for`

README.md
```diff
@@ -30,5 +30,5 @@ See [User Guide](http://docs.javaslang.io)
 Currently only basic microbenchmarks are available. To run
 
 ```
-mvn clean test -Pbenchmark
+mvn clean test -pl javaslang-benchmark
```
paplorinc: please explain

mping: The profile doesn't exist anymore, so I update the docs to run the benchmarks only.
`pl` stands for project list I guess.

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
mping: will update, didnt notice it was triggering CI.

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
mping: my bad, was a typo. `0` wasn't suppose to be there

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
mping: you're absolutely right, like I said just wanted to get a feel of it. In particular I never used the `for` comprehension so I would not know what would be the right way. Do you have any suggestions on how to properly test it? My rationale was that reduce code was the same in both cases (broken logic I know).

The for comprehension acts both as a map and as a consumer, so I dont exactly know what's the best way to test.

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
mping: True, the `0` was a typo.

mping:
Tests are failing because of OOM, since For comprehension essentially does a cross product I'll use smaller arrays.

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,122 @@
...
+            return result;
+        }
+
+        @Benchmark
+        public Object java_for() {
+            final List<Integer> result = new ArrayList<>(ELEMENTS_ARRAY.size() * ELEMENTS_ARRAY.size() * ELEMENTS_ARRAY.size());
+            for (Integer i : ELEMENTS_ARRAY) {
+                for (Integer j : ELEMENTS_ARRAY) {
+                    for (Integer k : ELEMENTS_ARRAY) {
+                        result.add(i * j * k * 0);
```
paplorinc: `* 0` ???

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,122 @@
...
+            final Iterable<Integer> result = For(ELEMENTS_ARRAY, ELEMENTS_ARRAY)
+                    .yield((i, j) -> i * j * 0);
+
+            assert Iterator.ofAll(result).forAll(e -> e == 0);
+            return result;
+        }
+
+        @Benchmark
+        public Object java_for() {
+            final List<Integer> result = new ArrayList<>(ELEMENTS_ARRAY.size() * ELEMENTS_ARRAY.size() * ELEMENTS_ARRAY.size());
```
paplorinc: that's a big array, why would you need something like that???

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,122 @@
...
+
+            assert Iterator.ofAll(result).forAll(e -> e == 0);
+            return result;
+        }
+
+        @Benchmark
+        public Object java_for() {
+            final List<Integer> result = new ArrayList<>(ELEMENTS_ARRAY.size());
+            for (Integer i : ELEMENTS_ARRAY) {
+                result.add(i * 0);
```
paplorinc: check the bytecode, this is probably 0 instantly (though certainly jitted to 0)

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,122 @@
...
+            final List<Integer> result = new ArrayList<>(ELEMENTS_ARRAY.size() * ELEMENTS_ARRAY.size() * ELEMENTS_ARRAY.size());
+            for (Integer i : ELEMENTS_ARRAY) {
+                for (Integer j : ELEMENTS_ARRAY) {
+                    for (Integer k : ELEMENTS_ARRAY) {
+                        result.add(i * j * k * 0);
+                    }
+                }
+            }
+
+            assert Iterator.ofAll(result).forAll(e -> e == 0);
```
paplorinc: I don't really like where all this is going, neither the fact that the build isn't even passing (again, please run it locally before committing).
These tests should be a *LOT* simpler and a lot more intuitive, otherwise we don't know what to say about the results.

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
mping: Just needed something to ensure the assert is simple.

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
mping: Gonna fix it.

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
mping: Forgot to commit/push, thus causing the OOM.

I'm open to suggestions on how to improve the tests, essentially from what I could see the `For` comprehension is sugar for nested for loops: http://static.javadoc.io/io.javaslang/javaslang/2.0.2/javaslang/API.html

Please feel free to suggest on how to test it properly, IMHO the pure java should have at least a nested `foreach`.

I made three examples because I imagine most code has 2 to 3 nested foreach.

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
mping: This is what I saw:
```
      45: astore_3
      46: aload_1
      47: aload_3
      48: invokevirtual #19                 // Method java/lang/Integer.intValue:()I
      51: iconst_0
      52: imul
      53: invokestatic  #20                 // Method java/lang/Integer.valueOf:(I)Ljava/lang/Integer;
      56: invokeinterface #21,  2           // InterfaceMethod java/util/List.add:(Ljava/lang/Object;)Z
```

I don't know how to check if the mult is elided, but like I said please feel free to point me in the right direction.

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,90 @@
...
+        Integer[] ELEMENTS;
+        List<Integer> ELEMENTS_ARRAY;
+
+        @Setup
+        public void setup() {
+            ELEMENTS = getRandomValues(CONTAINER_SIZE, 0);
+            ELEMENTS_ARRAY = Arrays.asList(ELEMENTS);
+        }
+
+        protected Object consume(Blackhole bh, Integer i) {
```
mping: `Blackhole` is from JMH and is designed for the case where you don't need/want to check the result but you do want to prevent DCE, so it's basically something to emulate a side-effect function.

mping:
Tests should be simpler now.

paplorinc:
Will check it out next week (I'm very busy, sorry), until then, please simplify everything as much as possible.
If you're testing the speed of for loops, don't include everything else in it.
Just iterate over the array and either consume it via the blackhole, or rather collect it via a cheap operation like xor so that we can have some assertions that fail if we mess it up (like they failed until now when you made an error, that's exactly why we need them :)

mping:
Will use a xor, thanks.

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+
+    @State(Scope.Benchmark)
+    public static class Base {
+        @Param({ "10", "100", "100" })
+        public int CONTAINER_SIZE;
+
+        Integer[] ELEMENTS;
+        List<Integer> ELEMENTS_ARRAY;
```
paplorinc: this doesn't seem like an array :)
Do you need both?

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,90 @@
...
+
+        @Setup
+        public void setup() {
+            ELEMENTS = getRandomValues(CONTAINER_SIZE, 0);
+            ELEMENTS_ARRAY = Arrays.asList(ELEMENTS);
+        }
+
+        protected Object consume(Blackhole bh, Integer i) {
+            bh.consume(i.intValue());
```
paplorinc: why would you unbox it by hand?
Please inline this, there's no need to do that (don't take my word for it, measure the difference)

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,90 @@
...
+
+        @Benchmark
+        public void java_for(Blackhole bh) {
+            for (Integer i : ELEMENTS_ARRAY) {
+                consume(bh, i);
+            }
+        }
+    }
+
+    public static class ForComprehension2Iterators extends Base {
```
paplorinc: I don't think we need this, you're already checking the speed of the same thing in the above example.
Is there any significant difference between the two approaches? (besides that this is `O(n^2)`)?

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
mping: I'm going to go for the `xor` because of added benefit of validity, as soon as I have some free time.

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
mping: You're right, I just need a list.

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
mping: My benchmarks seemed wildly different:

```
ForBenchmark  ForComprehension1Iterator   javaslang_for/java_for          0,88x
ForBenchmark  ForComprehension2Iterators  javaslang_for/java_for         11,89x
```

Gonna simplify and make sure there's nothing wrong, after the `xor`.

paplorinc: yes, the difference will be squared also

mping: Problem is that the `For` comprehension mostly makes sense for nested iterations, otherwise you could use a regular `map`. I'd like the benchmark to be similar to real-world usage, since it is comparing common idioms IMHO.

mping:
Geez, I'm getting horrible results with the XOR trick. @paplorinc can you share you idea with some code? I'm currently doing this:

 `Optional<Integer> res = For(ELEMENTS, ELEMENTS).yield((i, j) -> i ^ j).collect(XOR);`

But the results are horrible in comparison to the nested `for`. I'm guessing I am doing something wrong, feel free to point me in the right direction.

paplorinc:
I would be surprised if the simple for loop (which is often optimized by JIT anyway) wouldn't be at least `10x` faster ... maybe even `100x`.
However, you should get the same results with or without the `xor` (actually any constant size operation would only bring the values closer together).
Please try to debug it manually, I would do the same :).

mping:
Cool, I will proceed then, and add more benchmarks, for the rest of the idioms.
The `For().yield` implementation is quite simple via `flatmap` but I was wondering if it wouldn't be better to just nest the loops to squeeze more performance.

paplorinc:
Try it out and report your findings :)

mping:
Added a couple more benchmarks, now only pattern matching is missing.


```
Ratios <alternative_impl> / slang
Target        Operation                                                                          Ratio           

-- try -- 
TryBenchmark  TryThrowsCatchBenchmark         java_try_actual_exception/javaslang_try_actual_exception     2,20x
TryBenchmark  TryNoThrowsCatchBenchmark               java_try_no_exception/javaslang_try_no_exception     0,87x

-- tuple --
TupleBenchmark  Tuple2Benchmark       java_tuple2_creation/javaslang_tuple2_creation     1,00x


-- for comprehension  --
ForBenchmark  ForComprehensionNestedFor       java_for/javaslang_for    11,47x


```

README.md
```diff
@@ -27,8 +27,15 @@ See [User Guide](http://docs.javaslang.io)
 
 ### Benchmarks
 
-Currently only basic microbenchmarks are available. To run
+Currently only basic microbenchmarks are available. To run:
 
 ```
-mvn clean test -Pbenchmark
+mvn clean test -pl javaslang-benchmark -Dtest=TryBenchmark
```
danieldietrich: Do we already run a specific microbenchmark here? See line 37 and line 40.

javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
```diff
@@ -0,0 +1,111 @@
...
+            random = new Random().nextInt();
+        }
+    }
+
+    public static class TryThrowsCatchBenchmark extends Base {
+
+        @Benchmark
+        public Object javaslang_try_actual_exception() {
+            Integer res = Try.of(this::throwsException)
+                    .recover(x -> Match(x).of(Case(instanceOf(RuntimeException.class), () -> 1)))
```
danieldietrich: looking at this we should add a method `Try.recover(Class, Function)`:

```java
// recovers this Try with f if this is a failure of type X, otherwise returns this
default <X extends Throwable> Try<T> recover(Class<X> exception, Function<? super X, ? extends T> f) {
    if (isFailure()) {
        final Throwable cause = getCause();
        if (exception.isAssignableFrom(cause.getClass())) {
            return Try.of(() -> f.apply(cause));
        }
    }
    return this;
}
```

in order to get a way around the Match expression. Then we have

```java
Try.of(this::throwsException).recover(RuntimeException.class, x -> 1);
```

README.md
mping: Oops, too much copy/paste. Will fix.

javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
mping: Yeah looks nicer, will use that.

danieldietrich: Finally merged the new Try.recover syntax. Please use it. Then I will pull this PR in... Thx!

danieldietrich: I've simplified it even more. Please use `Try.recover(Class, T)` instead of `Try.recover(Class, Function)`. See #1501

javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+
+    @State(Scope.Benchmark)
+    public static class Base {
+
+        int random;
+
+        Integer throwsException() {
+            if ((random ^ random) == 0) throw new RuntimeException();
```
paplorinc: `random ^ random` == false ... always
https://en.wikipedia.org/wiki/Exclusive_or#Truth_table

javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
```diff
@@ -0,0 +1,111 @@
...
+            return 1;
+        }
+
+        Integer doesntThrow() {
+            if ((random ^ random) != 0) throw new RuntimeException();
```
paplorinc: same

javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
```diff
@@ -0,0 +1,111 @@
...
+            return 0;
+        }
+
+        @Setup
+        public void setup() {
+            random = new Random().nextInt();
```
paplorinc: * you can put this in the field instead
* please init the random with e.g. `0`, we want deterministic performance tests!!!

javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
```diff
@@ -0,0 +1,111 @@
...
+                    .recover(x -> Match(x).of(Case(instanceOf(RuntimeException.class), () -> 1)))
+                    .getOrElse(0);
+
+            assert res == 1;
+            return res;
+        }
+
+        @Benchmark
+        public Object java_try_actual_exception() {
+            Integer res = 0;
```
paplorinc: you could make this a `void` and andd a `Blackhole bh` as parameter
try {
	this.throwsException();
	assert false
} catch (RuntimeException e) {
        bh.consume(e);
}

(please apply these suggestions to the rest of the codebase also)

javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
```diff
@@ -0,0 +1,111 @@
...
+            Integer res = Try.of(this::doesntThrow)
+                    .recover(x -> Match(x).of(Case(instanceOf(RuntimeException.class), () -> 1)))
+                    .getOrElse(0);
+
+            assert res == 0;
+            return res;
+        }
+
+        @Benchmark
+        public Object java_try_no_exception() {
```
paplorinc: same here

javaslang-benchmark/src/test/java/javaslang/idiom/TupleBenchmark.java
```diff
@@ -0,0 +1,67 @@
...
+            string = new String(bytes);
+        }
+    }
+
+    public static class Tuple2Benchmark extends Base {
+
+        @Benchmark
+        public Object javaslang_tuple2_creation(Blackhole bh) {
+            Tuple2<String, String> tuple = Tuple.of(string, string);
+            bh.consume(tuple);
```
paplorinc: you should either have a `bh` or a `return`

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,79 @@
...
+            assert Iterator.ofAll(res).forAll(i -> i == 0);
+            return res;
+        }
+
+        @Benchmark
+        public Object java_for() {
+            List<Integer> result = new ArrayList<>(CONTAINER_SIZE * CONTAINER_SIZE);
+            for (Integer i : ELEMENTS) {
+                for (Integer j : ELEMENTS) {
+                    result.add(i ^ j);
```
paplorinc: FYI: you're measuring unboxing here also

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,79 @@
...
+            ELEMENTS = Arrays.asList(array);
+            XOR = Collectors.reducing((i, j) -> i ^ j);
+        }
+    }
+
+    public static class ForComprehensionNestedFor extends Base {
+
+        @Benchmark
+        public Object javaslang_for() {
+            List<Integer> res = For(ELEMENTS, ELEMENTS).yield((i, j) -> i ^ j).collect(Collectors.toList());
```
paplorinc: `.collect` ?
@danieldietrich?

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,79 @@
...
+
+        List<Integer> ELEMENTS;
+        Collector<Integer, ?, Optional<Integer>> XOR;
+
+        @Setup
+        public void setup() {
+            Integer[] array = new Integer[CONTAINER_SIZE];
+            Arrays.fill(array, 1);
+            ELEMENTS = Arrays.asList(array);
+            XOR = Collectors.reducing((i, j) -> i ^ j);
```
paplorinc: you're defining this, but not using it below, e.g. line `59`

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,79 @@
...
+        @Param({ "10" })
+        public int CONTAINER_SIZE;
+
+        List<Integer> ELEMENTS;
+        Collector<Integer, ?, Optional<Integer>> XOR;
+
+        @Setup
+        public void setup() {
+            Integer[] array = new Integer[CONTAINER_SIZE];
+            Arrays.fill(array, 1);
```
paplorinc: please use deterministic randoms here, otherwise `xor` could have a predictable outcome

javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
mping: Since it was merged, I can use it now.

javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
mping: I'm just trying to write a method that throws/doesn't throw in a way that the VM doesn't optimize away.
Not entirely sure that this is the best way though. As usual, open to suggestions.

javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
mping: Will do.

javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
mping: Will do.

mping:
Sorry for the delay guys, was out in vacation for a week. Will try to implement suggestions ASAP.

javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
paplorinc: not exactly sure what you would like to achieve, but this is certainly wrong :)

mping: I don't know if the JVM will determine that a method always throws an exception and inline/optimize away some `if`s in case it determines some branches aren't possible. Dunno if it can actually happen, though.

I guess I can measure it :)

paplorinc: Please do :)

Why would it be a problem if the method would always throw?
Also, the[ `JMH` samples](http://hg.openjdk.java.net/code-tools/jmh/file/tip/jmh-samples/src/main/java/org/openjdk/jmh/samples/) contain snippets of when it optimizes stuff and when it cannot :)

mping: Was afraid that the JVM would magically elide the unreachable branch. Will simplify, I guess the JVM is not that smart. The samples are really cool, thanks!

paplorinc: dead code is almost certainly eliminated, but why would you wanna' test that?

mping: Because in the "plain java" benchmark, if the jvm figures out one of the branches is not reachable due to a method never/always throwing an exception, the comparison is not really fair. In real life code, it would be much harder to figure out that a branch is not reachable.

Maybe I'm overreaching here :)

danieldietrich:
Do you know why the `ForBenchmark is failing`?

```
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.27 sec <<< FAILURE! - in javaslang.idiom.ForBenchmark
testAsserts(javaslang.idiom.ForBenchmark)  Time elapsed: 0.27 sec  <<< ERROR!
java.lang.RuntimeException: org.openjdk.jmh.runner.RunnerException: Benchmark caught the exception
	at javaslang.idiom.ForBenchmark.testAsserts(ForBenchmark.java:29)
Caused by: org.openjdk.jmh.runner.RunnerException: Benchmark caught the exception
	at javaslang.idiom.ForBenchmark.testAsserts(ForBenchmark.java:29)
Caused by: java.lang.AssertionError
```

mping:
No I don't. Will check it out today. Was an assert that failed :(
@danieldietrich is there a report for the benchmarks generated anywhere? would love to write a blog post on this.

danieldietrich:
A blog post would be awesome! I would appreciate to review it before publishing because interpreting benchmarks is a difficult topic. Four eyes see more than two.

Actually I don't know how to run the benchmarks :-/ The recent command `mvn test -Pbenchmark -pl javaslang` does not work any more. I asked @paplorinc for the details and will keep you up-to-date. We will document it, see #1527

paplorinc:
Sure, a blog post would be nice, but a lot of things are very hard to interpret here currently, I really urge you to understand JMH well first, it's a lot more difficult than it seams - and as I tried to point out in these benchmarks, they are still far from being finished.

mping:
I appreciate everyone's advice and input, I realize it's very tricky to interpret the results but my goal was never to provide an accurate measurement. I just want to get an idea of what to expect in terms of performance. I know using JMH is hard, but I think this is a good and practical way to learn

danieldietrich:
@mping Yes, we need that blog post. I would greatly appreciate if you would write a guest post on the Javaslang blog, including your Bio (/Photo etc) :-) The format is would be markdown.

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,81 @@
...
+        JmhRunner.runDebugWithAsserts(CLASSES);
+    }
+
+    public static void main(String... args) {
+        JmhRunner.runNormalNoAsserts(CLASSES);
+    }
+
+    @State(Scope.Benchmark)
+    public static class Base {
+        @Param({ "10" })
```
paplorinc: Please provide more params so that we can observe a trend

javaslang-benchmark/src/test/java/javaslang/idiom/ForBenchmark.java
```diff
@@ -0,0 +1,81 @@
...
+        public void setup() {
+            Integer[] array = getRandomValues(CONTAINER_SIZE, 0, true);
+            ELEMENTS = Arrays.asList(array);
+        }
+
+        Integer combineNumbers(Integer i, Integer j) {
+            //with random numbers, we need an op that is predictable,
+            //so we XOR both numbers and add them. It will always be zero.
+            //this way we can assert the end result
+            return (i ^ i) + (j ^ j);
```
paplorinc: Please stop doing this, `(i ^ i) + (j ^ j)` == 0 and the compiler knows this.
Measure, don't guess!

javaslang-benchmark/src/test/java/javaslang/idiom/PatternMatchingBenchmark.java
```diff
@@ -0,0 +1,74 @@
...
+
+
+    @State(Scope.Benchmark)
+    public static class Base {
+        int random;
+
+        @Setup
+        public void setup() {
+            random = new Random().nextInt();
+            random = random ^ random; //always 0, but compiler doesn't know
```
paplorinc: ... yes it does ...

javaslang-benchmark/src/test/java/javaslang/idiom/PatternMatchingBenchmark.java
```diff
@@ -0,0 +1,74 @@
...
+
+            assert "zero".equals(res);
+            return res;
+        }
+
+        @Benchmark
+        public Object java_switch() {
+            String res;
+
+            switch (random) {
```
paplorinc: you're testing binary branching, i.e. an if.
A switch is meant for more branches.
java's `switch` can be linear or can sometimes binary search or a lookup table.
Javaslang's is always linear.

Please test them properly.

paplorinc: `switch (random)` when `random` is actually a constant...

javaslang-benchmark/src/test/java/javaslang/idiom/TryBenchmark.java
```diff
@@ -0,0 +1,92 @@
...
+            bh.consume(res);
+        }
+
+        @Benchmark
+        public void java_try_no_exception(Blackhole bh) {
+            try {
+                Integer res = this.doesntThrow();
+                bh.consume(res);
+            } catch (RuntimeException e) {
+                assert false;
```
paplorinc: :+1:

javaslang-benchmark/src/test/java/javaslang/idiom/TupleBenchmark.java
```diff
@@ -0,0 +1,63 @@
...
+
+
+    @State(Scope.Benchmark)
+    public static class Base {
+        String string;
+
+        @Setup
+        public void setup() {
+            byte[] bytes = new byte[100];
+            new Random().nextBytes(bytes);
```
paplorinc: please use the provided random ... why do you need bytes here?

javaslang-benchmark/src/test/java/javaslang/idiom/TupleBenchmark.java
```diff
@@ -0,0 +1,63 @@
...
+
+        @Benchmark
+        public void javaslang_tuple2_creation(Blackhole bh) {
+            Tuple2<String, String> tuple = Tuple.of(string, string);
+            bh.consume(tuple);
+        }
+
+        @Benchmark
+        public void java_tuple2_creation(Blackhole bh) {
+            String[] pair = new String[] { string, string };
```
paplorinc: `Arrays.asList(string, string)`

javaslang-benchmark/src/test/java/javaslang/idiom/TupleBenchmark.java
```diff
@@ -0,0 +1,63 @@
...
+            string = new String(bytes);
+        }
+    }
+
+    public static class Tuple2Benchmark extends Base {
+
+        @Benchmark
+        public void javaslang_tuple2_creation(Blackhole bh) {
+            Tuple2<String, String> tuple = Tuple.of(string, string);
+            bh.consume(tuple);
```
paplorinc: single elements should rather be returned, not consumed

javaslang-benchmark/src/test/java/javaslang/idiom/TupleBenchmark.java
```diff
@@ -0,0 +1,63 @@
...
+            byte[] bytes = new byte[100];
+            new Random().nextBytes(bytes);
+            string = new String(bytes);
+        }
+    }
+
+    public static class Tuple2Benchmark extends Base {
+
+        @Benchmark
+        public void javaslang_tuple2_creation(Blackhole bh) {
```
paplorinc: the name used elsewhere is just `slang`.
Please do your homework, it's frustrating to point out so many mistakes that you should have caught...

mping: Sorry for mistakes, will fix all aforementioned issues. 
I was on vacation, so didn't have much time to work on this, but hopefully will fix the PR today.

Thanks again everyone for your patience.

danieldietrich: Hi @mping, good to hear. Take your time!

danieldietrich:
I will pull this PR in now - we need to burn down our issues and PRs in order to get forward. Optimizations of the Benchmarks can occur later.

@mping Many thanks!!

mping:
@danieldietrich yay thanks!

AnthonyKot:
JFYI, looks like the error was mentioned by danieldietrich on 24 Aug 2016 is still here:
-------------------------------------------------------------------------------
Test set: io.vavr.idiom.ForBenchmark
-------------------------------------------------------------------------------
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.475 sec <<< FAILURE! - in io.vavr.idiom.ForBenchmark
testAsserts(io.vavr.idiom.ForBenchmark)  Time elapsed: 0.475 sec  <<< ERROR!
java.lang.RuntimeException: org.openjdk.jmh.runner.RunnerException: Benchmark caught the exception
	at io.vavr.idiom.ForBenchmark.testAsserts(ForBenchmark.java:38)
Caused by: org.openjdk.jmh.runner.RunnerException: Benchmark caught the exception
	at io.vavr.idiom.ForBenchmark.testAsserts(ForBenchmark.java:38)
Caused by: org.openjdk.jmh.runner.BenchmarkException: Benchmark error during the run
	at io.vavr.idiom.ForBenchmark.testAsserts(ForBenchmark.java:38)

danieldietrich:
@AnthonyKot thank you, I will create an issue

PR #1473, opened 2016-08-03
Curried checked functions now returns the chain of Function1 followed by single CheckedFunction1 call - exactly what partial application would do.

SimY4:
Since the behaviour of partial application methods changed in 3.0.0 I assume that curried mathod call should match the new behaviour.
It is expected that curried function will only execute its body when all parameters are present therefore Function1 is a safe substitute for all except last interfaces in curried chain.

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -603,17 +603,10 @@ def generateMainClasses(): Unit = {
           case _ => ""
         }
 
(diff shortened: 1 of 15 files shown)


danieldietrich:
Great, this makes sense. Thank you.

danieldietrich:
This one was merged into 3.0.0 branch. Because 3.0.0 was deleted, we cherry-picked this change via #1609

PR #1475, opened 2016-08-04
Bugfix for Utils#isPrime

jorander:

PR #1476, opened 2016-08-05
126 euler number 37

jorander:
Contains(and requires) the bugfix for Utils#isPrime from PR 1475.

### DIFF
--- a/javaslang/src/test/java/javaslang/collection/euler/Utils.java
+++ b/javaslang/src/test/java/javaslang/collection/euler/Utils.java
@@ -17,6 +17,7 @@ import javaslang.API;
 import static javaslang.API.$;
 import static javaslang.API.Case;
 import javaslang.Function1;
+import static javaslang.collection.euler.PrimeNumbers.primes;
 
 public final class Utils {
 
@@ -48,9 +49,7 @@ public final class Utils {
     public static boolean isPrime(long val) {
         return API.Match(val).of(
                 Case($(2L), true),
-                Case($(3L), true),
-                Case($(n -> n > 3), n -> !Stream.rangeClosedBy(3, Math.sqrt(n), 2).exists(d -> n % d == 0)),
-                Case($(), false)
+                Case($(), n -> !primes().takeWhile(d -> d <= Math.sqrt(n)).exists(d -> n % d == 0))
         );
     }
 
--- a/javaslang/src/test/java/javaslang/collection/euler/Utils.java
+++ b/javaslang/src/test/java/javaslang/collection/euler/Utils.java
@@ -49,7 +49,10 @@ public final class Utils {
     public static boolean isPrime(long val) {
         return API.Match(val).of(
                 Case($(2L), true),
-                Case($(n -> n > 2), n -> !primes().takeWhile(d -> d <= Math.sqrt(n)).exists(d -> n % d == 0)),
+                Case($(n -> n > 2), n -> {
+                    final double upperLimitToCheck = Math.sqrt(n);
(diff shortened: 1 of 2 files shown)


javaslang/src/test/java/javaslang/collection/euler/Utils.java
```diff
@@ -48,8 +49,7 @@ public static BigInteger factorial(int n) {
     public static boolean isPrime(long val) {
         return API.Match(val).of(
                 Case($(2L), true),
-                Case($(3L), true),
-                Case($(n -> n > 3), n -> !Stream.rangeClosedBy(3, Math.sqrt(n), 2).exists(d -> n % d == 0)),
+                Case($(n -> n > 2), n -> !primes().takeWhile(d -> d <= Math.sqrt(n)).exists(d -> n % d == 0)),
```
paplorinc: * isn't `n` always bigger than 2 here?
* consider extracting `Math.sqrt(n)`, as we're recalculating it every time now

danieldietrich:
Thank you! Will close #1475 after merge.

javaslang/src/test/java/javaslang/collection/euler/Utils.java
jorander: Thanks for pointing out repeated calculation. Fix is pushed.

Since this is a Utils-method I don't want to assume val to be larger than 2. It might be 1, 0 or even negative.

javaslang/src/test/java/javaslang/collection/euler/Utils.java
```diff
@@ -49,7 +49,10 @@ public static BigInteger factorial(int n) {
     public static boolean isPrime(long val) {
         return API.Match(val).of(
                 Case($(2L), true),
-                Case($(n -> n > 2), n -> !primes().takeWhile(d -> d <= Math.sqrt(n)).exists(d -> n % d == 0)),
+                Case($(n -> n > 2), n -> {
+                    final double upperLimitToCheck = Math.sqrt(n);
```
paplorinc: I think you should cast it to `long`, otherwise it will do the conversion every time

javaslang/src/test/java/javaslang/collection/euler/Utils.java
```diff
@@ -49,7 +49,10 @@ public static BigInteger factorial(int n) {
     public static boolean isPrime(long val) {
         return API.Match(val).of(
                 Case($(2L), true),
-                Case($(n -> n > 2), n -> !primes().takeWhile(d -> d <= Math.sqrt(n)).exists(d -> n % d == 0)),
+                Case($(n -> n > 2), n -> {
```
paplorinc: if you're interested for values `<2`, I think those should come before the `$(2L)`

javaslang/src/test/java/javaslang/collection/euler/Utils.java
```diff
@@ -49,7 +49,10 @@ public static BigInteger factorial(int n) {
     public static boolean isPrime(long val) {
         return API.Match(val).of(
                 Case($(2L), true),
```
paplorinc: You could extend this to all numbers, divisible by `2`, like:
```java
Case($(n -> n % 2 == 0), n -> n == 2)
```

danieldietrich:
Huh!? I thought I've merged this because I already closed #1475. Strange. I'm still on vacation and the internet connection is not that good here on the beach. The merge request may have been ended in the nirvana...

I will merge it now to get the Utils bug-fix in. Further optimizations can be made later if needed.

PR #1477, opened 2016-08-06
Added solution for Euler number 38

jorander:


danieldietrich:
Great, thank you!

(Just saw, that we can improve CharSeq.mkString() and opened #1496)

PR #1480, opened 2016-08-09
Based `CharSeq` on `Vector` instead of `String`

paplorinc:
Depends on https://github.com/javaslang/javaslang/pull/1449

By changing the underlying delegate from `String` to `Vector<Character>`, `tail`, `update`, `prepend`, `append`, `subSequence`, `trim` etc. just became effectively constant (without memory leak), from `linear` :)

```java
Head    slang_persistent/java_persistent        1.19×    1.18×     1.24×
Head    slang_persistent/fjava_persistent       1.36×    1.36×     1.39×
Head    slang_persistent/old_slang_persistent   1.37×    1.31×     1.26×

Tail    slang_persistent/java_persistent        0.97×    0.95×     6.26×
Tail    slang_persistent/fjava_persistent       0.77×    0.50×     0.59×
Tail    slang_persistent/old_slang_persistent   1.15×    1.06×     6.23×

Get     slang_persistent/java_persistent        0.48×    0.12×     0.10×
Get     slang_persistent/fjava_persistent      19.31×  136.45×  1299.65×
Get     slang_persistent/old_slang_persistent   0.63×    0.15×     0.95×

Update  slang_persistent/java_persistent        2.10×    3.36×    22.06×
Update  slang_persistent/old_slang_persistent   2.56×    3.60×    22.08×

Prepend slang_persistent/java_persistent        0.89×    1.78×    11.72×
Prepend slang_persistent/fjava_persistent       2.94×    2.77×     3.00×
Prepend slang_persistent/old_slang_persistent   1.49×    2.47×    11.87×

Append  slang_persistent/java_persistent        0.33×    0.47×     3.35×
Append  slang_persistent/fjava_persistent       2.37×    1.90×     2.16×
Append  slang_persistent/old_slang_persistent   1.04×    2.19×    14.77×

Iterate slang_persistent/java_persistent        4.44×    3.34×     2.56×
Iterate slang_persistent/fjava_persistent       6.15×    5.07×     4.25×
Iterate slang_persistent/old_slang_persistent   0.91×    0.28×     1.50×
```

~~The `char`s are boxed now to `Character`, hence the speed difference in `Get`.~~
UPDATE: changing `Vector` to support primitives made the memory requirements very low, i.e.:
```
for 1000 elements
`java.lang.String` uses `2 KB` (`24 bytes` overhead, `0.0` bytes overhead per element)
`fj.data.LazyString` uses `228.6 KB` (`226.7 KB` overhead, `232.1` bytes overhead per element)
`javaslang.collection.CharSeq` uses `2.7 KB` (`712 bytes` overhead, `0.7` bytes overhead per element)

for 100 elements
`java.lang.String` uses `240 bytes` (`24 bytes` overhead, `0.2` bytes overhead per element)
`fj.data.LazyString` uses `23 KB` (`22.8 KB` overhead, `233.0` bytes overhead per element)
`javaslang.collection.CharSeq` uses `392 bytes` (`152 bytes` overhead, `1.5` bytes overhead per element)

for 10 elements
`java.lang.String` uses `64 bytes` (`24 bytes` overhead, `2.4` bytes overhead per element)
`fj.data.LazyString` uses `2.4 KB` (`2.4 KB` overhead, `241.6` bytes overhead per element)
`javaslang.collection.CharSeq` uses `176 bytes` (`112 bytes` overhead, `11.2` bytes overhead per element)
```

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/CharSeq.java
+++ b/javaslang/src/main/java/javaslang/collection/CharSeq.java
@@ -5,8 +5,10 @@
  */
 package javaslang.collection;
 
-import javaslang.*;
-import javaslang.collection.CharSeqModule.Combinations;
+import javaslang.Kind1;
+import javaslang.Tuple;
+import javaslang.Tuple2;
+import javaslang.Tuple3;
 import javaslang.control.Option;
 
 import java.io.Serializable;
@@ -18,23 +20,42 @@ import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collector;
 
+import static javaslang.collection.Arrays2.toPrimitiveArray;
+
 /**
  * The CharSeq (read: character sequence) collection essentially is a rich String wrapper having all operations
  * we know from the functional Javaslang collections.
  *
- * @author Ruslan Sennov, Daniel Dietrich
+ * @author Ruslan Sennov, Daniel Dietrich, Pap Lőrinc
  * @since 2.0.0
  */
 public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, IndexedSeq<Character>, Serializable {
-
     private static final long serialVersionUID = 1L;
 
-    private static final CharSeq EMPTY = new CharSeq("");
+    private static final CharSeq EMPTY = new CharSeq(Vector.empty());
 
-    private final String back;
+    final Vector<Character> delegate;
+
+    private CharSeq(Vector<Character> chars) {
+        if (chars.nonEmpty() && !chars.type.isPrimitive()) {
+            this.delegate = Vector.ofAll(Arrays2.<char[]> toPrimitiveArray(char.class, chars.toJavaArray()));
+        } else {
+            this.delegate = chars;
+        }
+    }
 
     private CharSeq(String javaString) {
-        this.back = javaString;
+        this(Vector.ofAll(javaString.toCharArray()));
+    }
+
+    private static CharSeq of(Vector<Character> chars) {
+        return chars.isEmpty() ? empty()
+                               : new CharSeq(chars);
+    }
+
+    private CharSeq from(Vector<Character> chars) {
+        return (chars == delegate) ? this
+                                   : of(chars);
     }
 
     public static CharSeq empty() {
@@ -70,7 +91,8 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
         if (sequence instanceof CharSeq) {
             return (CharSeq) sequence;
         } else {
-            return sequence.length() == 0 ? empty() : new CharSeq(sequence.toString());
+            return (sequence.length() == 0) ? empty()
+                                            : new CharSeq(sequence.toString());
         }
     }
 
@@ -81,7 +103,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @return A new {@code CharSeq} instance containing the given element
      */
     public static CharSeq of(char character) {
-        return new CharSeq(new String(new char[] { character }));
+        return of(Vector.ofAll(new char[] {character}));
     }
 
     /**
@@ -92,14 +114,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @throws NullPointerException if {@code elements} is null
      */
     public static CharSeq of(char... characters) {
-        Objects.requireNonNull(characters, "characters is null");
-        if (characters.length == 0) {
-            return empty();
-        } else {
-            final char[] chrs = new char[characters.length];
-            System.arraycopy(characters, 0, chrs, 0, characters.length);
-            return new CharSeq(new String(chrs));
-        }
+        return of(Vector.ofAll(characters));
     }
 
     /**
@@ -113,12 +128,8 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @throws NullPointerException if {@code elements} is null
      */
     public static CharSeq ofAll(Iterable<? extends Character> elements) {
-        Objects.requireNonNull(elements, "elements is null");
-        final StringBuilder sb = new StringBuilder();
-        for (Character character : elements) {
-            sb.append(character);
-        }
-        return sb.length() == 0 ? EMPTY : of(sb);
+        final char[] array = toPrimitiveArray(char.class, Vector.ofAll(elements).toJavaArray());
+        return of(Vector.ofAll(array));
     }
 
     /**
@@ -131,12 +142,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @throws NullPointerException if {@code f} is null
      */
     public static CharSeq tabulate(int n, Function<? super Integer, ? extends Character> f) {
-        Objects.requireNonNull(f, "f is null");
-        final StringBuilder sb = new StringBuilder();
-        for (int i = 0; i < n; i++) {
-            sb.append(f.apply(i));
-        }
-        return of(sb);
+        return of(Vector.tabulate(n, f));
     }
 
     /**
@@ -148,7 +154,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @throws NullPointerException if {@code s} is null
      */
     public static CharSeq fill(int n, Supplier<? extends Character> s) {
-        return tabulate(n, anything -> s.get());
+        return of(Vector.fill(n, s));
     }
 
     /**
@@ -167,11 +173,11 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @return a range of characters as specified or the empty range if {@code from >= toExclusive}
      */
     public static CharSeq range(char from, char toExclusive) {
-        return new CharSeq(Iterator.range(from, toExclusive).mkString());
+        return of(Vector.range(from, toExclusive));
     }
 
     public static CharSeq rangeBy(char from, char toExclusive, int step) {
-        return new CharSeq(Iterator.rangeBy(from, toExclusive, step).mkString());
+        return of(Vector.rangeBy(from, toExclusive, step));
     }
 
     /**
@@ -190,7 +196,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @return a range of characters as specified or the empty range if {@code from > toInclusive}
      */
     public static CharSeq rangeClosed(char from, char toInclusive) {
-        return new CharSeq(Iterator.rangeClosed(from, toInclusive).mkString());
+        return of(Vector.rangeClosed(from, toInclusive));
     }
 
     /**
@@ -214,7 +220,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @throws IllegalArgumentException if {@code step} is zero
      */
     public static CharSeq rangeClosedBy(char from, char toInclusive, int step) {
-        return new CharSeq(Iterator.rangeClosedBy(from, toInclusive, step).mkString());
+        return of(Vector.rangeClosedBy(from, toInclusive, step));
     }
 
     /**
@@ -241,7 +247,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @throws IllegalArgumentException if {@code f} is null
      */
     public static <T> CharSeq unfoldRight(T seed, Function<? super T, Option<Tuple2<? extends Character, ? extends T>>> f) {
-        return CharSeq.ofAll(Iterator.unfoldRight(seed, f));
+        return of(Vector.unfoldRight(seed, f));
     }
 
     /**
@@ -268,7 +274,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @throws IllegalArgumentException if {@code f} is null
      */
     public static <T> CharSeq unfoldLeft(T seed, Function<? super T, Option<Tuple2<? extends T, ? extends Character>>> f) {
-        return CharSeq.ofAll(Iterator.unfoldLeft(seed, f));
+        return of(Vector.unfoldLeft(seed, f));
     }
 
     /**
@@ -295,17 +301,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @throws IllegalArgumentException if {@code f} is null
      */
     public static CharSeq unfold(Character seed, Function<? super Character, Option<Tuple2<? extends Character, ? extends Character>>> f) {
-        return CharSeq.ofAll(Iterator.unfold(seed, f));
-    }
-
-    private Tuple2<CharSeq, CharSeq> splitByBuilder(StringBuilder sb) {
-        if (sb.length() == 0) {
-            return Tuple.of(EMPTY, this);
-        } else if (sb.length() == length()) {
-            return Tuple.of(this, EMPTY);
-        } else {
-            return Tuple.of(of(sb), of(back.substring(sb.length())));
-        }
+        return of(Vector.unfold(seed, f));
     }
 
     /**
@@ -316,10 +312,7 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @return A CharSeq representing {@code character * times}
      */
     public static CharSeq repeat(char character, int times) {
-        final int length = Math.max(times, 0);
-        final char[] characters = new char[length];
-        Arrays.fill(characters, character);
-        return new CharSeq(String.valueOf(characters));
+        return of(character).repeat(times);
     }
 
     /**
@@ -331,11 +324,11 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @return A CharSeq representing {@code this * times}
      */
     public CharSeq repeat(int times) {
-        final StringBuilder builder = new StringBuilder();
+        Vector<Character> result = Vector.empty();
         for (int i = 0; i < times; i++) {
-            builder.append(back);
+            result = result.appendAll(delegate);
         }
-        return of(builder);
+        return from(result);
     }
 
     //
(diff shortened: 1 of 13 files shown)


javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -22,19 +24,32 @@
...
  * @since 2.0.0
  */
 public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, IndexedSeq<Character>, Serializable {
-
     private static final long serialVersionUID = 1L;
 
-    private static final CharSeq EMPTY = new CharSeq("");
+    private static final CharSeq EMPTY = new CharSeq(Vector.empty());
+
+    private final Vector<Character> delegate; // TODO investigate the effect of storing bytes in case of ASCII chars
```
paplorinc: http://openjdk.java.net/jeps/254

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -1439,14 +1059,25 @@ public boolean contentEquals(CharSequence cs) {
...
-     * @param anotherString The {@code CharSeq} to compare this {@code CharSeq} against
+     * @param that The {@code CharSeq} to compare this {@code CharSeq} against
      * @return {@code true} if the argument is not {@code null} and it
      * represents an equivalent {@code CharSeq} ignoring case; {@code
      * false} otherwise
      * @see #equals(Object)
      */
-    public boolean equalsIgnoreCase(CharSeq anotherString) {
-        return back.equalsIgnoreCase(anotherString.back);
+    public boolean equalsIgnoreCase(CharSeq that) {
```
paplorinc: Could be `delegate.map(Character::toLowerCase).equals(that.delegate.map(Character::toLowerCase)) also, but this should be a lot more optimal (it can short circuit in case of non-equality, and only converts to lowercase if not equal)

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -2262,14 +1858,8 @@ public CharSeq replace(CharSequence target, CharSequence replacement) {
...
      */
-    @Deprecated(/* Use splitSeq instead, will be removed in 3.0.0 */)
-    @GwtIncompatible
-    public CharSeq[] split(String regex, int limit) {
-        return splitSeq(regex, limit).toJavaArray(CharSeq.class);
-    }
-
-    public Seq<CharSeq> splitSeq(String regex, int limit) {
-        final Seq<String> split = Array.wrap(back.split(regex, limit));
+    public Seq<CharSeq> split(String regex, int limit) {
```
paplorinc: this is the `3.0.0` branch

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -2039,7 +1635,7 @@ public String stringPrefix() {
      */
     @Override
     public String toString() {
-        return back;
+        return delegate.mkString();
```
paplorinc: should we memoize this?
Maybe with a `SoftReference` to avoid leak?

javaslang/src/main/java/javaslang/collection/package-info.java
```diff
@@ -17,13 +17,13 @@
  * </thead>
  * <tbody>
  * <tr><td>{@linkplain javaslang.collection.Array}</td><td><small>const</small></td><td><small>linear</small></td><td><small>const</small></td><td><small>const</small></td><td><small>linear</small></td><td><small>linear</small></td></tr>
- * <tr><td>{@linkplain javaslang.collection.CharSeq}</td><td><small>const</small></td><td><small>linear</small></td><td><small>const</small></td><td><small>linear</small></td><td><small>linear</small></td><td><small>linear</small></td></tr>
- * <tr><td><em>{@linkplain javaslang.collection.Iterator}</em></td><td><small>const</small></td><td><small>const</small></td><td>&mdash;</td><td>&mdash;</td><td>&mdash;</td><td>&mdash;</td></tr>
+ * <tr><td>{@linkplain javaslang.collection.CharSeq}</td><td><small>const</small></td><td><small>const<sup>eff</sup></small></td><td><small>const<sup>eff</sup></small></td><td><small>const<sup>eff</sup></small></td><td><small>const<sup>eff</sup></small></td><td><small>const<sup>eff</sup></small></td></tr>
```
paplorinc: oh yeah, `tail`, `update`, `prepend`, `append`, `subSequence` just became effectively constant (without memory leak) :)

javaslang/src/test/java/javaslang/collection/CharSeqTest.java
```diff
@@ -1341,7 +1341,7 @@ public void shouldStartsNilOfNonNilCalculate() {
 
     @Test
     public void shouldStartsNilOfNilWithOffsetCalculate() {
-        assertThat(empty().startsWith(empty(), 1)).isFalse();
+        assertThat(empty().startsWith(empty(), 1)).isTrue();
```
paplorinc: I think this test was wrong before, other `seq`s have the value `true` here

ruslansennov:
What about memory consumption in old/new realizations?

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -22,19 +24,32 @@
...
  * @since 2.0.0
  */
 public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, IndexedSeq<Character>, Serializable {
-
     private static final long serialVersionUID = 1L;
 
-    private static final CharSeq EMPTY = new CharSeq("");
+    private static final CharSeq EMPTY = new CharSeq(Vector.empty());
+
+    private final Vector<Character> delegate;
```
paplorinc: // TODO investigate the effect of storing bytes in case of ASCII chars
http://openjdk.java.net/jeps/254

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -1603,47 +1234,12 @@ public boolean regionMatches(int toffset, CharSeq other, int ooffset, int len) {
...
-     * at index {@code toffset}; {@code false} otherwise.
-     * The result is {@code false} if {@code toffset} is
-     * negative or greater than the length of this
-     * {@code CharSeq} object; otherwise the result is the same
-     * as the result of the expression
-     * <pre>
-     *          this.substring(toffset).startsWith(prefix)
-     *          </pre>
-     */
-    public boolean startsWith(CharSeq prefix, int toffset) {
```
paplorinc: already present in parent

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -2499,7 +2068,21 @@ public CharSeq toUpperCase() {
...
      */
     public CharSeq trim() {
-        return of(back.trim());
+        return trimLeft().trimRight();
+    }
+
+    /**
+     * Returns a string whose value is this {@code CharSeq}, with any leading whitespace removed.
+     */
+    public CharSeq trimLeft() {
```
paplorinc: could be useful

danieldietrich:
We should stay with the current implementation. CharSeq is effectively a rich string. Some operations are expected to be linear in time.

The memory footprint matters. Strings are beside numbers the most commonly used language features. In general I expect many small strings instead of many very big ones. For very small `n` `O(n)` is nearly `O(1)`.

If operations are performed that require heavy CPU utilization, the user might take another way to leverage the (new) Vector performance, e.g. by using `toVector()`.

Many operations are specific to the underlying String. It is important to us that these String operations perform the same as known from ordinary Strings. The current changes require us to utilize additional CPU/Mem by first calling `toString`. Example: `toString().offsetByCodePoints(index, codePointOffset)`

So please understand that CharSeq will stay as is.

paplorinc:
@danieldietrich, @ruslansennov, I'm working on making `Vector` a lot more memory efficient, bare with me :)

danieldietrich:
Let's keep CharSeq and BitSet as is and make Vector the best immutable Vector available on the JVM :-)

Don't just look at the `O(x)` complexity. E.g. if the tail operation is never called by users it does not matter. Also O(n) is nearly O(1) for small n.

We should not give up the special internal impls of CharSeq and BitSet.

Please focus on Vector solely.

paplorinc:
> Let's keep CharSeq and BitSet as is and make Vector the best immutable Vector available on the JVM :-)

That's what I'm working on, could you please [review it](https://github.com/javaslang/javaslang/pull/1449) :)?

paplorinc:
> Don't just look at the O(x) complexity.

I looked at the JMH tests, which showed that it's usually faster. Even for small arrays.

> Some operations are expected to be linear in time.

Why, if we can do better?

> The memory footprint matters.

I managed to store primitives in `Vector`, so now the memory footprint is basically the same:
```
for 1000 elements
`java.lang.String` uses `2 KB` (`24 bytes` overhead, `0.0` bytes overhead per element)
`javaslang.collection.CharSeq` uses `2.7 KB` (`712 bytes` overhead, `0.7` bytes overhead per element)
```
We could optimize it further by storing `byte`s instead of `char`s for ASCII Stings.
@danieldietrich, not sure I understand why you aren't more excited about this, we can get better performance, for less memory, with a lot less complicated codebase.

danieldietrich:
I [have the feeling](https://www.getyarn.io/yarn-clip/f2c295c6-ee1c-486e-a23c-dae0c129f634) that we are measuring what we want to see. My experience says that the whole picture has more facets than just the measurement of repeated method calls.

* I think that memory fragmentation may also play a role because Vector has a Tree structure
* I expect Strings to be copied continuous/coherent. Memory can be released relatively simple. (That's just my intention. I don't know how the G1 garbage collector works in particular.)
* I trust the JVM handling Strings efficient. JVM architects are still optimizing it.

Which method calls did we measure? Typical String operations are `substring`, `replaceAll`, ...  How do these operation perform now?

Our benchmarks run in theoretically infinite memory without garbage collection, right, or did we change that setting? A real world app also has GC timeouts. I know that collecting the garbage of many Vector operations will be definitely more cost intensive than freeing mem of Strings.

I don't think this PR will make it into master.

paplorinc:
@danieldietrich, @ruslansennov, @zsolt-donca

> Typical String operations are `substring`, `replaceAll`

1. If we have some concrete usage statistics, we should incorporate them into the JMH tests. 
If we don't, we shouldn't invent one.
2. If really these two are typical, why do we even have a wrapper around `String`?

> Our benchmarks run in theoretically infinite memory without garbage collection

We run `GC` between benchmarks - but then again, this impl can produce less garbage than `String`, not more (i.e. structural sharing, which is not enabled by default in String (i.e. 
`-XX:+UseG1GC` and `-XX:+UseStringDeduplication`)).

> I know that collecting the garbage of many Vector operations will be definitely more cost intensive than freeing mem of Strings.

GC traces live objects, not dead ones (it iterates through the GC roots).
Also, `String` literals are stored in the Constant Pool, they're not evicted by GC.

Instead of `Luke, trust your feelings`, my motto is `Measure, don't guess!`

i.e. memory storage with `byte`s for `ASCII` content would yield:
```java
for 10000 elements
`java.lang.String` uses `19.6 KB` (`24 bytes` overhead, `0.0` bytes overhead per element)
`javaslang.collection.Vector` uses `16.2 KB` (`6.4 KB` overhead, `0.7` bytes overhead per element)

for 1000 elements
`java.lang.String` uses `2 KB` (`24 bytes` overhead, `0.0` bytes overhead per element)
`javaslang.collection.Vector` uses `1.7 KB` (`720 bytes` overhead, `0.7` bytes overhead per element)
```
i.e. `80%` of the original memory usage (for very small `String`s it may be slightly bigger).

And the measurements for `replace` and `substring`:
```java
Operation  Ratio                               10    100   1000   10000
Replace    slang_persistent/java_persistent 0.57×  0.55×  1.44×   1.39×
Slice      slang_persistent/java_mutable    0.49×  0.58×  3.99×  34.95×
```
i.e. for very small `String`s it's about `50%` slower, for bigger ones it's a lot faster.

Conclusion: this implementation scales a lot better than the previous implementation!

ruslansennov:
> ```
> Operation  Ratio                               10    100   1000   10000
> Replace    slang_persistent/java_persistent 0.57×  0.55×  1.44×   1.39×
> Slice      slang_persistent/java_mutable    0.49×  0.58×  3.99×  34.95×
> ```
> i.e. for very small Strings it's about 50% slower, for bigger ones it's a lot faster.


* -43% and +39% for `replace` (the same values)
* -51% and +3500% for `slice` (ooops)

:smile: 

Also I believe there is no sense in benchmarks with &lt; 1000 elements

paplorinc:
@ruslansennov, not sure what's funny, or what your numbers mean.
If you have suggestions, I would appreciate them, if you're only here to mock, at least let me know what you're mocking so that I can improve it.

> Also I believe there is no sense in benchmarks with < 1000 elements

You may be right, that's why I provided bigger values also.

ruslansennov:
I did not want to offend you

> If you have suggestions, I would appreciate them, if you're only here to mock, at least let me know what you're mocking so that I can improve it.

My suggestion is to stay CharSeq as is, without changes.

paplorinc:
> My suggestion is to stay `CharSeq` as is, without changes.

@ruslansennov, @danieldietrich, could you please provide some measurable goals as to why you both would like to keep a more complicated codebase (most `CharSeq` methods are delegated now), using possibly more memory and with slower asymptotic speed?

It seems to me the decision was made before we investigated anything, and even when I provide 
measurable results, they're ignored for some reason I don't yet understand.

danieldietrich:
@paplorinc CharSeq IS DESIGNED TO BE AN ENHANCED STRING. It contains extension methods. The implementation is based on the delegate pattern.

---

**First question: Why do we not implement all other collections based on Vector?**

(The following is intentionally written in an exaggerated style to draw an expressionism-like picture.)

E.g. starting with Array? Your Vector implementation seems to be the most sophisticated collection in the code base.

We could even improve the time complexity of nearly all List methods. Why do we not implement List based on Vector? It may use a little more memory but I think in many cases it can be amortized. If we have List then we also have Queue and Stack.

Let's do Map based on Vector, then we have also Sets.

We could include some if-branches for the lazy case (like for primitive specializations). Then the Vector could also be used to implement the lazy collections.

I am convinced that the benchmarks will prove that _the new methods we measure_ perform way better than the old ones. **So why do we not implement our collections based on Vector?**

The answer is left to the reader.

---

**Second question: Why does the JVM does not use an immutable data structure like the new Vector?**

I don't know. There must be a reason.

ruslansennov:
I do not know what else to say.
I believe `CharSeq` is only needed for the expansion of the `java.lang.String` features to Javaslang level with all conveniences and inconveniences of `java.lang.String`. Nothing more. The realization of `CharSeq` should be plain and silly.
If user wants to use complex operations on characters sequence with good performance it **maybe** has to use a Vector<Character>

valery1707:
@paplorinc maybe you create separate library with this class?

danieldietrich:
@valery1707 I would love to see a real world application before. I have a compiler application in mind that reads source files (based on strings/CharSeq), transforms these into abstract syntax trees and further processes these trees by generating output files. This is a typical, real-world application heavily depending on string operations. This would us give the opportunity to see how the memory/GC behaves.

I've started to implement a parser framework as part of Javaslang in the very beginning (see the [parser](https://github.com/javaslang/javaslang/tree/parser) branch). CharSeq was intended to be a step towards this parser. The parser framework should be a replacement for Antlr. It is intended to be much simpler to use but at least similar powerful/feature rich. The key feature is that it combines the lexer and parser stages. I'm convinced it is possible. It all boils down to handle whitespace correctly...

danieldietrich:
Just one more word for clarification: CharSeq _is a_ String. We can't give up that constraint. In other words `toString` needs to perform in O(1).

In fact we would have named it javaslang.collection.String or javaslang.String but it would raise a conflict with the default import java.lang.String.

If Java had extension methods (or java.lang.String were not final) we would not have created the new type CharSeq. But creating a wrapper is the only solution.

Having the above in mind, changing the type of CharSeq would change the whole idea of a rich String. CharSeq would degrade to an ordinary collection of characters.

paplorinc:
`StringBuilder#toString` is linear also...

zsolt-donca:
`CharSeq` based on a `Vector` could be considered an immutable, rich `StringBuilder`. It could have similar, or even better, performance characteristics than those of `StringBuilder`. The use cases would be, also, the same.

With that parallel drawn in my mind, I do see value in having such `Vector`-based `CharSeq`. But, yes, it should be a separate class and not `CharSeq`.

danieldietrich:
@zsolt-donca interesting idea. A Builder would require to have intermediate operations, like in Java's Stream. The toString would finally create a String. If the Builder would be a Vector we would have problems implementing the original java.lang.String methods, such as `replaceAll(regexp, string)`.

@paplorinc yes, because it is mutable I think...

zsolt-donca:
> If the Builder would be a Vector we would have problems implementing the original java.lang.String methods, such as replaceAll(regexp, string).

It depends on what we consider "trouble". I haven't considered all String methods, but `replaceAll` seems doable (though not trivial). Given that our Vector-based char-sequence implementation implements `java.lang.CharSequence`, we can use [`java.util.regex.Pattern#matcher`](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#matcher-java.lang.CharSequence-) to create a `Matcher` instance. Then, we can progress through all matching ranges replacing them with the given string, and rebuild a new string/rich string/rich stringbuilder.

This would actually be a bonus when compared to [`StringBuilder`](https://docs.oracle.com/javase/8/docs/api/java/lang/StringBuilder.html), which doesn't have String's `replace`, `replaceAll`, and other, similar operations. StringBuilder seems to be limited to `append`, `insert`, and truncate (`setLength`) operations. We could do much more than that (`prepend`, `dropLeft`, `dropRight`) with better performance, even with a persistent data type.

@danieldietrich: when working on your parser library, what do you use for constructing strings (e.g. for error messages)? Wouldn't it feel weird to use something mutable like `StringBuilder` in a world of functional transformations, data types? I guess you could just build Strings or use `CharSeq`, but that would have it's performance drawbacks. Such a rich `StringBuilder` would be the best of both worlds (functional pureness & rich API + performance).

danieldietrich:
@zsolt-donca Some random thoughts while reading your ideas:

* Something like an immutable `CharSeqBuilder` with better performance characteristic (cpu/mem) sounds interesting.
* It could be even `CharSeq.Builder` which is instantiated with `CharSeq.builder()` `CharSeq.builder(CharSequence)`. A `CharSeq` could be obtained by `builder.toCharSeq()` then. Maybe we could make `CharSeq` an interface that is implemented by an internal `CharSeqImpl` and by `CharSeq.Builder`.
* While writing this I think we should change `CharSeq.of(String)` to `CharSeq.of(CharSequence)` and internally initialize the backing String with `this.back = charSequence.toString()`. For `String` it will remain an O(1) operation.
* The name 'Builder' is a little bit weird because it brings mutability to my mind. But that does not really matter. The pattern is the same, even if an immutable context. So `Builder` is the right name.

I have some thoughts on the current Vector implementation, also regarding some specializations - but that is another topic and should not impact the general O(x) performance of an imaginary CharSeq.Builder. I will write these thoughts in the Vector improvement PR...

paplorinc:
> It could be even CharSeq.Builder which is instantiated with CharSeq.builder()

In my opinion, every persistent datastructure is a builder to itself.

Besides the obvious solution of making a separate builder class, I think this implementation could become a builder **and** have an `O(1)` toString at the same time.

People could modify it at will (having a `builder` state) and when done, just call the `toString` method on it (`realized` state).

This could be achieved by having two impls (one storing a `String`, the other a `Vector`), or by simply memoizing the `toString` method.

danieldietrich:
> Besides the obvious solution of making a separate builder class, I think this implementation could become a builder and have an O(1) toString at the same time.

That's what I'm also thinking about. The CharSeq String impl and the CharSeq Vector impl would be roughly the same (both immutable, both the same interface) with different performance characteristics. I'm still not convinced if a builder makes really sense.

Having two impls and calling toString in the end is a solution. The CharSeq could be initialized with the String impl and only 'fall back' to Vector on specific operations. It would be a ping-pong between the impls. Just an idea from the top of my head. Does this make sense?

paplorinc:
FYI: https://www.youtube.com/watch?v=_evzaAkd594
I think it's worth checking out whether we can (or should) bring these Java 9 features to `Java 8` `CharsSeq` via thenew  primitivizable `Vector` impl

PR #1481, opened 2016-08-10
add Tuple withers

ronanM:
Add withers for tuples :

```java
/**
 * Wither of the 2nd element of this tuple.
 *
 * Note: The wither is NOT named {@code _2} due to possible ambiguity when using method reference (Tuple3::_2).
 *
 * @return a copy of this tuple with a new value for the 2nd element of this Tuple.
 */
public Tuple3<T1, T2, T3> w2(T2 t2) {
    return new Tuple3<>(_1, t2, _3);
}
```

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -1098,6 +1098,17 @@ def generateMainClasses(): Unit = {
               public T$j _$j() {
                   return _$j;
               }
+
+              /$javadoc
+               * Wither of the ${j.ordinal} element of this tuple.
+               *
+               * Note: The wither is NOT named {@code _$j} due to possible ambiguity when using method reference ($className::_$j).
+               *
+               * @return a copy of this tuple with a new value for the ${j.ordinal} element of this Tuple.
+               */
+              public $className$generics w$j(T$j t$j) {
+                  return new $className<>(${(1 to (j - 1)).gen(k => s"_$k")(", ")}${(j > 1).gen(", ")}t$j${(j < i).gen(", ")}${((j + 1) to i).gen(k => s"_$k")(", ")});
+              }
             """)("\n\n")}
 
             ${(i == 2).gen(xs"""
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -1100,14 +1100,12 @@ def generateMainClasses(): Unit = {
               }
 
               /$javadoc
-               * Wither of the ${j.ordinal} element of this tuple.
-               *
-               * Note: The wither is NOT named {@code _$j} due to possible ambiguity when using method reference ($className::_$j).
+               * Sets the ${j.ordinal} element of this tuple to the given {@code value}.
                *
                * @return a copy of this tuple with a new value for the ${j.ordinal} element of this Tuple.
                */
-              public $className$generics w$j(T$j t$j) {
-                  return new $className<>(${(1 to (j - 1)).gen(k => s"_$k")(", ")}${(j > 1).gen(", ")}t$j${(j < i).gen(", ")}${((j + 1) to i).gen(k => s"_$k")(", ")});
+              public $className$generics update$j(T$j value) {
+                  return new $className<>(${(1 to (j - 1)).gen(k => s"_$k")(", ")}${(j > 1).gen(", ")}value${(j < i).gen(", ")}${((j + 1) to i).gen(k => s"_$k")(", ")});
               }
             """)("\n\n")}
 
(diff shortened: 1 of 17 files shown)


javaslang/src-gen/main/java/javaslang/Tuple3.java
```diff
@@ -124,6 +124,17 @@ public T1 _1() {
     }
 
     /**
+     * Wither of the 1st element of this tuple.
+     *
+     * Note: The wither is NOT named {@code _1} due to possible ambiguity when using method reference (Tuple3::_1).
+     *
+     * @return a copy of this tuple with a new value for the 1st element of this Tuple.
+     */
+    public Tuple3<T1, T2, T3> w1(T1 t1) {
```
paplorinc: I don't like these abbreviations, why not `with1`, `with2` ... `with3 instead?

ronanM: To be as short as possible like `_1`, `_2`...

paplorinc: I think they're too short and quite confusing, I have to read the doc to understand what they do.

talios: To be honest, I've never liked _1 or _2 in javaslang ( or scala ).

ruslansennov:
What about `set1`, `set2`, etc?

javaslang/src-gen/main/java/javaslang/Tuple3.java
ruslansennov: What about set1, set2, etc?

paplorinc: The [new date-time API has with](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/java/time/ZonedDateTime.java#ZonedDateTime.withMinute%28int%29)s, I think they're less confusing than a `set` (which could imply that the Tuple is mutable)

danieldietrich: We use meaningful names, the _1, _2, ... are one of the few exceptions in Javaslang.

Naming updates on immutable collections are an interesting topic - and very controversial discussed. We will use the names `set1`, `set2`, ... here.

Additionally we should also add `set(index, value)` to `Seq`!

---

In general there are many naming candidates for setting an element at an index.

* `replaceAt(index, value)` (would go adhere with the existing `replace` methods)
* `set(index, value)` (known from mutable collections)
* `updated` (like in Scala)
* `with` (looks like a builder DSL to me)
* ...

We already have `insert(index, element)`, `put(key, value)`, `replace(oldValue, newValue)`, ... These also might be confused with mutable operation - because we are used to mutable collections. If we would only know immutable collections, it would be crystal clear that `seq.set(index, value)` would return a new instance. Therefore there is really no reason to not take the shortest name `set`.

javaslang/generator/Generator.scala
```diff
@@ -1098,6 +1098,17 @@ def generateMainClasses(): Unit = {
               public T$j _$j() {
                   return _$j;
               }
+
+              /$javadoc
+               * Wither of the ${j.ordinal} element of this tuple.
+               *
+               * Note: The wither is NOT named {@code _$j} due to possible ambiguity when using method reference ($className::_$j).
```
danieldietrich: Please remove this line (and the following empty line)

javaslang/generator/Generator.scala
```diff
@@ -1098,6 +1098,17 @@ def generateMainClasses(): Unit = {
               public T$j _$j() {
                   return _$j;
               }
+
+              /$javadoc
+               * Wither of the ${j.ordinal} element of this tuple.
```
danieldietrich: Please change it to `Sets the ${j.ordinal} element of this tuple to the given {@code value}.`

javaslang/generator/Generator.scala
```diff
@@ -1098,6 +1098,17 @@ def generateMainClasses(): Unit = {
...
+               *
+               * Note: The wither is NOT named {@code _$j} due to possible ambiguity when using method reference ($className::_$j).
+               *
+               * @return a copy of this tuple with a new value for the ${j.ordinal} element of this Tuple.
+               */
+              public $className$generics w$j(T$j t$j) {
```
danieldietrich: please change `t$j` to `value` here

danieldietrich: please rename `w1`, ... to `set1`, ...

danieldietrich: **Update:** please rename `w1`, ... to `update1`, ... This will align the names to the existing `Seq.update(int index, T element)` function

danieldietrich:
Awesome, thank you!

Issue #1488, opened 2016-08-13
Unify conversion method names


danieldietrich:
The initial conversion method names followed different naming schemes. While adding new collections the naming schemes diverged even more.

* Conversion methods of Seq impls have impl names in general, e.g. `toArray`, `toCharSeq`, `toList`, `toStream`, `toVector`, `toQueue`
* Conversion methods of Sets/SortedSets, Maps/SortedMaps have interface names, e.g. `toSet`, `toSortedSet` (for `toTreeSet`), ...
* Conversion methods regarding new collections have names of non-existing types, e.g. `toLinkedSet` (for `toLinkedHashSet`), `toSortedQueue` (for `toPriorityQueue`), ...

The task of this issue is to unify the names following the 'what you type is what you get' scheme.

E.g. `toTreeSet` instead of `toSortedSet`.

This will also resolve the issue that we want to return the same instance, if the actual type matches the conversion type. Strictly speaking we would need sealed interfaces/traits in order to prevent `List` to be extended by code outside of Javaslang. But this restriction should be documented at the interface as long as we cannot seal interfaces.

We need to deprecate `toSortedSet` et al instead of removing them because otherwise we would break the 2.x line (but of course only the methods that are present in 2.0.2). Deprecated methods are removed in the next major release (3.0).

danieldietrich:
PR #1485 is a first step

danieldietrich:
This is about naming. We should not break backward-compatibility. I will close this ticket.

PR #1500, opened 2016-08-16
Add aliases in API

### LINKED ISSUE #1431: Provide shorter aliases for static factories
From
```
Tuple.of(1,2,3)
List.of(1,2,3)
```

To

```
tuple(1,2,3)
list(1,2,3)
```

By providing a sort of `Aliases` like this :

```
public final class Aliases {

    public static <A, B> Tuple2<A, B> tuple(A a, B b) {
        return Tuple.of(a, b);
    }

    public static <A, B, C> Tuple3<A, B, C> tuple(A a, B b, C c) {
        return Tuple.of(a, b, c);
    }

    public static <T> List<T> list(T element) {
        return List.of(element);
    }

    public static <T> List<T> list(T... elements) {
        return List.of(elements);
    }

   // many more...
}
```

valery1707:
Fixes #1431

Add aliases for:
* Functions
* Tuples
* Either
* Future
* Lazy
* Option
* Try
* Validation
* CharSeq
* Traversable
* Sorted traversable
* Maps

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -57,6 +57,7 @@ def generateMainClasses(): Unit = {
       val ExecutorServiceType = im.getType("java.util.concurrent.ExecutorService")
       val TryType = im.getType("javaslang.control.Try")
       val ValidationType = im.getType("javaslang.control.Validation")
+      val CharSeqType = im.getType("javaslang.collection.CharSeq")
 
       def genAliases(im: ImportManager, packageName: String, className: String): String = {
         xs"""
@@ -333,6 +334,37 @@ def generateMainClasses(): Unit = {
            */
           public static <E, T> $ValidationType<E, T> Invalid(E error) {
               return $ValidationType.invalid(error);
+          }
+
+          /$javadoc
+           * Alias for {@link $CharSeqType#of(char)}
+           *
+           * @param character A character.
+           * @return A new {@link $CharSeqType} instance containing the given element
+           */
+          public static $CharSeqType CharSeq(char character) {
+              return $CharSeqType.of(character);
+          }
+
+          /$javadoc
+           * Alias for {@link $CharSeqType#of(char...)}
+           *
+           * @param characters Zero or more characters.
+           * @return A new {@link $CharSeqType} instance containing the given characters in the same order.
+           * @throws NullPointerException if {@code elements} is null
+           */
+          public static $CharSeqType CharSeq(char... characters) {
+              return $CharSeqType.of(characters);
+          }
+
+          /$javadoc
+           * Alias for {@link $CharSeqType#of(CharSequence)}
+           *
+           * @param sequence {@code CharSequence} instance.
+           * @return A new {@link $CharSeqType} instance
+           */
+          public static $CharSeqType CharSeq(CharSequence sequence) {
+              return $CharSeqType.of(sequence);
           }
 
         """
@@ -1829,6 +1861,12 @@ def generateTestClasses(): Unit = {
           ${genSimpleAliasTest("Valid", "1")}
 
           ${genSimpleAliasTest("Invalid", "new Error()")}
+
+          ${genMediumAliasTest("Char", "(Iterable<Character>) CharSeq", "'1'")}
+
+          ${genMediumAliasTest("CharArray", "(Iterable<Character>) CharSeq", "'1', '2', '3'")}
+
+          ${genMediumAliasTest("CharSeq", "(Iterable<Character>) CharSeq", "\"123\"")}
 
         """
       }
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -55,6 +55,7 @@ def generateMainClasses(): Unit = {
       val FutureType = im.getType("javaslang.concurrent.Future")
       val CheckedSupplierType = im.getType("javaslang.control.Try.CheckedSupplier")
       val ExecutorServiceType = im.getType("java.util.concurrent.ExecutorService")
+      val TryType = im.getType("javaslang.control.Try")
 
       def genAliases(im: ImportManager, packageName: String, className: String): String = {
         xs"""
@@ -271,6 +272,40 @@ def generateMainClasses(): Unit = {
            */
           public static <T> $OptionType<T> None() {
               return $OptionType.none();
+          }
+
+          /$javadoc
+           * Alias for {@link $TryType#of($CheckedSupplierType)}
+           *
+           * @param <T>      Component type
+           * @param supplier A checked supplier
+           * @return {@link $TryType.Success} if no exception occurs, otherwise {@link $TryType.Failure} if an
+           * exception occurs calling {@code supplier.get()}.
+           */
+          public static <T> $TryType<T> Try($CheckedSupplierType<? extends T> supplier) {
+              return $TryType.of(supplier);
+          }
+
+          /$javadoc
+           * Alias for {@link $TryType#success(Object)}
+           *
+           * @param <T>   Type of the given {@code value}.
+           * @param value A value.
+           * @return A new {@link $TryType.Success}.
+           */
+          public static <T> $TryType<T> Success(T value) {
+              return $TryType.success(value);
+          }
+
+          /$javadoc
+           * Alias for {@link $TryType#failure(Throwable)}
+           *
+           * @param <T>       Component type of the {@code Try}.
+           * @param exception An exception.
+           * @return A new {@link $TryType.Failure}.
+           */
+          public static <T> $TryType<T> Failure(Throwable exception) {
+              return $TryType.failure(exception);
           }
 
         """
@@ -1653,6 +1688,7 @@ def generateTestClasses(): Unit = {
       val FutureType = im.getType("javaslang.concurrent.Future")
       val ExecutorServiceType = im.getType("java.util.concurrent.Executors")
       val ExecutorService = s"${ExecutorServiceType}.newSingleThreadExecutor()"
+      val TryType = im.getType("javaslang.control.Try")
 
       val d = "$";
 
@@ -1691,6 +1727,18 @@ def generateTestClasses(): Unit = {
       def genMediumAliasTest(name: String, func: String, value: String): String = genExtAliasTest(s"${name}ReturnNotNull", func, value, "isNotNull()")
 
       def genSimpleAliasTest(name: String, value: String): String = genMediumAliasTest(name, name, value)
+
+      def genTryTests(func: String, value: String, success: Boolean): String = {
+        val check = if (success) "isSuccess" else "isFailure"
+        xs"""
+          @$test
+          public void should${func.firstUpper}ReturnNotNull() {
+              final $TryType<?> t = $func($value);
+              assertThat(t).isNotNull();
+              assertThat(t.$check()).isTrue();
+          }
+        """
+      }
 
       def genAliasesTests(im: ImportManager, packageName: String, className: String): String = {
         xs"""
@@ -1744,6 +1792,12 @@ def generateTestClasses(): Unit = {
           ${genSimpleAliasTest("Some", "1")}
 
           ${genSimpleAliasTest("None", "")}
+
+          ${genTryTests("Try", "() -> 1", success = true)}
+
+          ${genTryTests("Success", "1", success = true)}
+
+          ${genTryTests("Failure", "new Error()", success = false)}
 
         """
       }
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -230,6 +230,17 @@ def generateMainClasses(): Unit = {
               return $FutureType.failed(executorService, exception);
           }
 
+          /$javadoc
+           * Alias for {@link Lazy#of($SupplierType)}
+           *
+           * @param <T>      type of the lazy value
+           * @param supplier A supplier
+           * @return A new instance of {@link Lazy}
+           */
+          public static <T> Lazy<T> Lazy($SupplierType<? extends T> supplier) {
+              return Lazy.of(supplier);
+          }
+
         """
       }
 
@@ -1636,6 +1647,19 @@ def generateTestClasses(): Unit = {
         """
       }
 
+      def genExtAliasTest(name: String, func: String, value: String, check: String): String = {
+        xs"""
+          @$test
+          public void should${name}() {
+              assertThat($func($value)).$check;
+          }
+        """
+      }
+
+      def genMediumAliasTest(name: String, func: String, value: String): String = genExtAliasTest(s"${name}ReturnNotNull", func, value, "isNotNull()")
+
+      def genSimpleAliasTest(name: String, value: String): String = genMediumAliasTest(name, name, value)
+
       def genAliasesTests(im: ImportManager, packageName: String, className: String): String = {
         xs"""
           ${(0 to N).gen(i => {
@@ -1680,6 +1704,8 @@ def generateTestClasses(): Unit = {
           ${genFutureTests("Value", "1", success = true)}
 
           ${genFutureTests("Error", "new Error()", success = false)}
+
+          ${genSimpleAliasTest("Lazy", "() -> 1")}
 
         """
       }
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -69,6 +69,10 @@ def generateMainClasses(): Unit = {
       val TreeSetType = im.getType("javaslang.collection.TreeSet")
       val PriorityQueueType = im.getType("javaslang.collection.PriorityQueue")
       val JavaComparatorType = im.getType("java.util.Comparator")
+      val LinkedHashMapType = im.getType("javaslang.collection.LinkedHashMap")
+      val HashMapType = im.getType("javaslang.collection.HashMap")
+      val TreeMapType = im.getType("javaslang.collection.TreeMap")
+      val JavaMapType = im.getType("java.util.Map")
 
       def genTraversableAliases(traversableType: String, name: String): String = {
         xs"""
@@ -249,6 +253,72 @@ def generateMainClasses(): Unit = {
            */
           public static <T> $traversableType<T> $name($JavaComparatorType<? super T> comparator, $JavaStreamType<? extends T> elements) {
               return $traversableType.ofAll(comparator, elements);
+          }
+        """
+      }
+
+      def genMapAliases(mapType: String, name: String): String = {
+        xs"""
+          /$javadoc
+           * Alias for {@link $mapType#empty()}
+           *
+           * @param <K> The key type.
+           * @param <V> The value type.
+           * @return A singleton instance of empty {@link $mapType}
+           */
+          public static <K, V> $mapType<K, V> $name() {
+              return $mapType.empty();
+          }
+
+          /$javadoc
+           * Alias for {@link $mapType#of(Object, Object)}
+           *
+           * @param <K>   The key type.
+           * @param <V>   The value type.
+           * @param key   A singleton map key.
+           * @param value A singleton map value.
+           * @return A new {@link $mapType} instance containing the given entry
+           */
+          public static <K, V> $mapType<K, V> $name(K key, V value) {
+              return $mapType.of(key, value);
+          }
+
+          /$javadoc
+           * Alias for {@link $mapType#ofEntries(Tuple2...)}
+           *
+           * @param <K>     The key type.
+           * @param <V>     The value type.
+           * @param entries Map entries.
+           * @return A new {@link $mapType} instance containing the given entries
+           */
+          @SuppressWarnings("varargs")
+          @SafeVarargs
+          public static <K, V> $mapType<K, V> $name(Tuple2<? extends K, ? extends V>... entries) {
+              return $mapType.ofEntries(entries);
+          }
+
+          /$javadoc
+           * Alias for {@link $mapType#ofAll($JavaMapType)}
+           *
+           * @param <K> The key type.
+           * @param <V> The value type.
+           * @param map A map entry.
+           * @return A new {@link $mapType} instance containing the given map
+           */
+          public static <K, V> $mapType<K, V> $name($JavaMapType<? extends K, ? extends V> map) {
+              return $mapType.ofAll(map);
+          }
+
+          /$javadoc
+           * Alias for {@link $mapType#of(Object...)}
+           *
+           * @param <K>   The key type.
+           * @param <V>   The value type.
+           * @param pairs A list of key-value pairs.
+           * @return A new {@link $mapType} instance containing the given entries
+           */
+          public static <K, V> $mapType<K, V> $name(Object... pairs) {
+              return $mapType.of(pairs);
           }
         """
       }
@@ -582,6 +652,113 @@ def generateMainClasses(): Unit = {
           ${genSortedTraversableAliases(TreeSetType, "SortedSet", "extends Comparable<? super T>")}
 
           ${genSortedTraversableAliases(PriorityQueueType, "PriorityQueue", "extends Comparable<T>")}
+
+          ${genMapAliases(LinkedHashMapType, "LinkedMap")}
+
+          ${genMapAliases(HashMapType, "Map")}
+
+          /$javadoc
+           * Alias for {@link $TreeMapType#empty()}
+           *
+           * @param <K> The key type.
+           * @param <V> The value type.
+           * @return A new empty {@link $TreeMapType} instance
+           */
+          public static <K extends Comparable<? super K>, V> $TreeMapType<K, V> SortedMap() {
+              return $TreeMapType.empty();
+          }
+
+          /$javadoc
+           * Alias for {@link $TreeMapType#empty($JavaComparatorType)}
+           *
+           * @param <K>           The key type.
+           * @param <V>           The value type.
+           * @param keyComparator The comparator used to sort the entries by their key
+           * @return A new empty {@link $TreeMapType} instance
+           */
+          public static <K, V> $TreeMapType<K, V> SortedMap($JavaComparatorType<? super K> keyComparator) {
+              return $TreeMapType.empty(keyComparator);
+          }
+
+          /$javadoc
+           * Alias for {@link $TreeMapType#of(Comparable, Object)}
+           *
+           * @param <K>   The key type.
+           * @param <V>   The value type.
+           * @param key   A singleton map key.
+           * @param value A singleton map value.
+           * @return A new {@link $TreeMapType} instance containing the given entry
+           */
+          public static <K extends Comparable<? super K>, V> $TreeMapType<K, V> SortedMap(K key, V value) {
+              return $TreeMapType.of(key, value);
+          }
+
+          /$javadoc
+           * Alias for {@link $TreeMapType#of(Comparator, Object, Object)}
+           *
+           * @param <K>           The key type.
+           * @param <V>           The value type.
+           * @param keyComparator The comparator used to sort the entries by their key
+           * @param key           A singleton map key.
+           * @param value         A singleton map value.
+           * @return A new {@link $TreeMapType} instance containing the given entry
+           */
+          public static <K, V> $TreeMapType<K, V> SortedMap(Comparator<? super K> keyComparator, K key, V value) {
+              return $TreeMapType.of(keyComparator, key, value);
+          }
+
+          /$javadoc
+           * Alias for {@link $TreeMapType#ofEntries(Tuple2...)}
+           *
+           * @param <K>     The key type.
+           * @param <V>     The value type.
+           * @param entries Map entries.
+           * @return A new {@link $TreeMapType} instance containing the given entries
+           */
+          @SuppressWarnings("varargs")
+          @SafeVarargs
+          public static <K extends Comparable<? super K>, V> $TreeMapType<K, V> SortedMap(Tuple2<? extends K, ? extends V>... entries) {
+              return $TreeMapType.ofEntries(entries);
+          }
+
+          /$javadoc
+           * Alias for {@link $TreeMapType#ofEntries($JavaComparatorType, Tuple2...)}
+           *
+           * @param <K>           The key type.
+           * @param <V>           The value type.
+           * @param keyComparator The comparator used to sort the entries by their key
+           * @param entries       Map entries.
+           * @return A new {@link $TreeMapType} instance containing the given entry
+           */
+          @SuppressWarnings("varargs")
+          @SafeVarargs
+          public static <K, V> $TreeMapType<K, V> SortedMap($JavaComparatorType<? super K> keyComparator, Tuple2<? extends K, ? extends V>... entries) {
+              return $TreeMapType.ofEntries(keyComparator, entries);
+          }
+
+          /$javadoc
+           * Alias for {@link $TreeMapType#ofAll($JavaMapType)}
+           *
+           * @param <K> The key type.
+           * @param <V> The value type.
+           * @param map A map entry.
+           * @return A new {@link $TreeMapType} instance containing the given map
+           */
+          public static <K extends Comparable<? super K>, V> $TreeMapType<K, V> SortedMap($JavaMapType<? extends K, ? extends V> map) {
+              return $TreeMapType.ofAll(map);
+          }
+
+          /$javadoc
+           * Alias for {@link $TreeMapType#of(Object...)}
+           *
+           * @param <K>   The key type.
+           * @param <V>   The value type.
+           * @param pairs A list of key-value pairs.
+           * @return A new {@link $TreeMapType} instance containing the given entries
+           */
+          public static <K extends Comparable<? super K>, V> $TreeMapType<K, V> SortedMap(Object... pairs) {
+              return $TreeMapType.of(pairs);
+          }
         """
       }
 
@@ -1965,6 +2142,7 @@ def generateTestClasses(): Unit = {
       val TryType = im.getType("javaslang.control.Try")
       val JavaStreamType = im.getType("java.util.stream.Stream")
       val JavaComparatorType = im.getType("java.util.Comparator")
+      val JavaCollectionsType = im.getType("java.util.Collections")
 
       val d = "$";
 
@@ -2040,6 +2218,21 @@ def generateTestClasses(): Unit = {
           ${genMediumAliasTest(s"${func}WithStream", func, s"$JavaStreamType.of('1', '2', '3')")}
 
           ${genMediumAliasTest(s"${func}WithStreamAndComparator", func, s"Character::compareTo, $JavaStreamType.of('1', '2', '3')")}
+
+        """
+      }
+
+      def genMapTests(func: String): String = {
+        xs"""
+          ${genMediumAliasTest(s"Empty${func}", func, "")}
+
+          ${genMediumAliasTest(s"${func}FromSingle", func, "1, '1'")}
+
+          ${genMediumAliasTest(s"${func}FromTuples", func, "Tuple(1, '1'), Tuple(2, '2'), Tuple(3, '3')")}
+
+          ${genMediumAliasTest(s"${func}FromMap", func, s"$JavaCollectionsType.singletonMap(1, '1')")}
+
+          ${genMediumAliasTest(s"${func}FromPairs", func, "1, '1', 2, '2', 3, '3'")}
 
         """
       }
@@ -2137,6 +2330,15 @@ def generateTestClasses(): Unit = {
 
           ${genSortedTraversableTests("SortedSet")}
           ${genSortedTraversableTests("PriorityQueue")}
+
+          ${genMapTests("LinkedMap")}
+          ${genMapTests("Map")}
+          ${genMapTests("SortedMap")}
+          ${genMediumAliasTest("EmptySortedMapFromComparator", "SortedMap", "Integer::compareTo")}
+
+          ${genMediumAliasTest("SortedMapFromSingleAndComparator", "SortedMap", s"($JavaComparatorType<Integer>)Integer::compareTo, 1, '1'")}
+
+          ${genMediumAliasTest("SortedMapFromTuplesAndComparator", "SortedMap", s"($JavaComparatorType<Integer>)Integer::compareTo, Tuple(1, '1'), Tuple(2, '2'), Tuple(3, '3')")}
         """
       }
 
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -92,6 +92,36 @@ def generateMainClasses(): Unit = {
             """
           })("\n\n")}
 
+          /$javadoc
+           * Alias for {@link Tuple#empty()}
+           *
+           * @return the empty tuple.
+           */
+          public static Tuple0 Tuple() {
+              return Tuple.empty();
+          }
+
+          ${(1 to N).gen(i => {
+            val generics = (1 to i).gen(j => s"T$j")(", ")
+            val links = (1 to i).gen(j => s"Object")(", ")
+            val params = (1 to i).gen(j => s"T$j t$j")(", ")
+            val args = (1 to i).gen(j => s"t$j")(", ")
+            xs"""
+              /$javadoc
+               * Alias for {@link Tuple#of($links)}
+               *
+               * Creates a tuple of ${i.numerus("element")}.
+               *
+               ${(1 to i).gen(j => s"* @param <T$j> type of the ${j.ordinal} element")("\n")}
+               ${(1 to i).gen(j => s"* @param t$j   the ${j.ordinal} element")("\n")}
+               * @return a tuple of ${i.numerus("element")}.
+               */
+              public static <$generics> Tuple$i<$generics> Tuple($params) {
+                  return Tuple.of($args);
+              }
+            """
+          })("\n\n")}
+
         """
       }
 
@@ -1487,6 +1517,17 @@ def generateTestClasses(): Unit = {
               @$test
               public void shouldCheckedFunction${i}ReturnNotNull() {
                   assertThat(CheckedFunction$i(($params) -> null)).isNotNull();
+              }
+
+            """
+          })("\n\n")}
+
+          ${(0 to N).gen(i => {
+            val params = (1 to i).gen(j => s"$j")(", ")
+            xs"""
+              @$test
+              public void shouldTuple${i}ReturnNotNull() {
+                  assertThat(Tuple($params)).isNotNull();
               }
 
             """
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -52,6 +52,9 @@ def generateMainClasses(): Unit = {
       val SupplierType = im.getType("java.util.function.Supplier")
       val IteratorType = im.getType("javaslang.collection.Iterator")
       val EitherType = im.getType("javaslang.control.Either")
+      val FutureType = im.getType("javaslang.concurrent.Future")
+      val CheckedSupplierType = im.getType("javaslang.control.Try.CheckedSupplier")
+      val ExecutorServiceType = im.getType("java.util.concurrent.ExecutorService")
 
       def genAliases(im: ImportManager, packageName: String, className: String): String = {
         xs"""
@@ -145,6 +148,86 @@ def generateMainClasses(): Unit = {
            */
           public static <L, R> $EitherType<L, R> Left(L left) {
               return $EitherType.left(left);
+          }
+
+          /$javadoc
+           * Alias for {@link $FutureType#of($CheckedSupplierType)}
+           *
+           * @param <T>         Type of the computation result.
+           * @param computation A computation.
+           * @return A new {@link $FutureType} instance.
+           * @throws NullPointerException if computation is null.
+           */
+          @GwtIncompatible
+          public static <T> $FutureType<T> Future($CheckedSupplierType<? extends T> computation) {
+              return $FutureType.of(computation);
+          }
+
+          /$javadoc
+           * Alias for {@link $FutureType#of($ExecutorServiceType, $CheckedSupplierType)}
+           *
+           * @param <T>             Type of the computation result.
+           * @param executorService An executor service.
+           * @param computation     A computation.
+           * @return A new {@link $FutureType} instance.
+           * @throws NullPointerException if one of executorService or computation is null.
+           */
+          @GwtIncompatible
+          public static <T> $FutureType<T> Future($ExecutorServiceType executorService, $CheckedSupplierType<? extends T> computation) {
+              return $FutureType.of(executorService, computation);
+          }
+
+          /$javadoc
+           * Alias for {@link $FutureType#successful(Object)}
+           *
+           * @param <T>    The value type of a successful result.
+           * @param result The result.
+           * @return A succeeded {@link $FutureType}.
+           */
+          @GwtIncompatible
+          public static <T> $FutureType<T> Future(T result) {
+              return $FutureType.successful(result);
+          }
+
+          /$javadoc
+           * Alias for {@link $FutureType#successful($ExecutorServiceType, Object)}
+           *
+           * @param <T>             The value type of a successful result.
+           * @param executorService An {@code ExecutorService}.
+           * @param result          The result.
+           * @return A succeeded {@link $FutureType}.
+           * @throws NullPointerException if executorService is null
+           */
+          @GwtIncompatible
+          public static <T> $FutureType<T> Future($ExecutorServiceType executorService, T result) {
+              return $FutureType.successful(executorService, result);
+          }
+
+          /$javadoc
+           * Alias for {@link $FutureType#failed(Throwable)}
+           *
+           * @param <T>       The value type of a successful result.
+           * @param exception The reason why it failed.
+           * @return A failed {@link $FutureType}.
+           * @throws NullPointerException if exception is null
+           */
+          @GwtIncompatible
+          public static <T> $FutureType<T> Future(Throwable exception) {
+              return $FutureType.failed(exception);
+          }
+
+          /$javadoc
+           * Alias for {@link $FutureType#failed($ExecutorServiceType, Throwable)}
+           *
+           * @param <T>             The value type of a successful result.
+           * @param executorService An executor service.
+           * @param exception       The reason why it failed.
+           * @return A failed {@link $FutureType}.
+           * @throws NullPointerException if executorService or exception is null
+           */
+          @GwtIncompatible
+          public static <T> $FutureType<T> Future($ExecutorServiceType executorService, Throwable exception) {
+              return $FutureType.failed(executorService, exception);
           }
 
         """
@@ -1524,10 +1607,34 @@ def generateTestClasses(): Unit = {
       val AssertionsExtensions = im.getType("javaslang.AssertionsExtensions")
       val ListType = im.getType("javaslang.collection.List")
       val OptionType = im.getType("javaslang.control.Option")
+      val FutureType = im.getType("javaslang.concurrent.Future")
+      val ExecutorServiceType = im.getType("java.util.concurrent.Executors")
+      val ExecutorService = s"${ExecutorServiceType}.newSingleThreadExecutor()"
 
       val d = "$";
 
       im.getStatic("javaslang.API.*")
+
+      def genFutureTests(name: String, value: String, success: Boolean): String = {
+        val check = if (success) "isSuccess" else "isFailure"
+        xs"""
+          @$test
+          public void shouldFutureWith${name}ReturnNotNull() {
+              final $FutureType<?> future = Future($value);
+              future.await();
+              assertThat(future).isNotNull();
+              assertThat(future.$check()).isTrue();
+          }
+
+          @$test
+          public void shouldFutureWithinExecutorWith${name}ReturnNotNull() {
+              final $FutureType<?> future = Future($ExecutorService, $value);
+              future.await();
+              assertThat(future).isNotNull();
+              assertThat(future.$check()).isTrue();
+          }
+        """
+      }
 
       def genAliasesTests(im: ImportManager, packageName: String, className: String): String = {
         xs"""
@@ -1567,6 +1674,13 @@ def generateTestClasses(): Unit = {
 
             """
           })("\n\n")}
+
+          ${genFutureTests("Supplier", "() -> 1", success = true)}
+
+          ${genFutureTests("Value", "1", success = true)}
+
+          ${genFutureTests("Error", "new Error()", success = false)}
+
         """
       }
 
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -241,6 +241,38 @@ def generateMainClasses(): Unit = {
               return Lazy.of(supplier);
           }
 
+          /$javadoc
+           * Alias for {@link $OptionType#of(Object)}
+           *
+           * @param <T>   type of the value
+           * @param value A value
+           * @return {@link $OptionType.Some} if value is not {@code null}, {@link $OptionType.None} otherwise
+           */
+          public static <T> $OptionType<T> Option(T value) {
+              return $OptionType.of(value);
+          }
+
+          /$javadoc
+           * Alias for {@link $OptionType#some(Object)}
+           *
+           * @param <T>   type of the value
+           * @param value A value
+           * @return {@link $OptionType.Some}
+           */
+          public static <T> $OptionType<T> Some(T value) {
+              return $OptionType.some(value);
+          }
+
+          /$javadoc
+           * Alias for {@link $OptionType#none()}
+           *
+           * @param <T> component type
+           * @return the singleton instance of {@link $OptionType.None}
+           */
+          public static <T> $OptionType<T> None() {
+              return $OptionType.none();
+          }
+
         """
       }
 
@@ -1706,6 +1738,12 @@ def generateTestClasses(): Unit = {
           ${genFutureTests("Error", "new Error()", success = false)}
 
           ${genSimpleAliasTest("Lazy", "() -> 1")}
+
+          ${genSimpleAliasTest("Option", "1")}
+
+          ${genSimpleAliasTest("Some", "1")}
+
+          ${genSimpleAliasTest("None", "")}
 
         """
       }
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -52,6 +52,49 @@ def generateMainClasses(): Unit = {
       val SupplierType = im.getType("java.util.function.Supplier")
       val IteratorType = im.getType("javaslang.collection.Iterator")
 
+      def genAliases(im: ImportManager, packageName: String, className: String): String = {
+        xs"""
+          //
+          // Aliases for static factories
+          //
+
+          ${(0 to N).gen(i => {
+            val generics = (1 to i).gen(j => s"T$j")(", ")
+            val fullGenerics = s"<${(i > 0).gen(s"$generics, ")}R>"
+            xs"""
+              /$javadoc
+               * Alias for {@link Function$i#of(Function$i)}
+               *
+               ${(0 to i).gen(j => if (j == 0) "* @param <R>             return type" else s"* @param <T$j>            type of the ${j.ordinal} argument")("\n")}
+               * @param methodReference A method reference
+               * @return A {@link Function$i}
+               */
+              public static $fullGenerics Function$i$fullGenerics Function$i(Function$i$fullGenerics methodReference) {
+                  return Function$i.of(methodReference);
+              }
+            """
+          })("\n\n")}
+
+          ${(0 to N).gen(i => {
+            val generics = (1 to i).gen(j => s"T$j")(", ")
+            val fullGenerics = s"<${(i > 0).gen(s"$generics, ")}R>"
+            xs"""
+              /$javadoc
+               * Alias for {@link CheckedFunction$i#of(CheckedFunction$i)}
+               *
+               ${(0 to i).gen(j => if (j == 0) "* @param <R>             return type" else s"* @param <T$j>            type of the ${j.ordinal} argument")("\n")}
+               * @param methodReference A method reference
+               * @return A {@link CheckedFunction$i}
+               */
+              public static $fullGenerics CheckedFunction$i$fullGenerics CheckedFunction$i(CheckedFunction$i$fullGenerics methodReference) {
+                  return CheckedFunction$i.of(methodReference);
+              }
+            """
+          })("\n\n")}
+
+        """
+      }
+
       im.getStatic("javaslang.API.Match.*")
 
       def genJavaTypeTweaks(im: ImportManager, packageName: String, className: String): String = {
@@ -571,6 +614,8 @@ def generateMainClasses(): Unit = {
 
             private API() {
             }
+
+            ${genAliases(im, packageName, className)}
 
             ${genJavaTypeTweaks(im, packageName, className)}
 
@@ -1429,6 +1474,26 @@ def generateTestClasses(): Unit = {
 
       im.getStatic("javaslang.API.*")
 
+      def genAliasesTests(im: ImportManager, packageName: String, className: String): String = {
+        xs"""
+          ${(0 to N).gen(i => {
+            val params = (1 to i).gen(j => s"v$j")(", ")
+            xs"""
+              @$test
+              public void shouldFunction${i}ReturnNotNull() {
+                  assertThat(Function$i(($params) -> null)).isNotNull();
+              }
+
+              @$test
+              public void shouldCheckedFunction${i}ReturnNotNull() {
+                  assertThat(CheckedFunction$i(($params) -> null)).isNotNull();
+              }
+
+            """
+          })("\n\n")}
+        """
+      }
+
       xs"""
         public class $className {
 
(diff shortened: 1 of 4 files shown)


javaslang/generator/Generator.scala
```diff
@@ -51,6 +51,302 @@ def generateMainClasses(): Unit = {
...
+           * Alias for {@code $traversableType.of()}
+           *
+           * @see $traversableType#of(T...)
+           */
+          @SuppressWarnings("varargs")
+          @SafeVarargs
+          public static <T> $traversableType<T> $name(T... elements) {
+            return $traversableType.of(elements);
+          }
+          """
```
danieldietrich: minor/indentation: delete 2 spaces

javaslang/generator/Generator.scala
```diff
@@ -51,6 +51,302 @@ def generateMainClasses(): Unit = {
...
+              /$javadoc
+               * Alias for {@code Function$i.of()}
+               *
+               * @see Function$i#of(Function$i)
+               */
+              public static $fullGenerics Function$i$fullGenerics f$i(Function$i$fullGenerics methodReference) {
+                  return Function$i.of(methodReference);
+              }
+            """
+        })("\n\n")}
```
danieldietrich: minor/indentation: line 100-112: add two spaces, but line 111 needs to stay as-is :-)

javaslang/generator/Generator.scala
```diff
@@ -51,6 +51,302 @@ def generateMainClasses(): Unit = {
...
+              /$javadoc
+               * Alias for {@code CheckedFunction$i.of()}
+               *
+               * @see CheckedFunction$i#of(CheckedFunction$i)
+               */
+              public static $fullGenerics CheckedFunction$i$fullGenerics cf$i(CheckedFunction$i$fullGenerics methodReference) {
+                  return CheckedFunction$i.of(methodReference);
+              }
+            """
+        })("\n\n")}
```
danieldietrich: minor/indentation: same as above (also applies to the following blocks)

javaslang/src-gen/main/java/javaslang/API.java
```diff
@@ -84,6 +90,573 @@ private API() {
     }
 
     //
+    // Aliases for static factories
+    //
+
+    /**
+     * Alias for {@code Function0.of()}
```
danieldietrich: We need to generate valid Javadoc (or leave it completely away but that is no viable option for us) - otherwise `mvn javadoc:javadoc` will fail. We need that maven target for a release, so please test it (ideally on every PR).

I suggest Javadoc of the following form. Using `@link` instead of `@code` in the first line might come handy when adding more aliases.

```java
/**
 * Alias for {@link Function0#of(Function0)}.
 *
 * @param methodReference A method reference
 * @return A {@code Function0}
 */
public static <R> Function0<R> f0(Function0<R> methodReference) {
    return Function0.of(methodReference);
}
```

javaslang/generator/Generator.scala
```diff
@@ -51,6 +51,302 @@ def generateMainClasses(): Unit = {
...
+          public static <T> $OptionType<T> none() {
+            return $OptionType.none();
+          }
+
+          /$javadoc
+           * Alias for {@code Try.of()}
+           *
+           * @see $TryType#of($CheckedSupplierType)
+           */
+          public static <T> $TryType<T> Try($CheckedSupplierType<? extends T> supplier) {
```
danieldietrich: Good! `Try` is better than `try_` or `_try`

javaslang/generator/Generator.scala
```diff
@@ -51,6 +51,302 @@ def generateMainClasses(): Unit = {
...
+           * Alias for {@code $FutureType.successful()}
+           *
+           * @see $FutureType#successful($ExecutorServiceType, T)
+           */
+          public static <T> $FutureType<T> future($ExecutorServiceType executorService, T result) {
+            return $FutureType.successful(executorService, result);
+          }
+
+          /$javadoc
+           * Alias for {@code $FutureType.of()}
```
danieldietrich: needs to be `failed` instead of `of`

javaslang/generator/Generator.scala
```diff
@@ -51,6 +51,302 @@ def generateMainClasses(): Unit = {
...
+           *
+           * @see $FutureType#failed(Throwable)
+           */
+          public static <T> $FutureType<T> future(Throwable exception) {
+            return $FutureType.failed(exception);
+          }
+
+          /$javadoc
+           * Alias for {@code $FutureType.of()}
```
danieldietrich: needs to be `failed` instead of `of`

javaslang/generator/Generator.scala
```diff
@@ -51,6 +51,302 @@ def generateMainClasses(): Unit = {
...
+          public static <T> $FutureType<T> future($ExecutorServiceType executorService, $CheckedSupplierType<? extends T> computation) {
+            return $FutureType.of(executorService, computation);
+          }
+
+          /$javadoc
+           * Alias for {@code $FutureType.successful()}
+           *
+           * @see $FutureType#successful(T)
+           */
+          public static <T> $FutureType<T> future(T result) {
```
danieldietrich: I think `future(Throwable)` and `future(T)` are ambiguous. Please try if the compiler knows what to do when using `future(new Error("oh"))`

javaslang/generator/Generator.scala
```diff
@@ -51,6 +51,302 @@ def generateMainClasses(): Unit = {
...
+
+          /$javadoc
+           * Alias for {@code Option.none()}
+           *
+           * @see $OptionType#none()
+           */
+          public static <T> $OptionType<T> none() {
+            return $OptionType.none();
+          }
+
```
danieldietrich: `Option.nothing()` is useful, too (I know, it is still in progress)

danieldietrich:
Thx @valery1707, I like it. I think the aliases will come handy to make code more concise!

valery1707:
@danieldietrich , I update JavaDoc - please check.
Later I create tests and aliases for `TreeSet` and `Map`-s

danieldietrich:
Great work so far, thank you. I can't just pull it in because I'm still not sure about the function names.

In near future there will be also CheckedConsumer, CheckedBiConsumer, CheckedRunnable, CheckedSupplier, CheckedPredicate, CheckedBiPredicate. Or maybe CheckedConsumer1..8, CheckedPredicate1..8 - not sure yet. But we need abbreviations for these, too.

I think these names are better:

* function0..8 - Function0..8
* checkedFunction0..8 - CheckedFunction0..8

and later:

* checkedSupplier - CheckedSupplier
* checkedRunnable - CheckedRunnable
* checkedPredicate1..8 - CheckedPredicate1..8
* checkedConsumer1..8 - CheckedConsumer1..8

Example (future Java version):

```java
val foo = checkedFunction2(this::bar);
```

danieldietrich:
@valery1707 please rename `f0..8` to `function0..8` and `cf0..8` to `checkedFunction0..8`.

valery1707:
@danieldietrich ok, soon. Anyway PR is not completed yet, I still need create tests and aliases for all `Map`-s and `TreeSet`

danieldietrich:
Great, thanks! These will be really useful to users. Take your time.
\- Daniel

javaslang/src-gen/main/java/javaslang/API.java
```diff
@@ -84,6 +90,865 @@ private API() {
...
+     */
+    public static <T1, T2, T3, T4, T5, T6, R> Function6<T1, T2, T3, T4, T5, T6, R> f6(Function6<T1, T2, T3, T4, T5, T6, R> methodReference) {
+        return Function6.of(methodReference);
+    }
+
+    /**
+     * Alias for {@link Function7#of(Function7)}
+     *
+     * @param <R>             return type
+     * @param <T1>            type of the 1st argument
```
paplorinc: @danieldietrich, @valery1707, do you think these docs have any value?
It's like putting a javadoc to a `getValue`: "this function returns the value"

javaslang/generator/Generator.scala
```diff
@@ -51,6 +51,377 @@ def generateMainClasses(): Unit = {
       val PredicateType = im.getType("java.util.function.Predicate")
       val SupplierType = im.getType("java.util.function.Supplier")
       val IteratorType = im.getType("javaslang.collection.Iterator")
+      val EitherType = im.getType("javaslang.control.Either")
+      val TryType = im.getType("javaslang.control.Try")
+      val ValidationType = im.getType("javaslang.control.Validation")
+      val FutureType = im.getType("javaslang.concurrent.Future")
+      val CheckedSupplierType = im.getType("javaslang.control.Try.CheckedSupplier")
+      val ExecutorServiceType = im.getType("java.util.concurrent.ExecutorService")
+      val ArrayType = im.getType("javaslang.collection.Array")
```
paplorinc: FYI: https://github.com/javaslang/javaslang/issues/1528

javaslang/src-gen/main/java/javaslang/API.java
```diff
@@ -84,6 +90,865 @@ private API() {
...
+    }
+
+    /**
+     * Alias for {@link LinkedHashSet#of(Object)}
+     *
+     * @param <T>     Component type of element.
+     * @param element An element.
+     * @return A new LinkedHashSet instance containing the given element
+     */
+    public static <T> LinkedHashSet<T> linkedHashSet(T element) {
```
paplorinc: @danieldietrich is the word `hash` part important here?
The user should care about the properties of the set (i.e. sorted, keeps insertion order etc), not whether we're using hashing or not

emmanueltouzery:
FWIW I do regret a little the shorter `f1` compared to `function1`. Maybe `fun` or `func` could be options?

paplorinc:
Good job @valery1707!

I also vote for no abbreviations, they always seem like a good idea at the time and always turn out to be a bad investment. I usually search for shorter synonyms, when the names get too long.

What I currently miss from the syntax is `ranges`, for which I guess the ones from `Iterator` should be used (every other class just uses those anyway, and we definitely need laziness there by default).

opinions, @danieldietrich?

paplorinc:
@valery1707, would it be possible for you to fix mistakes in the same commit there were made?
Pull requests are meant to fix all mistakes that were found, without confusing the future generation with confusing commit histories. Also, new reviewers shouldn't go over redundant commits that modify the same line multiple times.

You can edit previous commits by doing an interactive rebase (in Idea there's a GUI for that - `VCS` / `Git`).

In the end, the commits should be divided logically, not historically (i.e. added `checked function aliases`, `added collection static factories` etc :)

valery1707:
@paplorinc OK

valery1707:
@paplorinc currently I already mixed commits with adding different parts.
After this PR will good for everyone, I can create new PR with splitted commits for different parts of API, or with merged all commits in one, for future simplicity

valery1707:
@danieldietrich @paplorinc @emmanueltouzery What alias use for `Function1..8`/`CheckedFunction1..8`?

paplorinc:
> After this PR will good for everyone, I can create new PR with splitted commits for different part

In git you can modify the history, and PRs are meant for that (i.e. it's ok to force push BEFORE it's merged).
You can force push (http://stackoverflow.com/questions/5509543/how-do-i-properly-force-a-git-push), in order for the commit history to make sense: otherwise it just represents your mistakes, not a logical order of steps.

valery1707:
@paplorinc I known about history rewrite and force push. I don't use force push with GitHub PR. If PR will correctly work with force push - I will rewrite history within this PR

javaslang/src-gen/main/java/javaslang/API.java
danieldietrich: The important part is `Alias for {@link Function7#of(Function7)}`, which links to the method we provide an alias for. We should not duplicate the Javadoc. Instead the user should follow the link.

Unfortunately doclint will error when the parameters etc. are not documented in the javadoc. How could we provide better documentation here?

javaslang/generator/Generator.scala
danieldietrich: yes, but as long as Array is here we need to include it in our changes. it is not clear yet, if we will remove it.

javaslang/src-gen/main/java/javaslang/API.java
danieldietrich: Our names are currently inconsistent, see #1488. In Scala we use either interface names or class names for object creation but never names of non-existing types.

Examples:

```scala
scala> Seq(1, 2, 3)
res0: Seq[Int] = List(1, 2, 3)

scala> Set(1, 2, 3)
res1: scala.collection.immutable.Set[Int] = Set(1, 2, 3)

scala> scala.collection.immutable.SortedSet.empty[String] + ("a", "c", "b")
res3: scala.collection.immutable.SortedSet[String] = TreeSet(a, b, c)
```

These are all interfaces. The is no such thing like `LinkedSet`. The name would be inconsistent.

danieldietrich:
@valery1707

> What alias use for `Function1..8`/`CheckedFunction1..8`?

We go with the long names `function1..8()`/`checkedFunction1..8()`. Otherwise we will regret it when introducing more function types (Consumer, Runnable, Supplier, Predicate etc.)

javaslang/src-gen/main/java/javaslang/API.java
paplorinc: > Unfortunately doclint will error when the parameters etc. are not documented in the javadoc. How could we improve the documentation here?

I've checked the Javadoc maven plugin and couldn't find an option to disable this warning.
Is the warning also present of there is no `@param` at all (there are lots of methods in Javaslang that don't have param docs :/)?

I think it would enhance readability if **very** obvious things weren't documented, like a description AND a `@return` that say the same things, or when self-describing parameters are spelled out as documentation also.

javaslang/generator/Generator.scala
paplorinc: I agree, just pointing it out :)

javaslang/src-gen/main/java/javaslang/API.java
danieldietrich: I agree that it would increase readability / would turn focus on non-trivial things. But if we want formally ensure that no documentation is missing, we need to enable Java's doclint option.

Btw - I'm a little bit proud that we use doclint. Not everyone is able to do it because tons of Javadoc are broken and they are not able to effort the time to fix it.

javaslang/generator/Generator.scala
```diff
@@ -51,6 +51,377 @@ def generateMainClasses(): Unit = {
...
+          }
+
+          /$javadoc
+           * Alias for {@link Option#some(Object)}
+           *
+           * @param <T>   type of the value
+           * @param value A value
+           * @return {@code Some(value)}
+           */
+          public static <T> $OptionType<T> some(T value) {
```
paplorinc: do we need these?
What's the difference between importing statically `Option` or `API`?
I mean these are already available to us by importing `Option` statically.

@danieldietrich ?

danieldietrich: This is a valid point. Several types already have the static factory methods with similar names:

* [x] controls (Either, Option, Try, Validation)
* [x] Future has successful() and failed() (Promise has the same method names...)
* [ ] Collections
* [ ] Tuples
* [ ] Functions

We still have additional problems here - some API is ambiguous, e.g.

```java
// returns a success
public static <T> $FutureType<T> future(T result) { ... }

// returns a failure
public static <T> $FutureType<T> future(Throwable exception) { ... }
```

It was a design decision to use consequently use `of` throughout our whole codebase, e.g.

* `Tuple.of(1, true, "nice")`
* `List.of(1, 2, 3)`
* `Try.of(() -> computation())`

* Java also uses the `of` API. It looks familiar to the user.
* Lower case 'constructors' look alien to Java developers.
* Scala has companion objects which give use upper case constructors without the use of `new` - that is what we really want. But the upper-case names are already taken by the Pattern Match API in order to deconstruct objects.
* Minor: Static imports are always a few keystrokes more for the developer, even with an IDE.

**But: The `of` notation is not concise enough.**

Like @paplorinc said, we do not need to duplicate the existing API. What we really want are shortcuts for the `of` factory methods only.

I would like to have first char **upper-case** static factory method names within `javaslang.API`, e.g.

* `Tuple(1, 2)`, `Tuple(1, 2, 3, 4)`
* `Function1(obj::methodRef)`, `CheckedFunction1(obj::methodRef)`
* `Try(() -> computation())`, `Option(value)`, `Future(() -> computation())`
* `List(1, 2, 3)`
* `HashMap(1, "a", 2, "b", 3, "c")`

This would go adhere with `API.Match()` and `API.Case()` etc.

To this day static factory methods for (most) collection _interfaces_ were out of scope, e.g.

* `Seq(1, 2, 3)`
* `Set(1, 2, 3)`
* `Map(1, "a", 2, "b", 3, "c")`

But maybe it would be nice to have these too.

---

**Q:** So, what to do with the names of generated patterns? They will clash with the new API.

**A:** We will break backward compatibility and change the annotation processor of `javaslang-match`. We will generate a preceding `$` for patterns. This goes adhere with the use of `$` for standard patterns like

* `$()` - any
* `$(value)` - equality
* `$(predicate)` - condition

Example:

```java
T result = Match(value).of(
    Case($Tuple($(1), $(true), $("nice")), (v1, v2, v3) -> ...),
    Case($List($(1), $()), (head, tail) -> ...)
);
```

We determine these _specialities_:

* `$Tuple($(Tuple(1, 2, 3))` will contain a `$Tuple` pattern and a `Tuple` constructor. Hard to read?
* `$List($(), $())` deconstructs _head_ element and _tail_ list
* `List(v1, v2)` constructs a List containing two values `v1`, `v2`

---

# The bottom line is that this will be a breaking change that will be targeted for 3.0.0

However, the first char upper case constructors can be already introduced but they will raise problems when used in conjunction with the Match API (, e.g. `import static javaslang.API.*;`).

danieldietrich:
@paplorinc

> do we need these?

There is one thing that still bothers me - we will have mixed upper case and lower case factory methods, e.g.

* upper case `API.Option()` but lower case `Option.some()`, `Option.none()`

This idea is still unfinished...

paplorinc:
> There is one thing that still bothers me - we will have mixed upper case and lower case factory methods, e.g.

Could you please explain the need for upper-camel? If it's simply to model `Scala`'s companion constructor, we could simply translate that to statically-importable lowercase static factories, e.g. `option`, `some`, `checkedFunction`.
About `API`, well, it could have a prefix of e.g. `$` or `_`, but still lowercase, e.g. `$option`.

danieldietrich:
Yes, that's a valid solution I also currently think about.

Deconstructors (read: unapply patterns) and constructors are _dual_. This should be reflected by the name. I think we have two options: both upper- or both lower case. Distinguishing them is mandatory. A prefix should work fine ('$' would integrate well for the patterns).

These are our options:

* All methods upper or all lower case, e.g. `Some()`/`$Some()` vs `some()`/`$some()`
* Define all methods within `javaslang.API` or all within related types or a hybrid solution

Each type has several `of` and `ofAll` methods in general. When adding new factory methods we concentrate on `of(T)` and `of(T...)`.

Looking at `Option` we will take one of these approaches:

* `API.Option(T)`, `API.Option(T...)` and `Patterns.$Option()`
* `API.option(T)`, `API.option(T...)` and `Patterns.$option()`
* `Option.Option(T)`, `Option.Option(T...)` and `Patterns.$Option()`
* `Option.option(T)`, `Option.option(T...)` and `Patterns.$option()`

I like to just `import static javaslang.API.*` and then start to write Javaslang using `Match`, `For` and almost all of the Values, Tuples and (Checked)Functions.

**We could also _define_ the following in _idiomatic Javaslang_:**

* Static factory methods related to types are lower case and not intended to be imported statically, e.g. we always use `Option.some(...)` instead of `some(...)`. This would go adhere with `of` and `ofAll`.
* The `javaslang.API` is intended to be used as static import _only_. E.g. we write `For` and `Match` instead of `API.For` and `API.Match`.
* API contains the keywords of the Java_slang_. These keywords are all upper-case in order to circumvent name-clashes with default Java. E.g. we have `For` and `Try(() -> computation)` instead of `for` and `try(() -> computation())`.
* We would then add also `API.Some()` and `API.None()` in addition to `Option.some()` and `Option.none()`.

In other words `Some(...)` and `Option.some()` are different in the way they are used, the former as static import, the latter full qualified. The case indicates the usage.

valery1707:
@danieldietrich PR ready for check

javaslang/src-gen/main/java/javaslang/API.java
```diff
@@ -84,6 +84,280 @@ private API() {
     }
 
     //
+    // Aliases for static factories
```
paplorinc: Interesting :). You're using `//` for delimiting groups in the same class, right?

javaslang/src-gen/main/java/javaslang/API.java
```diff
@@ -84,6 +84,280 @@ private API() {
...
+     */
+    public static <R> Function0<R> function0(Function0<R> methodReference) {
+        return Function0.of(methodReference);
+    }
+
+    /**
+     * Alias for {@link Function1#of(Function1)}
+     *
+     * @param <R>             return type
+     * @param <T1>            type of the 1st argument
```
paplorinc: technically `R` is also the type of the first argument

javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -25,6 +25,101 @@ public void shouldNotBeInstantiable() {
...
     }
 
+    //
+    // Alias should return not null.
+    // More specific test for each aliased class implemented in separate test class
+    //
+
+    @Test
+    public void shouldFunction0ReturnNotNull() {
+        assertThat(function0(() -> null)).isNotNull();
```
paplorinc: you could be more specific and check instance of `Function0` maybe?

javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -165,6 +165,16 @@ public void shouldTuple8ReturnNotNull() {
         assertThat(tuple(1, 2, 3, 4, 5, 6, 7, 8)).isNotNull();
     }
 
+    @Test
+    public void shouldRightReturnNotNull() {
+        assertThat(right(null)).isNotNull();
```
paplorinc: isn't this the same as if I would have imported the method from `Either`?
@danieldietrich, what's the take on that, is this meant as a shortcut for only importing the `API`?

javaslang/generator/Generator.scala
```diff
@@ -147,6 +150,80 @@ def generateMainClasses(): Unit = {
...
+          }
+
+          /$javadoc
+           * Alias for {@link $FutureType#of($ExecutorServiceType, $CheckedSupplierType)}
+           *
+           * @param <T>             Type of the computation result.
+           * @param executorService An executor service.
+           * @param computation     A computation.
+           * @return A new Future instance.
+           * @throws NullPointerException if one of executorService of computation is null.
```
paplorinc: `executorService of computation` should probably be `executorService or computation`

javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -175,6 +177,54 @@ public void shouldLeftReturnNotNull() {
...
+    public void shouldFutureWithSupplierReturnNotNull() {
+        Future<Object> future = future(() -> 1);
+        future.await();
+        assertThat(future).isNotNull();
+        assertThat(future.isSuccess()).isTrue();
+    }
+
+    @Test
+    public void shouldFutureWithinExecutorWithSupplierReturnNotNull() {
+        Future<Object> future = future(Executors.newSingleThreadExecutor(), () -> 1);
```
paplorinc: these tests are **very** useful as documentation! :)

javaslang/generator/Generator.scala
```diff
@@ -1558,6 +1569,15 @@ def generateTestClasses(): Unit = {
         """
       }
 
+      def genSimpleAliasTest(name: String, value: String): String = {
+        xs"""
+          @$test
+          public void should${name}ReturnNotNull() {
+              assertThat(${name.toLowerCase}($value)).isNotNull();
```
paplorinc: this seemed fishy at first (the `toLowerCase`, instead of specifying it separately, like in `Either`), but it represents very well how a human expects it to work, good job) :)

javaslang/generator/Generator.scala
paplorinc: @valery1707, @danieldietrich, so what's the final decision, do we need aliases for already good enough naming (e.g. `Either` or `Optional`)?

javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -230,6 +230,26 @@ public void shouldLazyReturnNotNull() {
         assertThat(lazy(() -> 1)).isNotNull();
     }
 
+    @Test
+    public void shouldOptionReturnNotNull() {
+        assertThat(option(1)).isNotNull();
```
paplorinc: could you provide a test for `option(null)` also, I'm not sure how that should behave differently from `some` (I guess it should represent a `none` intuitively)

javaslang/src-gen/main/java/javaslang/API.java
```diff
@@ -680,6 +681,40 @@ public static Tuple0 tuple() {
...
 
+    /**
+     * Alias for {@link Try#of(CheckedSupplier)}
+     *
+     * @param <T>      Component type
+     * @param supplier A checked supplier
+     * @return {@code Success(supplier.get())} if no exception occurs, otherwise {@code Failure(throwable)} if an
+     * exception occurs calling {@code supplier.get()}.
+     */
+    public static <T> Try<T> Try(CheckedSupplier<? extends T> supplier) {
```
paplorinc: hmm, i don't find it intuitive that we've uppercased only this one (for obvious reasons, thouh).
@danieldietrich, how should we proceed in this case?

javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -250,6 +251,27 @@ public void shouldNothingReturnNotNull() {
...
+    @Test
+    public void shouldSuccessReturnNotNull() {
+        Try<Object> _try = success(1);
+        assertThat(_try).isNotNull();
+        assertThat(_try.isSuccess()).isTrue();
+    }
+
+    @Test
+    public void shouldFailureReturnNotNull() {
+        Try<Object> _try = failure(new Error("error"));
```
paplorinc: I think the "error" is redundant

javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -272,6 +272,16 @@ public void shouldFailureReturnNotNull() {
...
     }
 
+    @Test
+    public void shouldValidReturnNotNull() {
+        assertThat(valid(1)).isNotNull();
+    }
+
+    @Test
+    public void shouldInvalidReturnNotNull() {
+        assertThat(invalid(new Error("error"))).isNotNull();
```
paplorinc: same, `"error"` seems redundant here

javaslang/src-gen/main/java/javaslang/API.java
```diff
@@ -742,6 +743,37 @@ public static Tuple0 tuple() {
...
+        return CharSeq.of(characters);
+    }
+
+    /**
+     * Alias for {@link CharSeq#of(CharSequence)}
+     *
+     * @param sequence {@code CharSequence} instance.
+     * @return A new {@code javaslang.String}
+     */
+    public static CharSeq charSeq(CharSequence sequence) {
```
paplorinc: I think we could/should have a general `Object` static factory here also/instead.
It could make sense to construct a `CharSeq` from a number also, e.g. `charSeq(3.14)`
@danieldietrich?

javaslang/generator/Generator.scala
```diff
@@ -370,6 +431,20 @@ def generateMainClasses(): Unit = {
               return $CharSeqType.of(sequence);
           }
 
+          ${genTraversableAliases(ArrayType, "array")}
```
paplorinc: :+1:

javaslang/generator/Generator.scala
```diff
@@ -370,6 +431,20 @@ def generateMainClasses(): Unit = {
...
+
+          ${genTraversableAliases(ListType, "list")}
+
+          ${genTraversableAliases(StreamType, "stream")}
+
+          ${genTraversableAliases(QueueType, "queue")}
+
+          ${genTraversableAliases(LinkedHashSetType, "linkedHashSet")}
+
+          ${genTraversableAliases(HashSetType, "hashSet")}
```
paplorinc: `set`?

javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -297,6 +297,146 @@ public void shouldCharSeqReturnNotNull() {
...
+    }
+
+    @Test
+    public void shouldArrayWithIterableReturnNotNull() {
+        assertThat(array(charSeq('1', '2', '3'))).isNotNull();
+    }
+
+    @Test
+    public void shouldArrayWithStreamReturnNotNull() {
+        assertThat(array(charSeq('1', '2', '3').toJavaList().stream())).isNotNull();
```
paplorinc: why are you testing Java's stream?

javaslang/src-gen/main/java/javaslang/API.java
```diff
@@ -1213,6 +1214,95 @@ public static CharSeq charSeq(CharSequence sequence) {
...
+    /**
+     * Alias for {@link HashMap#of(Object, Object)}
+     *
+     * @param <K>   The key type.
+     * @param <V>   The value type.
+     * @param key   A singleton map key.
+     * @param value A singleton map value.
+     * @return A new HashMap instance containing the given entry
+     */
+    public static <K, V> HashMap<K, V> hashMap(K key, V value) {
```
paplorinc: `map`?
@danieldietrich

paplorinc:
@valery1707, wow, **really** nice job, congrats!
Neat, organized commits, nice generator, tests ... can't wait to start using it! :)

Once all is accepted, it would be nice to apply these to Javaslang itself (including tests), to serve as documentation! :)

javaslang/generator/Generator.scala
danieldietrich: If name is denoted in camel case, we should only convert the first character to lower case.

javaslang/src-gen/test/java/javaslang/APITest.java
danieldietrich: Likewise to Scala

* option(null) should be None
* some(null) should be Some(null)

javaslang/src-gen/main/java/javaslang/API.java
danieldietrich: > I think we could/should have a general Object static factory here also/instead.

No here. These are only aliases. The factory methods / the functionality reside in the Value implementations.

For now I see no need to add additional factory methods. The user can decide how to convert an object to a String. The purpose of CharSeq is to enrich a String by being a Wrapper with manifold methods.

javaslang/src-gen/main/java/javaslang/API.java
danieldietrich: We (will) have conversion methods `toHashMap()` instead of `toMap()`, etc.

But the aliases in fact should be the simple names, e.g.

* `Map` instead of `hashMap`
* `SortedMap` instead of `TreeMap`
* ...

danieldietrich:
Thank you @valery1707! This looks great! We still will have some changes regarding naming.

I have an idea regarding the Patterns vs the aliases. Maybe we can **fuse** these concepts.

**Example:** Option

The Option patterns look like this:

```java
    public static <T, _1 extends T> Pattern1<Option.Some<T>, _1> Some(Pattern<_1, ?> p1) {
        return Pattern1.of(Option.Some.class, p1, javaslang.$::Some);
    }

    public static <T> Pattern0<Option.None<T>> None() {
        return Pattern0.of(Option.None.class);
    }
```

where Pattern0 and Pattern1 are defined like this:

```java
public interface Pattern<T, R> extends java.util.function.Function<T, javaslang.control.Option<R>> {}

public static abstract class Pattern0<T> implements Pattern<T, T> { ... }

public static abstract class Pattern1<T, T1> implements Pattern<T, T1> { ... }
```

I'm not sure why I made Pattern0 .. Pattern1 abstract classes instead of interfaces. (There was a reason, I have to investigate...). But if these could be interfaces, Option.None could directly implement Pattern0 and Option.Some could implement Pattern1.

Then the aliases `Some(...)` and `None()` would be both, instances of `Option` and also patterns that can be used with pattern matching.

The borderline is, that Objects have a constructor and also an unapply method. It is still just an idea - I need to play around with it...

I will continue the review tomorrow, need some sleep :)

javaslang/src-gen/main/java/javaslang/API.java
valery1707: Right, but it's not my decision - I am only follow current code style: 
* https://github.com/javaslang/javaslang/blob/master/javaslang/src-gen/main/java/javaslang/API.java#L87
* https://github.com/javaslang/javaslang/blob/master/javaslang/src-gen/main/java/javaslang/API.java#L109
* https://github.com/javaslang/javaslang/blob/master/javaslang/src-gen/main/java/javaslang/API.java#L598

javaslang/src-gen/main/java/javaslang/API.java
valery1707: `R` is not return type of this function, but it return type in the lambda. I am use same text as in JavaDoc of `javaslang.Function1#of`.
All types in this JavaDoc are refers to the argument types
I will be glad if you offer a good option

javaslang/src-gen/test/java/javaslang/APITest.java
valery1707: If this can add more significance to test - yes I can, but how `function1` can return something other than `Function1`?

javaslang/src-gen/test/java/javaslang/APITest.java
valery1707: API, as described before, can be used as single static import `import static javaslang.API.*` - and you receive access to almost functions from Javaslang, without needed to import something else

javaslang/generator/Generator.scala
valery1707: This is not best implementation, I am really want only `uncapitalize`, but currently simple `toLowerCase` work to, but not in all cases.
I will add function `uncapitalize` today

javaslang/src-gen/test/java/javaslang/APITest.java
valery1707: Need I create more specific test for both variant with check of result classes?

javaslang/generator/Generator.scala
valery1707: @danieldietrich what aliases we need for access for `LinkedHashSet`, `HashSet`, `LinkedHashMap`, `HashMap`?
1. `linkedHashSet`, `hashSet`, `linkedHashMap`, `hashMap`
2. `linkedSet`, `set`, `linkedMap`, `map`

javaslang/src-gen/test/java/javaslang/APITest.java
valery1707: Most test for aliases is only for get 100% test coverage

javaslang/generator/Generator.scala
valery1707: Fixed

javaslang/src-gen/test/java/javaslang/APITest.java
valery1707: Fixed

javaslang/src-gen/test/java/javaslang/APITest.java
valery1707: Fixed

javaslang/src-gen/test/java/javaslang/APITest.java
valery1707: And in this case I am testing alias which expect Java's Stream as input, not Java's Stream itself

javaslang/generator/Generator.scala
valery1707: I found exists functions `firstLower`/`firstUpper`

valery1707:
@danieldietrich @paplorinc now I have some questions:

What aliases we need for access for `LinkedHashSet`, `HashSet`, `LinkedHashMap`, `HashMap`?
1. `linkedHashSet`, `hashSet`, `linkedHashMap`, `hashMap`
2. `linkedSet`, `set`, `linkedMap`, `map`

What aliases we need for access for `TreeSet`, `TreeMap`?
1. `treeSet`, `treeMap`
2. `sortedSet`, `sortedMap`

In alias for `TreeSet`/`TreeMap` we need return `TreeSet`/`TreeMap` or `SortedSet`/`SortedMap`?

Should all aliases start with lower case or upper case character?

javaslang/src-gen/test/java/javaslang/APITest.java
danieldietrich: The _semantics_ are checked in OptionTest.

Here we should check if the _aliases_ work right. E.g.

* `assertThat(Option(1)).isEqualTo(Option.of(1));`
* `assertThat(Some(1)).isEqualTo(Option.some(1));`
* `assertThat(None()).isEqualTo(Option.none());`

Bonus points:

* `assertThat(Option(null)).isEqualTo(Option.of(null));

(I used first-upper case methods here because it _might_ be our API, but before changing it let me investigate my idea further of letting value types implement Pattern*)

javaslang/src-gen/main/java/javaslang/API.java
danieldietrich: Stay tuned. I'm analyzing to integrate Pattern into the Value type hierarchy. Then we will get Values that are decomposable out of the box (by having an `unapply` method). Such a change can still be backward compatible I think...

javaslang/src-gen/main/java/javaslang/API.java
```diff
@@ -84,6 +84,280 @@ private API() {
...
+     * @return A {@code Function0}
+     */
+    public static <R> Function0<R> function0(Function0<R> methodReference) {
+        return Function0.of(methodReference);
+    }
+
+    /**
+     * Alias for {@link Function1#of(Function1)}
+     *
+     * @param <R>             return type
```
paplorinc: I mean this is the return type of the parameter, ot of the wrapper method

javaslang/generator/Generator.scala
paplorinc: Right, I didn't know about camel case methods, but you're right :).

javaslang/src-gen/main/java/javaslang/API.java
paplorinc: Sure, I just meant it to reduce the number of additional methods.

javaslang/src-gen/test/java/javaslang/APITest.java
paplorinc: I see, consider using either `.toJavaStream()` or rather `java.util.stream.Stream.of('1', '2', '3')` instead :)

javaslang/generator/Generator.scala
```diff
@@ -1357,6 +1402,26 @@ def generateTestClasses(): Unit = {
...
       im.getStatic("javaslang.API.*")
 
+      def genAliasesTests(im: ImportManager, packageName: String, className: String): String = {
+        xs"""
+          ${(0 to N).gen(i => {
+            val params = (1 to i).gen(j => s"v$j")(", ")
+            xs"""
+              @$test
+              public void shouldFunction${i}ReturnNotNull() {
+                  assertThat(function$i(($params) -> null)).isNotNull();
```
paplorinc: sleeping on it, it might make sense to only test that it returns anything, without repeating the production logic of how it should behave exactly.

javaslang/src-gen/test/java/javaslang/APITest.java
danieldietrich: There exists no such think as function equality (in limited time/mem). Two functions `f1: T -> R` and `f2: T -> R` are equal, if

1. f1(t) is defined if and only if f2(t) is defined for all t &in; T
2. f1(t) is defined => f1(t) == f2(t), for all t &in; T

I think `isNotNull` is the best we can do here.

paplorinc: We could check the return for `instanceof`, but I don't mind either way :)

danieldietrich:
@valery1707 

> What aliases we need for access for LinkedHashSet, HashSet, LinkedHashMap, HashMap?

See [my comment](https://github.com/javaslang/javaslang/pull/1500#discussion_r81289780)

> Should all aliases start with lower case or upper case character?

We target upper case. (This will solve also the problem with 'Try'  vs 'try')

javaslang/src-gen/test/java/javaslang/APITest.java
danieldietrich: We already have Value.toJavaStream(). But maybe we will integrate better with Java by deprecating it and adding Value.stream() and Value.parallelStream() :-)

paplorinc: Good idea, but how would we handle the difference between slang Stream and java stream?
`toStream` vs `stream` is not very desciptive in my opinion

javaslang/generator/Generator.scala
valery1707: Almost all current tests created only for test what alias return something, and only few tests check that alias return exactly that he need to return (Future, Try).

javaslang/src-gen/test/java/javaslang/APITest.java
danieldietrich: Yes. We could also stay with toJavaStream() and add toJavaParallelStream().

javaslang/src-gen/test/java/javaslang/APITest.java
paplorinc: This will be especially important, if they'll start with uppercase :)

valery1707:
@danieldietrich @paplorinc 
Already:
* Removed `hash` from aliases for `HashSet`, `LinkedHashSet`, `HashMap`, `LinkedHashMap`
* In name for `TreeMap` aliases used `sorted` prefix
* Build Java's Stream directly
* Add alias `Seq`, `IndexedSeq`

Soon:
* All aliases started with upper char

javaslang/generator/Generator.scala
valery1707: I think - yes we need aliases and for this classes to: user can add one import and obtain access to most useful functions

valery1707:
Now almost all aliases started with upper case char.
`API#None` clashed with `$#None` and `Patterns#None` - what I need to do?
Currently API have only one lower case alias: `none`.

@danieldietrich and @paplorinc please review

danieldietrich:
@valery1707 Thank you! Before we go on I need to prepare some things.

I don't think that `API#None` clashes with `$#None` and `Patterns#None`. This is only the case then both are imported. I will deprecate `$`and `Patterns`, not with the intend to remove them (which would break existing apps) but with the intend not to use them any more. API.* will be a replacement.

I thought about letting all Values extend some sort of pattern, e.g.

```java
interface HashSet<T> implements Pattern... {
    Tuple... unapply() {
        return ...;
    }
}
```

But this will cause problems within the type hierarchy I think. We do want to unapply a `Set`, a `SortedSet`, etc. But only one of them can implement a Pattern.

There might be another solution: Maybe it is practicable to return joint union in type parameters:

```java
<T extends Set<T> & Pattern0> T Set() {
    return ...
}

<T extends Set<T> & Pattern2> T Set(T head, T... tail) {
    return ...
}
```

Option(t), None() and Some(t) all need to return an Option. This is important for correct type-inference in other expressions. I have to ensure that inference still works with the joint union solution above.

By using only interfaces (e.g. Set and Pattern*), we are able to provide runtime implementations (Proxies). But I don't want to have this overhead. Maybe there is another solution.

There's still something to do for me before we can pull this change in. It is important to get the basis right because if we release something we cannot change it without breaking existing code.

Stay tuned! I think I will first pull in @paplorinc's Vector optimizations. He's waiting for so long now...

valery1707:
> I don't think that `API#None` clashes with `$#None` and `Patterns#None`. This is only the case then both are imported.

Yes, this happens only if both methods are imported. But at current time `MatchTest` use full static import from both `API` and `Patterns`.

I change imports in `MatchTest` to import only required parts of `API`.
Now all aliases in `API` start with upper case char.

@danieldietrich and @paplorinc PR again ready for review.

javaslang/src-gen/main/java/javaslang/API.java
```diff
@@ -84,6 +92,1316 @@ private API() {
...
+    public static <T> Option<T> None() {
+        return Option.none();
+    }
+
+    /**
+     * Alias for {@link Option#nothing()}
+     *
+     * @return {@code Some<Void>}
+     */
+    public static Option<Void> Nothing() {
```
danieldietrich: There was some discussion about Option.nothing() and I think you also mentioned it.
Let's leave it away for now. We can add it later if really needed. (If we add it we can't get rid of it any more without breaking backward compatibility.)

javaslang/src-gen/main/java/javaslang/API.java
```diff
@@ -84,6 +92,1316 @@ private API() {
...
+     *
+     * @param <K> The key type.
+     * @param <V> The value type.
+     * @param map A map entry.
+     * @return A new TreeMap instance containing the given map
+     */
+    public static <K extends Comparable<? super K>, V> TreeMap<K, V> SortedMap(Map<? extends K, ? extends V> map) {
+        return TreeMap.ofAll(map);
+    }
+
```
danieldietrich: This looks really great! When reading this file I can't wait to get my hands on the new API :-))

Please add also all empty collection methods:

* List(), Set(), Map(), ...

danieldietrich:
Great work @valery1707 - the users will definitely love this one :-)

danieldietrich:
@ruslansennov The build currently fails - the GWT module says it has no sources for Future etc. Could you take a look please? I have no idea what to do...
Does @valery1707 need to merge with master again because we pulled in GWT support yesterday?
Thank you :-)

ruslansennov:
> The build currently fails - the GWT module says it has no sources for Future etc. Could you take a look please? I have no idea what to do...

The whole package `javaslang.concurrent` was excluded from GWT sources because of its incompatible. So, GWT-compiler knows nothing about `Future` etc. All new methods that uses GWT-incompatible objects should be marked with `@GwtIncompatible` annotation.

For example, this new `API` method

```java
public static <T> Future<T> Future(CheckedSupplier<? extends T> computation) {
    return Future.of(computation);
}
```

should be modified:

```java
@GwtIncompatible
public static <T> Future<T> Future(CheckedSupplier<? extends T> computation) {
    return Future.of(computation);
}
```

> Does valery1707 need to merge with master again because we pulled in GWT support yesterday?

Definitely yes. @valery1707 sorry :)

javaslang/src-gen/main/java/javaslang/API.java
paplorinc: @valery1707, I still think the documentation should be updated to make it obvious that we're not referring to the wrapper method's parameters and return value. If you think it's ok this way also, I don't mind, I care more about good code than good documentation :)

javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -25,6 +25,101 @@ public void shouldNotBeInstantiable() {
...
     }
 
+    //
+    // Alias should return not null.
+    // More specific test for each aliased class implemented in separate test class
+    //
+
+    @Test
+    public void shouldFunction0ReturnNotNull() {
+        assertThat(Function0(() -> null)).isNotNull();
```
paplorinc: the decision is no `instanceof` check here, right? :)

danieldietrich: For me this is ok. When this PR is pulled I will get my hands dirty and add more tests.

Example:

```java
assertThat(Valid(1)).isEqualTo(Validation.valid(1));

File: javaslang/generator/Generator.scala [hunk 35]
```diff
@@ -1452,11 +1529,35 @@ def generateTestClasses(): Unit = {
       val AssertionsExtensions = im.getType("javaslang.AssertionsExtensions")
       val ListType = im.getType("javaslang.collection.List")
       val OptionType = im.getType("javaslang.control.Option")
+      val FutureType = im.getType("javaslang.concurrent.Future")
+      val ExecutorServiceType = im.getType("java.util.concurrent.Executors")
+      val ExecutorService = ExecutorServiceType + ".newSingleThreadExecutor()"
```
paplorinc: minor: 
```scala
val ExecutorService = s"${ExecutorServiceType}.newSingleThreadExecutor()"
```

javaslang/generator/Generator.scala
```diff
@@ -1452,11 +1529,35 @@ def generateTestClasses(): Unit = {
...
 
       im.getStatic("javaslang.API.*")
 
+      def genFutureTests(name: String, value: String, success: Boolean): String = {
+        val check = if (success) "isSuccess" else "isFailure"
+        xs"""
+          @$test
+          public void shouldFutureWith${name}ReturnNotNull() {
+              $FutureType<Object> future = Future($value);
+              future.await();
```
paplorinc: @danieldietrich, if `await` would return the parent `Future`, we could eliminate the need for this line, i.e. 
```java
$FutureType<Object> future = Future($value).await();
```

javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -175,6 +177,54 @@ public void shouldLeftReturnNotNull() {
         assertThat(Left(null)).isNotNull();
     }
 
+    @Test
+    public void shouldFutureWithSupplierReturnNotNull() {
+        Future<Object> future = Future(() -> 1);
+        future.await();
+        assertThat(future).isNotNull();
+        assertThat(future.isSuccess()).isTrue();
```
paplorinc: :+1: for additional, simple test criteria, without repeating the impl :)

javaslang/generator/Generator.scala
```diff
@@ -1452,11 +1529,35 @@ def generateTestClasses(): Unit = {
...
       val d = "$";
 
       im.getStatic("javaslang.API.*")
 
+      def genFutureTests(name: String, value: String, success: Boolean): String = {
+        val check = if (success) "isSuccess" else "isFailure"
+        xs"""
+          @$test
+          public void shouldFutureWith${name}ReturnNotNull() {
+              $FutureType<Object> future = Future($value);
```
paplorinc: Minor: I think we could use `<?>` also, i.e.
```java
$FutureType<?> future = Future($value);
```

javaslang/generator/Generator.scala
```diff
@@ -1558,6 +1569,19 @@ def generateTestClasses(): Unit = {
...
+      def genExtAliasTest(name: String, func: String, value: String, check: String): String = {
+        xs"""
+          @$test
+          public void should${name}() {
+              assertThat($func($value)).$check;
+          }
+        """
+      }
+
+      def genMediumAliasTest(name: String, func: String, value: String): String = genExtAliasTest(name + "ReturnNotNull", func, value, "isNotNull()")
```
paplorinc: minor: `name + "ReturnNotNull"` could also be written as interpolated Scala string, I think

javaslang/generator/Generator.scala
```diff
@@ -235,6 +235,47 @@ def generateMainClasses(): Unit = {
...
+          public static <T> $OptionType<T> None() {
+              return $OptionType.none();
+          }
+
+          /$javadoc
+           * Alias for {@link $OptionType#nothing()}
+           *
+           * @return {@code Some<Void>}
+           */
+          public static $OptionType<Void> Nothing() {
```
paplorinc: This should be removed, right?

javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -250,6 +251,27 @@ public void shouldNothingReturnNotNull() {
         assertThat(Nothing()).isNotNull();
     }
 
+    @Test
+    public void shouldTryReturnNotNull() {
+        Try<Object> _try = Try(() -> 1);
```
paplorinc: Minor suggestions:
* consider making local variables `final` (@danieldietrich likes it that way)
* you could name this `t` instead of starting with a weird underscore, since the scope is very small

javaslang/generator/Generator.scala
```diff
@@ -58,6 +58,67 @@ def generateMainClasses(): Unit = {
       val TryType = im.getType("javaslang.control.Try")
       val ValidationType = im.getType("javaslang.control.Validation")
       val CharSeqType = im.getType("javaslang.collection.CharSeq")
+      val ArrayType = im.getType("javaslang.collection.Array")
+      val VectorType = im.getType("javaslang.collection.Vector")
+      val ListType = im.getType("javaslang.collection.List")
+      val StreamType = im.getType("javaslang.collection.Stream")
+      val QueueType = im.getType("javaslang.collection.Queue")
```
paplorinc: maybe add the new `PriorityQueue` also? :)
@danieldietrich ?

javaslang/generator/Generator.scala
```diff
@@ -58,6 +58,67 @@ def generateMainClasses(): Unit = {
...
+           */
+          public static <T> $traversableType<T> $name(T element) {
+              return $traversableType.of(element);
+          }
+
+          /$javadoc
+           * Alias for {@link $traversableType#of(Object...)}
+           *
+           * @param <T>      Component type of elements.
+           * @param elements Zero of more elements.
```
paplorinc: `or`

javaslang/generator/Generator.scala
```diff
@@ -58,6 +58,67 @@ def generateMainClasses(): Unit = {
...
+          @SafeVarargs
+          public static <T> $traversableType<T> $name(T... elements) {
+              return $traversableType.of(elements);
+          }
+
+          /$javadoc
+           * Alias for {@link $traversableType#ofAll(Iterable)}
+           *
+           * @param <T>      Component type of elements.
+           * @param elements Zero of more elements.
```
paplorinc: same

javaslang/generator/Generator.scala
```diff
@@ -58,6 +58,67 @@ def generateMainClasses(): Unit = {
...
+           * @return A new $traversableType instance containing the given elements
```
paplorinc: could be a link, i.e.
```
A new {@code $traversableType} instance
```

javaslang/generator/Generator.scala
```diff
@@ -448,6 +450,102 @@ def generateMainClasses(): Unit = {
           ${genTraversableAliases(ListType, "Seq")}
 
           ${genTraversableAliases(VectorType, "IndexedSeq")}
+
+          /$javadoc
+           * Alias for {@link $TreeSetType#of(Comparable)}
+           *
+           * @param <T>     Component type of element.
+           * @param element An element.
+           * @return A new $TreeSetType instance containing the given element
```
paplorinc: `{@code $TreeSetType}`?

javaslang/generator/Generator.scala
```diff
@@ -448,6 +450,102 @@ def generateMainClasses(): Unit = {
...
+              return $TreeSetType.of(element);
+          }
+
+          /$javadoc
+           * Alias for {@link $TreeSetType#of($JavaComparatorType, Object)}
+           *
+           * @param <T>     Component type of element.
+           * @param comparator The comparator used to sort the elements
+           * @param element An element.
+           * @return A new $TreeSetType instance containing the given element
```
paplorinc: same

javaslang/generator/Generator.scala
```diff
@@ -448,6 +450,102 @@ def generateMainClasses(): Unit = {
...
+           */
+          public static <T> $TreeSetType<T> SortedSet($JavaComparatorType<? super T> comparator, T element) {
+              return $TreeSetType.of(comparator, element);
+          }
+
+          /$javadoc
+           * Alias for {@link $TreeSetType#of(Comparable...)}
+           *
+           * @param <T>      Component type of element.
+           * @param elements Zero of more elements.
```
paplorinc: `or` (note, I don't mind if you put these in a separate commit, as I know it can be a lot of work to rebase everything)

javaslang/generator/Generator.scala
```diff
@@ -122,6 +126,35 @@ def generateMainClasses(): Unit = {
...
+      def genMapAliases(mapType: String, name: String): String = {
+        xs"""
+          /$javadoc
+           * Alias for {@link $mapType#of(Object, Object)}
+           *
+           * @param <K>   The key type.
+           * @param <V>   The value type.
+           * @param key   A singleton map key.
+           * @param value A singleton map value.
+           * @return A new $mapType instance containing the given entry
```
paplorinc: `{@code $mapType}`

javaslang/generator/Generator.scala
```diff
@@ -546,6 +579,49 @@ def generateMainClasses(): Unit = {
           public static <T> $TreeSetType<T> SortedSet($JavaComparatorType<? super T> comparator, $JavaStreamType<? extends T> elements) {
               return $TreeSetType.ofAll(comparator, elements);
           }
+
+          ${genMapAliases(LinkedHashMapType, "LinkedMap")}
+
+          ${genMapAliases(HashMapType, "Map")}
+
+          /$javadoc
+           * Alias for {@link $TreeMapType#of(Comparable, Object)}
```
paplorinc: I think this could also be updated to `#of(K, V)`

javaslang/src-gen/main/java/javaslang/API.java
valery1707: Yes I remember discussion. But in that time we decide to include this method.
Will be removed

javaslang/generator/Generator.scala
valery1707: Yes

javaslang/src/test/java/javaslang/MatchTest.java
```diff
@@ -16,7 +16,10 @@
 import java.time.Year;
 import java.util.function.Predicate;
 
-import static javaslang.API.*;
+import static javaslang.API.$;
```
valery1707: Yes

javaslang/generator/Generator.scala
valery1707: `{@code Something}` is not generate link. Links generate `{@link ClassName}`

javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -297,6 +298,186 @@ public void shouldCharSeqReturnNotNull() {
         assertThat((Iterable<Character>) CharSeq("123")).isNotNull();
     }
 
+    @Test
+    public void shouldArrayWithSingleReturnNotNull() {
+        assertThat(Array('1')).isNotNull();
```
valery1707: After create alias for empty collections I create tests for this aliases

javaslang/generator/Generator.scala
paplorinc: I don't mind either way, but since you used them previously, I thought it could be more consistent if you unify it. Whichever you prefer, I'm just expressing observations, the choice is yours :)

javaslang/generator/Generator.scala
valery1707: I will `edit` all commits - and it will be relatively simple

javaslang/generator/Generator.scala
```diff
@@ -122,6 +126,35 @@ def generateMainClasses(): Unit = {
         """
       }
 
+      def genMapAliases(mapType: String, name: String): String = {
+        xs"""
+          /$javadoc
+           * Alias for {@link $mapType#of(Object, Object)}
```
paplorinc: Idea seems to accept it, even navigation is working for the params :/

javaslang/src-gen/test/java/javaslang/APITest.java
paplorinc: k, thanks for clarifying :)

danieldietrich:
@valery1707 One more thing: for Maps (Map, SortedMap, LinkedMap) we need also the following, which is one of the main initializers I currently use because of its conciseness:

```java
Map(
  1, "a",
  2, "b",
  3, "c"
);
```

**Example implementation:**

```java
public static <K, V> HashMap<K, V> Map(Object... pairs) {
    return HashMap.of(pairs);
}
```

valery1707:
@danieldietrich , I don't implement initializer for pairs because of they is not type-safe.
Will be implemented.
P.S.
Sorry for bad english :(

paplorinc:
@danieldietrich, I agree with @valery1707, I don't like the unsafe version either (e.g. `HashMap<Integer> = Map('a');`), but now that we have compact tuples, it could be:
```java
Map(
  Tuple(1, "a"),
  Tuple(2, "b"),
  Tuple(3, "c")
);
```

or maybe we should think about making Tuple construction special (as in Scala) and create a method called `_` or `$` or `T` in the API instead, making it typesafe, number-of-arguments-safe like:
```java
Map(
  T(1, "a"),
  T(2, "b"),
  T(3, "c")
);
```

danieldietrich:
I know that Map pair initializers are unsafe. But they will most probably come to the JDK also! See [JEP 269](http://openjdk.java.net/jeps/269)

javaslang/generator/Generator.scala
```diff
@@ -68,6 +68,10 @@ def generateMainClasses(): Unit = {
       val JavaStreamType = im.getType("java.util.stream.Stream")
       val TreeSetType = im.getType("javaslang.collection.TreeSet")
       val JavaComparatorType = im.getType("java.util.Comparator")
+      val LinkedHashMapType = im.getType("javaslang.collection.LinkedHashMap")
```
danieldietrich: Multimaps: Not now, maybe in another PR. They are created with a builder, not with a single constructor or factory method. We need to think about it. Maybe it makes not sense at all...

javaslang/generator/Generator.scala
danieldietrich: For one str concatenation an interpolated String looks like too much overhead. I would not do it... But the question is valid.

javaslang/generator/Generator.scala
valery1707: `{@code Something}` in JavaDoc will generate text with style `code` - only text without any link.
In IDEA no links too.
Now I set `{@link ClassName}` in all `return` JavaDoc parts

danieldietrich:
@valery1707 @paplorinc Let's keep `Map(Object...)`.
I know, it is unsafe, but it is a static constructor. Code that fails will most probably fail on first run and can be fixed. Let's do it this way.

valery1707:
@danieldietrich maybe create alias `Map(Object...)` with `@Deprecated` annotation?

javaslang/generator/Generator.scala
paplorinc: https://github.com/javaslang/javaslang/issues/1607

danieldietrich:
@valery1707 no `@Deprecated` for now...

valery1707:
Completed:
* Rebase to current master. @ruslansennov all good, don't worry
* All future's aliases excluded from GWT
* All local variables are final
* Minor fixes in Generator for `Either`, `Try` and `Future`
* Remove `Nothing`
* Fix typos in some JavaDoc
* Use `{@link ClassName}` inplace of `{@code ClassName}`
* Add alias for `PriorityQueue`
* Add aliases for empty collections and maps
* Add aliases for maps of tuple vararg
* Add unsafe alias for maps with Object vararg

@danieldietrich and @paplorinc please review.

paplorinc:
@valery1707, awesome job, please update the description if you consider it done now :)!

javaslang/generator/Generator.scala
valery1707: Yes, I remember correctly: IDEA accept JavaDoc link as `{@link HashMap#of(K, V)}`, but javadoc linter in stage `javadoc:javadoc` generate warnings `warning - Tag @link: can't find of(K, V) in javaslang.collection.HashMap` for all links to methods with generic classes in arguments.

danieldietrich:
@valery1707 Yes, awesome job!!

---

After the PR is before the PR :) Therefore I want to share some thoughts with you.

@paplorinc  talked to me because I rejected a change that was unsafe (`toJavaArray(Class<? super T>)`).

And Lorinc is also right when saying that `Map(Object...)` is unsafe. Stuart Marks explained in his JavaOne 2016 Session [Collections Refueled](https://www.youtube.com/watch?v=LgR9ByD1dEw) how and why the Java language architects implemented [JEP 269](http://openjdk.java.net/jeps/269).

The essence regarding `Map.of()`is best described by this image:

jep 269 api overview

Javaslang needs to be safe. We will follow the decisions made in JEP 269 and therefore we will deprecate or `*Map.of(Object...)` methods and strongly discourage using them. Instead we will add methods

* `of(K k1, V v1)`
* `of(K k1, V v1, K k2, V v2)`
* ...

up to 10 key/value pairs. For more than 10 pairs we need to use tuples:

* `of(Tuple2<? extends K, ? extends V>...)`

Accordingly we will also deprecate `ofEntries(Tuple2<? extends K,? extends V>...)`.

(See also #1608)
---

Another pain-point is the long name 'Tuple(...)'. Lorinc asked to shorten the alias, say to `T(...)`. I share Lorinc's opinion that `Tuple` is not concise. But `T` (or any other abbrev.) is too cryptic. In pure Java we write `entry(...)` to obtain a `Map.Entry`. We will also write the full name `Tuple`.

---

Please don't think that some of the changes we made here are unnecessary. From the local viewpoint, `HashMap.of(Object...)` is what we currently have. We did not release the new API, so we are able to improve things afterwards. I think it should be ok when we change API#Map methods together with our Hash/Tree/LinkedHash-Map.of(Object...) methods in one PR.

Again, awesome work!

Thank you

Daniel

Issue #1479, opened 2016-08-09
Add `dropRightWhile` and `dropRightUntil` to `Traversable`


paplorinc:
`drop` and `dropRight` should be symmetric, i.e. if we have:

``` java
@Override
public Vector<T> dropWhile(Predicate<? super T> predicate) {
    Objects.requireNonNull(predicate, "predicate is null");
    return dropUntil(predicate.negate());
}

@Override
public Vector<T> dropUntil(Predicate<? super T> predicate) {
    Objects.requireNonNull(predicate, "predicate is null");
    for (int i = 0; i < length(); i++) {
        if (predicate.test(get(i))) {
            return drop(i);
        }
    }
    return empty();
}
```

we should have the opposite also, for `dropRight`:

``` java
public Vector<T> dropRightWhile(Predicate<? super T> predicate) {
    Objects.requireNonNull(predicate, "predicate is null");
    return dropRightUntil(predicate.negate());
}

public Vector<T> dropRightUntil(Predicate<? super T> predicate) {
    Objects.requireNonNull(predicate, "predicate is null");
    for (int i = length() - 1; i >= 0; i--) {
        if (predicate.test(get(i))) {
            return take(i + 1);
        }
    }
    return empty();
}
```

valery1707:
Can I implement this?

danieldietrich:
Yes, that would be great :-))

Issue #1413, opened 2016-07-03
can we revert to benchmark profile?


ruslansennov:
Benchmark profile was deleted in #1398
Now maven goal `test` runs about 12 min.
I would revert this deletion and create special profile again.

cc @paplorinc

danieldietrich:
The benchmarks should not run on a regular basis in our CI environments because subsequent benchmarks are not comparable (CI has no fixed hardware).

We once decided to run them manually. We need a distinct maven profile!!

paplorinc:
This was highlighted in the commit message you both reviewed: assertions are checked for JMH benchmarks, no performance is measured, just correctness.

paplorinc:
It's only ~3 minutes overhead (will be less after an optimization I created yesterday).
Please don't revert this too...

danieldietrich:
We will revert nothing! It is just the `<module>javaslang-benchmark</module>` in the parent pom and the missing benchmark profile we discuss. We need fast builds, let's not execute the benchmark assertions on every CI run.

paplorinc:
Let's not execute the rest of the tests either, we will have even faster builds...

paplorinc:
I've found several bugs after I introduced the assertions (mentioned in the code review also). 
Benchmarks aren't tests (they don't validate behavior automatically), they do need tests.

danieldietrich:
Yes, I've looked again at the assertions, they make sense.

90% of the assertions are related to 3rd party libraries (collections are constructed as excpected etc). If we have assertions that check the behavior of Javaslang, e.g. the head is the first element etc., this test should be part of the Javaslang unit tests. There is no reason to run the benchmark assertions on a CI build triggered by a Javaslang source code change.

I'm still convinced that we should create a separate profile for the benchmarks.

@paplorinc could you do that please?

paplorinc:
> this test should be part of the Javaslang unit tests

Again, I'm testing with the asserts whether I'm bechmarking what I think I am (e.g. `cons` in `Clojure` `Vector` adds to the front (as in List), or to the end? Am I accidentally comparing a `prepend` to an `append`?). It would be naive to think this is any different from why we're testing production code -- i.e. for documentation purposes and to make everything modifiable.

If we don't run all tests for every build, they are useless.

If @ruslansennov feels that 3 extra minutes are unacceptable, I won't block him from separating it into another `Travis` job that is run in parallel with the unit tests.

ruslansennov:
[...]

PR #1503, opened 2016-08-17
Vector maintenance

paplorinc:
Leaving the old implementation, but prepping the ground for https://github.com/javaslang/javaslang/pull/1449

### DIFF
--- a/javaslang-benchmark/src/test/java/javaslang/JmhRunner.java
+++ b/javaslang-benchmark/src/test/java/javaslang/JmhRunner.java
@@ -4,9 +4,14 @@ import javaslang.collection.*;
 import org.openjdk.jmh.annotations.Mode;
 import org.openjdk.jmh.results.RunResult;
 import org.openjdk.jmh.runner.Runner;
-import org.openjdk.jmh.runner.options.*;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.ChainedOptionsBuilder;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.TimeValue;
+import org.openjdk.jmh.runner.options.VerboseMode;
 
-import java.util.*;
+import java.util.Collection;
+import java.util.Random;
 import java.util.concurrent.TimeUnit;
 
 public class JmhRunner {
@@ -15,7 +20,7 @@ public class JmhRunner {
      * Note: it takes about 3 hours.
      */
     public static void main(String[] args) {
-        JmhRunner.runSlowNoAsserts(Array.of(
+        final Array<Class<?>> CLASSES = Array.of(
                 ArrayBenchmark.class,
                 BitSetBenchmark.class,
                 CharSeqBenchmark.class,
@@ -23,68 +28,83 @@ public class JmhRunner {
                 ListBenchmark.class,
                 PriorityQueueBenchmark.class,
                 VectorBenchmark.class
-        ));
+        );
+        runDebugWithAsserts(CLASSES);
+        runSlowNoAsserts(CLASSES);
     }
 
     /** enables debugging and assertions for benchmarks and production code - the speed results will be totally unreliable */
     public static void runDebugWithAsserts(Array<Class<?>> groups) {
         final Array<String> classNames = groups.map(Class::getCanonicalName);
-        run(classNames, 0, 1, 1, 0, VerboseMode.SILENT, Assertions.Enable);
+        run(classNames, 0, 1, 1, ForkJvm.DISABLE, VerboseMode.SILENT, Assertions.ENABLE, PrintInlining.DISABLE);
 
         MemoryUsage.printAndReset();
     }
 
     @SuppressWarnings("unused")
     public static void runQuickNoAsserts(Array<Class<?>> groups) {
-        runAndReport(groups, 10, 10, 10, 1, VerboseMode.NORMAL, Assertions.Disable);
+        runAndReport(groups, 5, 5, 10, ForkJvm.ENABLE, VerboseMode.NORMAL, Assertions.DISABLE, PrintInlining.DISABLE);
     }
 
     @SuppressWarnings("unused")
     public static void runNormalNoAsserts(Array<Class<?>> groups) {
-        runAndReport(groups, 15, 10, 100, 1, VerboseMode.NORMAL, Assertions.Disable);
+        runAndReport(groups, 10, 10, 200, ForkJvm.ENABLE, VerboseMode.NORMAL, Assertions.DISABLE, PrintInlining.DISABLE);
     }
 
     @SuppressWarnings("unused")
     public static void runSlowNoAsserts(Array<Class<?>> groups) {
-        runAndReport(groups, 15, 15, 300, 1, VerboseMode.EXTRA, Assertions.Disable);
+        runAndReport(groups, 25, 15, 500, ForkJvm.ENABLE, VerboseMode.EXTRA, Assertions.DISABLE, PrintInlining.DISABLE);
     }
 
-    public static void runAndReport(Array<Class<?>> groups, int warmupIterations, int measurementIterations, int millis, int forks, VerboseMode silent, Assertions assertions) {
+    public static void runAndReport(Array<Class<?>> groups, int warmupIterations, int measurementIterations, int millis, ForkJvm forkJvm, VerboseMode silent, Assertions assertions, PrintInlining printInlining) {
         final Array<String> classNames = groups.map(Class::getCanonicalName);
-        final Array<RunResult> results = run(classNames, warmupIterations, measurementIterations, millis, forks, silent, assertions);
+        final Array<RunResult> results = run(classNames, warmupIterations, measurementIterations, millis, forkJvm, silent, assertions, printInlining);
         BenchmarkPerformanceReporter.of(classNames, results).print();
-
-        MemoryUsage.printAndReset();
     }
 
-    private static Array<RunResult> run(Array<String> classNames, int warmupIterations, int measurementIterations, int millis, int forks, VerboseMode verboseMode, Assertions assertions) {
-        final ChainedOptionsBuilder builder = new OptionsBuilder()
-                .shouldDoGC(true)
-                .verbosity(verboseMode)
-                .shouldFailOnError(true)
-                .mode(Mode.Throughput)
-                .timeUnit(TimeUnit.SECONDS)
-                .warmupTime(TimeValue.milliseconds(millis))
-                .warmupIterations(warmupIterations)
-                .measurementTime(TimeValue.milliseconds(millis))
-                .measurementIterations(measurementIterations)
-                .forks(forks)
-                // We are using 2Gb and setting NewGen to 100% to avoid GC during testing.
-                // Any GC during testing will destroy the iteration (i.e. introduce unreliable noise in the measurement), which should get ignored as an outlier
-                .jvmArgsAppend("-XX:+UseG1GC", "-Xss100m", "-Xms2g", "-Xmx2g", "-XX:MaxGCPauseMillis=1000", "-XX:+UnlockExperimentalVMOptions", "-XX:G1NewSizePercent=100", "-XX:G1MaxNewSizePercent=100", assertions.vmArg);
-
-        classNames.forEach(builder::include);
-
+    private static Array<RunResult> run(Array<String> classNames, int warmupIterations, int measurementIterations, int millis, ForkJvm forkJvm, VerboseMode verboseMode, Assertions assertions, PrintInlining printInlining) {
         try {
+            final ChainedOptionsBuilder builder = new OptionsBuilder()
+                    .shouldDoGC(true)
+                    .verbosity(verboseMode)
+                    .shouldFailOnError(true)
+                    .mode(Mode.Throughput)
+                    .timeUnit(TimeUnit.SECONDS)
+                    .warmupTime(TimeValue.milliseconds(millis))
+                    .warmupIterations(warmupIterations)
+                    .measurementTime(TimeValue.milliseconds(millis))
+                    .measurementIterations(measurementIterations)
+                    .forks(forkJvm.forkCount)
+                  /* We are using 4Gb and setting NewGen to 100% to avoid GC during testing.
+                     Any GC during testing will destroy the iteration (i.e. introduce unreliable noise in the measurement), which should get ignored as an outlier */
+                    .jvmArgsAppend("-XX:+UseG1GC", "-Xss100m", "-Xms4g", "-Xmx4g", "-XX:MaxGCPauseMillis=1000", "-XX:+UnlockExperimentalVMOptions", "-XX:G1NewSizePercent=100", "-XX:G1MaxNewSizePercent=100", assertions.vmArg);
+            classNames.forEach(builder::include);
+
+            if (printInlining == PrintInlining.ENABLE) {
+                builder.jvmArgsAppend("-XX:+UnlockDiagnosticVMOptions", "-XX:+PrintInlining"); /* might help in deciding when the JVM is properly warmed up - or where to optimize the code */
+            }
+
             return Array.ofAll(new Runner(builder.build()).run());
-        } catch (Exception e) {
-            throw new RuntimeException(e.getMessage(), e);
+        } catch (RunnerException e) {
+            throw new RuntimeException(e);
         }
     }
 
-    public enum Assertions {
-        Enable("-enableassertions"),
-        Disable("-disableassertions");
+    /* Options */
+    private enum ForkJvm {
+        ENABLE(1),
+        DISABLE(0);
+
+        final int forkCount;
+
+        ForkJvm(int forkCount) {
+            this.forkCount = forkCount;
+        }
+    }
+
+    private enum Assertions {
+        ENABLE("-enableassertions"),
+        DISABLE("-disableassertions");
 
         final String vmArg;
 
@@ -93,13 +113,22 @@ public class JmhRunner {
         }
     }
 
+    private enum PrintInlining {
+        ENABLE,
+        DISABLE;
+    }
+
+    /* Helper methods */
+
     public static Integer[] getRandomValues(int size, int seed) {
         return getRandomValues(size, seed, false);
     }
 
     public static Integer[] getRandomValues(int size, int seed, boolean nonNegative) {
-        final Random random = new Random(seed);
+        return getRandomValues(size, nonNegative, new Random(seed));
+    }
 
+    public static Integer[] getRandomValues(int size, boolean nonNegative, Random random) {
         final Integer[] results = new Integer[size];
         for (int i = 0; i < size; i++) {
             final int value = random.nextInt(size) - (nonNegative ? 0 : (size / 2));
--- a/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
@@ -6,14 +6,19 @@ import org.openjdk.jmh.annotations.*;
 import org.openjdk.jmh.infra.Blackhole;
 import scala.compat.java8.JFunction;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Objects;
+import java.util.Random;
 import java.util.stream.Collectors;
 
 import static java.util.Arrays.asList;
-import static javaslang.JmhRunner.*;
+import static javaslang.JmhRunner.create;
+import static javaslang.JmhRunner.getRandomValues;
 import static javaslang.collection.Collections.areEqual;
-import static scala.collection.JavaConversions.*;
+import static scala.collection.JavaConversions.asJavaCollection;
+import static scala.collection.JavaConversions.asScalaBuffer;
 
+@SuppressWarnings({"ALL", "unchecked"})
 public class VectorBenchmark {
     static final Array<Class<?>> CLASSES = Array.of(
             Create.class,
@@ -34,30 +39,34 @@ public class VectorBenchmark {
     }
 
     public static void main(String... args) {
+        JmhRunner.runDebugWithAsserts(CLASSES);
         JmhRunner.runNormalNoAsserts(CLASSES);
     }
 
     @State(Scope.Benchmark)
     public static class Base {
-        @Param({ "32", "1024", "32768"/*, "1048576"*/ }) // i.e. depth 1,2,3(,4) for a branching factor of 32
+        @Param({"32", "1024", "32768" /*, "1048576", "33554432", "1073741824" */}) /* i.e. depth 1,2,3(,4,5,6) for a branching factor of 32 */
         public int CONTAINER_SIZE;
 
         int EXPECTED_AGGREGATE;
         Integer[] ELEMENTS;
+        int[] RANDOMIZED_INDICES;
 
         /* Only use this for non-mutating operations */
         java.util.ArrayList<Integer> javaMutable;
 
-        fj.data.Seq<Integer> fjavaPersistent = fj.data.Seq.empty();
-        org.pcollections.PVector<Integer> pcollectionsPersistent = org.pcollections.TreePVector.empty();
         scala.collection.immutable.Vector<Integer> scalaPersistent;
         clojure.lang.PersistentVector clojurePersistent;
+        fj.data.Seq<Integer> fjavaPersistent;
+        org.pcollections.PVector<Integer> pcollectionsPersistent;
         javaslang.collection.Vector<Integer> slangPersistent;
 
         @Setup
-        @SuppressWarnings("unchecked")
         public void setup() {
-            ELEMENTS = getRandomValues(CONTAINER_SIZE, 0);
+            final Random random = new Random(0);
+            ELEMENTS = getRandomValues(CONTAINER_SIZE, false, random);
+            RANDOMIZED_INDICES = Arrays2.shuffle(Array.range(0, CONTAINER_SIZE).toJavaStream().mapToInt(Integer::intValue).toArray(), random);
+
             EXPECTED_AGGREGATE = Array.of(ELEMENTS).reduce(JmhRunner::aggregate);
 
             javaMutable = create(java.util.ArrayList::new, asList(ELEMENTS), v -> areEqual(v, asList(ELEMENTS)));
@@ -69,6 +78,7 @@ public class VectorBenchmark {
         }
     }
 
+    /** Bulk creation from array based, boxed source */
     public static class Create extends Base {
         @Benchmark
         public Object java_mutable() {
@@ -109,7 +119,7 @@ public class VectorBenchmark {
         public Object slang_persistent() {
             final javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.ofAll(javaMutable);
             assert areEqual(values, javaMutable);
-            return values.head();
+            return values;
         }
     }
 
@@ -157,7 +167,6 @@ public class VectorBenchmark {
         }
     }
 
-    @SuppressWarnings("Convert2MethodRef")
     public static class Tail extends Base {
         @State(Scope.Thread)
         public static class Initialized {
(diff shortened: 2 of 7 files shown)


javaslang-benchmark/src/test/java/javaslang/JmhRunner.java
```diff
@@ -15,76 +20,91 @@
...
+                    .mode(Mode.Throughput)
+                    .timeUnit(TimeUnit.SECONDS)
+                    .warmupTime(TimeValue.milliseconds(millis))
+                    .warmupIterations(warmupIterations)
+                    .measurementTime(TimeValue.milliseconds(millis))
+                    .measurementIterations(measurementIterations)
+                    .forks(forkJvm.forkCount)
+                  /* We are using 4Gb and setting NewGen to 100% to avoid GC during testing.
+                     Any GC during testing will destroy the iteration (i.e. introduce unreliable noise in the measurement), which should get ignored as an outlier */
+                    .jvmArgsAppend("-XX:+UseG1GC", "-Xss100m", "-Xms4g", "-Xmx4g", "-XX:MaxGCPauseMillis=1000", "-XX:+UnlockExperimentalVMOptions", "-XX:G1NewSizePercent=100", "-XX:G1MaxNewSizePercent=100", assertions.vmArg);
```
paplorinc: changed it back to 4gb, to enable bigger collections also

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -34,30 +39,34 @@ public void testAsserts() {
...
 
     public static void main(String... args) {
+        JmhRunner.runDebugWithAsserts(CLASSES);
         JmhRunner.runNormalNoAsserts(CLASSES);
     }
 
     @State(Scope.Benchmark)
     public static class Base {
-        @Param({ "32", "1024", "32768"/*, "1048576"*/ }) // i.e. depth 1,2,3(,4) for a branching factor of 32
+        @Param({"32", "1024", "32768" /*, "1048576", "33554432", "1073741824" */}) /* i.e. depth 1,2,3(,4,5,6) for a branching factor of 32 */
```
paplorinc: somebody should run it for the disabled values also (I don't have enough memory)

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -177,9 +186,9 @@ public void tearDown() {
 
         @Benchmark
         public Object java_mutable(Initialized state) {
-            final java.util.ArrayList<Integer> values = state.javaMutable;
+            java.util.List<Integer> values = state.javaMutable;
             for (int i = 0; i < CONTAINER_SIZE; i++) {
-                values.remove(0);
+                values = values.subList(1, values.size()); // remove(0) would copy everything, but this will slow access down because of nesting
```
paplorinc: `remove(0)` might be cheating (i.e. making `ArrayList` look worst), but `subList` has its drawbacks also, i.e. access will depend on the number of `subList` wrappers

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -196,16 +205,6 @@ public Object scala_persistent() {
         }
 
         @Benchmark
-        public Object clojure_persistent() {
```
paplorinc: Clojure doesn't have a `tail`

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -236,11 +235,12 @@ public Object slang_persistent() {
         }
     }
 
+    /** Aggregated, randomized access to every element */
     public static class Get extends Base {
         @Benchmark
         public int java_mutable() {
             int aggregate = 0;
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
+            for (int i : RANDOMIZED_INDICES) {
```
paplorinc: `Scala` optimizes for sequential access (i.e. it extracts the most recently used leaf node, providing actual constant time access to it). I personally find it weird that an immutable data structure stores part of its history...

What we want is to simulate random access, hence the iteration.

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -399,6 +399,20 @@ public Object scala_persistent() {
         }
 
         @Benchmark
+        public Object clojure_persistent() {
```
paplorinc: Clojure's Vector doesn't have an prepend, so I'm reconstructing everything every time to simulate a prepend

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -445,7 +458,9 @@ public Object java_mutable() {
         public Object scala_persistent() {
             scala.collection.immutable.Vector<Integer> values = scala.collection.immutable.Vector$.MODULE$.empty();
             for (Integer element : ELEMENTS) {
-                values = values.appendBack(element);
+                final scala.collection.immutable.Vector<Integer> newValues = values.appendBack(element);
+                assert values != newValues;
```
paplorinc: just a quick validation to be sure the data structure is persistent

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -509,36 +532,50 @@ public Object slang_persistent() {
...
 
+    /** Consume the vector one-by-one, from the front and back */
     public static class Slice extends Base {
         @Benchmark
-        public void java_mutable(Blackhole bh) {
-            for (int i = 0; i < CONTAINER_SIZE; i++) {
-                for (int j = i; j < Math.min(CONTAINER_SIZE, 100); j++) {
-                    bh.consume(javaMutable.subList(i, j));
-                }
+        public void java_mutable(Blackhole bh) { // stores the whole list underneath
```
paplorinc: has memory leak :/

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -509,36 +532,50 @@ public Object slang_persistent() {
...
+            java.util.List<Integer> values = javaMutable;
+            while (!values.isEmpty()) {
+                values = values.subList(1, values.size());
+                values = values.subList(0, values.size() - 1);
+                bh.consume(values);
+            }
+        }
+
+        @Benchmark
+        public void clojure_persistent(Blackhole bh) { // stores the whole list underneath
```
paplorinc: has memory leak :/

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -103,13 +107,9 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
     @SafeVarargs
+    @SuppressWarnings("varargs")
     public static <T> Vector<T> of(T... elements) {
-        Objects.requireNonNull(elements, "elements is null");
-        HashArrayMappedTrie<Integer, T> result = HashArrayMappedTrie.empty();
-        for (T element : elements) {
-            result = result.put(result.size(), element);
-        }
-        return wrap(result);
+        return ofAll(List.of(elements));
```
paplorinc: simplify `Vector` to make the review of its replacement simpler

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -594,59 +600,70 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
-        HashArrayMappedTrie<Integer, T> trie = HashArrayMappedTrie.empty();
-        for (int i = 0; i < length() - n; i++) {
-            trie = trie.put(trie.size(), get(i));
-        }
-        return wrap(trie);
     }
 
-    @Override
-    public Vector<T> dropUntil(Predicate<? super T> predicate) {
+    public Vector<T> dropRightWhile(Predicate<? super T> predicate) {
```
paplorinc: https://github.com/javaslang/javaslang/issues/1479

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1013,44 +974,27 @@ public int length() {
     }
 
     @Override
-    public IndexedSeq<T> removeAll(Predicate<? super T> predicate) {
+    public Vector<T> removeAll(Predicate<? super T> predicate) {
```
paplorinc: The testing framework should have caught this, right @danieldietrich?

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1075,30 +1015,24 @@ public int length() {
...
-        if (beginIndex >= endIndex || beginIndex >= length() || isEmpty()) {
-            return Vector.empty();
-        }
-        if (beginIndex <= 0 && endIndex >= length()) {
+        if ((beginIndex >= endIndex) || (beginIndex >= size()) || isEmpty()) {
+            return empty();
+        } else if ((beginIndex <= 0) && (endIndex >= length())) {
             return this;
+        } else {
+            return take(endIndex).drop(beginIndex);
```
paplorinc: slice = drop + dropRight

javaslang/src/test/java/javaslang/collection/VectorTest.java
```diff
@@ -238,10 +238,32 @@ public void shouldUnfoldToEmpty() {
...
+                Vector.unfold(10, x -> x == 0
+                                       ? Option.none()
+                                       : Option.of(new Tuple2<>(x - 1, x))))
+                .isEqualTo(of(1, 2, 3, 4, 5, 6, 7, 8, 9, 10));
+    }
+
+    // -- dropRightWhile
+
+    @Test
+    public void shouldDropRightWhileNoneOnNil() {
```
paplorinc: https://github.com/javaslang/javaslang/issues/1479

javaslang/src/test/java/javaslang/collection/VectorTest.java
```diff
@@ -263,40 +285,6 @@ public void shouldNotSerializeEnclosingClass() throws Throwable {
...
     }
 
-    @Test(expected = InvalidObjectException.class)
-    public void shouldNotDeserializeListWithSizeLessThanOne() throws Throwable {
-        try {
-            /*
-             * This implementation is stable regarding jvm impl changes of object serialization. The index of the number
-             * of List elements is gathered dynamically.
-             */
-            final byte[] listWithOneElement = Serializables.serialize(List.of(0));
```
paplorinc: was testing `List`, not `Vector`

javaslang/src/main/java/javaslang/collection/Vector.java
ruslansennov: I believe this is bad idea for `{length=1_000_000; beginIndex=300_000; endIndex=600_000}`
With your changes two big loops will be executed (but only one in previous implementation)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -22,15 +24,19 @@
  * @author Ruslan Sennov
  * @since 2.0.0
  */
-public final class Vector<T> implements Kind1<Vector<?>, T>, IndexedSeq<T>, Serializable {
-
+@SuppressWarnings({"SuspiciousArrayCast", "unchecked"})
+public class Vector<T> implements Kind1<Vector<?>, T>, IndexedSeq<T> {
```
ruslansennov: Why `Vector` doesn't implement `Serializable`?

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: This will be constant time in the next version, I'm optimizing now for simpler review.
Would you like me to introduce this only when the complexity is also better? (i.e. the next PR)

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: It was redundant, already defined in `interface λ<R> extends Serializable`.
I could still leave it there if you think it's more readable, it just makes no difference to the compiler.

javaslang/src/main/java/javaslang/collection/Vector.java
ruslansennov: But now complexity is O(n)

paplorinc: It was O(n) before also, but I will change it temporarily back, if it bothers you

javaslang/src/main/java/javaslang/collection/Vector.java
ruslansennov: > It was redundant, already defined in interface λ<R> extends Serializable.
> I could still leave it there if you think it's more readable, it just makes no difference to the compiler.

Thank you, it's OK for me.
And second question: why `Vector` is not final now? I don't remember exactly, but this keyword very important here... /cc @danieldietrich

paplorinc: It was put back in a later commit, I just realized after pushing that I forgot to rebase the `final` back here :)
Force pushed a new version :)

javaslang/src/main/java/javaslang/collection/Vector.java
ruslansennov: Two loops again :(

ruslansennov:
IMHO as first step in `Vector` improvement it is enough to implement the BitMappedTrie (I guess it will be Vector.VectorTree with some modifications maybe) and just replace HAMT with BMT in current Vector implementation.

paplorinc:
would it complicate the review for you?
I don't mind putting it in a separate PR :)

ruslansennov:
> would it complicate the review for you?
I don't mind putting it in a separate PR :)

In ideal it will be two changed classes: new class `BitMappedTrie.java` and modified (with just replaced HAMT to BMT) `Vector.java`. Nothing complicated :)

paplorinc:
Sure, should I do them in this PR, in separate commits, or in another PR?

ruslansennov:
> Sure, should I do them in this PR, in separate commits, or in another PR?

I'm sorry for your wasted time, but I believe it should be in separated PR.
For Vector purposes BMT is **obviously** better choice than HAMT.
All other improvements are discusable, but not this one.
@danieldietrich ?

paplorinc:
> In ideal it will be two changed classes: new class BitMappedTrie.java and modified (with just replaced HAMT to BMT) Vector.java. Nothing complicated :)

In HAMT you can put negative indices and add an offset to it later. In BMT you need to shift the whole tree by a full half tree (i.e. add another parent and pretend the whole new left side is filled with elements) and store it in an offset so that only the actual nodes are accessed.

It's not a problem, but I do consider it a lot more complicated than a search and replace :).

ruslansennov:
> In HAMT you can put negative indices and add an offset to it later. In BMT you need to shift the whole tree by a full half tree (i.e. add another parent and pretend the whole new left side is filled with elements) and store it in an offset so that only the actual nodes are accessed.

It's not such a huge change like #1449 or this PR.
And main goal: this change can be compact and observable

ruslansennov:
> In BMT you need to shift the whole tree by a full half tree (i.e. add another parent and pretend the whole new left side is filled with elements) and store it in an offset so that only the actual nodes are accessed.

Sorry, maybe I don't understand something, but why this logic can't be moved to BMT?

paplorinc:
It can, it's just not as trivial as for a HAMT.
I thought a lot about how to make `prepend` constant time (Clojure simply skipped it).

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: I had to find an intermediary representation that removed the `HAMT` (but was still working correctly).
In a later commit it's replaced again with `BMT`, but will replace it with `appendAll` or something here :)

paplorinc:
@ruslansennov, @danieldietrich, pushed a new version and following Ruslan's advice, I made the simplest `BMT` I could :).
Will push that once this is merged and once the benchmarks are run, so that we can compare them against the optimized one.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -22,15 +24,19 @@
  * @author Ruslan Sennov
  * @since 2.0.0
  */
-public final class Vector<T> implements Kind1<Vector<?>, T>, IndexedSeq<T>, Serializable {
-
+@SuppressWarnings({"SuspiciousArrayCast", "unchecked"})
```
ruslansennov: I would revert this back. It should be done deliberately, in place of code where it is needed. Somebody can make a mistake in the future because warnings switched off in whole class

paplorinc: Done :)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -655,19 +676,14 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
-            if (predicate.test(t)) {
-                filtered = filtered.put(filtered.size(), t);
-            }
-        }
 
-        if (filtered.isEmpty()) {
-            return Vector.empty();
-        } else if (filtered.size() == size()) {
+        final Vector<T> results = ofAll(iterator().filter(predicate));
+        if (results.isEmpty()) {
```
ruslansennov: This is redundant. If `results.isEmpty()` then we have empty singleton and can return `results`

paplorinc: if we can assume that we won't construct any other empty instance, than indeed, it can be removed, thanks :)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -677,28 +693,23 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
+                                     : ofAll(results);
         }
     }
 
     @Override
     public T get(int index) {
-        if (index < 0 || index >= length()) {
-            throw new IndexOutOfBoundsException("get(" + index + ")");
-        }
+        if ((index < 0) || (index >= length())) throw new IndexOutOfBoundsException("get(" + index + ")");
```
ruslansennov: Please, revert this. Too hard to read...

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -744,61 +756,28 @@ public int indexOf(T element, int from) {
...
-            }
-            if (i < length()) {
-                trie = trie.put(trie.size(), get(i));
-            }
-        }
-        return new Vector<>(trie);
+
+        final Vector<T> begin = take(index);
+        final Vector<T> end = drop(index);
+        return begin.appendAll(elements).appendAll(end);
```
ruslansennov: Here we have four (!) loops against one loop in previous implementation

paplorinc: k, will put this in the next commit then, where `take` and `drop` are `O(1)`

paplorinc:
@ruslansennov, force pushed a new version

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -842,37 +816,39 @@ public int length() {
...
-            trie = trie.put(i, mapper.apply(get(i)));
+
+        if (isEmpty()) {
+            return empty();
+        } else {
+            final Iterator<? extends U> results = iterator().map(mapper);
+            return results.isEmpty() ? empty()
+                                     : ofAll(results);
         }
-        return wrap(trie);
```
ruslansennov: I believe this code is enough
```java
public <U> Vector<U> map(Function<? super T, ? extends U> mapper) {
    return ofAll(iterator().map(mapper));
}
```

Method `ofAll` returns singleton if parameter is empty

paplorinc: Yes, that's basically what I just pushed:
```java
public <U> Vector<U> map(Function<? super T, ? extends U> mapper) {

    final Iterator<? extends U> results = iterator().map(mapper);
    return ofAll(results);
}
```

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -902,76 +868,65 @@ public int length() {
...
-            newTrie = newTrie.put(i, list.head());
-            list = list.tail();
+        Objects.requireNonNull(elements, "elements is null");
+        if (!elements.iterator().hasNext()) {
+            return this;
+        } else if (isEmpty()) {
+            return ofAll(elements);
+        } else {
+            Vector<T> result = this;
+            for (T element : seq(elements).reverse()) {
```
ruslansennov: This is more obvious code, thank you

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -902,76 +868,65 @@ public int length() {
...
-            if (found) {
-                trie = trie.put(trie.size(), value);
-            } else {
-                if (element.equals(value)) {
-                    found = true;
-                } else {
-                    trie = trie.put(trie.size(), value);
-                }
+            if (Objects.equals(get(i), element)) {
+                return removeAt(i);
```
ruslansennov: :+1:

ruslansennov:
Thanks, @paplorinc 
On first sight I haven't anymore significant comments on this PR.

@danieldietrich your move!

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -594,59 +599,70 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
-        final java.util.Set<T> seen = new java.util.TreeSet<>(comparator);
+        final Set<T> seen = new java.util.TreeSet<>(comparator);
         return filter(seen::add);
     }
 
     @Override
     public <U> Vector<T> distinctBy(Function<? super T, ? extends U> keyExtractor) {
         Objects.requireNonNull(keyExtractor, "keyExtractor is null");
-        final java.util.Set<U> seen = new java.util.HashSet<>();
+        final Set<U> seen = new java.util.HashSet<>();
```
danieldietrich: Hint: In Javaslang code I almost always fully qualify Java collection types to make the difference clear, even if there is no conflict.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -22,15 +24,18 @@
  * @author Ruslan Sennov
  * @since 2.0.0
  */
-public final class Vector<T> implements Kind1<Vector<?>, T>, IndexedSeq<T>, Serializable {
-
+public final class Vector<T> implements Kind1<Vector<?>, T>, IndexedSeq<T> {
```
danieldietrich: I see it is final again (can't comment on out-dated diffs). It is important because the return types cannot be adjusted by sub-classes of Vector any more. 

IndexedSeq has `Option<? extends IndexedSeq<T>> initOption();`

Vector has `Option<Vector<T>> initOption();`.

A subclass `X`cannot have `Option<X<T>>` any more.

danieldietrich: Oohhh, Vector.initOption() currently has return type `Option<? extends Vector<T>>`. That is not correct. Our final collections do have `? extends` any more. Could you please correct it @paplorinc?

danieldietrich: It would be better to explicitly state that Vector is Serializable by declaring the interface in the implements section. I've checked the other collection types and we did it not consequently. But I think we should do it for all collections (in a separate PR), it is more robust regarding changes of super-interfaces.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -882,12 +848,12 @@ public int length() {
     @Override
     public Tuple2<Vector<T>, Vector<T>> partition(Predicate<? super T> predicate) {
         Objects.requireNonNull(predicate, "predicate is null");
-        final java.util.List<T> left = new ArrayList<>(), right = new ArrayList<>();
+        final ArrayList<T> left = new ArrayList<>(), right = new ArrayList<>();
```
danieldietrich: Did you change it to ArrayList because of the full qualified name? It is ok for me here because it is internal.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -902,76 +868,65 @@ public int length() {
...
-                    found = true;
-                } else {
-                    trie = trie.put(trie.size(), value);
-                }
+            if (predicate.test(get(i))) {
+                return removeAt(i);
             }
         }
-        return trie.size() == length() ? this : wrap(trie);
+        return this;
```
danieldietrich: nice

javaslang/src/main/java/javaslang/collection/Vector.java
danieldietrich: I removed that kind of tests (then re-added it after a few months and finally removed it again) - we have to take care of our own (or re-add the tests again).

The tests were too complicated and I had to maintain a white-list of invalid return types. This was error prone.

danieldietrich:
Thanks, finished my review of this PR, please find my comments attached. I like the simplifications!

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: yes

paplorinc:
Thank you @ruslansennov and @danieldietrich, applied the rest of the comments via force push.

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
danieldietrich: how much mem do you have (I have 16 GB)

danieldietrich:
Thanks! Merging now the first step towards the new Vector :)

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
paplorinc: only 8 :/

danieldietrich: Is the command still valid / do you use the same to start our benchmarks?

```bash
mvn test -Pbenchmark -pl javaslang
```

Is it possible to run only `VectorBenchmark` (command line and/or within IntelliJ IDEA)?

paplorinc: yes, run the `main` from `VectorBenchmark` :)

danieldietrich: will run the disabled values later today. I'm preparing release 2.0.3 just now - will still take some time

danieldietrich: @paplorinc 

Regarding my question:

> mvn test -Pbenchmark -pl javaslang
> Is the command still valid / do you use the same to start our benchmarks?

I did run

```java
mvn test -Pbenchmark -pl javaslang
```

The result:

```java
[WARNING] The requested profile "benchmark" could not be activated because it does not exist.
```

**How do I run the bechmarks? How do I get the formatted reports?**

ruslansennov: Benchmark profile removed. To run performance benchmarks you should run static `main()` method in specific class. Memory consumption benchmarks starts with `maven test` automatically.
Also see #1413

danieldietrich: Thank you, Ruslan! We could add a class RunAllBenchmarks if needed.

paplorinc: > yes, run the main from VectorBenchmark :)

paplorinc: > We could add a class RunAllBenchmarks if needed.

javaslang.JmhRunner#main:
```java
/**
 * Runs all the available benchmarks in precision mode.
 * Note: it takes about 3 hours.
 */
    final Array<Class<?>> CLASSES = Array.of(
            ArrayBenchmark.class,
            BitSetBenchmark.class,
            CharSeqBenchmark.class,
            HashSetBenchmark.class,
            ListBenchmark.class,
            PriorityQueueBenchmark.class,
            VectorBenchmark.class
    );
    runDebugWithAsserts(CLASSES);
    runSlowNoAsserts(CLASSES);
}
```

PR #1504, opened 2016-08-17
Changed `Vector` from `HAMT` to `BMT`

paplorinc:
Fixes https://github.com/javaslang/javaslang/issues/568.
Dependency for: https://github.com/javaslang/javaslang/pull/1449 and https://github.com/javaslang/javaslang/pull/1522

This is the non-optimized version that I only compared to `Scala` for now, feel free to compare it against the [optimized version](https://github.com/javaslang/javaslang/pull/1522).

Memory:
```java
for 32768 elements
 `scala.collection.immutable.Vector` uses `659 KB` (`20.7 KB` overhead, `0.6` bytes overhead per element)
 `javaslang.collection.Vector` uses `659 KB` (`20.6 KB` overhead, `0.6` bytes overhead per element)

for 1024 elements
 `scala.collection.immutable.Vector` uses `19.3 KB` (`672 bytes` overhead, `0.7` bytes overhead per element)
 `javaslang.collection.Vector` uses `19.3 KB` (`664 bytes` overhead, `0.6` bytes overhead per element)

for 32 elements
 `scala.collection.immutable.Vector` uses `536 bytes` (`32 bytes` overhead, `1.0` bytes overhead per element)
 `javaslang.collection.Vector` uses `528 bytes` (`24 bytes` overhead, `0.8` bytes overhead per element)
```

Speed:
```java
Operation Ratio                                 32   1024  32768 
Create    slang_persistent/scala_persistent  0.33×  0.11×  0.07×
Head      slang_persistent/scala_persistent  0.85×  0.64×  0.49×
Tail      slang_persistent/scala_persistent  1.04×  1.06×  0.96×
Get       slang_persistent/scala_persistent  1.22×  0.90×  0.85×
Update    slang_persistent/scala_persistent  1.04×  1.14×  1.24×
Prepend   slang_persistent/scala_persistent  0.64×  0.58×  0.43×
Append    slang_persistent/scala_persistent  0.94×  0.37×  0.29×
GroupBy   slang_persistent/scala_persistent  0.96×  0.31×  0.22×
Slice     slang_persistent/scala_persistent  0.79×  1.23×  1.64×
Iterate    slang_persistent/scala_persistent 1.72×  1.03×  0.95×
```

i.e. for small number of elements it often has the same speed, for larger ones it's often `2-3x` slower.
Compared to the optimized version this is roughly:
* create: `~30x` slower
* head: `~2x` slower
* tail: `~4x` slower
* get:  `~0.7x` slower (i.e. `~1.4x` faster)
* update:  `~0.95x` slower (i.e. `~1.05x` faster)
* prepend: `~2.5x` slower
* append: `~2.5x` slower
* groupby:  `~3.5x` slower
* slice: `~3.5x` slower
* iterate: `~1.1x` slower


javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -11,33 +11,40 @@
...
+ * Many other operations (e.g. `tail`, `drop`, `slice`) are also effectively constant.
+ * <p>
+ * It's implemented using a `bit-mapped trie`, i.e. a tree where each node has `32` children.
+ * For integer indices this means that the tree's depth will be ≤ 6.
  *
  * @param <T> Component type of the Vector.
- * @author Ruslan Sennov
- * @since 2.0.0
+ * @author Pap Lőrinc
+ * @since 3.0.0
```
danieldietrich: maybe it will make it to 2.1.0, the interface did not change (only methods were added)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1387,23 +1265,158 @@ public String toString() {
...
+
+            return new VectorTree<T>(array, offset, newSize, depthShift);
+        }
+
+        VectorTree<T> update(int index, T element) {
+            final Object[] root = recursiveSet(array, offset + index, depthShift, element, 0);
+            return new VectorTree<T>(root, offset, length(), depthShift);
+        }
+
+        /** Since the depth of the tree is always small, a non-tailrecursive call is actually faster than pre-calculating the path (is probably unrolled) */
```
danieldietrich: +1 for using recursion here

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1387,23 +1265,158 @@ public String toString() {
...
             }
         }
+
+        VectorTree<T> drop(int n) {
+            final int index = offset + n;
+
+            int shiftValue = depthShift;
+            int previousOffset = firstDigit(index, shiftValue);
+            final Object[] root = copyDrop(array, previousOffset);
+            for (Object[] array = root; (array != null) && (shiftValue >= BRANCHING_BASE); ) {
```
danieldietrich: Can array ever be `null` here?

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1387,23 +1265,158 @@ public String toString() {
...
+                shiftValue -= BRANCHING_BASE;
+                final int offset = digit(index, shiftValue);
+
+                final Object[] newNode = copyDrop(getAt(array, previousOffset), offset);
+                array[previousOffset] = newNode;
+                array = newNode;
+
```
danieldietrich: I don't understand line 1361 and 1362. First we assign an array at a specific index to an object, then the array itself. How's that!?

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1387,23 +1265,158 @@ public String toString() {
...
+        private VectorTree(Object[] array, int offset, int length, int depthShift) {
+            this.array = array;
+            this.offset = offset;
+            this.length = length;
+            this.depthShift = depthShift;
+
+            assert length() <= treeSize(branchingFactor(), depthShift);
+        }
+
+        static int treeSize(int branchCount, int depthShift) {
```
danieldietrich: I would restrict all methods and fields to `private` access if possible. In most cases white-box unit tests are not really necessary. All branches can also be tested by using public API.

javaslang/src/main/java/javaslang/collection/Arrays2.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Internal class, containing Java array manipulation helpers.
+ * Many arrays are represented as simple Objects, to avoid casts at the client site, and to work with all types of arrays, including primitive ones.
+ *
+ * @author Pap Lőrinc
+ * @since 3.0.0
+ */
+@SuppressWarnings({"unchecked", "SuspiciousArrayCast"})
+final class Arrays2 { // TODO reuse these in `Array` also
```
danieldietrich: I would name it `Arrays` because original Java collection(-related) classes are full qualified per convention. We also have `Collections` in Javaslang (and Java).

javaslang/src/main/java/javaslang/collection/Arrays2.java
```diff
@@ -0,0 +1,92 @@
...
+        return (isNullOrEmpty(array) || (i < 0) || (i >= array.length))
+               ? defaultValue
+               : (T) getAt(array, i);
+    }
+
+    /** Create a single element array */
+    static Object[] asArray(Object element) {
+        final Object[] copy = arrayCopy(1, null, 0, 0, 0);
+        copy[0] = element;
+        return copy;
```
danieldietrich: I would remove this method completely and use

```java
new Object[] { element }
```

directly instead of

```java
asArray(element);
```

To many calls increase the cyclomatic complexity and make it hard for humans to read the code. I know that you like meaningful names. If the compiler optimizes it 'away' this method is ok for me but the method body is really hard to read.

javaslang/src/main/java/javaslang/collection/Arrays2.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    static <T> T getAt(Object arrayObject, int index) {
+        final Object[] array = (Object[]) arrayObject;
+        return (T) array[index];
+    }
+
+    /** Randomly mutate array positions */
+    static <T> int[] shuffle(int[] array, Random random) {
```
danieldietrich: Please move `shuffle` and `swap` out of this class to the benchmark package. These (mutating) methods are not needed in production code.

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: Valid question, it could in the past, lemme' recheck  :)

paplorinc: Nope, it cannot, fixed

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: We've made a copy (and a drop) of `array[previousOffset]` and a insert it back there.
After that, we dive deeper and make the newly modified node the current array (to which we'll do the same).
I.e. in `drop` we don't have to go to the leaf and come back again (as in `update` ... though we could do both in the same style, I will investigate), we just have to go down, the new root is already stored.

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: sure, we can open it up later, if needed

javaslang/src/main/java/javaslang/collection/Arrays2.java
paplorinc: done

javaslang/src/main/java/javaslang/collection/Arrays2.java
paplorinc: will update the method body for now, if that's ok :)

javaslang/src/main/java/javaslang/collection/Arrays2.java
paplorinc: Sure, will move it. The rest of the methods (e.g. `getAt`) are there to avoid all the casts the code would need, as it's not typesafe by nature (we're storing array of arrays and array of values in the same tree).
Later these methods may also help in the primitivization commit (it's the same abstraction).

javaslang/src/main/java/javaslang/collection/Arrays2.java
paplorinc: though I've done it in a separate commit, as now all completely unrelated classes have to have a fully qualified `java.util.Arrays`

javaslang/src/main/java/javaslang/collection/Vector.java
danieldietrich: I understand, thx, looks good to me

javaslang/src/main/java/javaslang/collection/Arrays2.java
danieldietrich: great!

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: ... you did point out an asymmetry though :)
`drop` and `take` should have had a similar structure (one was iterative, the other recursive).
By changing `take` to be iterative (i.e. the equivalent of a tail recursive call) it became more unified and almost twice as fast for big values :).
Thanks.

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: Thanks, though this iterated twice over the depth.
I managed to unify depth traversal in a single method, hope you like that also :).

paplorinc:
Force pushed a new version.
Thanks to @danieldietrich, I managed to have a single depth modification method, extended with some lambdas. This way most methods are unified now!

We don't need to go down the tree and back again (saves time), but the lambdas slow it down a bit, so the end performance is basically the same as before (updated it in the description, though).

Now the non-optimized `BMT` is only 150 lines :).

@danieldietrich, @ruslansennov, @zsolt-donca, waiting eagerly for your reviews :)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1283,9 +1254,162 @@ public String stringPrefix() {
...
+            }
+        }
+
+        private static Object[] modifyLeafValue(Object[] array, int depthShift, int index, Object element) {
+            final NodeModifier node = (o, i) -> copy(o, i + 1);
+            final NodeModifier leaf = (o, i) -> copyUpdate(o, i, element);
+            return modifyLeaf(array, depthShift, index, node, leaf);
+        }
+
+        private static Object[] modifyLeaf(Object[] array, int depthShift, int index, NodeModifier nodeModifier, NodeModifier leafModifier) {
```
paplorinc: `prepend`, `append`, `update`, `drop` and `take` now use this iterative method instead

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -1283,9 +1254,162 @@ public String stringPrefix() {
     public String toString() {
         return mkString(stringPrefix() + "(", ", ", ")");
     }
+
+    static final class VectorTree<T> implements Serializable {
```
ruslansennov: It is possible to rename this class to BMT and move outside of `Vector` with package visibility? Just thought

paplorinc: If @danieldietrich doesn't mind, I would love to do that

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,160 @@
...
+    BitMappedTrie<T> prependLeaf(T leading) {
+        final int newSize = length() + 1;
+        if (length() == 0) {
+            return new BitMappedTrie<>(asArray(leading), offset, newSize, depthShift);
+        } else {
+            Object[] array = this.array;
+            int depthShift = this.depthShift, offset = this.offset;
+            if (isFullLeft()) {
+                array = copyUpdate(emptyArray(), Vector.branchingFactor() - 1, array);
+                depthShift += Vector.BRANCHING_BASE;
```
ruslansennov: Can `Vector.BRANCHING_BASE` (and other BMT-related stuff) be moved to BMT?

paplorinc: Of course, thanks.

javaslang/src/main/java/javaslang/collection/Vector.java
danieldietrich: I think we can save lambda instances by creating singletons. I'm not sure if the curried NODE version helps...

```java
// names can be changed
static final NodeModifier NODE = (o, i) -> copy(o, i + 1);
static final Function<Object, NodeModifier> LEAF = element -> (o, i) -> copyUpdate(o, i, element);

// ...

private static Object[] modifyLeafValue(Object[] array, int depthShift, int index, Object element) {
    return modifyLeaf(array, depthShift, index, NODE, LEAF.apply(element));
}
```

paplorinc: Lemme' benchmark it :)

paplorinc: There's a measurable difference this way :dancer: 
Thanks @danieldietrich!

before:
```java
Update    slang_persistent/scala_persistent  1.03×  0.93×  1.17×
Prepend   slang_persistent/scala_persistent  0.64×  0.58×  0.44×
Append    slang_persistent/scala_persistent  0.84×  0.33×  0.26×
Slice     slang_persistent/scala_persistent  0.74×  1.19×  1.47×
```
after:
```java
Operation  Ratio                                 32   1024  32768 
Update     slang_persistent/scala_persistent  1.04×  1.14×  1.24×
Prepend    slang_persistent/scala_persistent  0.64×  0.58×  0.43×
Append     slang_persistent/scala_persistent  0.94×  0.37×  0.29×
Slice      slang_persistent/scala_persistent  0.79×  1.23×  1.64×
```

danieldietrich: Mmhh, interesting - do I interpret it right that creating more instances performs better?

How frequently `modifyLeafValue()` is called on an operation, round about? Zero or Once / logarithmic times?

paplorinc: It's called once per operation, e.g. one `update` will call `modifyLeaf` once.
But don't worry about `GC`, through escape analysis short-lived lambdas might end up on the stack, instead of the heap (https://dzone.com/articles/java-lambdas-and-low-latency).

javaslang/src/main/java/javaslang/collection/Arrays.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /** System.arrayCopy with same source and destination */
+    static Object[] copyRange(Object array, int from, int to) {
+        final int length = to - from;
+        return arrayCopy(length, array, from, 0, length);
+    }
+
+    /** clone the source and set the value at the given position */
+    static Object[] copyUpdate(Object arrayObject, int index, Object element) {
+        final Object[] array = (Object[]) arrayObject;
```
paplorinc: as stated in the above JavaDoc, untyped `Object` array, to get rid of call-site casts (and make room for non-Object based arrays also)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+
+    int offset() { return offset; }
+    int length() { return length; }
+}
+
+@FunctionalInterface
+interface NodeModifier {
+    Object[] apply(Object arrayObject, int index);
```
paplorinc: introduced, to avoid boxing (i.e. `Function2<Object, Integer, Object[]>`)

javaslang/src/main/java/javaslang/collection/Arrays.java
```diff
@@ -0,0 +1,80 @@
...
+        return copyRange(array, 0, lastIndex + 1);
+    }
+
+    /** for performance reasons the array allocation and the System.arraycopy must be next to each other - even an if (size > 0) will make it slower */
+    static Object[] arrayCopy(int arraySize, Object source, int sourceFrom, int destinationFrom, int size) {
+        if (size == 0) {
+            return new Object[arraySize];
+        } else {
+            final Object[] result = new Object[arraySize];
+            System.arraycopy(source, sourceFrom, result, destinationFrom, size);
```
paplorinc: this was a weird one ... if I put an `if` statement in front of the `arraycopy`, the performance will be halved (i.e. `if (size > 0)`, which should only simplify things...)

danieldietrich: wow, interesting!

javaslang/src/main/java/javaslang/collection/Vector.java
danieldietrich: sounds good!

danieldietrich:
Build fails with

```bash
[ERROR] /home/travis/build/javaslang/javaslang/javaslang-benchmark/src/test/java/javaslang/collection/CharSeqBenchmark.java:[54,98] method equals in class java.lang.Object cannot be applied to given types;
  required: java.lang.Object
  found: char[],char[]
  reason: actual and formal argument lists differ in length
```

javaslang/src/main/java/javaslang/collection/Arrays.java
paplorinc: it probably `null`s out the array otherwise...

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,164 @@
...
+
+        assert length() <= treeSize(branchingFactor(), depthShift);
+    }
+
+    private static int treeSize(int branchCount, int depthShift) {
+        final int fullBranchSize = 1 << depthShift;
+        return branchCount * fullBranchSize;
+    }
+
+    BitMappedTrie<T> prependLeaf(T leading) {
```
paplorinc: should be optimized by appending whole arrays instead...

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,164 @@
...
+            }
+
+            offset -= 1;
+            array = modifyLeaf(array, depthShift, offset, NODE_MODIFIER, LEAF_MODIFIER.apply(leading));
+            return new BitMappedTrie<>(array, offset, newSize, depthShift);
+        }
+    }
+    private boolean isFullLeft() { return offset == 0; }
+
+    BitMappedTrie<T> appendLeaf(T trailing) {
```
paplorinc: same

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,164 @@
...
+
+    private static final NodeModifier NODE_MODIFIER = (o, i) -> copy(o, i + 1);
+    private static final Function<Object, NodeModifier> LEAF_MODIFIER = element -> ((o, i) -> copyUpdate(o, i, element));
+
+    BitMappedTrie<T> update(int index, T element) {
+        final Object[] root = modifyLeaf(array, depthShift, offset + index, NODE_MODIFIER, LEAF_MODIFIER.apply(element));
+        return new BitMappedTrie<>(root, offset, length(), depthShift);
+    }
+
+    BitMappedTrie<T> drop(int n) {
```
paplorinc: `drop` and `tail` could be optimized to only modifying the tree, if anything other than the leaf should be changed (i.e. something like `if (n - offset > branchingFactor())`)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -561,20 +566,22 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
-        HashArrayMappedTrie<Integer, T> result = trie;
-        for (T element : elements) {
-            result = result.put(result.size() + indexShift, element);
+        Objects.requireNonNull(elements, "elements is null");
+        if (isEmpty()) {
+            return ofAll(elements);
+        } else {
+            Vector<T> result = this;
+            for (T element : elements) {
+                result = result.append(element);
```
paplorinc: this could be optimized via constructing mutable leafs and inserting those into the tree

javaslang/src/main/java/javaslang/collection/Arrays.java
```diff
@@ -0,0 +1,80 @@
...
+        copy[index] = element;
+        return copy;
+    }
+
+    static Object[] copy(Object arrayObject, int minLength) {
+        final Object[] array = (Object[]) arrayObject;
+        final int arrayLength = (array == null) ? 0 : array.length;
+        final int length = Math.max(arrayLength, minLength);
+        if (length == arrayLength) {
+            return array.clone();
```
paplorinc: this distinction is important for `update`, where the length isn't changed

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -690,7 +687,9 @@ public T get(int index) {
         if ((index < 0) || (index >= length())) {
```
paplorinc: `iterator` should be optimized by getting the corresponding leaf array and iterating over that (instead of finding it every time)

javaslang/src/main/java/javaslang/collection/Vector.java
danieldietrich: Would memory sharing of the immutable data structure still work when using mutable leaves?

paplorinc: Sure, that's how `create` worked previously also: https://github.com/javaslang/javaslang/pull/1449/commits/d0b782f386271cb2fa6c2dcc7d2bcbe2c79a57bc#diff-c6c1876105d111251f38e0916d4d99a4R24

i.e. create the array locally, populate it, put it in the tree and don't keep any other references to it :)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -561,20 +566,22 @@ private Vector(int indexShift, HashArrayMappedTrie<Integer, T> trie) {
...
+            return ofAll(elements);
+        } else {
+            Vector<T> result = this;
+            for (T element : elements) {
+                result = result.append(element);
+            }
+            return result.isEmpty() ? empty()
+                                    : result;
         }
-        return new Vector<>(indexShift, result);
```
danieldietrich: If we internally leverage mutability to gain performance we should add unit tests that ensure that memory sharing of our persistent versions still works. This test makes our implementation robust regarding changes by other persons that may not know these internal properties.

paplorinc: I did, please check https://github.com/javaslang/javaslang/pull/1504/files/a07dbff7904d85c9d5c7a457dc913d88a681ddff..dcc87bc47d9541d2e07e96cf42f174c23211a932#diff-c89f0d7cf85723aceb8e16223092d4f1R240

danieldietrich: Great! Just to get sure - `history.forEach(...)` also checks that previous versions remain unmodified? I.e. `t._1` is independent of `t._2`

ruslansennov:
Sorry guys, I can't take part in discussion, too busy until Monday.
I'll review this PR later (if it will be actual)

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -0,0 +1,264 @@
...
+    public void shouldIterate() {
+        for (byte depth = 1; depth <= 6; depth++) {
+            for (int i = 0; i < getMaxSizeForDepth(depth); i++) {
+                final Seq<Integer> expected = Array.range(0, i);
+                final Vector<Integer> actual = Vector.ofAll(expected);
+
+                assertAreEqual(actual, null, constant(actual), expected);
+            }
+
+            System.out.println("Depth " + depth + " ok!");
```
danieldietrich: we should remove all System.out calls (and assert if necessary)

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -0,0 +1,264 @@
...
+                actual = assertAreEqual(actual, drop, Vector::drop, expected);
+
+                for (int i = 0; i < actual.length(); i++) {
+                    final Integer newValue = mapper.apply(actual.get(i));
+                    actual = actual.update(i, newValue);
+                }
+
+                assertAreEqual(actual, 0, (a, p) -> a, expected.map(mapper));
+            }
+            System.out.println("Depth " + depth + " ok!");
```
danieldietrich: dito

danieldietrich:
For me it looks good so far. I can pull it in if the System.outs are removed...

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
paplorinc: Sure, it was mainly for debugging (I've seen that many other Javaslang tests print out content, but I agree that they shouldn't) :)

danieldietrich: we will clean up the others also (in another PR)

paplorinc: (I will keep `System.out.println("using seed " + seed);` as it's the only way to reproduce it in case of failure)

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: sure, `_1` is an `Array`, `_2` is a `Vector` (https://github.com/javaslang/javaslang/pull/1504/files/a07dbff7904d85c9d5c7a457dc913d88a681ddff..dcc87bc47d9541d2e07e96cf42f174c23211a932#diff-c89f0d7cf85723aceb8e16223092d4f1R168)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,164 @@
...
+final class BitMappedTrie<T> implements Serializable {
+    static int BRANCHING_BASE = 5;
+    static int branchingFactor()                   { return 1 << BRANCHING_BASE; }
+    static int firstDigit(int num, int depthShift) { return num >> depthShift; }
+    static int digit(int num, int depthShift)      { return lastDigit(firstDigit(num, depthShift)); }
+    static int lastDigit(int num)                  { return num & (-1 >>> -BRANCHING_BASE); }
+
+    private static final long serialVersionUID = 1L;
+
+    private static final BitMappedTrie<?> EMPTY = new BitMappedTrie<>(emptyArray(), 0, 0, 0);
```
paplorinc: will rename this to `empty`

paplorinc:
Pushed a fixed version.
I think we should wait for @zsolt-donca's review first (or even @ruslansennov's), I don't mind.

danieldietrich:
I received an email

> \> +
> \> +import java.io.Serializable;
> \> +import java.util.function.Function;
> \> +
> \> +import static javaslang.collection.Arrays.*;
> \> +
> \> +final class BitMappedTrie<T> implements Serializable {
> \> +    static int BRANCHING_BASE = 5;
> \> +    static int branchingFactor()                   { return 1 << BRANCHING_BASE; }
> \> +    static int firstDigit(int num, int depthShift) { return num >> depthShift; }
> \> +    static int digit(int num, int depthShift)      { return lastDigit(firstDigit(num, depthShift)); }
> \> +    static int lastDigit(int num)                  { return num & (-1 >>> -BRANCHING_BASE); }
> \> +
> \> +    private static final long serialVersionUID = 1L;
> \> +
> \> +    private static final BitMappedTrie<?> EMPTY = new BitMappedTrie<>(emptyArray(), 0, 0, 0);
> will rename this to empty

Can't find it any more on the PR but have one question. What do you mean - renaming `EMPTY` to `empty`? Just asking because `static final` fields are by convention upper-case...

paplorinc:
yeah, I posted it to the wrong place and deleted it :p
I renamed `BMT.emptyTree()` to `BMT.empty()` (https://github.com/javaslang/javaslang/pull/1504/commits/98b6f26ce0c368e792b1d4dc6812bae77116d68c#diff-fc685519cd17ce4aee692e5275289289R24)

paplorinc:
@eduardmanas, I would be interested in your opinion also, you've had valuable suggestions before :)

danieldietrich:
Would be so great if @djspiewak could take a look. Could be of interest for #Scala also!

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,153 @@
...
+                array = newNode;
+            }
+
+            array[previousIndex] = leaf.apply(array[previousIndex], lastDigit(index));
+            return root;
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    T[] getLeaf(int index) {
```
paplorinc: Simplified it further...

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            array = modifyLeaf(array, depthShift, offset + length(), NODE_MODIFIER, LEAF_MODIFIER.apply(trailing));
+            return new BitMappedTrie<>(array, offset, newSize, depthShift);
+        }
+    }
+    private boolean isFullRight(int newSize) { return (offset + newSize) > treeSize(branchingFactor(), depthShift); }
+
+    private static final NodeModifier NODE_MODIFIER = (o, i) -> copy(o, i + 1);
+    private static final Function<Object, NodeModifier> LEAF_MODIFIER = element -> ((o, i) -> copyUpdate(o, i, element));
```
zsolt-donca: You could replace this with a simple method:
```
    private NodeModifier leafModifier(T element) {
        return (o, i) -> copyUpdate(o, i, element);
    }
```

Invocations of the function `LEAF_MODIFIER.apply(foo)` could be transformed to simple method calls: `leafModifier(foo)`.

Since you don't actually store any of the `NodeModifier` instances, there is no caching in place (as opposed to `NODE_MODIFIER`), so I don't see any such performance disadvantages for what I'm proposing.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
zsolt-donca: Why is this method called `prependLeaf` as opposed to just `prepend` ? Does it make sense to have, or can you imagine, an operation in which we prepend anyplace else but to the first leaf?

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
zsolt-donca: Same :)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: Good idea, let's see how it affects performance :)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: Until now you could only append full leafs (proposed here as optimization) and didn't see the point in renaming it back and forth.
Renamed both :)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,153 @@
...
+                depthShift += BRANCHING_BASE;
+            }
+
+            array = modifyLeaf(array, depthShift, offset + length(), NODE_MODIFIER, LEAF_MODIFIER.apply(trailing));
+            return new BitMappedTrie<>(array, offset, newSize, depthShift);
+        }
+    }
+    private boolean isFullRight(int newSize) { return (offset + newSize) > treeSize(branchingFactor(), depthShift); }
+
+    private static final NodeModifier NODE_MODIFIER = (o, i) -> copy(o, i + 1);
```
zsolt-donca: You could give the field `NODE_MODIFIER` a name that reflects *what it does* as opposed to *what it is*. You would do the same if, instead of a field, you would have a no-arg method that returned you the  `NodeModifier` instance.

I understand that this `NodeModifier` copies the given array and makes sure that `i + 1` becomes a valid index (increasing the array size if needed). One suggestion would be to name it `COPY_NODE` (ignoring the part about the array size, as I'm not sure if that's really important).

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
zsolt-donca: Continuing the idea on naming (see comment on `NODE_MODIFIER`), you could rename this field/method to something like `updateLeafWith`, as this `NodeModifier` updates the leaf to have the given element.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,153 @@
...
+        } else if (n <= 0) {
+            return empty();
+        } else {
+            final NodeModifier take = Arrays::copyTake;
+            final Object[] root = modifyLeaf(array, depthShift, (offset + n) - 1, take, take);
+            return new BitMappedTrie<>(root, offset, n, depthShift);
+        }
+    }
+
+    private static Object[] modifyLeaf(Object[] root, int depthShift, int index, NodeModifier node, NodeModifier leaf) {
```
zsolt-donca: +1 to this method: it's a really great idea to have a single method to traverse the tree, with the behavior for modifying the nodes and the leaf being parametric.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -11,39 +11,43 @@
...
+ * Many other operations (e.g. `tail`, `drop`, `slice`) are also effectively constant.
+ * <p>
+ * It's implemented using a `bit-mapped trie`, i.e. a tree where each node has `32` children.
+ * For integer indices this means that the tree's depth will be ≤ 6.
  *
  * @param <T> Component type of the Vector.
- * @author Ruslan Sennov
- * @since 2.0.0
+ * @author Ruslan Sennov, Pap Lőrinc
+ * @since 2.0.0, 2.1.0
```
zsolt-donca: What is the reason behind adding `2.1.0`?

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: It affected performance slightly, so I kept both fields, renamed them and added a convenience method to avoid the call-site `apply`.

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: Not sure what the protocol is in these cases, @danieldietrich?

paplorinc:
Pushed a new version with the applied recommendations. Keep them coming :)

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -578,6 +581,7 @@ public void slang_persistent(Blackhole bh) {
             for (int i = 1; !values.isEmpty(); i++) {
                 values = values.slice(1, values.size());
                 values = values.slice(0, values.size() - 1);
+                assert sizeOf(values.trie) < (branchingFactor() * sizeOf(copyRange(ELEMENTS, i, javaMutable.size() - i))); /* detect memory leak */
```
paplorinc: added leak detection, i.e. if `slice` only keeps a few elements from a lot, its size should be small

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -0,0 +1,260 @@
...
+        }
+    }
+
+    @Test
+    public void shouldIterate() {
+        for (byte depth = 1; depth <= 6; depth++) {
+            for (int i = 0; i < getMaxSizeForDepth(depth); i++) {
+                final Seq<Integer> expected = Array.range(0, i);
+                final Vector<Integer> actual = Vector.ofAll(expected);
+                assertAreEqual(actual, null, constant(actual), expected);
```
zsolt-donca: I believe you could simply write `assertAreEqual(expected, actual);`

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -0,0 +1,260 @@
...
+    public void shouldCreateAndGet() {
+        for (int i = 0; i < 2000; i++) {
+            final Seq<Integer> expected = Array.range(0, i);
+            final Vector<Integer> actual = Vector.ofAll(expected);
+            assertAreEqual(expected, actual);
+        }
+    }
+
+    @Test
+    public void shouldIterate() {
```
zsolt-donca: It seems to me that this test and `shouldCreateAndGet` test the same thing, but with different array sizes.

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -0,0 +1,260 @@
...
+        for (byte depth = 1; depth <= 6; depth++) {
+            for (int i = 0; i < getMaxSizeForDepth(depth); i++) {
```
zsolt-donca: Since `getMaxSizeForDepth` currently has a cap of `10 000`, which requires depth = 3, you are iterating over the very same interval for depths `3 .. 6`.

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
paplorinc: Indeed, thanks :)

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
paplorinc: Yup, stupid leftover code, fixed :)

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -0,0 +1,260 @@
...
+import java.util.function.Function;
+
+import static javaslang.Function2.constant;
+import static javaslang.collection.BitMappedTrie.BRANCHING_BASE;
+import static javaslang.collection.BitMappedTrie.branchingFactor;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class VectorPropertyTest {
+    @Before
+    public void setUp() { BRANCHING_BASE = 2; }
```
zsolt-donca: It seems risky to me to not change this back to its original value. This can effect other tests run in the same JVM. For example, running `VectorPropertyTest` and `VectorTest` subsequently will mean that `VectorTest` runs with a branching base of `2`, and branching factor of `4`. However, running `VectorTest` individually will use the default values, such as base `5` and factor `32`.

Therefore, I'd suggest adding an `@After` method with setting `BRANCHING_BASE` back to `5`.

paplorinc: `gradle` would run them in forked JVMs and all tests should run with `BRANCHING_BASE = 2` also, but you're right that this introduces unwanted statefulness (and the fix is easy)
Thanks!.

javaslang/src/main/java/javaslang/collection/Vector.java
danieldietrich: We do not protocal updates. Just leave only 2.0.0 please - the interface will remain the same. Changes to implementations are continuous.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -690,7 +684,9 @@ public T get(int index) {
         if ((index < 0) || (index >= length())) {
             throw new IndexOutOfBoundsException("get(" + index + ")");
         } else {
-            return trie.get(index + indexShift).get();
+            final T[] leaf = trie.getLeaf(index);
+            final int leafIndex = lastDigit(trie.offset() + index);
+            return leaf[leafIndex];
```
ruslansennov: I believe these 3 lines should be moved to BMT, it's his logic

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -752,18 +748,10 @@ public int indexOf(T element, int from) {
...
-                    trie = trie.put(trie.size(), element);
-                }
-            }
-            if (i < length()) {
-                trie = trie.put(trie.size(), get(i));
-            }
-        }
-        return new Vector<>(trie);
+
+        final Vector<T> begin = take(index);
```
ruslansennov: here complexity is O(log(n)) with quite small constant, but not O(1) like you told, am I right?

paplorinc: it's effectively constant, yes

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: I could put this particular logic inside `BMT`, but it wouldn't scale.
I think that would hinder reusability, as `leaf` and `index` will be used by the optimized iterator also.
In addition you can't primitivize a getter (it would have to return an `Object`), but you can cast an array and get the value out without boxing.
If you still think I should extract it, I will.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -11,39 +11,41 @@
...
+    private static final Vector<?> EMPTY = new Vector<>(BitMappedTrie.empty());
 
-    private final HashArrayMappedTrie<Integer, T> trie;
-    private final int indexShift;
-
-    private static <T> Vector<T> wrap(HashArrayMappedTrie<Integer, T> trie) {
-        return trie.isEmpty() ? empty() : new Vector<>(trie);
+    final BitMappedTrie<T> trie;
+    private Vector(BitMappedTrie<T> trie) {
+        assert (EMPTY == null) || (trie.length() > 0);
```
danieldietrich: Wow, this is tricky :-) During initialization of EMPTY the constructor is called and EMPTY == null is true.

In all other calls of the constructor we assert that the empty trie is not passed to ensure the EMPTY Vector is a singleton.

I had to proof it to understand the assert statement:

```java
    (EMPTY == null) || (trie.length() > 0)   // applying De Morgan
<=> !((EMPTY != null) && (trie.length <= 0)) // my assumption: (EMPTY != null) is true
<=> !(true && (trie.length <= 0))            // true && A <=> A
<=> !(trie.length <= 0)                      // negation
<=> trie.length > 0                          // not correct for EMPTY Vector
```

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -11,39 +11,41 @@
...
+        this.trie = trie;
     }
 
-    private Vector(HashArrayMappedTrie<Integer, T> trie) {
-        this(0, trie);
+    @SuppressWarnings("ObjectEquality")
+    private Vector<T> wrap(BitMappedTrie<T> trie) {
+        return (trie == this.trie) ? this
```
danieldietrich: Currently we have only one case where `trie == this.trie` and `trie.length() > 0`, namely `trie.drop(n)` where `n <= 0`:

```java
wrap(trie.append(element));  // trie.append(element) != trie
wrap(trie.drop(n));          // trie.drop(n) == trie <=> n <= 0
wrap(trie.prepend(element)); // trie.prepend(element) != trie
wrap(trie.take(n));          // trie.take(n) == trie <=> trie is empty
```

I thought of merging `wrap(BMT)` and `ofAll(BMT)` to one static method `wrap(BMT self, BMT that)` but that is not possible for the other static `ofAll` methods (without passing `null` as self).

I would like to rename `ofAll(BMT)` to `wrap(BMT)`. In order to resolve the name clash we make the current `wrap(BMT)` static and pass the `this` instance as `self` parameter: `wrap(self, that)`.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,164 @@
...
+
+    private static final BitMappedTrie<?> EMPTY = new BitMappedTrie<>(Arrays.empty(), 0, 0, 0);
+    @SuppressWarnings("unchecked")
+    static <T> BitMappedTrie<T> empty() { return (BitMappedTrie<T>) EMPTY; }
+
+    private final Object[] array;
+    private final int offset, length;
+    private final int depthShift;
+
+    BitMappedTrie(Object[] array, int offset, int length, int depthShift) {
```
danieldietrich: Can we assert here s.th. similar to Vector? E.g.

```java
assert (EMPTY == null) || (array.length > 0)
```

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,164 @@
...
+final class BitMappedTrie<T> implements Serializable {
+    static int BRANCHING_BASE = 5;
+    static int branchingFactor()                   { return 1 << BRANCHING_BASE; }
+    static int firstDigit(int num, int depthShift) { return num >> depthShift; }
+    static int digit(int num, int depthShift)      { return lastDigit(firstDigit(num, depthShift)); }
+    static int lastDigit(int num)                  { return num & (-1 >>> -BRANCHING_BASE); }
+
+    private static final long serialVersionUID = 1L;
+
+    private static final BitMappedTrie<?> EMPTY = new BitMappedTrie<>(Arrays.empty(), 0, 0, 0);
```
danieldietrich: Why is `Arrays.empty()` a method and not a constant?

Why do we not initialize `EMPTY` with `new BitMappedTrie<>(new Object[0], 0, 0, 0);` and get the empty array like so when needed within BitMappedTrie: `EMPTY.array`?

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,164 @@
...
+            array = modifyLeaf(array, shift, offset, COPY_LEAF, updateLeafWith(leading));
+            return new BitMappedTrie<>(array, offset, newSize, shift);
+        }
+    }
+    private boolean isFullLeft() { return offset == 0; }
+
+    BitMappedTrie<T> append(T trailing) {
+        final int newSize = length() + 1;
+        if (length() == 0) {
+            return new BitMappedTrie<>(asArray(trailing), offset, newSize, depthShift);
```
danieldietrich: while reading `asArray(element)` I have a question in mind (without exactly knowing what I'm talking about because I hadn't the time to read enough):

Is it possible to allocate the whole array (of size 32) without using all 'slots'? Would that safe us array copy operations when modifying the trie?

javaslang/src/main/java/javaslang/collection/Vector.java
danieldietrich: Please put the whole get() method into BMT and replace this with

```java
@Override
public T get(int index) {
    return trie.get(index);
}
```

Additionally the optimized Iterator should also be moved to BMT, and not be part of Vector.

> In addition you can't primitivize a getter (it would have to return an Object), but you can cast an array and get the value out without boxing.

That's right. But we start to mix up our code here in order to do low level optimizations. The primitive optimizations are not part of this PR. We can talk about that later...

javaslang/src/main/java/javaslang/collection/Vector.java
danieldietrich: minor: in line 669 is an unnecessary newline.
IntelliJ IDEA does not preserve your format of the ternary operator:
```java
         final Vector<T> results = ofAll(iterator().filter(predicate));
         return (results.length() == length()) ? this
                                               : results;
```

After formatting the file with our presets, the lines look like this:

```java
        return (results.length() == length()) ? this
                : results;
```

Please format ternary operators like this:

```java
        return (results.length() == length()) ? this : results;
```

danieldietrich:
We can optimize the partition method, which currently first fills an ArrayList, then calls  `ofAll(Iterable)`, which internally calls `trie.append`:

```java
    @Override
    public Tuple2<Vector<T>, Vector<T>> partition(Predicate<? super T> predicate) {
        final ArrayList<T> left = new ArrayList<>(), right = new ArrayList<>();
        for (int i = 0; i < length(); i++) {
            final T t = get(i);
            (predicate.test(t) ? left : right).add(t);
        }
        return Tuple.of(ofAll(left), ofAll(right));
    }
```

I think we could save CPU + Mem by calling `trie.append` directly:

```java
    @Override
    public Tuple2<Vector<T>, Vector<T>> partition(Predicate<? super T> predicate) {
        final BitMappedTrie<T> left = BitMappedTrie.empty(), right = BitMappedTrie.empty();
        for (int i = 0; i < length(); i++) {
            final T t = get(i);
            if (predicate.test(t)) {
                left = left.append(t);
            } else {
                right = right.append(t);
            }
        }
        return Tuple.of(ofAll(left), ofAll(right));
    }
```

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: Done: 
```java
assert (EMPTY == null) || ((length > 0) && (array.length > 0));
```

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: Currently everything is an `Object[]` but the `trailing` and `leading` arrays they would be `T[]`s and would need casts.
I could convert them to `Object[]` also (slightly reducing type safety), but they still wouldn't work for the primitivized alternative (though I may be able to work around that, as it's an empty instance).
However, I consider empty array creation the responsibility of the `Arrays` helper method, not `Vector`'s.
If you don't agree, I will change it like you asked :)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: Do you mean treating all non-full arrays (i.e. first and last, the rest should be full) as semi-mutable arrays (i.e. masked out parts to be modifiable)?
In simple cases it could probably be done, but not if we allow valid `null` values too (If I `prepend`, `drop` and `prepend`, while keeping all the intermediary states, the second prepend could only override values, if the value was `null`, but that might have been inserted by the first `prepend`), i.e.

* `of(1)` -> (1)
* `prepend(null)` -> (null, 1)
* `drop(1)` -> null, (1)
* `prepend(1)` -> (1, 0)

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: I could move the iterator at this stage, but when I'll introduce the `leading` and `trailing`, I would need to put it back.
I've had a separate `get` method at first also, but later decided that it's the same.
Will put both in `BMT` and move it back when the `PR` needs it, if that's ok.

paplorinc:
> I think we could save CPU + Mem by calling trie.append directly:

Again, you're basing these optimizations on the unoptimized version.
In the new optimized version `ofAll` is `~90x` faster (I'm working on that PR in parallel), so this method would become a **lot** slower if I changed it to append directly.
I suggest doing the optimizations later, when we can measure something concrete :).

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: Would it be ok if I rather pushed an updated formatter config instead, that supports aligning ternaries?
```File | Settings | Editor | Code Style | Java | Wrapping and Braces | Align when multiline ☑```

Comment (paplorinc) [inline javaslang/src/main/java/javaslang/collection/Vector.java]: https://github.com/javaslang/javaslang/pull/1517

**Comment by paplorinc** [comment]
@danieldietrich, @ruslansennov, @zsolt-donca, @eduardmanas, did I answer or apply all your requests?
Can we merge it :)?

File: javaslang/src/main/java/javaslang/collection/Vector.java [hunk 42]
```diff
@@ -11,39 +11,42 @@
...
 /**
- * Vector is the default Seq implementation. It provides the best performance in between Array (with constant time element access)
- * and List (with constant time element addition).
+ * Vector is the default Seq implementation that provides effectively constant time access to any element.
+ * Many other operations (e.g. `tail`, `drop`, `slice`) are also effectively constant.
  *
  * @param <T> Component type of the Vector.
- * @author Ruslan Sennov
+ * @author Ruslan Sennov, Pap Lőrinc
+ * @see BitMappedTrie
```
ruslansennov: Maybe last my comment :) BMT has package visibility, I believe this line should be removed (?)

paplorinc: I think this can be useful when someone's viewing the sources (I commented `BMT` thoroughly).
@danieldietrich?

danieldietrich: @paplorinc Please remove the line. Instead you could modify the first line and mention that our Vector is based on a Bit Mapped Trie implementation.

paplorinc: Done, let's merge it, as I have at least 2 more PRs to prepare :)

zsolt-donca:
@paplorinc Mine are all addressed!

Keep up the good work!

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,178 @@
...
+ * Access to a given position is done by converting the index to a base 32 number and using each digit to descend down the tree.
+ * Modifying the tree is done similarly, but along the way the path is copied, returning a new root every time.
+ * Append inserts in the last leaf, or if the tree is full from the right, it adds another layer on top of it (the old root will be the first of the new one).
+ * Prepend is done similarly, but an offset is needed, because adding a new top node (where the current root would be the last node of the new root)
+ * shifts the indices by half of the current tree's full size. The `offset` shifts them back to the correct index.
+ *
+ * @author Pap Lőrinc
+ * @since 2.1.0
+ */
+final class BitMappedTrie<T> implements Serializable {
```
danieldietrich: Hi @paplorinc, great work so far. This implementation is smaller than I expected!

Before pulling it I want to see just one more thing: Please make `BitMappedTrie` `Iterable` and implement the `iterator()` method in the way that it does not use `get(index)`. Instead it should use the internal BMT structure to efficiently iterate the elements.

This is inevitable because otherwise either the internal structure would leak to the outside or we could not leverage the knowledge about the internal structure and have to fall back to less efficient ways to access elements. The latter case is currently implemented.

I think that it will be a better base for comparing the upcoming optimizations.

It shouldn't be much effort, is it?

paplorinc: Sure, will do, but will have to delete it later when the `leading` and `trailing` are introduced to optimize `append` and `prepend` (in order to be able to iterate over all arrays, cannot combine 3 iterators efficiently otherwise)

paplorinc: Force pushed a new version.
@danieldietrich, I thought you were interested in the simplest solution in this commit.
I consider this explicit `iterator` an optimization.

danieldietrich: We will see. In the design of types it is the right place for iterator(). We will take a look at leading and trailing next.

In Scala Vector is in between Array and List. List is best at prepend() & tail(), Array is best at random access get() and update(). This is the goal for Vector optimizations, it should be good at both and be in between Array and List. That's what I'm looking at, not the optimization of specific methods like groupBy() and slice() for example.

I'm really looking forward to review leading() and trailing(), especially on the performance changes of get(), update(), append(), prepend() and tail().

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,209 @@
...
+        Object[] leaf = array;
+        for (int shift = depthShift; shift > 0; shift -= BRANCHING_BASE) {
+            leaf = getAt(leaf, digit(offset + index, shift));
+        }
+        return (T[]) leaf;
+    }
+
+    Iterator<T> iterator() {
+        return new Iterator<T>() {
+            int globalIndex = 0;
```
danieldietrich: Please create `final length = BitMappedTrie.this.length();`

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -0,0 +1,209 @@
...
+            int index = lastDigit(offset());
+            T[] leaf = getLeaf(globalIndex);
+            int length = leaf.length;
+
+            @Override
+            public boolean hasNext() { return globalIndex < BitMappedTrie.this.length(); }
```
danieldietrich: Please use `globalIndex < length` here.

javac does not inline the call. I was curious and tested 3 variants, please take a look at [this Gist](https://gist.github.com/danieldietrich/d004eb5c5432c384f3bb8ec429810a65).

danieldietrich: Also javac does not inline fields that are return by methods like length().

See [this Gist](https://gist.github.com/danieldietrich/138b4201ed34932b95e545fae8c6c5e9)

**With other words we should use not method calls within Javaslang classes when there are instance fields present.**

paplorinc: @danieldietrich, you're checking the bytecode, not the JITted result.
If you're interested whether they're inlined or not, create a JMH test, enable print inlining and run the tests :). But then again, it may be JDK dependent ... maybe even processor dependent.

> Please use globalIndex < length here.

Ok, I will, but I though you asked me to do optimizations in another commit.

paplorinc:
Pushed, could we please merge this and move on the the optimizations?

danieldietrich:
ok, thank you, pulling it in now

paplorinc:
> ok, thank you, pulling it in now

Please review https://github.com/javaslang/javaslang/pull/1522 also.

Issue #1326, opened 2016-05-11
[type hierarchy] Method references of public classes refer to non-visible methods of sub-classes


danieldietrich:
@tcn reported the following:

``` java
import javaslang.collection.Array;
import javaslang.collection.HashMap;

import java.util.function.BiFunction;

public class Test {

    public static void main(String[] args) {

        // prints HashMap((a, AA), (b, BB), (c, CC))
        test((l, r) -> l.merge(r));

        // Exception in thread "main" java.lang.IllegalAccessError: tried to access class javaslang.collection.AbstractMap from class Test
        test(HashMap::merge);

    }

    static void test(BiFunction<HashMap<String, String>, HashMap<String, String>, HashMap<String, String>> merge) {
        final HashMap<String, String> reduced = Array.of("a", "b", "c")
                .map(t -> HashMap.of(t, t.concat(t).toUpperCase()))
                .reduce(merge);
        System.out.println(reduced);
    }
}
```

`HashMap` has no declared method `merge`. It inherits `merge` from `AbstractMap`. The method reference `HashMap::merge` internally points to `AbstractMap::merge`. Because our class `AbstractMap` is package private, we get a `java.lang.IllegalAccessError`.

**We need to check sub-classes of all package private classes!**

The solution which solves this bug should **not** make internal classes like `AbstractMap` visible. It is not intended to be public API, we added it to reduce code duplication. We could override the methods in the sub-classes and delegate to the methods of the package-private super-class. Maybe it is also an option not to inherit from package-private classes, so we are forced to delegate (to static methods).

ruslansennov:
First of all we can move all tests to another package

danieldietrich:
Mmhh, interesting idea. We could do that!

nicolaiparlog:
Came here from Twitter, sorry for barging in.

Are you sure this is the correct analysis? It sounds very "un-Java" and I can not reproduce it in a simple setup (Gist pending).

nicolaiparlog:
Here's [the Gist](https://gist.github.com/nicolaiparlog/264e9a705cba15766d66eed4d3549389). I can't see anything unusual.

nicolaiparlog:
Gonna keep spamming... :)

This looks like the same case: [IllegalAccessError when using a public method reference of a package-private class through a public subclass from another package](http://stackoverflow.com/q/36100552/2525313)

danieldietrich:
Hi Nicolai, I like that kind of spam! More of that :)
[...]

PR #1512, opened 2016-08-20
Replaces AbstractMap with Maps. Towards #1326

danieldietrich:
This is a workaround for [JDK Bug 8141122](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8141122).

In this PR step-by-step all package private super classes of Javaslang types will be substituted with package private helper classes that contain static functions.


danieldietrich:
I will replace every abstract, package-private super class in a separate PR in order to make it easier to merge changes into the 2.0.3 branch.

PR #1513, opened 2016-08-20
Added solution for Euler number 43

jorander:
This is a straight-forward solution to Euler 43. But it is rather slow, takes almost 30 sec to run on my laptop. In profiling I see that about 80% of the time is spent in CharSeq#permutations.
I've tried to use Stream and List as alternatives, but that is even slower. The main problem is that this solutions creates lots of permutations (about 3.5 millions) out of which most can be discarded already by seeing that the number formed by digit 2,3 and 4 is not divisible by 2. I'll try to come up with an implementation not using permutations but rather breaking it down to check if it is meaningful to produce further permutations after checking divisibility with 2, then 3 and so on. I think CharSeq#combinations(k) can be a good starting-point.

### DIFF
--- a/javaslang/src/test/java/javaslang/collection/euler/Euler43Test.java
+++ b/javaslang/src/test/java/javaslang/collection/euler/Euler43Test.java
@@ -6,7 +6,10 @@
 package javaslang.collection.euler;
 
 import javaslang.collection.CharSeq;
-import javaslang.collection.Stream;
+import javaslang.Tuple;
+import javaslang.collection.List;
+import javaslang.collection.Seq;
+
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


danieldietrich:
Interesting - looking forward to see an optimization!

danieldietrich:
Wow - sophisticated. I was curious and changed the collections to Stream (instead of CharSeq and List). The tests then ran in 428 ms on my machine. Your version is better, it runs in 286 ms on my machine.

Great, thx!

jorander:
Yes, I think CharSeq and List are generally faster than Stream,  no hard evidence, just empirical data from trying them all on different Euler problems.. I've come to the conclusion to use Stream only when the problem can benefit from the laziness to avoid evaluating unnecessarily.

PR #1514, opened 2016-08-21
Making Euler number 43 a bit simpler.

jorander:
I found a few changes to make the code simpler and easier to read (perhaps also faster).

I really like this problem/solution because it shows the power of the persistent data types. I do CharSeq#remove lots of times on ALL_DIGITS without affecting the original list. It's really great to be able to declare a common set of data and then use it in different ways without having to think about shared state.

### DIFF
--- a/javaslang/src/test/java/javaslang/collection/euler/Euler43Test.java
+++ b/javaslang/src/test/java/javaslang/collection/euler/Euler43Test.java
@@ -6,7 +6,6 @@
 package javaslang.collection.euler;
 
 import javaslang.collection.CharSeq;
-import javaslang.Tuple;
 import javaslang.collection.List;
 import javaslang.collection.Seq;
 
@@ -64,7 +63,7 @@ public class Euler43Test {
                                         .removeAll(digitsSoFar)
                                         .map(nextDigit -> digitsSoFar.append(nextDigit))
                                 )
-                                .filter(digitsToTest -> Integer.valueOf(digitsToTest.takeRight(3).mkString()) % divisor == 0)
+                                .filter(digitsToTest -> digitsToTest.takeRight(3).parseInt() % divisor == 0)
                         )
                 )
                 .map(tailDigitsWithProperty -> tailDigitsWithProperty
@@ -73,7 +72,6 @@ public class Euler43Test {
                                 .head()
                         )
                 )
-                .map(CharSeq::mkString)
-                .map(Long::valueOf);
+                .map(CharSeq::parseLong);
     }
 }
--- a/javaslang/src/test/java/javaslang/collection/euler/Euler43Test.java
+++ b/javaslang/src/test/java/javaslang/collection/euler/Euler43Test.java
@@ -54,25 +54,22 @@ public class Euler43Test {
     private static Seq<Long> tenDigitPandigitalsWithProperty() {
         final CharSeq ALL_DIGITS = CharSeq.of("0123456789");
         final List<Integer> DIVISORS = List.of(2, 3, 5, 7, 11, 13, 17);
-        final char DUMMY_CHAR = Character.MIN_VALUE;
 
         return ALL_DIGITS
                 .combinations(2)
                 .flatMap(CharSeq::permutations)
-                .map(cs -> Tuple.of(cs, cs.prepend(DUMMY_CHAR), Integer.valueOf(cs.mkString())))
                 .flatMap(firstTwoDigits -> DIVISORS
                         .foldLeft(List.of(firstTwoDigits), (accumulator, divisor) -> accumulator
                                 .flatMap(digitsSoFar -> ALL_DIGITS
-                                        .removeAll(digitsSoFar._1)
-                                        .map(nextDigit -> Tuple.of(digitsSoFar._1.append(nextDigit), digitsSoFar._2.tail().append(nextDigit)))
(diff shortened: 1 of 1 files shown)


javaslang/src/test/java/javaslang/collection/euler/Euler43Test.java
```diff
@@ -54,25 +54,22 @@ public void shouldSolveProblem43() {
...
                         .foldLeft(List.of(firstTwoDigits), (accumulator, divisor) -> accumulator
                                 .flatMap(digitsSoFar -> ALL_DIGITS
-                                        .removeAll(digitsSoFar._1)
-                                        .map(nextDigit -> Tuple.of(digitsSoFar._1.append(nextDigit), digitsSoFar._2.tail().append(nextDigit)))
-                                        .map(digitsToTest -> Tuple.of(digitsToTest._1, digitsToTest._2, Integer.valueOf(digitsToTest._2.mkString())))
+                                        .removeAll(digitsSoFar)
+                                        .map(nextDigit -> digitsSoFar.append(nextDigit))
                                 )
-                                .filter(digitsToTest -> digitsToTest._3 % divisor == 0)
+                                .filter(digitsToTest -> Integer.valueOf(digitsToTest.takeRight(3).mkString()) % divisor == 0)
```
danieldietrich: I think about adding some syntactic sugar/numeric conversion methods to CharSeq, e.g.

```java
Integer.valueOf(digitsToTest.take(3).mkString())
```

could be

```java
digitsToTest.take(3).toInt()
```

jorander: That would be nice. It is pretty common doing that type of conversions in
the Euler problems.

danieldietrich: Tempted to do it before breakfast :)

danieldietrich: @jorander the conversion methods are here now. Please merge your branch with current master.

CharSeq.parse*, CharSeq.decode* and additional CharSeq.toXxx methods.

See also #1515

danieldietrich: To avoid boxing/unboxing

```java
Integer.valueOf(digitsToTest.takeRight(3).mkString()) % divisor == 0
```

is converted best using

```java
digitsToTest.takeRight(3).parseInt() % divisor == 0
```

javaslang/src/test/java/javaslang/collection/euler/Euler43Test.java
```diff
@@ -54,25 +54,22 @@ public void shouldSolveProblem43() {
     private static Seq<Long> tenDigitPandigitalsWithProperty() {
         final CharSeq ALL_DIGITS = CharSeq.of("0123456789");
         final List<Integer> DIVISORS = List.of(2, 3, 5, 7, 11, 13, 17);
-        final char DUMMY_CHAR = Character.MIN_VALUE;
 
         return ALL_DIGITS
                 .combinations(2)
                 .flatMap(CharSeq::permutations)
-                .map(cs -> Tuple.of(cs, cs.prepend(DUMMY_CHAR), Integer.valueOf(cs.mkString())))
```
danieldietrich: the Tuple import (at the top of the file) is not used any more - it can be removed

javaslang/src/test/java/javaslang/collection/euler/Euler43Test.java
```diff
@@ -54,25 +54,22 @@ public void shouldSolveProblem43() {
...
                 .flatMap(firstTwoDigits -> DIVISORS
                         .foldLeft(List.of(firstTwoDigits), (accumulator, divisor) -> accumulator
                                 .flatMap(digitsSoFar -> ALL_DIGITS
-                                        .removeAll(digitsSoFar._1)
-                                        .map(nextDigit -> Tuple.of(digitsSoFar._1.append(nextDigit), digitsSoFar._2.tail().append(nextDigit)))
-                                        .map(digitsToTest -> Tuple.of(digitsToTest._1, digitsToTest._2, Integer.valueOf(digitsToTest._2.mkString())))
+                                        .removeAll(digitsSoFar)
+                                        .map(nextDigit -> digitsSoFar.append(nextDigit))
```
danieldietrich: here we can use the shorted method reference: `.map(digitsSoFar::append)`

javaslang/src/test/java/javaslang/collection/euler/Euler43Test.java
danieldietrich: I applied the changes on my local machine. The `.parseInt()` makes no measurable difference regarding this example. It is just a bit more concise.

javaslang/src/test/java/javaslang/collection/euler/Euler43Test.java
jorander: Yes, but I think this a bit more verbose version naming what is being
appended (nextDigit) helps understanding/readability. If shortened to a
method reference the whole concept of "nextdigit" is implicit.

danieldietrich: you're right

danieldietrich:
Very nice solution, thanks!

PR #1518, opened 2016-08-22
Fixes Future bug with escaping RejectedExecutionException

danieldietrich:
@mattjtodd wrote:

I'm looking at approaches to apply back pressure using spring's `DeferredResult` and a limited-size Thread pool. In the case where all available Threads in the pools are in use, and a `RejectedExecutionException` is thrown, how this can be expressed using as little spring as possible.

Here's the a basic `CompletableFuture` version:

```java
private DeferredResult<ResponseEntity<HashResponse>> eexample1(@RequestBody HashRequest request) {
        DeferredResult<ResponseEntity<HashResponse>> deferredResult = new DeferredResult<>(Long.MAX_VALUE);

        try {
            CompletableFuture
                .supplyAsync(() -> {
                    try {
                        return encode(request, randomSource);
                    } catch (InvalidKeySpecException | NoSuchAlgorithmException e) {
                        throw new IllegalStateException(e);
                    }
                }, executor)
                .thenApply(ResponseEntity::ok)
                .whenComplete((success, thrown) -> {
                    if (Objects.nonNull(thrown)) {
                        deferredResult.setErrorResult(ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build());
                    } else {
                        deferredResult.setResult(success);
                    }

                });
        } catch (RejectedExecutionException e) {
            deferredResult.setErrorResult(ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).build());
        }

        return deferredResult;
    }
```

and the `Future` version. much neater :grinning:

```java
    private DeferredResult<ResponseEntity<HashResponse>> hash2(@RequestBody HashRequest request) {
        DeferredResult<ResponseEntity<HashResponse>> deferredResult = new DeferredResult<>(Long.MAX_VALUE);

        Try
            .of(() -> Future.of(executor, () -> encode(request, randomSource)))
            .getOrElseGet(Future::failed)
            .map(ResponseEntity::ok)
            .onFailure(thrown -> deferredResult.setErrorResult(ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build()))
            .onSuccess(deferredResult::setResult);

        return deferredResult;

    }
```

The second example doesn't offer the `SERVICE_UNAVAILABLE 503` response but it would be easy with `Match` / `Case` constructs. What's puzzling me now is that the first example handles the `RejectedExecutionException` as expected. The second example blocks forever on the first request once a second request is made which is then rejected. Unless there's something I'm missing, there could be a subtle bug in there somewhere.

### DIFF
--- a/javaslang/src/main/java/javaslang/concurrent/FutureImpl.java
+++ b/javaslang/src/main/java/javaslang/concurrent/FutureImpl.java
@@ -6,10 +6,10 @@
 package javaslang.concurrent;
 
 import javaslang.collection.Queue;
-import javaslang.control.*;
+import javaslang.control.Option;
+import javaslang.control.Try;
 import javaslang.control.Try.CheckedSupplier;
 
-import java.io.Serializable;
 import java.util.Objects;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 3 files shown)


danieldietrich:
For the sake of completeness: My suggested solution for the above example is:

```java
Future.of(executor, () -> encode(request, randomSource))
    .onComplete(result -> Match(result).of(
        Case(Success($()), deferredResult::setResult),
        Case(Failure($(instanceOf(RejectedExecutionException.class)), thrown -> deferredResult.setErrorResult(...)),
        Case(Failure($()), thrown -> deferredResult.setErrorResult(...))
    ));
```

PR #1520, opened 2016-08-23
Removed chained method alignment from formatter

paplorinc:
See the first method for an alternative layout, with aligned chained methods (though method extraction might not always be applicable).
See second method for new (old) formatting, after this PR.
formatting


danieldietrich:
Thank you Lorinc!

PR #1522, opened 2016-08-23
Vector optimizations

paplorinc:
Optimizes #1504

Every commit introduces a new type of optimization, i.e. 
* **Optimized `ofAll` for `Vector`**
  * `~150×` faster by building the tree from bottom up
* ~~**Optimized `prepend` and `append` for `Vector`**~~
  * ~~`~6×` faster by adding the `trailing` and `leading` arrays as depthless staging areas (i.e. only modify the tree if these two are full)~~  - **removed, as it hinders maintainability**
* **Optimized `drop` and `take` for `Vector`**
  * `>6×` faster by keeping the leafs intact on drops, as the middle tree already has an `offset` and a `length` that can mask them
* **Optimized `map` for `Vector`** and **Optimized `filter` for `Vector`**
  * `>14×` faster by using the internal `visitor` with `ofAll`

---
[Benchmarks](https://github.com/paplorinc/javaslang/blob/VectorOptimizations/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java#L49):

```java
Operation Ratio                                          32     1024    32768 
Create    slang_persistent/java_mutable               1.59×    0.93×    1.11×
Create    slang_persistent/fjava_persistent         198.49×  187.60×  179.09×
Create    slang_persistent/pcollections_persistent  115.43×  220.62×  351.42×
Create    slang_persistent/ecollections_persistent    1.93×    0.91×    1.12×
Create    slang_persistent/clojure_persistent         0.98×   13.66×   11.09×
Create    slang_persistent/scala_persistent          13.76×    8.23×    6.02×
          
Head      slang_persistent/java_mutable               0.80×    0.58×    0.44×
Head      slang_persistent/fjava_persistent           1.02×    0.74×    0.56×
Head      slang_persistent/pcollections_persistent    2.50×    4.21×    4.80×
Head      slang_persistent/ecollections_persistent    0.94×    0.68×    0.52×
Head      slang_persistent/clojure_persistent         0.85×    0.97×    1.00×
Head      slang_persistent/scala_persistent           0.96×    0.69×    0.52×
          
Tail      slang_persistent/java_mutable               0.98×    0.70×    0.52×
Tail      slang_persistent/fjava_persistent          69.30×   61.65×   64.54×
Tail      slang_persistent/pcollections_persistent    7.94×   10.77×   15.30×
Tail      slang_persistent/ecollections_persistent   23.08×  178.94× 5353.95×
Tail      slang_persistent/clojure_persistent         1.23×    0.88×    0.75×
Tail      slang_persistent/scala_persistent           4.15×    6.05×    7.77×
          
Get       slang_persistent/java_mutable               0.77×    0.46×    0.38×
Get       slang_persistent/fjava_persistent          61.60×  114.97×   88.71×
Get       slang_persistent/pcollections_persistent    8.43×   15.06×   18.61×
Get       slang_persistent/ecollections_persistent    0.85×    0.38×    0.36×
Get       slang_persistent/clojure_persistent         0.97×    1.43×    1.49×
Get       slang_persistent/scala_persistent           1.20×    1.20×    0.99×
          
Update    slang_persistent/java_mutable               0.10×    0.05×    0.03×
Update    slang_persistent/fjava_persistent         115.25×  241.74×  245.64×
Update    slang_persistent/pcollections_persistent    2.32×    4.11×    5.46×
Update    slang_persistent/ecollections_persistent   13.21×  154.34× 3096.27×
Update    slang_persistent/clojure_persistent         0.83×    1.25×    1.25×
Update    slang_persistent/scala_persistent           0.98×    1.05×    1.37×
          
Map       slang_persistent/java_mutable_loop          0.69×    0.70×    0.62×
Map       slang_persistent/java_mutable               2.82×    2.47×    2.11×
Map       slang_persistent/ecollections_persistent    1.45×    1.17×    1.17×
Map       slang_persistent/scala_persistent           1.34×    1.56×    1.95×
          
Filter    slang_persistent/java_mutable               2.44×    1.92×    1.45×
Filter    slang_persistent/ecollections_persistent    1.91×    1.37×    1.04×
Filter    slang_persistent/scala_persistent           1.63×    1.68×    1.72×
          
Prepend   slang_persistent/java_mutable               0.38×    0.72×   12.15×
Prepend   slang_persistent/fjava_persistent           1.59×    1.64×    1.18×
Prepend   slang_persistent/pcollections_persistent    1.06×    2.58×    3.10×
Prepend   slang_persistent/ecollections_persistent    2.74×   28.35×  615.48×
Prepend   slang_persistent/clojure_persistent         6.39×  132.62× 3502.23×
Prepend   slang_persistent/scala_persistent           0.26×    0.24×    0.18×
         
Append    slang_persistent/java_mutable               0.25×    0.06×    0.03×
Append    slang_persistent/fjava_persistent           4.99×    1.68×    1.04×
Append    slang_persistent/pcollections_persistent    2.64×    2.06×    2.23×
Append    slang_persistent/ecollections_persistent    8.74×   43.27×  562.24×
Append    slang_persistent/clojure_persistent         1.05×    0.27×    0.19×
Append    slang_persistent/scala_persistent           0.99×    0.27×    0.17×
                                                    
GroupBy   slang_persistent/java_mutable               0.40×    0.91×    0.83×
GroupBy   slang_persistent/scala_persistent           1.47×    1.09×    1.21×
                                                    
Slice     slang_persistent/java_mutable               0.65×    0.56×    0.57×
Slice     slang_persistent/pcollections_persistent    0.57×    0.48×    0.51×
Slice     slang_persistent/clojure_persistent         0.63×    0.53×    0.55×
Slice     slang_persistent/scala_persistent           2.66×    6.17×   10.82×
                                                    
Iterate   slang_persistent/java_mutable               1.10×    0.32×    0.49×
Iterate   slang_persistent/fjava_persistent         422.22×  306.40×  363.16×
Iterate   slang_persistent/pcollections_persistent   12.27×    9.77×   11.87×
Iterate   slang_persistent/ecollections_persistent    2.62×    1.18×    1.68×
Iterate   slang_persistent/clojure_persistent         0.88×    0.84×    1.06×
Iterate   slang_persistent/scala_persistent           1.97×    1.01×    1.12×
```

i.e. this PR made the following methods faster (for `32768` elements, compared to `Scala`)
```java
Create   6.02 / 0.04 = ~150.50× faster
Head     0.52 / 0.45 =   ~1.15× faster
Tail     7.77 / 1.02 =   ~7.61× faster
Get      0.99 / 0.85 =   ~1.16× faster
Update   1.37 / 0.50 =   ~2.74× faster
Map      1.95 / 0.06 =  ~32.50× faster
Filter   1.72 / 0.12 =  ~14.33× faster
Slice   10.82 / 1.61 =   ~6.72× faster
Iterate  1.12 / 1.08 =   ~1.03× faster
```

and memory usages:
```java
for `32` elements
  Java mutable @ ArrayList                            →   504 bytes
  Functional Java persistent @ Seq                    → 3,064 bytes
  PCollections persistent @ TreePVector               → 1,712 bytes
  Eclipse Collections persistent @ ImmutableArrayList →   496 bytes
  Clojure persistent @ PersistentVector               →   704 bytes
  Scala persistent @ Vector                           →   536 bytes
  Javaslang persistent @ Vector                       →   528 bytes

for `1024` elements
  Java mutable @ ArrayList                            →  19,064 bytes
  Functional Java persistent @ Seq                    → 104,760 bytes
  PCollections persistent @ TreePVector               →  55,984 bytes
  Eclipse Collections persistent @ ImmutableArrayList →  19,056 bytes
  Clojure persistent @ PersistentVector               →  20,504 bytes
  Scala persistent @ Vector                           →  19,736 bytes
  Javaslang persistent @ Vector                       →  19,728 bytes

for `32768` elements
  Java mutable @ ArrayList                            →   653,672 bytes
  Functional Java persistent @ Seq                    → 3,408,624 bytes
  PCollections persistent @ TreePVector               → 1,833,376 bytes
  Eclipse Collections persistent @ ImmutableArrayList →   653,664 bytes
  Clojure persistent @ PersistentVector               →   700,168 bytes
  Scala persistent @ Vector                           →   674,824 bytes
  Javaslang persistent @ Vector                       →   674,816 bytes
```

Note: only very simple optimizations were kept, favoring maintainability over raw speed.
Note2: all collections store boxed, non-primitive values for now.

### DIFF
--- a/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
@@ -11,8 +11,6 @@ import scala.compat.java8.functionConverterImpls.FromJavaPredicate;
 
 import java.util.Objects;
 import java.util.Random;
-import java.util.function.Function;
-import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import static java.util.Arrays.asList;
@@ -440,111 +438,77 @@ public class VectorBenchmark {
     }
 
     public static class Map extends Base {
-        private final Function<Integer, Integer> MAPPER = i -> i + 1;
+        private static int mapper(int i) { return i + 1; }
 
         @Benchmark
         public Object java_mutable_loop() {
             final Integer[] values = ELEMENTS.clone();
             for (int i = 0; i < CONTAINER_SIZE; i++) {
-                values[i] = MAPPER.apply(values[i]);
+                values[i] = mapper(values[i]);
             }
-            assert areEqual(Array.of(values), Array.of(ELEMENTS).map(MAPPER));
+            assert areEqual(Array.of(values), Array.of(ELEMENTS).map(Map::mapper));
             return values;
         }
 
         @Benchmark
         public Object java_mutable() {
-            final java.util.List<Integer> values = javaMutable.stream().map(MAPPER).collect(Collectors.toList());
-            assert areEqual(values, Array.of(ELEMENTS).map(MAPPER));
+            final java.util.List<Integer> values = javaMutable.stream().map(Map::mapper).collect(Collectors.toList());
+            assert areEqual(values, Array.of(ELEMENTS).map(Map::mapper));
             return values;
         }
 
         @Benchmark
         public Object ecollections_persistent() {
-            org.eclipse.collections.api.list.ImmutableList<Integer> values = eCollectionsPersistent.collect(MAPPER::apply);
-            assert areEqual(values, Array.of(ELEMENTS).map(MAPPER));
+            final org.eclipse.collections.api.list.ImmutableList<Integer> values = eCollectionsPersistent.collect(Map::mapper);
+            assert areEqual(values, Array.of(ELEMENTS).map(Map::mapper));
             return values;
         }
 
         @Benchmark
         public Object scala_persistent() {
             final CanBuildFrom canBuildFrom = scala.collection.immutable.Vector.canBuildFrom();
-            final scala.collection.immutable.Vector<Integer> values = (scala.collection.immutable.Vector<Integer>) scalaPersistent.map(new FromJavaFunction<>(MAPPER), canBuildFrom);
-            assert areEqual(asJavaCollection(values), Array.of(ELEMENTS).map(MAPPER));
+            final scala.collection.immutable.Vector<Integer> values = (scala.collection.immutable.Vector<Integer>) scalaPersistent.map(new FromJavaFunction<>(Map::mapper), canBuildFrom);
+            assert areEqual(asJavaCollection(values), Array.of(ELEMENTS).map(Map::mapper));
             return values;
         }
 
         @Benchmark
         public Object slang_persistent() {
-            final javaslang.collection.Vector<Integer> values = slangPersistent.map(MAPPER);
-            assert areEqual(values, Array.of(ELEMENTS).map(MAPPER));
+            final javaslang.collection.Vector<Integer> values = slangPersistent.map(Map::mapper);
+            assert areEqual(values, Array.of(ELEMENTS).map(Map::mapper));
             return values;
         }
     }
 
     public static class Filter extends Base {
-        private final Predicate<Integer> ALL = i -> true;
-        private final Predicate<Integer> SOME = i -> (i & 1) == 1;
-        private final Predicate<Integer> NONE = i -> false;
+        private static boolean isOdd(int i) { return (i & 1) == 1; }
 
         @Benchmark
-        public void java_mutable(Blackhole bh) {
-            final java.util.List<Integer> allValues = javaMutable.stream().filter(ALL).collect(Collectors.toList());
-            assert areEqual(allValues, Array.of(ELEMENTS).filter(ALL));
-            bh.consume(allValues);
-
-            final java.util.List<Integer> someValues = javaMutable.stream().filter(SOME).collect(Collectors.toList());
-            assert areEqual(someValues, Array.of(ELEMENTS).filter(SOME));
-            bh.consume(someValues);
-
-            final java.util.List<Integer> noValues = javaMutable.stream().filter(NONE).collect(Collectors.toList());
-            assert areEqual(noValues, Array.of(ELEMENTS).filter(NONE));
-            bh.consume(noValues);
+        public Object java_mutable() {
+            final java.util.List<Integer> someValues = javaMutable.stream().filter(Filter::isOdd).collect(Collectors.toList());
+            assert areEqual(someValues, Array.of(ELEMENTS).filter(Filter::isOdd));
+            return someValues;
         }
 
         @Benchmark
-        public void ecollections_persistent(Blackhole bh) {
-            org.eclipse.collections.api.list.ImmutableList<Integer> allValues = eCollectionsPersistent.select(ALL::test);
-            assert areEqual(allValues, Array.of(ELEMENTS).filter(ALL));
-            bh.consume(allValues);
-
-            org.eclipse.collections.api.list.ImmutableList<Integer> someValues = eCollectionsPersistent.select(SOME::test);
-            assert areEqual(someValues, Array.of(ELEMENTS).filter(SOME));
-            bh.consume(someValues);
-
-            org.eclipse.collections.api.list.ImmutableList<Integer> noValues = eCollectionsPersistent.select(NONE::test);
-            assert areEqual(noValues, Array.of(ELEMENTS).filter(NONE));
-            bh.consume(noValues);
+        public Object ecollections_persistent() {
+            final org.eclipse.collections.api.list.ImmutableList<Integer> someValues = eCollectionsPersistent.select(Filter::isOdd);
+            assert areEqual(someValues, Array.of(ELEMENTS).filter(Filter::isOdd));
+            return someValues;
         }
 
         @Benchmark
-        public void scala_persistent(Blackhole bh) {
-            final scala.collection.immutable.Vector<Integer> allValues = (scala.collection.immutable.Vector<Integer>) scalaPersistent.filter(new FromJavaPredicate(ALL));
-            assert areEqual(asJavaCollection(allValues), Array.of(ELEMENTS).filter(ALL));
-            bh.consume(allValues);
-
-            final scala.collection.immutable.Vector<Integer> someValues = (scala.collection.immutable.Vector<Integer>) scalaPersistent.filter(new FromJavaPredicate(SOME));
-            assert areEqual(asJavaCollection(someValues), Array.of(ELEMENTS).filter(SOME));
-            bh.consume(someValues);
-
-            final scala.collection.immutable.Vector<Integer> noValues = (scala.collection.immutable.Vector<Integer>) scalaPersistent.filter(new FromJavaPredicate(NONE));
-            assert areEqual(asJavaCollection(noValues), Array.of(ELEMENTS).filter(NONE));
-            bh.consume(noValues);
+        public Object scala_persistent() {
+            final scala.collection.immutable.Vector<Integer> someValues = (scala.collection.immutable.Vector<Integer>) scalaPersistent.filter(new FromJavaPredicate<>(Filter::isOdd));
+            assert areEqual(asJavaCollection(someValues), Array.of(ELEMENTS).filter(Filter::isOdd));
+            return someValues;
         }
 
         @Benchmark
-        public void slang_persistent(Blackhole bh) {
-            final javaslang.collection.Vector<Integer> allValues = slangPersistent.filter(ALL);
-            assert areEqual(allValues, Array.of(ELEMENTS).filter(ALL));
-            bh.consume(allValues);
-
-            final javaslang.collection.Vector<Integer> someValues = slangPersistent.filter(SOME);
-            assert areEqual(someValues, Array.of(ELEMENTS).filter(SOME));
-            bh.consume(someValues);
-
-            final javaslang.collection.Vector<Integer> noValues = slangPersistent.filter(NONE);
-            assert areEqual(noValues, Array.of(ELEMENTS).filter(NONE));
-            bh.consume(noValues);
+        public Object slang_persistent() {
+            final javaslang.collection.Vector<Integer> someValues = slangPersistent.filter(Filter::isOdd);
+            assert areEqual(someValues, Array.of(ELEMENTS).filter(Filter::isOdd));
+            return someValues;
         }
     }
 
@@ -626,6 +590,7 @@ public class VectorBenchmark {
         }
     }
 
+    /** Add all elements (one-by-one, as we're not testing bulk operations) */
     public static class Append extends Base {
         @Benchmark
         public Object java_mutable() {
@@ -641,9 +606,7 @@ public class VectorBenchmark {
         public Object fjava_persistent() {
             fj.data.Seq<Integer> values = fj.data.Seq.empty();
             for (Integer element : ELEMENTS) {
-                final fj.data.Seq<Integer> newValues = values.snoc(element);
-                assert values != newValues;
-                values = newValues;
+                values = values.snoc(element);
             }
             assert areEqual(values, javaMutable);
             return values;
@@ -665,9 +628,7 @@ public class VectorBenchmark {
         public Object pcollections_persistent() {
             org.pcollections.PVector<Integer> values = org.pcollections.TreePVector.empty();
             for (Integer element : ELEMENTS) {
-                final org.pcollections.PVector<Integer> newValues = values.plus(element);
-                assert values != newValues;
-                values = newValues;
+                values = values.plus(element);
             }
             assert areEqual(values, javaMutable);
             return values;
@@ -677,9 +638,7 @@ public class VectorBenchmark {
         public Object clojure_persistent() {
             clojure.lang.PersistentVector values = clojure.lang.PersistentVector.EMPTY;
             for (Integer element : ELEMENTS) {
-                final clojure.lang.PersistentVector newValues = values.cons(element);
-                assert values != newValues;
-                values = newValues;
+                values = values.cons(element);
             }
             assert areEqual(values, javaMutable);
             return values;
@@ -689,9 +648,7 @@ public class VectorBenchmark {
         public Object scala_persistent() {
             scala.collection.immutable.Vector<Integer> values = scala.collection.immutable.Vector$.MODULE$.empty();
             for (Integer element : ELEMENTS) {
-                final scala.collection.immutable.Vector<Integer> newValues = values.appendBack(element);
-                assert values != newValues;
-                values = newValues;
+                values = values.appendBack(element);
             }
             assert areEqual(asJavaCollection(values), javaMutable);
             return values;
@@ -701,9 +658,7 @@ public class VectorBenchmark {
         public Object slang_persistent() {
             javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.empty();
             for (Integer element : ELEMENTS) {
-                final javaslang.collection.Vector<Integer> newValues = values.append(element);
-                assert values != newValues;
-                values = newValues;
+                values = values.append(element);
             }
             assert areEqual(values, javaMutable);
             return values;
--- a/javaslang-gwt/src/test/java/client/CollectionsTestGwt.java
+++ b/javaslang-gwt/src/test/java/client/CollectionsTestGwt.java
@@ -60,7 +60,8 @@ public class CollectionsTestGwt extends GWTTestCase {
         applyCollection(chars -> TreeSet.ofAll(Iterator.ofAll(chars)));
     }
 
+    @SuppressWarnings("Convert2MethodRef")
     public void testCompileVector() {
-        applyCollection(Vector::ofAll);
+        applyCollection(chars -> Vector.ofAll(chars));
     }
 }
--- a/javaslang/src/main/java/javaslang/collection/Arrays.java
+++ b/javaslang/src/main/java/javaslang/collection/Arrays.java
@@ -16,15 +16,44 @@ final class Arrays {
     private static final Object[] EMPTY = {};
     static Object[] empty() { return EMPTY; }
 
+    @SuppressWarnings("unchecked")
+    static <T> T newInstance(int length) { return (T) copy(empty(), length); }
+
     /** Create a single element array */
     static Object[] asArray(Object element) {
-        return copyUpdate(empty(), 0, element);
+        final Object[] copy = newInstance(1);
+        copy[0] = element;
+        return copy;
+    }
+
+    /** Store the content of an iterable in an array */
+    static Object[] asArray(java.util.Iterator<?> it, int length) {
+        final Object[] array = newInstance(length);
+        for (int i = 0; i < length; i++) {
+            array[i] = it.next();
+        }
+        return array;
     }
 
     /** System.arrayCopy with same source and destination */
     static Object[] copyRange(Object array, int from, int to) {
         final int length = to - from;
         return arrayCopy(length, array, from, 0, length);
+    }
+
+    /** Repeatedly group an array into equal sized sub-trees */
+    static Object[] grouped(Object[] array, int groupSize) {
+        assert array.length > groupSize;
+        final Object[] results = newInstance(1 + ((array.length - 1) / groupSize));
+        results[0] = copyRange(array, 0, groupSize);
+
+        for (int start = groupSize, i = 1; start < array.length; i++) {
+            final int nextLength = Math.min(groupSize, array.length - (i * groupSize));
+            results[i] = copyRange(array, start, start + nextLength);
+            start += nextLength;
+        }
+
+        return results;
     }
 
     /** clone the source and set the value at the given position */
--- a/javaslang/src/main/java/javaslang/collection/Arrays.java
+++ b/javaslang/src/main/java/javaslang/collection/Arrays.java
@@ -12,7 +12,7 @@ package javaslang.collection;
  * @author Pap Lőrinc
  * @since 2.1.0
  */
-final class Arrays { // TODO reuse these in `Array` also
+final class Arrays {
     private static final Object[] EMPTY = {};
     static Object[] empty() { return EMPTY; }
 
--- a/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
+++ b/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
@@ -7,10 +7,10 @@ package javaslang.collection;
 
 import java.io.Serializable;
 import java.util.Objects;
+import java.util.function.Predicate;
 
 import static javaslang.collection.Arrays.*;
-import static javaslang.collection.NodeModifier.COPY_NODE;
-import static javaslang.collection.NodeModifier.updateLeafWith;
+import static javaslang.collection.NodeModifier.*;
 
 /**
  * A `bit-mapped trie` is a very wide and shallow tree (for integer indices the depth will be `≤6`).
@@ -121,8 +121,11 @@ final class BitMappedTrie<T> implements Serializable {
         } else if (n >= length()) {
             return empty();
         } else {
-            final Object[] root = modifyLeaf(array, depthShift, offset + n, Arrays::copyDrop, Arrays::copyDrop);
-            return new BitMappedTrie<>(root, offset + n, length() - n, depthShift);
+            final int index = offset + n;
+            final Object[] root = arePointingToSameLeaf(0, n)
+                                  ? array
+                                  : modifyLeaf(array, depthShift, index, Arrays::copyDrop, IDENTITY);
+            return collapsed(root, index, length() - n, depthShift);
         }
     }
 
@@ -132,9 +135,31 @@ final class BitMappedTrie<T> implements Serializable {
         } else if (n <= 0) {
             return empty();
         } else {
-            final Object[] root = modifyLeaf(array, depthShift, (offset + n) - 1, Arrays::copyTake, Arrays::copyTake);
-            return new BitMappedTrie<>(root, offset, n, depthShift);
+            final int index = n - 1;
+            final Object[] root = arePointingToSameLeaf(index, length() - 1)
+                                  ? array
+                                  : modifyLeaf(array, depthShift, offset + index, Arrays::copyTake, IDENTITY);
+            return collapsed(root, offset, n, depthShift);
         }
+    }
+
+    private boolean arePointingToSameLeaf(int i, int j) {
+        final boolean result = firstDigit(offset + i, BRANCHING_BASE) == firstDigit(offset + j, BRANCHING_BASE);
+        assert result == (getLeaf(i) == getLeaf(j));
+        return result;
+    }
+
+    /* drop root node while it has a single element */
+    private static <T> BitMappedTrie<T> collapsed(Object[] array, int offset, int length, int shift) {
+        for (; shift > 0; shift -= BRANCHING_BASE) {
+            final int skippedElements = array.length - 1;
+            if (skippedElements != digit(offset, shift)) {
+                break;
+            }
+            array = getAt(array, skippedElements);
+            offset -= treeSize(skippedElements, shift);
+        }
+        return new BitMappedTrie<>(array, offset, length, shift);
     }
 
     /* descend the tree from root to leaf, applying the given modifications along the way, returning the new root */
--- a/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
+++ b/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
@@ -8,6 +8,7 @@ package javaslang.collection;
 import java.io.Serializable;
 import java.util.Objects;
 import java.util.function.Function;
+import java.util.function.Predicate;
 
 import static javaslang.collection.Arrays.*;
 import static javaslang.collection.NodeModifier.*;
@@ -262,6 +263,25 @@ final class BitMappedTrie<T> implements Serializable {
         return globalIndex;
     }
 
+    BitMappedTrie<T> filter(Predicate<? super T> predicate) {
+        Objects.requireNonNull(predicate, "predicate is null");
+
+        final Object[] results = new Object[length()];
+        final int length = this.<T> visit((index, leaf, start, end) -> {
+            for (int i = start; i < end; i++) {
+                final T value = leaf[i];
+                if (predicate.test(value)) {
+                    results[index++] = value;
+                }
+            }
+            return index;
+        });
+
+        return (length() == length)
+               ? this
+               : BitMappedTrie.ofAll(copyRange(results, 0, length));
+    }
+
     <U> BitMappedTrie<U> map(Function<? super T, ? extends U> mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
 
@@ -273,7 +293,7 @@ final class BitMappedTrie<T> implements Serializable {
             return index;
         });
 
-        return BitMappedTrie.ofAll(results, length);
+        return BitMappedTrie.ofAll(results);
     }
 
     int length() { return length; }
--- a/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
+++ b/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
@@ -6,29 +6,33 @@
 package javaslang.collection;
 
 import java.io.Serializable;
+import java.util.Objects;
 
 import static javaslang.collection.Arrays.*;
-import static javaslang.collection.NodeModifier.copyLeaf;
+import static javaslang.collection.NodeModifier.COPY_NODE;
 import static javaslang.collection.NodeModifier.updateLeafWith;
 
 /**
- * A `bit-mapped trie` is a very wide and shallow tree (for integer indices the depth will be ≤ 6).
+ * A `bit-mapped trie` is a very wide and shallow tree (for integer indices the depth will be `≤6`).
  * Each node has a maximum of `32` children (configurable).
  * Access to a given position is done by converting the index to a base 32 number and using each digit to descend down the tree.
  * Modifying the tree is done similarly, but along the way the path is copied, returning a new root every time.
- * Append inserts in the last leaf, or if the tree is full from the right, it adds another layer on top of it (the old root will be the first of the new one).
- * Prepend is done similarly, but an offset is needed, because adding a new top node (where the current root would be the last node of the new root)
+ * `Append` inserts in the last leaf, or if the tree is full from the right, it adds another layer on top of it (the old root will be the first of the new one).
+ * `Prepend` is done similarly, but an offset is needed, because adding a new top node (where the current root would be the last node of the new root)
  * shifts the indices by half of the current tree's full size. The `offset` shifts them back to the correct index.
+ * `Slice` is done by trimming the path from the root and discarding any `leading`/`trailing` values in effectively constant time (without memory leak, as in `Java`/`Clojure`).
  *
  * @author Pap Lőrinc
  * @since 2.1.0
  */
 final class BitMappedTrie<T> implements Serializable {
-    static int BRANCHING_BASE = 5;
-    static int branchingFactor()                   { return 1 << BRANCHING_BASE; }
+    static final int BRANCHING_BASE = 5;
+    static final int BRANCHING_FACTOR = 1 << BRANCHING_BASE;
+    static final int BRANCHING_MASK = -1 >>> -BRANCHING_BASE;
+
     static int firstDigit(int num, int depthShift) { return num >> depthShift; }
     static int digit(int num, int depthShift)      { return lastDigit(firstDigit(num, depthShift)); }
-    static int lastDigit(int num)                  { return num & (-1 >>> -BRANCHING_BASE); }
+    static int lastDigit(int num)                  { return num & BRANCHING_MASK; }
 
     private static final long serialVersionUID = 1L;
 
@@ -40,13 +44,13 @@ final class BitMappedTrie<T> implements Serializable {
     private final int offset, length;
     private final int depthShift;
 
-    BitMappedTrie(Object[] array, int offset, int length, int depthShift) {
+    private BitMappedTrie(Object[] array, int offset, int length, int depthShift) {
         this.array = array;
         this.offset = offset;
         this.length = length;
         this.depthShift = depthShift;
 
-        assert length <= treeSize(branchingFactor(), depthShift);
+        assert length <= treeSize(BRANCHING_FACTOR, depthShift);
         assert (EMPTY == null) || ((length > 0) && (array.length > 0));
     }
 
@@ -63,13 +67,13 @@ final class BitMappedTrie<T> implements Serializable {
             Object[] array = this.array;
             int shift = depthShift, offset = this.offset;
             if (isFullLeft()) {
-                array = copyUpdate(Arrays.empty(), branchingFactor() - 1, array);
+                array = copyUpdate(Arrays.empty(), BRANCHING_FACTOR - 1, array);
                 shift += BRANCHING_BASE;
-                offset = treeSize(branchingFactor() - 1, shift);
+                offset = treeSize(BRANCHING_FACTOR - 1, shift);
             }
 
             offset -= 1;
-            array = modifyLeaf(array, shift, offset, copyLeaf, updateLeafWith(leading));
+            array = modifyLeaf(array, shift, offset, COPY_NODE, updateLeafWith(leading));
             return new BitMappedTrie<>(array, offset, newSize, shift);
         }
     }
@@ -87,14 +91,14 @@ final class BitMappedTrie<T> implements Serializable {
                 shift += BRANCHING_BASE;
             }
 
-            array = modifyLeaf(array, shift, offset + length(), copyLeaf, updateLeafWith(trailing));
+            array = modifyLeaf(array, shift, offset + length(), COPY_NODE, updateLeafWith(trailing));
             return new BitMappedTrie<>(array, offset, newSize, shift);
         }
     }
-    private boolean isFullRight(int newSize) { return (offset + newSize) > treeSize(branchingFactor(), depthShift); }
+    private boolean isFullRight(int newSize) { return (offset + newSize) > treeSize(BRANCHING_FACTOR, depthShift); }
 
     BitMappedTrie<T> update(int index, T element) {
-        final Object[] root = modifyLeaf(array, depthShift, offset + index, copyLeaf, updateLeafWith(element));
+        final Object[] root = modifyLeaf(array, depthShift, offset + index, COPY_NODE, updateLeafWith(element));
         return new BitMappedTrie<>(root, offset, length(), depthShift);
     }
 
@@ -104,8 +108,7 @@ final class BitMappedTrie<T> implements Serializable {
         } else if (n >= length()) {
             return empty();
         } else {
-            final NodeModifier drop = Arrays::copyDrop;
-            final Object[] root = modifyLeaf(array, depthShift, offset + n, drop, drop);
+            final Object[] root = modifyLeaf(array, depthShift, offset + n, Arrays::copyDrop, Arrays::copyDrop);
             return new BitMappedTrie<>(root, offset + n, length() - n, depthShift);
         }
     }
@@ -116,14 +119,13 @@ final class BitMappedTrie<T> implements Serializable {
         } else if (n <= 0) {
             return empty();
         } else {
-            final NodeModifier take = Arrays::copyTake;
-            final Object[] root = modifyLeaf(array, depthShift, (offset + n) - 1, take, take);
+            final Object[] root = modifyLeaf(array, depthShift, (offset + n) - 1, Arrays::copyTake, Arrays::copyTake);
             return new BitMappedTrie<>(root, offset, n, depthShift);
         }
     }
 
     /* descend the tree from root to leaf, applying the given modifications along the way, returning the new root */
-    private static Object[] modifyLeaf(Object[] root, int depthShift, int index, NodeModifier node, NodeModifier leaf) {
+    private Object[] modifyLeaf(Object[] root, int depthShift, int index, NodeModifier node, NodeModifier leaf) {
         if (depthShift == 0) {
             return leaf.apply(root, index);
         } else {
@@ -134,7 +136,8 @@ final class BitMappedTrie<T> implements Serializable {
             for (int shift = depthShift - BRANCHING_BASE; shift >= BRANCHING_BASE; shift -= BRANCHING_BASE) {
                 final int offset = digit(index, shift);
 
-                final Object[] newNode = node.apply(array[previousIndex], offset);
+                final Object previous = array[previousIndex];
+                final Object[] newNode = node.apply(previous, offset);
                 array[previousIndex] = newNode;
 
                 previousIndex = offset;
@@ -148,19 +151,21 @@ final class BitMappedTrie<T> implements Serializable {
 
     T get(int index) {
         final T[] leaf = getLeaf(index);
-        final int leafIndex = lastDigit(offset() + index);
+        final int leafIndex = lastDigit(offset + index);
         return leaf[leafIndex];
     }
 
     /**
      * fetch the leaf, corresponding to the given index.
      * Node: the offset and length should be taken into consideration as there may be leading and trailing garbage.
+     * Also, the returned array is mutable, but should not be mutated!
      */
     @SuppressWarnings("unchecked")
     T[] getLeaf(int index) {
+        index += offset;
         Object[] leaf = array;
         for (int shift = depthShift; shift > 0; shift -= BRANCHING_BASE) {
-            leaf = getAt(leaf, digit(offset + index, shift));
+            leaf = getAt(leaf, digit(index, shift));
         }
         return (T[]) leaf;
     }
@@ -170,7 +175,7 @@ final class BitMappedTrie<T> implements Serializable {
             final int globalLength = BitMappedTrie.this.length();
             int globalIndex = 0;
 
-            int index = lastDigit(offset());
+            int index = lastDigit(offset);
             T[] leaf = getLeaf(globalIndex);
             int length = leaf.length;
 
@@ -182,6 +187,7 @@ final class BitMappedTrie<T> implements Serializable {
                 if (index == length) { setCurrentArray(); }
 
                 final T next = leaf[index];
+                assert Objects.equals(next, BitMappedTrie.this.get(globalIndex));
 
                 index++;
                 globalIndex++;
@@ -193,12 +199,10 @@ final class BitMappedTrie<T> implements Serializable {
                 index = 0;
                 leaf = getLeaf(globalIndex);
                 length = leaf.length;
-                assert leaf != null;
             }
         };
     }
 
-    int offset() { return offset; }
     int length() { return length; }
 }
 
@@ -206,6 +210,7 @@ final class BitMappedTrie<T> implements Serializable {
 interface NodeModifier {
     Object[] apply(Object arrayObject, int index);
 
-    static NodeModifier copyLeaf = (o, i) -> copy(o, i + 1);
-    static NodeModifier updateLeafWith(Object element) { return (o, i) -> copyUpdate(o, i, element); }
+    static <T> NodeModifier updateLeafWith(T element) { return (o, i) -> copyUpdate(o, i, element); }
+    NodeModifier COPY_NODE = (o, i) -> copy(o, i + 1);
+    NodeModifier IDENTITY = (o, i) -> (Object[]) o;
 }
--- a/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
+++ b/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
@@ -59,6 +59,19 @@ final class BitMappedTrie<T> implements Serializable {
         return branchCount * fullBranchSize;
     }
 
+    static <T> BitMappedTrie<T> ofAll(Object[] array) {
+        final int size = array.length;
+        if (size == 0) {
+            return empty();
+        } else {
+            int shift = 0;
+            for (; array.length > BRANCHING_FACTOR; shift += BRANCHING_BASE) {
+                array = grouped(array, BRANCHING_FACTOR);
+            }
+            return new BitMappedTrie<>(array, 0, size, shift);
+        }
+    }
+
     BitMappedTrie<T> prepend(T leading) {
         final int newSize = length() + 1;
         if (length() == 0) {
--- a/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
+++ b/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
@@ -173,24 +173,30 @@ final class BitMappedTrie<T> implements Serializable {
      * Node: the offset and length should be taken into consideration as there may be leading and trailing garbage.
      * Also, the returned array is mutable, but should not be mutated!
      */
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({ "unchecked", "WeakerAccess" })
     T[] getLeaf(int index) {
-        index += offset;
-        Object[] leaf = array;
-        for (int shift = depthShift; shift > 0; shift -= BRANCHING_BASE) {
-            leaf = getAt(leaf, digit(index, shift));
+        if (depthShift == 0) {
+            return (T[]) array;
+        } else if (depthShift == BRANCHING_BASE) {
+            return getAt(array, firstDigit(offset + index, depthShift));
+        } else {
+            index += offset;
+            Object[] leaf = getAt(array, firstDigit(index, depthShift));
+            for (int shift = depthShift - BRANCHING_BASE; shift > 0; shift -= BRANCHING_BASE) {
+                leaf = getAt(leaf, digit(index, shift));
+            }
+            return (T[]) leaf;
         }
-        return (T[]) leaf;
     }
 
     Iterator<T> iterator() {
         return new Iterator<T>() {
-            final int globalLength = BitMappedTrie.this.length();
-            int globalIndex = 0;
+            private final int globalLength = BitMappedTrie.this.length();
+            private int globalIndex = 0;
 
-            int index = lastDigit(offset);
-            T[] leaf = getLeaf(globalIndex);
-            int length = leaf.length;
+            private int index = lastDigit(offset);
+            private T[] leaf = getLeaf(globalIndex);
+            private int length = leaf.length;
 
             @Override
             public boolean hasNext() { return globalIndex < globalLength; }
@@ -208,7 +214,7 @@ final class BitMappedTrie<T> implements Serializable {
                 return next;
             }
 
-            void setCurrentArray() {
+            private void setCurrentArray() {
                 index = 0;
                 leaf = getLeaf(globalIndex);
                 length = leaf.length;
--- a/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
+++ b/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
@@ -7,7 +7,7 @@ package javaslang.collection;
 
 import java.io.Serializable;
 import java.util.Objects;
-import java.util.function.Predicate;
+import java.util.function.Function;
 
 import static javaslang.collection.Arrays.*;
 import static javaslang.collection.NodeModifier.*;
@@ -247,6 +247,35 @@ final class BitMappedTrie<T> implements Serializable {
         };
     }
 
+    @SuppressWarnings("unchecked")
+    <T2> int visit(LeafVisitor<T2[]> visitor) {
+        int globalIndex = 0, start = lastDigit(offset);
+        for (int index = 0; index < length; ) {
+            final T2[] leaf = (T2[]) getLeaf(index);
+            final int end = Math.min(leaf.length, start + length - index);
+
+            globalIndex = visitor.visit(globalIndex, leaf, start, end);
+
+            index += end - start;
+            start = 0;
+        }
+        return globalIndex;
+    }
+
+    <U> BitMappedTrie<U> map(Function<? super T, ? extends U> mapper) {
+        Objects.requireNonNull(mapper, "mapper is null");
+
+        final Object[] results = newInstance(length);
+        this.<T> visit((index, leaf, start, end) -> {
+            for (int i = start; i < end; i++) {
+                results[index++] = mapper.apply(leaf[i]);
+            }
+            return index;
+        });
+
+        return BitMappedTrie.ofAll(results, length);
+    }
+
     int length() { return length; }
 }
 
@@ -257,4 +286,9 @@ interface NodeModifier {
     static <T> NodeModifier updateLeafWith(T element) { return (o, i) -> copyUpdate(o, i, element); }
     NodeModifier COPY_NODE = (o, i) -> copy(o, i + 1);
     NodeModifier IDENTITY = (o, i) -> (Object[]) o;
+}
+
+@FunctionalInterface
+interface LeafVisitor<T> {
+    int visit(int index, T leaf, int start, int end);
 }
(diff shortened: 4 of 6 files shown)


paplorinc:
@danieldietrich, @ruslansennov, @zsolt-donca, @eduardmanas these are the optimizations for [`Vector`](https://github.com/javaslang/javaslang/pull/1504), please dig in :)

danieldietrich:
Interesting table of measures :-))

I've removed functionaljava and PCollections from the following list because they are too slow.

Also I've reduced the not-so-common methods.

Get and Update are most important to us. Vector is a random-access (index sequence) collection and has to be as good as possible in these operations.

Head, Tail, Prepend and Append are typical linear sequence operations. These are not so interesting because users might choose a linear sequence collection (List or Stream) for that purpose.

This is what remains:

```java
Operation  Ratio                                         32     1024    32768
Create    slang_persistent/java_mutable               0.87×    0.44×      0.58×
Create    slang_persistent/scala_persistent          14.73×    8.65×      6.38×

Get       slang_persistent/java_mutable               0.86×    0.34×      0.30×
Get       slang_persistent/scala_persistent           1.26×    0.86×      0.81×

Update    slang_persistent/java_mutable               0.10×    0.05×      0.02×
Update    slang_persistent/scala_persistent           1.17×    0.98×      1.17×

Prepend   slang_persistent/java_mutable               1.47×    3.99×    100.43×
Prepend   slang_persistent/scala_persistent           1.33×    1.41×      1.44×

Append    slang_persistent/java_mutable               0.26×    0.14×      0.10×
Append    slang_persistent/scala_persistent           0.84×    0.59×      0.60×
```

Update is slow compared to Java mutable. But I think that is not a surprise. Compared to Scala we are head to head here.

Why Get and Append are relatively slower than in Scala?

danieldietrich:
Java mutable, are you drunk?

```java
Prepend   slang_persistent/java_mutable               1.47×    3.99×    100.43×
```

paplorinc:
> Java mutable, are you drunk?

This would be a case where `ArrayDeque` would shine :)
Maybe I should compare against it instead :/

paplorinc:
> Why Get and Append are relatively slower than in Scala?

Not sure yet, but I'm planning on investigating it

paplorinc:
@danieldietrich, @ruslansennov, @zsolt-donca, @eduardmanas, these are the optimizations for the new `Vector`.

Please review it [commit-by-commit](https://github.com/javaslang/javaslang/pull/1522/commits) and take a look at the commit messages to see how much faster it got.

Added `map` to the optimizations also, now it's somehow `3x` faster than `ArrayList#stream` ;)

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -42,7 +43,7 @@ public void testAsserts() {
 
     public static void main(String... args) {
         JmhRunner.runDebugWithAsserts(CLASSES);
-        JmhRunner.runNormalNoAsserts(CLASSES);
+        JmhRunner.runNormalNoAsserts(CLASSES, JAVA, FUNCTIONAL_JAVA, PCOLLECTIONS, CLOJURE, SCALA, JAVASLANG);
```
paplorinc: During optimizations I don't want to wait for `pcollections` and `fjava`, so I just delete these temporarily:
```java
JmhRunner.runNormalNoAsserts(CLASSES, SCALA, JAVASLANG);
```

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -74,10 +73,11 @@ private static int treeSize(int branchCount, int depthShift) {
         return new BitMappedTrie<>(array, 0, size, shift);
     }
 
-    BitMappedTrie<T> prepend(T leading) {
-        final int newSize = length() + 1;
+    BitMappedTrie<T> prepend(Object[] leading) {
```
paplorinc: now we only modify the tree (via append and prepend), if we have full arrays, not one-by-one

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -191,39 +192,6 @@ T get(int index) {
         }
     }
 
-    Iterator<T> iterator() {
```
paplorinc: since we have `trailing` and `leading` arrays, we move the iterator to include those also

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -124,9 +124,11 @@ private static int treeSize(int branchCount, int depthShift) {
         } else if (n >= length()) {
             return empty();
         } else {
-            final NodeModifier drop = Arrays::copyDrop;
-            final Object[] root = modifyLeaf(array, depthShift, offset + n, drop, drop);
-            return new BitMappedTrie<>(root, offset + n, length() - n, depthShift);
+            final int index = offset + n;
```
paplorinc: BMT nodes are dropped, but leafs aren't: they're masked by `offset` and `length`.
The first array was always prepended by `null` anyway, so only the last node might be bigger this way by at most `31` elements. However, since it's not copied at all, it will usually save a lot more memory.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -136,12 +138,37 @@ private static int treeSize(int branchCount, int depthShift) {
...
     }
 
+    private boolean arePointingToSameLeaf(int i, int j) {
+        final boolean result = firstDigit(offset + i, BRANCHING_BASE) == firstDigit(offset + j, BRANCHING_BASE);
+        assert result == (getLeaf(i) == getLeaf(j));
+        return result;
+    }
+
+    /* drop root node while it has a single element */
+    private BitMappedTrie<T> collapseHeight() {
```
paplorinc: if we have a tree that has unnecessary height, compact it

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -760,7 +760,7 @@ public T head() {
         if (isEmpty()) {
             throw new NoSuchElementException("head of empty Vector");
         } else {
-            return get(0);
+            return getAt(leading, leadingOffset);
```
paplorinc: the `leading` has elements always, making the `get(0)` slightly faster (~30%, compared to `get(0)`)

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -389,6 +394,41 @@ public Object slang_persistent() {
...
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values[i] = MAPPER.apply(values[i]);
+            }
+            assert areEqual(Array.of(values), Array.of(ELEMENTS).map(MAPPER));
+            return values;
+        }
+
+        @Benchmark
+        public Object java_mutable() {
+            final java.util.List<Integer> values = javaMutable.stream().map(MAPPER).collect(Collectors.toList());
```
paplorinc: this is surprisingly slow, i.e. 3x slower than `Vector`

javaslang/src/main/java/javaslang/collection/Arrays.java
```diff
@@ -71,6 +73,15 @@
         return copy;
     }
 
+    @SuppressWarnings("unchecked")
+    static <T, U> Object[] copyMap(Function<? super T, ? extends U> mapper, T[] array, int from, int to) {
+        final Object[] copy = array.clone();
+        for (int i = from; i < to; i++) {
+            copy[i] = mapper.apply((T) copy[i]);
```
paplorinc: it's faster to write where I read from, instead of creating an empty array and writing the values from the source

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -880,6 +880,50 @@ void setCurrentArray() {
         };
     }
 
+    /* TODO unify te two iterators somehow */
```
paplorinc: ... as the TODO states ...
This method will help with the primitivization iteration also, avoiding boxing

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -899,8 +943,22 @@ public int length() {
     public <U> Vector<U> map(Function<? super T, ? extends U> mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
 
-        final Iterator<? extends U> results = iterator().map(mapper);
-        return ofAll(results);
+        final AtomicReference<Object[]> newLeading = new AtomicReference<>(Arrays.empty());
+        final Object[] newMiddle = new Object[middle.leafCount()];
+        final AtomicReference<Object[]> newTrailing = new AtomicReference<>(Arrays.empty());
+        unsafeLeafIterator((leafCount, array, arrayStart, arrayLength) -> {
```
paplorinc: after the leaf iterator this impl is quite simple

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -55,10 +55,29 @@ private static int treeSize(int branchCount, int depthShift) {
         return branchCount * fullBranchSize;
     }
 
-    BitMappedTrie<T> prepend(T leading) {
-        final int newSize = length() + 1;
+    static <T> BitMappedTrie<T> ofAll(Iterable<? extends T> iterable, int size) {
+        if (size == 0) { return empty(); }
+
+        Object[] array = (iterable instanceof Collection<?>) ? ((Collection<?>) iterable).toArray()
```
ruslansennov: Let's call `BitMappedTrie.ofAll(veryLongCollection)`.
Here is the first copy loop, and in `Arrays.grouped()` is second one, without any changes of array between them.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -11,9 +11,12 @@
 
 import java.io.Serializable;
 import java.util.*;
+import java.util.concurrent.atomic.AtomicReference;
```
ruslansennov: this is GWT-incompatible

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: Sorry, I don't really understand what you're implying.
Are you suggesting an optimization?

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: K, I will introduce a Javaslang `MutableWrapper` for this then, so that we can assign values from inside a lambda.
@danieldietrich?

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
ruslansennov: Yes, I believe this can be improve performance on long source collections

paplorinc: That would be nice :)
Are you referring to the fact that `toArray` already copies everything? That's actually an optimization, since it enabled `System.arrayCopy` usage, which uses `SIMD` instructions to speed up copies.
Could you please try your idea out, I would be very happy if we could speed up things further :)

ruslansennov: OK

javaslang/src/main/java/javaslang/collection/Vector.java
danieldietrich: @paplorinc to your question above:

> K, I will introduce a Javaslang MutableWrapper for this then, so that we can assign values from inside a lambda.
@danieldietrich?

Please do not introduce the MutableWrapper. Maybe the `unsafeLeafIterator` is the wrong abstraction. I will come back with a suggestion, please hold your horses.

**Currently I need to plan the next release. This is top priority because our users have serious problems.**

paplorinc: > I will come back with a suggestion, please hold your horses.

I made some other modifications also, will push so that you can give the suggestions based on the latest codebase :).

> This is top priority because our users have serious problems.

Understood.

paplorinc:
Pushed new version.
Optimized `filter` also (as I think `map` and `filter` are used most often), it's `~2x` faster than `Java` now (finally some values that are better than Java).

@ruslansennov, I might have found what you meant by double copying, simplified the grouping algorithm to avoid it, thanks!

@danieldietrich, the `MutableWrapper` is only visible from `Vector` now :).

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -883,6 +884,16 @@ void setCurrentArray() {
         }
     }
 
+    void unsafeLeafIterator(LeafVisitor<T> visitor) {
```
paplorinc: simplified this significantly

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -429,6 +432,57 @@ public Object slang_persistent() {
         }
     }
 
+    public static class Filter extends Base {
+        private final Predicate<Integer> ALL = i -> true;
+        private final Predicate<Integer> SOME = i -> (i & 1) == 1;
+        private final Predicate<Integer> NONE = i -> false;
+
+        @Benchmark
+        public void java_mutable(Blackhole bh) {
```
paplorinc: testing `filter` is tricky: how many elements should it keep???
Therefore I added 3 variants: keeps everything, keeps around half and deletes everything.

paplorinc:
The branching base is selected to be `5` (i.e. `2⁵ = 32`), as it seems to be the best compromise between read and write operations:
```java
Operation                         Ratio  base=2  base=3  base=4  base=5  base=6  base=7
Get       java_mutable/slang_persistent   8.84×   5.36×   4.23×   3.40×   3.39×   3.32×
Update    java_mutable/slang_persistent  82.37×  57.74×  53.07×  44.47×  49.36×  77.90×
Iterate   java_mutable/slang_persistent   8.01×   7.06×   3.05×   2.74×   2.83×   2.38×
```
(it's surprising that `update` is this slow compared to the mutable counterpart, not sure we can change that, though)

danieldietrich:
@paplorinc thanks for the great overview. I like it that you not guess but measure!

Could you create the same overview for the previous/not optimized Vector version? Would be nice for comparison.

paplorinc:
There's no difference in this regard.

danieldietrich:
Update is a little faster than in Scala, I think it is fine.

I've recognized your [speed-fix](https://github.com/javaslang/javaslang/pull/1522#issuecomment-241910863) regarding `append()`. Do you have an explanation? `trailingLength` and `array.length` are the same. How can `int length = trailingLength` be slower than `int length = array.length`??

paplorinc:
I installed a new linux which will give me proper disassembly (http://psy-lob-saw.blogspot.ro/2015/07/jmh-perfasm.html), which will probably say that this way the copied array is not nulled out for some reason (i.e. if it sees `array.length` being used, it might conclude that the whole array is overridden) ... dunno yet, it's just a hunch

javaslang-benchmark/src/test/java/javaslang/BenchmarkPerformanceReporter.java
```diff
@@ -162,15 +171,15 @@ public void print() {
...
-                    alternativeImplHeader
+                                                padLeft("Target", groupSize),
+                                                padLeft("Operation", nameSize),
+                                                padLeft("Impl", implSize),
+                                                padRight("Params", paramKeySize),
+                                                padRight("Count", countSize),
+                                                padRight("Score", scoreSize),
+                                                padRight("Error", errorSize),
+                                                padRight("Unit", unitSize),
+                                                alternativeImplHeader
```
danieldietrich: Oh - I've overseen that the new formatter also adds here more whitespace. Could you please revert the formatter option in an additional PR. We want generally not add additional horizontal whitespace (like we had also before for method chaining).

It is minor for this PR. I will pull it in as-is (formatted). Once we have the new formatter I will format the whole code-base in a separate PR.

Sorry for being off-topic again :)

danieldietrich: This commit looks fine (beside formatting as stated above) - will move to the review of the next commit now...

javaslang/src/main/java/javaslang/collection/Arrays.java
```diff
@@ -36,11 +60,10 @@
     }
 
     static Object[] copy(Object arrayObject, int minLength) {
```
danieldietrich: would a `assert minLength >= 0` (or something similar) make sense?

javaslang/src/main/java/javaslang/collection/Arrays.java
```diff
@@ -36,11 +60,10 @@
...
-        final Object[] array = (Object[]) arrayObject;
-        final int arrayLength = (array == null) ? 0 : array.length;
-        final int length = Math.max(arrayLength, minLength);
-        return (length == arrayLength) ? array.clone()
-                                       : arrayCopy(length, array, 0, 0, arrayLength);
+        final Object[] array = (arrayObject == null) ? EMPTY : (Object[]) arrayObject;
+        return (array.length >= minLength)
```
danieldietrich: What if array == EMPTY? Would it make sense to just return EMPTY then instead of calling `close()` (can minLength == 0?) or `arrayCopy()`?

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -55,6 +56,24 @@ private static int treeSize(int branchCount, int depthShift) {
         return branchCount * fullBranchSize;
     }
 
+    static <T> BitMappedTrie<T> ofAll(Iterable<? extends T> iterable, int size) {
+        if (size == 0) { return empty(); }
+
+        Object[] array = (iterable instanceof Collection<?>) ? ((Collection<?>) iterable).toArray()
+                                                             : asArray(iterable.iterator(), size);
+        return ofAll(array, size, 0);
```
danieldietrich: Style-alert :) `else` branch missing (two related disjunct cases) and `final` keyword at `Object[] array`

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -55,6 +56,24 @@ private static int treeSize(int branchCount, int depthShift) {
...
+    }
+
+    static <T> BitMappedTrie<T> ofAll(Object[] array, int size, int depthShift) {
+        if (size == 0) { return empty(); }
```
danieldietrich: dito (else-branch)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -55,6 +56,24 @@ private static int treeSize(int branchCount, int depthShift) {
         return branchCount * fullBranchSize;
     }
 
+    static <T> BitMappedTrie<T> ofAll(Iterable<? extends T> iterable, int size) {
```
danieldietrich: Is always clear that Iterable isn't null? We could also do the check here and not in Vector (but don't necessarily have to...)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -55,6 +56,24 @@ private static int treeSize(int branchCount, int depthShift) {
...
+        Object[] array = (iterable instanceof Collection<?>) ? ((Collection<?>) iterable).toArray()
+                                                             : asArray(iterable.iterator(), size);
+        return ofAll(array, size, 0);
+    }
+
+    static <T> BitMappedTrie<T> ofAll(Object[] array, int size, int depthShift) {
+        if (size == 0) { return empty(); }
+
+        int shift = depthShift;
+        for (; array.length > branchingFactor(); shift += BRANCHING_BASE) {
```
danieldietrich: Because `branchingFactor()` is referential transparent and has no arguments it is effectively constant. I know that the call will be optimized away at runtime but I would write it as constant:

```java
static int BRANCHING_FACTOR = 1 << BRANCHING_BASE;
```

It isn't final because of your branching base tests, right?

We should make these fields/methods `private static final` and use [a hack](http://stackoverflow.com/questions/3301635/change-private-static-final-field-using-java-reflection) for the tests instead.

The tests could overwrite both fields, the BRANCHING_FACTOR does not need to be a method!

We should not weaken code because of unit tests!

File: javaslang/src/main/java/javaslang/collection/Arrays.java [hunk 22]
```diff
@@ -27,6 +27,30 @@
...
     }
 
+    /** Repeatedly group an array into equal sized sub-trees */
+    static Object[] grouped(Object[] array, int groupSize) {
+        assert array.length > groupSize;
+        final Object[] results = copy(empty(), 1 + ((array.length - 1) / groupSize));
+        results[0] = copyRange(array, 0, groupSize);
+
+        for (int start = groupSize, i = 1; start < array.length; i++) {
+            final int nextLength = Math.min(groupSize, array.length - (i * groupSize));
```
danieldietrich: Traversable.grouped() behaves like this:

```java
[1,2,3,4,5].grouped(2) = [[1,2],[3,4],[5]]
```

Here we would have `[[1,2],[3,4],[5,?]]` instead, right? What is the content of the last element of the last subtree? null?

javaslang/src/main/java/javaslang/collection/Arrays.java
```diff
@@ -27,6 +27,30 @@
...
+            results[i] = copyRange(array, start, start + nextLength);
+            start += nextLength;
+        }
+
+        return results;
+    }
+
+    /** Store the content of an iterable in an array */
+    static Object[] asArray(java.util.Iterator<?> it, int length) {
+        final Object[] array = copy(empty(), length);
```
danieldietrich: Why do we not use

```
final Object[] array = new Object[length];
```

?

javaslang/src/main/java/javaslang/collection/Arrays.java
```diff
@@ -27,6 +27,30 @@
...
+            final int nextLength = Math.min(groupSize, array.length - (i * groupSize));
+            results[i] = copyRange(array, start, start + nextLength);
+            start += nextLength;
+        }
+
+        return results;
+    }
+
+    /** Store the content of an iterable in an array */
+    static Object[] asArray(java.util.Iterator<?> it, int length) {
```
danieldietrich: We could change it to use `Iterable`:

```java
static Object[] asArray(java.util.Iterable<?> it, int length)
```

That saves us `iterator()` calls within our code-base, right? I can't find here more uses of this method but maybe there exist already other uses where we only have an Iterator at hand instead of an Iterable - just a question.

javaslang/src/main/java/javaslang/collection/Arrays.java
paplorinc: Here we could, since it's already boxed, but wanted the allocations to be in a single place ...
Will change it :)

javaslang/src/main/java/javaslang/collection/Arrays.java
paplorinc: it should be the same as for `Traversable`'s `groupes`, though it's applied recursively until the root array is small enough (<=32)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -164,14 +164,21 @@ private Vector(BitMappedTrie<T> trie) {
...
-            for (T element : elements) {
-                trie = trie.append(element);
-            }
-            return ofAll(trie);
+            final Seq<? extends T> seq = seq(elements);
+            return ofAll(seq, seq.size());
         }
     }
 
+    public static <T> Vector<T> ofAll(Collection<? extends T> elements) {
```
danieldietrich: Do we really need that extra method? Why use the `ofAll(Iterable)` method and check `instanceof Collection` in order to call `size()`? Would it really be significant slower? Other collections do not have that `ofAll` method.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -164,14 +164,21 @@ private Vector(BitMappedTrie<T> trie) {
...
+        Objects.requireNonNull(elements, "elements is null");
+        return ofAll(elements, elements.size());
+    }
+
+    public static <T> Vector<T> ofAll(Iterable<? extends T> iterable, int size) {
```
danieldietrich: Why do we need this public method? The user should not be bothered providing a size, right? If it is needed internally I would reduce the visibility.

danieldietrich: What if size < 0?

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: Why do you think a constant method is weakening?
I need `BRANCHING_BASE` to be changeable because it's very difficult to test and debug a tree that huge.

Also, constants could be inlined in other places (e.g. maybe a `2 * BRANCHING_FACTOR` would be inlined) and changing the final variable from the test wouldn't change this particular usage.

I would prefer keeping it a method, or making the `BRANCHING_BASE` for the tests also (would only be able to test 2-3 levels that way)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: I thought of this as a guard statement, but I don't mind `else`-ing it either :)

javaslang/src/main/java/javaslang/collection/Arrays.java
danieldietrich: Yep, I've confused min with max - you are right :)

javaslang-benchmark/src/test/java/javaslang/BenchmarkPerformanceReporter.java
paplorinc: I couldn't anticipate every benchmark modification need, so every PR contained some modifications here also.
Now that it's mostly done, I will separate the unrelated benchmarking modifications to a separate PR :)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
danieldietrich: Ok, your call.

Mutable static fields + concurrency = 💣 Are we sure our benchmarks / tests are correct?

javaslang-benchmark/src/test/java/javaslang/BenchmarkPerformanceReporter.java
danieldietrich: Ok. But removing whitespace again (aligning newlines do `(`) would be cool.

paplorinc: I reimported the settings, the formattings seems ok now in the separated PR:
* https://github.com/javaslang/javaslang/pull/1558

Will rebase this, once that's merged.
Thanks for the comments, I don't mind these minor ones either :)

danieldietrich: Ok. Are you sure the settings are the most recent or might they be old, according to that version of the branch / rebased changes?

paplorinc: Not sure what you mean, please check the separated PR if you think the formatting is ok there: https://github.com/javaslang/javaslang/pull/1558/files

danieldietrich: Yes, looks good to me. Asking myself why my local settings are different... Will re-import the settings then also.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -783,7 +824,51 @@ public boolean isTraversableAgain() {
 
     @Override
     public Iterator<T> iterator() {
-        return isEmpty() ? Iterator.empty() : trie.iterator();
+        return isEmpty() ? Iterator.empty()
+                         : new VectorIterator();
+    }
+
```
danieldietrich: My main pain point in previous discussions was here that we leak BMT internals into the Vector impl.

How about leaving the BMT.iterator() method as is and implementing Vector.iterator() like that:

```java
@Override
public Iterator<T> iterator() {
    return (Iterator<T>) Iterator.concat(
            Iterator.of(leading),
            trie.iterator(),
            Iterator.of(trailing)
    );
}
```

_(Did not test it)_

paplorinc:
Rebased, will apply the rest of the comments shortly :)
Keep them coming!

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: Thanks, but I did test it and it was surprisingly slow.
Will try to come up with something better, if you don't like it.

danieldietrich: > Thanks, but I did test it and it was surprisingly slow.

Really?

There might be a little overhead for creating 4 Iterators instead of 1 but this is what I expect:

For Iterating a small amount of elements the constant factor of creating 4 Iterators will be weighted much.

I.e. when Iteration takes 8 ms for one Iterator and 16 ms using 4 Iterators the _relative_ diff looks huge, but _really_ it isn't!

I.e. when Iteration takes 10008 ms for one Iterator and 100016 ms using 4 Iterators the _relative_ diff is near zero (+/- measure error).

---

Because the BMT.iterator() is nearly perfect and also the Object[] iterators are very efficient loops, the diff can't be huge. Our test cases are not right (interpreted).

danieldietrich: Will go to sleep now and go on tomorrow - first 'working' day after 4 weeks vacation :) I'm mostly offline during the day. My mobile phone is broken (I wait until the release of iPhone 7), also I have no Internet (for private use) at $work.

Great work so far!!!

paplorinc: > Our test cases are not right (interpreted).

I'm surprised you think so. Feel free to investigate, I would be happy to find a better solution :)

danieldietrich: I know this sounds weird but it is the only explanation. Creating the Iterators is constant effort. Their implementation is as efficient as the solution with one Iterator. Beside Iterator creation there could be only one more reason - the concat Iterator is implemented inefficient. But it is only a simple iteration and switch of existing Iterators. That can't be the bottleneck.

We should investigate it. It is a good opportunity to get my hands dirty regarding the benchmarks.

javaslang/src/main/java/javaslang/collection/Arrays.java
paplorinc: empty can be called with `minLength` 10 also, resulting in a simple allocation, no copying.
This might be counter-intuitive, but allocation has to be in a single place, not spread all over the codebase.

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: Good catch, thanks :)

javaslang/src/main/java/javaslang/collection/Vector.java
danieldietrich: I've thought about it. The time-penalty of the simple, combined Iterator may come from the delegated Iterator calls:

* first the combined Iterator is called
* then the combined Iterator delegates to the current underlying Iterator

When I get some time (hopefully tomorrow) I will benchmark it, comparing a simple Iterator and a combined Iterator that delegates to a simple Iterator.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -116,14 +146,39 @@ private static int treeSize(int branchCount, int depthShift) {
...
+            array = getAt(array, skippedElements);
+            offset -= treeSize(skippedElements, shift);
+        }
+        return (shift == depthShift) ? this
+                                     : new BitMappedTrie<>(array, offset, length(), shift);
+    }
+
     /* descend the tree from root to leaf, applying the given modifications along the way, returning the new root */
-    private static Object[] modifyLeaf(Object[] root, int depthShift, int index, NodeModifier node, NodeModifier leaf) {
+    private Object[] modifyLeaf(Object[] root, int depthShift, int index, NodeModifier node, NodeModifier leaf) {
```
danieldietrich: It could stay static. Is there a reason to make it an instance method? Then implicit `this` is also passed...

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -155,57 +211,35 @@ T get(int index) {
...
 @FunctionalInterface
 interface NodeModifier {
     Object[] apply(Object arrayObject, int index);
 
-    static NodeModifier copyLeaf = (o, i) -> copy(o, i + 1);
-    static NodeModifier updateLeafWith(Object element) { return (o, i) -> copyUpdate(o, i, element); }
+    static <T> NodeModifier updateLeafWith(T element)  { return (o, i) -> copyUpdate(o, i, element); }
+    static NodeModifier replaceLeafWith(Object[] leaf) { return (o, i) -> leaf; }
+    NodeModifier copyNode = (o, i) -> copy(o, i + 1);
+    NodeModifier Identity = (o, i) -> (Object[]) o;
```
danieldietrich: These interface fields are implicitly `static final`, i.e. constants. By naming convention the names have to be upper-case:

```java
NodeModifier COPY_NODE = (o, i) -> copy(o, i + 1);
NodeModifier IDENTITY = (o, i) -> (Object[]) o;
```

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -104,9 +132,11 @@ private static int treeSize(int branchCount, int depthShift) {
...
             return empty();
         } else {
-            final NodeModifier drop = Arrays::copyDrop;
-            final Object[] root = modifyLeaf(array, depthShift, offset + n, drop, drop);
-            return new BitMappedTrie<>(root, offset + n, length() - n, depthShift);
+            final int index = offset + n;
+            final Object[] root = arePointingToSameLeaf(0, n)
+                                  ? array
+                                  : modifyLeaf(array, depthShift, index, Arrays::copyDrop, Identity);
+            return new BitMappedTrie<T>(root, index, length() - n, depthShift).collapseHeight();
```
danieldietrich: `collapseHeight()` is always called directly after constructing the BMT - which is a smell :)

We can save one BMT instance (in the worst case) by changing `collapseHeight()` to be a static factory method:

```java
/* drop root node while it has a single element */
private static BitMappedTrie<T> collapseHeight(Object[] array, int offset, int length, int shift) {
    for (; shift > 0; shift -= BRANCHING_BASE) {
        final int skippedElements = array.length - 1;
        if (skippedElements != digit(offset, shift)) {
            break;
        }
        array = getAt(array, skippedElements);
        offset -= treeSize(skippedElements, shift);
    }
    return new BitMappedTrie<>(array, offset, length, shift);
}
```

Then we replace

```java
return new BitMappedTrie<T>(root, index, length() - n, depthShift).collapseHeight();
```

by

```java
return collapseHeight(root, index, length() - n, depthShift);
```

Maybe we should rename `collapseHeight(...)` then to `of(...)`? But not necessarily.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -855,6 +854,18 @@ void setCurrentArray() {
         }
     }
 
+    @SuppressWarnings("unchecked")
+    boolean leafVisitor(LeafVisitor<T> visitor) {
+        final VectorIterator it = new VectorIterator();
+        for (int ordinal = 0; it.hasNext(); ordinal++) {
+            visitor.visit(ordinal, it.leaf, it.index, it.length);
+
+            it.globalIndex += it.length - it.index;
```
danieldietrich: this is evil - modifying the VectorIterator from the outside!?

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -872,8 +883,22 @@ public int lastIndexOf(T element, int end) {
     public <U> Vector<U> map(Function<? super T, ? extends U> mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
 
-        final Iterator<? extends U> results = iterator().map(mapper);
-        return ofAll(results);
+        final MutableWrapper<Object[]> newLeading = new MutableWrapper<>(Arrays.empty());
```
danieldietrich: MutableWrapper is unnecessary overhead. Instead we should remove the `final` keyword and delete MutableWrapper.

danieldietrich:
Oh Lorinc, please sit down, this one will be hard.

After investigating your code for some hours I came to a decision. I share Ruslan's opinion that the 'leading' & 'trailing' optimization significantly increases the complexity of the implementation.

I want to pull the first 3 commits (Refactored 'Vector', Optimized 'ofAll', Optimized 'Interator'). I can't pull the other optimizations (append/prepend, drop/take, head, map, filter) of this PR because they depend on the 'leading' & 'trailing' optimization.

The fact that implementation details of the BitMappedTrie leak into Vector is also a no-go but the increasing complexity of the implementation is the main pain point.

You've put so much time & effort into your optimizations that I really struggled with myself - I thought I have no other choice than pulling all changes into the repository. But to be honest, I have more than a bad feeling regarding the maintenance.

For the health of Javaslang I decide for maintainability over performance. But if we take a closer look - do we really decide agaist performance?

**BEFORE:**

```java
    @Override
    public T get(int index) {
            throw new IndexOutOfBoundsException("get(" + index + ")");
        } else {
            return trie.get(index);
        }
    }
```

**AFTER:**

```java
    @Override
    public T get(int index) {
            throw new IndexOutOfBoundsException("get(" + index + ")");
        } else {
            if (index < leadingLength()) {
                return getAt(leading, leadingOffset + index);
            } else {
                index -= leadingLength();
                if (index < middle.length()) {
                    return middle.get(index);
                } else {
                    return getAt(trailing, index - middle.length());
                }
            }
        }
    }
```

Users expect **Vector** mainly to be as good as possible in `get()` and `update()`.

From the [Scala documentation](http://docs.scala-lang.org/overviews/collections/concrete-immutable-collection-classes.html):

> "(...) vectors strike a good balance between fast random selections and fast random functional updates (...)"

In my opinion we optimize the wrong methods. Vector is an 'IndexedSeq' with accentuation on _indexed_. `append`, `prepend` and `head` are typical for `LinearSeq` implementations.

Again, from the Scala docs:

> Lists are very efficient when the algorithm processing them is careful to only process their heads. Accessing, adding, and removing the head of a list takes only constant time (...)

I think the right place for optimized versions of `map` and `filter` is BitMappedTrie`. We can't leak implementations details to the outside of BMT.

I *really* hope you understand my points. Your work is really awesome.

paplorinc:
Thanks for the comments @danieldietrich, let's see how it affects the codebase and the benchmarks of `prepend` and `append`.

danieldietrich:
I'm very interested to take all possibilities into account. Do you have time for a hangout during this week?

paplorinc:
Sure, is @zsolt-donca also interested?

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: I agree, will think of a better solution, thanks.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: I tried so many alternatives here (including the one you suggested), I didn't even know which one was committed in the end. This is why I needed an extra pair of eyes, thanks :).

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: Will need parent instance in the future, i.e. it's conceptually part of the instance.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
danieldietrich: Lorinc, could you please explain to me the internal meaning of `compactHeight()`. I understand that unnecessary parts of the BMT are released. But which modification of the BMT made it necessary? Does it make sense for the initial/minimal BMT also or do we need it because we introduced some sort of change to the BMT?

Thanks for 'sharpening me up' (in german: 'Aufschärfen') ;-)

paplorinc: Valid question (and awesome new `GitHub` changes).
`compactHeight` is needed in cases when e.g. you have a tree of height `5` and you `drop` most of it, that could fit on a tree of e.g. height `2`. In that case, the root will have a single child, and its child will also have a single child. Maybe even the next one, but that cannot be guaranteed (imagine a `1`, prepended by `0` ... you will have two levels, though it could fit on a single one ... note also, that this doesn't escalate, the biggest difference because of these is one level difference).
In this case (i.e. after `drop` and `take`) there's just no point in keeping all those extra depth, hence the compactor :).

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -443,111 +441,77 @@ public Object slang_persistent() {
     }
 
     public static class Map extends Base {
-        private final Function<Integer, Integer> MAPPER = i -> i + 1;
+        private static int mapper(int i) { return i + 1; }
```
paplorinc: instead of defining a `Functon` and calling `::apply`, where the method expects a different `SAM` (or inlining it, violating `SRP`), a mapper function does both :)

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -443,111 +441,77 @@ public Object slang_persistent() {
...
+            assert areEqual(values, Array.of(ELEMENTS).map(Map::mapper));
             return values;
         }
     }
 
     public static class Filter extends Base {
-        private final Predicate<Integer> ALL = i -> true;
-        private final Predicate<Integer> SOME = i -> (i & 1) == 1;
-        private final Predicate<Integer> NONE = i -> false;
+        private static boolean isOdd(int i) { return (i & 1) == 1; }
```
paplorinc: same with filter

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -443,111 +441,77 @@ public Object slang_persistent() {
...
-
-            final java.util.List<Integer> someValues = javaMutable.stream().filter(SOME).collect(Collectors.toList());
-            assert areEqual(someValues, Array.of(ELEMENTS).filter(SOME));
-            bh.consume(someValues);
-
-            final java.util.List<Integer> noValues = javaMutable.stream().filter(NONE).collect(Collectors.toList());
-            assert areEqual(noValues, Array.of(ELEMENTS).filter(NONE));
-            bh.consume(noValues);
+        public Object java_mutable() {
+            final java.util.List<Integer> someValues = javaMutable.stream().filter(Filter::isOdd).collect(Collectors.toList());
```
paplorinc: simplified `filter`, based on the recommendation of @RichardWarburton. Thanks.

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -629,6 +593,7 @@ public Object slang_persistent() {
         }
     }
 
+    /** Add all elements (one-by-one, as we're not testing bulk operations) */
```
paplorinc: @RichardWarburton, we're using loops here, as we're simulating rebuilding the given data structure, one-by-one.
We're interested in the amortized cost of the operations across different edge cases (e.g. where append needs to add a new level to the tree).
If we could use bulk operations, we could simply return the already created instance.
Does it make sense or do you recommend a better approach :)?

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -6,48 +6,52 @@
...
  * shifts the indices by half of the current tree's full size. The `offset` shifts them back to the correct index.
+ * `Slice` is done by trimming the path from the root and discarding any `leading`/`trailing` values in effectively constant time (without memory leak, as in `Java`/`Clojure`).
  *
  * @author Pap Lőrinc
  * @since 2.1.0
  */
 final class BitMappedTrie<T> implements Serializable {
-    static int BRANCHING_BASE = 5;
-    static int branchingFactor()                   { return 1 << BRANCHING_BASE; }
+    static final int BRANCHING_BASE = 5;
```
paplorinc: let's extract these to constants, we don't need to change them programatically from tests anymore :)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -6,48 +6,52 @@
...
     @SuppressWarnings("unchecked")
-    static <T> BitMappedTrie<T> empty() { return (BitMappedTrie<T>) EMPTY; }
+    static <T> BitMappedTrie<T> empty() { return (BitMappedTrie<T>) EMPTY_BMT; }
 
     private final Object[] array;
     private final int offset, length;
     private final int depthShift;
 
-    BitMappedTrie(Object[] array, int offset, int length, int depthShift) {
+    private BitMappedTrie(Object[] array, int offset, int length, int depthShift) {
```
paplorinc: @danieldietrich, no more `BMT` leaks into `Vector` :)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -190,7 +186,7 @@ private Vector(BitMappedTrie<T> trie) {
      * @param array a boolean array
      * @return A new Vector of Boolean values
      */
-    public static Vector<Boolean> ofAll(boolean[] array) {
+    public static Vector<Boolean> ofAll(boolean... array) {
```
paplorinc: @danieldietrich, is this ok?

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -159,12 +160,13 @@ private Vector(BitMappedTrie<T> trie) {
         Objects.requireNonNull(iterable, "iterable is null");
         if (iterable instanceof Vector) {
             return (Vector<T>) iterable;
+        } else if (iterable instanceof Collection<?>) {
```
paplorinc: @danieldietrich, good advice for unifying these :)

javaslang/src/main/java/javaslang/collection/Arrays.java
```diff
@@ -16,16 +16,19 @@
     private static final Object[] EMPTY = {};
     static Object[] empty() { return EMPTY; }
 
+    @SuppressWarnings("unchecked")
+    static <T> T newInstance(int length) { return (T) copy(empty(), length); }
+
     /** Create a single element array */
     static Object[] asArray(Object element) {
-        final Object[] copy = copy(empty(), 1);
+        final Object[] copy = newInstance(1);
```
paplorinc: @danieldietrich, unified these also, as per your suggestion

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -246,6 +246,34 @@ void setCurrentArray() {
...
+            final int end = Math.min(leaf.length, start + length - index);
+
+            visitor.visit(leaf, start, end);
+
+            index += end - start;
+            start = 0;
+        }
+    }
+
+    <U> BitMappedTrie<U> map(Function<? super T, ? extends U> mapper) {
```
paplorinc: could be optimized to take advantage of the fast that map keeps the original structure (i.e. almost `2x` speedup), but would result in slightly more complicated code.
@danieldietrich?

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -260,6 +261,27 @@ void setCurrentArray() {
         }
     }
 
+    BitMappedTrie<T> filter(Predicate<? super T> predicate) {
+        Objects.requireNonNull(predicate, "predicate is null");
+
+        final Object[] results = new Object[length()];
+        final MutableInt index = new MutableInt();
+        this.<T> visit((leaf, start, end) -> {
+            int resultIndex = index.val;
```
paplorinc: mutating a local variable and writing it back to `MutableInt` at the end makes it `~5%` faster

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -131,11 +134,33 @@ private static int treeSize(int branchCount, int depthShift) {
...
     }
 
+    private boolean arePointingToSameLeaf(int i, int j) {
+        final boolean result = firstDigit(offset + i, BRANCHING_BASE) == firstDigit(offset + j, BRANCHING_BASE);
+        assert result == (getLeaf(i) == getLeaf(j));
+        return result;
+    }
+
+    /* drop root node while it has a single element */
+    private static <T> BitMappedTrie<T> collapsed(Object[] array, int offset, int length, int shift) {
```
paplorinc: we could also drop leading `null`s, but that wouldn't result in any speed advantage, just aesthetics...

paplorinc:
@danieldietrich, @zsolt-donca I pushed a new version with the applied recommendations.
The code is **A LOT** more readable now, simplified every optimization as much as I could, though that meant that `map` is `2x` slower (still faster than `java`/`scala`), `prepend` is almost `10x` slower (`6x` slower than `Scala`) and `append` is `~5x` slower than before.
We could apply those optimizations at a later stage, if needed.
Some methods have become slightly faster also, e.g. `get`, `update` and `iterate`.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -148,29 +189,37 @@ private static int treeSize(int branchCount, int depthShift) {
...
         return leaf[leafIndex];
     }
 
     /**
      * fetch the leaf, corresponding to the given index.
      * Node: the offset and length should be taken into consideration as there may be leading and trailing garbage.
+     * Also, the returned array is mutable, but should not be mutated!
      */
     @SuppressWarnings("unchecked")
     T[] getLeaf(int index) {
```
paplorinc: this is also `private` now, but since it's accessed from `iterator`, this way we can avoid the bridge

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: Could also unroll the middle manually, resulting in a measurable change, but distorting the code significantly, i.e.
```java
int i = start;
for (; i < end - 3; i += 4) {
    final int val = index.val;
    elems[val] = mapper.apply(leaf[i]);
    elems[val + 1] = mapper.apply(leaf[i + 1]);
    elems[val + 2] = mapper.apply(leaf[i + 2]);
    elems[val + 3] = mapper.apply(leaf[i + 3]);
    index.val = val + 4;
}
for (; i < end; i++) {
    elems[index.val++] = mapper.apply(leaf[i]);
}
```
before:
```java
Map slang_persistent/scala_persistent 1.26×  1.53×  1.88×
```

after:
```java
Map slang_persistent/scala_persistent 1.40×  1.54×  1.92×
```

i.e. not worth it

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
danieldietrich: looks good, also we save the boxing of ints here

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
danieldietrich: 👍

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -6,48 +6,52 @@
...
+
     static int firstDigit(int num, int depthShift) { return num >> depthShift; }
     static int digit(int num, int depthShift)      { return lastDigit(firstDigit(num, depthShift)); }
-    static int lastDigit(int num)                  { return num & (-1 >>> -BRANCHING_BASE); }
+    static int lastDigit(int num)                  { return num & BRANCHING_MASK; }
 
     private static final long serialVersionUID = 1L;
 
-    private static final BitMappedTrie<?> EMPTY = new BitMappedTrie<>(Arrays.empty(), 0, 0, 0);
+    private static final BitMappedTrie<?> EMPTY_BMT = new BitMappedTrie<>(Arrays.empty(), 0, 0, 0);
```
danieldietrich: Minor: I would it name it `EMPTY` to be consistent with all other Javaslang types that have an EMPTY instance

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -89,9 +87,7 @@ private Vector(BitMappedTrie<T> trie) {
      * @return the given {@code vector} instance as narrowed type {@code Vector<T>}.
      */
     @SuppressWarnings("unchecked")
-    public static <T> Vector<T> narrow(Vector<? extends T> vector) {
-        return (Vector<T>) vector;
-    }
+    public static <T> Vector<T> narrow(Vector<? extends T> vector) { return (Vector<T>) vector; }
```
danieldietrich: Methods should be no on-liners (for readability)

javaslang/src/main/java/javaslang/collection/Vector.java
danieldietrich: I tested it with Stream (short version):

```java
interface Stream<T> {

    static Stream<Integer> ofAll(int... array) {
        return Stream.ofAll(Iterator.ofAll(array));
    }

    static Stream<Long> ofAll(long... array) {
        return Stream.ofAll(Iterator.ofAll(array));
    }
}
```

The test:

```java
public class Test {{

    // ok
    Stream<Integer> ints = Stream.ofAll();

    // IntelliJ IDEA complains: Required: Stream<Long>, Found: Stream<Integer>
    Stream<Long> longs = Stream.ofAll();

    // does also not work
    Stream<Long> longs = Stream.<Long> ofAll();
}}
```

However, unexpected behavior with varargs can only occur when we have reference types, e.g.

```java
// reference type T
public <T> void foo(T... params) { ... }

int[] arr = {1, 2, 3};

foo(arr); // passes an int[][] array containing a single int[] element
```

See also [this StackOverflow question](http://stackoverflow.com/questions/5405673/java-varags-method-param-list-vs-array).

I see no problem to switch to varargs for primitive types. There is no risk that the array is confused with a single object.

We should apply that change then to all other Javaslang types that have static factory methods `ofAll(<primitive array>)` - of course in another PR.

That's a nice API change, simplifying things! :-)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -584,24 +580,16 @@ private Vector(BitMappedTrie<T> trie) {
     }
 
     @Override
-    public Vector<Vector<T>> combinations() {
-        return rangeClosed(0, length()).map(this::combinations).flatMap(Function.identity());
-    }
+    public Vector<Vector<T>> combinations() { return rangeClosed(0, length()).map(this::combinations).flatMap(Function.identity()); }
```
danieldietrich: Just a question :) Did you change your local formatter? Now we have many single-line methods. It does not really hurt but I think you may have a reason to reformat. Generally ok for me!

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: Indeed, now I can rename it back. Previously it was used from `Vector`, which had its own `EMPTY`

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: I formatted these manually, the formatter should keep them intact (mine does).
I wanted the formatting to reflect how it would look in e.g. `Scala` or in a Java 8 lambda: very simple methods in the same line, complex ones on multiple, to increase SNR.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
danieldietrich: I think we do not need `MutableInt`. Instead we could add a `visit` method that takes a param `index` and returns the updated `index`:

```java
@FunctionalInterface
interface LeafVisitor<T> {
    int visit(T leaf, int start, int end, int index);
}
```

The `visit` method then looks like this:

```java
    <T2> int visit(LeafVisitor<T2[]> visitor) {
        int start = lastDigit(offset);
        int currentIndex = 0;
        for (int index = 0; index < length; ) {
            final T2[] leaf = (T2[]) getLeaf(index);
            final int end = Math.min(leaf.length, start + length - index);

            currentIndex = visitor.visit(leaf, start, end, currentIndex);

            index += end - start;
            start = 0;
        }
        return currentIndex;
    }
```

BEFORE:

```java
        ...
        final MutableInt index = new MutableInt();
        this.<T> visit((leaf, start, end) -> {
            int resultIndex = index.val;
            for (int i = start; i < end; i++) {
                final T value = leaf[i];
                if (predicate.test(value)) {
                    results[resultIndex++] = value;
                }
            }
            index.val = resultIndex;
        });

        return (length() == index.val) ? ...
```

AFTER:

```java
        ...
        final int resultLength = this.<T> visit((leaf, start, end, index) -> {
            for (int i = start; i < end; i++) {
                final T value = leaf[i];
                if (predicate.test(value)) {
                    results[index++] = value;
                }
            }
            return index;
        });

        return (length() == resultLength) ? ...
```

Note: Similar changes could be applied to all other methods that use `visit`.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
danieldietrich: Please explain to me how nulls can occur. I have questions:

* Does array contain these nulls?
* What is the meaning of these nulls?
* Is more and more memory consumed over time be nulled array slots?

The BMT should only contain valid nodes.

Before changing the implementation please explain to me how the tree structure changes over time.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
danieldietrich: Yep, `EMPTY` is better.

danieldietrich:
@paplorinc 

> I pushed a new version with the applied recommendations.

Many thanks for doing it, I really appreciate it! As you said - further optimizations can be applied later.

Let's finalize this PR. I will still take a look. Maybe we need another iteration...

Thx!

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
danieldietrich: I also think we do not need the further improvement (for now).

With the new visitor, the `map` method looks like this:

```java
    <U> BitMappedTrie<U> map(Function<? super T, ? extends U> mapper) {
        final Object[] results = new Object[length()];
        this.<T> visit((leaf, start, end, index) -> {
            for (int i = start; i < end; i++) {
                elems[index++] = mapper.apply(leaf[i]);
            }
            return index;
        });
        return BitMappedTrie.ofAll(elems, length);
    }
```

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: Since the tree is conceptually not completely symmetric (i.e. a `prepend` changes existing indices, an `append` doesn't), we don't need to store every array as if it were a full array, i.e. if we have only `20` elements, the array can be of length < `32`.
If you have a full array and drop `10` from the end, we can create a new array of size `22`.
If however we drop from the beginning, it's easier to make a full array, and only copy the kept elements (and store the difference in the offset), having the effect of `prepend` with `null`s.
If we would want to eliminate all leading `null`s, the computations would be more difficult (though I can try again, if this constant overhead (that's applicable only to non-leaf nodes) bothers you).

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: I hesitated at first, but trying it out seems indeed like a good idea, thanks :)

paplorinc: The benchmark speeds are basically the same, pushed a new version.
Did I address all your concerns?

danieldietrich:
(Note: I will focus now on this PR again - starting tomorrow)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
danieldietrich: it is fine as is!

danieldietrich:
Hi @paplorinc, now I can't merge because of the conflicts. My fault - I waited too long...
Could you please merge/resolve the conflicts?

paplorinc:
@ruslansennov, the build seems to be failing because of GWT incompatibility (https://travis-ci.org/javaslang/javaslang/builds/169480221), could you please help (I presume I need some annotations somewhere)

ruslansennov:
Hi @paplorinc 
Right now I can't help you because I'm very busy until Sunday evening.
Most GWT-related problems can be solved by removing (or marking by `@GwtIncompatible` annotation) all reflection stuff like `isAssignable` method etc

paplorinc:
Ok, but how do I find out which method is causing the problem (I can't run the GWT tests locally)?

ruslansennov:
Now I see that error log shows not enough info
```
[INFO] java.lang.ClassCastException
[INFO] 	at java.lang.Throwable.Throwable(Throwable.java:61)
[INFO] 	at java.lang.Exception.Exception(Exception.java:25)
[INFO] 	at java.lang.RuntimeException.RuntimeException(RuntimeException.java:25)
[INFO] 	at java.lang.ClassCastException.ClassCastException(ClassCastException.java:23)
[INFO] 	at javaemul.internal.InternalPreconditions.checkCriticalType(InternalPreconditions.java:141)
[INFO] 	at com.google.gwt.lang.Cast.castTo(InternalPreconditions.java:129)
[INFO] 	at client.CollectionsTestGwt$6methodref$ofAll$Type.$apply(Character.java:485)
[INFO] 	at client.CollectionsTestGwt.testCompileVector(CollectionsTestGwt.java:23)
[INFO] 	at Unknown.anonymous(GWTTestMetadataImpl.java:14)
```

> I can't run the GWT tests locally

Why? Just run `mvn test` and you will see results of JDT tests **and** GWT tests

paplorinc:
Cannot debug it from Idea (e.g. `debug testCompileVector`), and maven gives the same error as on server (class cast exception).

paplorinc:
@ruslansennov, it seems to be a GWT bug, i.e. apparently method references don't convert arrays to varargs: https://github.com/javaslang/javaslang/pull/1522/commits/5102b4d8268b443f1bc37b8b226c67088e77d1ae

javaslang-gwt/src/test/java/client/CollectionsTestGwt.java
```diff
@@ -60,7 +60,8 @@ public void testCompileTreeSet() {
         applyCollection(chars -> TreeSet.ofAll(Iterator.ofAll(chars)));
     }
 
+    @SuppressWarnings("Convert2MethodRef")
     public void testCompileVector() {
-        applyCollection(Vector::ofAll);
+        applyCollection(chars -> Vector.ofAll(chars));
```
danieldietrich: nice finding!

danieldietrich:
Great, thank you!!!

javaslang-gwt/src/test/java/client/CollectionsTestGwt.java
paplorinc: @ruslansennov, should we open a bug report for this?

danieldietrich: At least there are chances that it will be fixed near-time. Our GWT dependency is still an RC (release candidate).

Issue #1524, opened 2016-08-23
Extend `Vector` to implement `Queue` and `Stack`, similarly to Java's `ArrayDeque`


paplorinc:
Since `Vector` has symmetrical `prepend`-`append` and `head`-`last` operations, it's well suited for an `ArrayDeque` alternative.

The only required change is to add a `peekFirst` and `peekLast` operations to the corresponding interfaces -- otherwise `peek` would be ambiguous.
It's how the `Deque` interface does it also: 
http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/java/util/Deque.java#Deque.peekFirst%28%29

I personally would unify the two interfaces slightly more: Java's [`Deque`](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/java/util/Deque.java) would be a good place to start.
See also: https://github.com/javaslang/javaslang/issues/1523

A draft PR has been provided for convenience: https://github.com/javaslang/javaslang/pull/1525

danieldietrich:
This would mix up our collections too much and confuse our users.

Our design goal is: choose the right collection for the right job. The collections are designed to do one thing best. There already is a Queue and List as Stack (where a linked list is a stack by definition, therefore Stack is not a Traversable).

We should close this issue.

PR #1525, opened 2016-08-23
Extend `Vector` to implement `Queue` and `Stack`, similarly to Java's `ArrayDeque`

paplorinc:
See: https://github.com/javaslang/javaslang/issues/1524

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/AbstractsQueue.java
+++ b/javaslang/src/main/java/javaslang/collection/AbstractsQueue.java
@@ -6,11 +6,14 @@
 
 package javaslang.collection;
 
-import javaslang.*;
+import javaslang.Tuple;
+import javaslang.Tuple2;
 import javaslang.control.Option;
 
-import java.util.*;
-import java.util.function.*;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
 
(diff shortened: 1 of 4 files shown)


danieldietrich:
I'm no sure if Vector should implement these interfaces.

We then have the following AbstractQueue impls:

* PriorityQueue
* Queue
* Vector

PriorityQueue is sorted, Queue and Vector not. Queues are used to enqueue/dequeue elements, other use-cases are rare. When to use Queue, when to use Vector?

The same with Stack. Common operations are push, pop, peek. When to use Vector over List?

I think Vector should not implement Queue/Stack - it is more confusing than it helps. Keep it simple!

paplorinc:
> When to use Queue, when to use Vector?

I think the usecase would be exactly the same as when choosing between `ArrayList` or `ArrayDeque` in Java.

danieldietrich:
As I said, Vector should not extends AbstractQueue nor implement Stack.

That's the reason Stack is not a Traversable. We even thought about removing that interface completely. I think manifesting it tighter within the type hierarchy is the wrong direction for Javaslang.

> I think the usecase would be exactly the same as when choosing between ArrayList or ArrayDeque in Java.

ArrayList does not implement Queue or Dequeue. Our Vector should not implement Queue operations. If a Queue is needed the user should use the appropriate collection for that purpose.

Issue #1530, opened 2016-08-24
Execute Future operations on the same Thread by default


danieldietrich:
UPDATE: We can't do execute all action on one thread. That would be inefficient. In general, the actions are executed in parallel.

However, the issues mentioned below (currently @Ignored in our unit tests) should be solved. Therefore I will leave this ticket open.

---

OBSOLETE:

It is interesting that Scala's Future does not reuse threads. Spawning new threads is a waste of resources - more context switches and parallel threads than necessary. Some are even completed and wait to be returned to the pool. This makes it intransparent for the call-site/user how resources are managed. The risk of a deadlock increases when having bounded pools.
I see only bebefits in reusing threads that already idle because the Future already completed on subsequent operations.

It is common to add an onComplete handler or an onSuccess and an onFailure handler. We rarely see adding many complete handlers. This means that only one handler is executed when the underlying Future is complete and does idle. In other words the theead can be reused.

If another behavior is needed, an ExecutorService should be explicitly specified when calling an operation or registering a handler. We will provide additional methods for that.

This change of behavior is expected to be binary backward compatible.

danieldietrich:
The Future core (FutureImpl) is small and well organized. A by changing a single line we managed to execute all actions on the same Future Thread:

```diff
     private void perform(Consumer<? super Try<T>> action) {
-        Try.run(() -> executorService.execute(() -> action.accept(value.get())));
+        Try.run(() -> value.forEach(action));
     }
```

But this is only for testing purposes and not the solution for the problem. We still need to execute the actions by an ExecutorService. It has to be a new single thread ExecutorService (singleton) by default with the option to exchange that:

```java
final Future<T> future = Future.of(someExecutor, () -> computation());

 // these use the same thread of future to process result
future.onComplete(Handler::processResult);
future.map(mapper);
...

// these use a new thread provided by fooExecutor to process result
future.onComplete(fooExecutor, Handler::processResult);
future.map(fooExecutor, mapper);
...
```

danieldietrich:
TODO: Re-enable `@Ignore`d unit tests in `FutureTest.java`

danieldietrich:
This issue should solve the following problem (tests by @mattjtodd):

```java
        BlockingQueue<Runnable> queue = new SynchronousQueue<>();
        ExecutorService service = new ThreadPoolExecutor(1, 1 , 0L, TimeUnit.MILLISECONDS, queue);

        Future
[...]

PR #1533, opened 2016-08-26
Minor changes to/housekeeping of FutureImpl. Added internal GuardedBy…

danieldietrich:
… annotation for documentation purpose. Added @Ignored unit tests towards #1530.


danieldietrich:
Not intended to be pulled yet. Locally the tests still fail, most probably because of the changes to exception handling. But these were made already in 2.0.3!? Why did the tests not fail in the CI tests of 2.0.3? Will update the tests accordingly and investigate further.

danieldietrich:
> Why did the tests not fail in the CI tests of 2.0.3?

Because I changed internals of `FutureImpl` that depended on the dummy impls of `java.util.concurrent.Future` for our unit tests.

In particular the underlying `java.util.concurrent.Future` contains now a `null` value when completed.  Before the value of the computation was stored but never used because we store the value of the computation on our own variable which is guarded by a lock. This is important because we need to synchronize the state of our Future ourselves.

This change is backward compatible because the `java.util.concurrent.Future` was not accessible from the outside. It is still used to control the asynchronous computation in the same way as before.

PR #1547, opened 2016-08-31
Gwt unit tests

ruslansennov:
```
Running client.GwtTestCollections
Compiling module TestModule.JUnit
   Ignored 90 units with compilation errors in first pass.
Compile with -strict or with -logLevel set to TRACE or DEBUG to see all errors.
   Computing all possible rebind results for 'com.google.gwt.logging.client.LogConfiguration'
      Rebinding com.google.gwt.logging.client.LogConfiguration
         Checking rule <generate-with class='com.google.gwt.editor.rebind.SimpleBeanEditorDriverGenerator'/>
            [WARN] Detected warnings related to 'com.google.gwt.editor.client.SimpleBeanEditorDriver'.   Are validation-api-<version>.jar and validation-api-<version>-sources.jar on the classpath?
            Specify -logLevel DEBUG to see all errors.
            [WARN] Unknown type 'com.google.gwt.editor.client.SimpleBeanEditorDriver' specified in deferred binding rule
   Compiling 1 permutation
      Compiling permutation 0...
   Compile of permutations succeeded
   Compilation succeeded -- 83.026s
Linking into /home/ruslan/proj/javaslang/javaslang-gwt/target/gwt-tests/www/TestModule.JUnit
   Link succeeded
   Linking succeeded -- 0.123s
Ignored 90 units with compilation errors in first pass.
Compile with -strict or with -logLevel set to TRACE or DEBUG to see all errors.
Starting http://localhost:38576/TestModule.JUnit/junit.html on browser FF38
200 - GET /TestModule.JUnit/junit.html (127.0.0.1)
200 - GET /TestModule.JUnit/TestModule.JUnit.nocache.js (127.0.0.1)
200 - GET /TestModule.JUnit/65D34B77DDE093B981B36E80D84B5CE9.cache.js (127.0.0.1)
200 - POST /TestModule.JUnit/junithost (127.0.0.1) 217 bytes
200 - POST /TestModule.JUnit/junithost (127.0.0.1) 178 bytes
200 - POST /TestModule.JUnit/junithost (127.0.0.1) 182 bytes
200 - POST /TestModule.JUnit/junithost (127.0.0.1) 184 bytes
200 - POST /TestModule.JUnit/junithost (127.0.0.1) 185 bytes
200 - POST /TestModule.JUnit/junithost (127.0.0.1) 182 bytes
200 - POST /TestModule.JUnit/junithost (127.0.0.1) 184 bytes
200 - POST /TestModule.JUnit/junithost (127.0.0.1) 182 bytes
200 - POST /TestModule.JUnit/junithost (127.0.0.1) 189 bytes
200 - POST /TestModule.JUnit/junithost (127.0.0.1) 181 bytes
200 - POST /TestModule.JUnit/junithost (127.0.0.1) 182 bytes
Tests run: 11, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 87.312 sec - in client.GwtTestCollections
```


javaslang/src/main/java/javaslang/collection/Arrays.java
```diff
@@ -39,8 +39,7 @@
         final Object[] array = (Object[]) arrayObject;
         final int arrayLength = (array == null) ? 0 : array.length;
         final int length = Math.max(arrayLength, minLength);
-        return (length == arrayLength) ? array.clone()
-                                       : arrayCopy(length, array, 0, 0, arrayLength);
+        return arrayCopy(length, array, 0, 0, arrayLength);
```
ruslansennov: `Object[].clone()` is GWT-incompatible
/cc @paplorinc

javaslang/src/main/java/javaslang/collection/RedBlackTree.java
```diff
@@ -68,7 +68,7 @@
 
     static <T extends Comparable<? super T>> RedBlackTree<T> ofAll(Iterable<? extends T> values) {
         Objects.requireNonNull(values, "values is null");
-        return ofAll((Comparator<? super T> & Serializable) T::compareTo, values);
+        return ofAll(Comparators.naturalComparator(), values);
```
ruslansennov: Something wrong with cast to `(Comparator<> & Serializable)`, need to be investigated

javaslang/src/main/java/javaslang/collection/Arrays.java
paplorinc: Thanks, it's ok, in my local primitivization commit I changed it too, to:
```java
static Object copy(Class<?> type, Object array, int minLength) {
    final int length = lengthOf(array);
    return arrayCopy(type, Math.max(minLength, length), array, 0, 0, length);
}
```

javaslang/src/main/java/javaslang/collection/RedBlackTree.java
paplorinc: Yes, `Java 9` complains also...

javaslang/src/main/java/javaslang/control/GwtIncompatible.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright 2013 Google Inc.
```
paplorinc: ?

javaslang-gwt/pom.xml
```diff
@@ -28,6 +28,18 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>javaslang</artifactId>
+            <version>${project.version}</version>
+            <classifier>sources</classifier>
```
paplorinc: hmmm, why sources?

javaslang-gwt/src/test/java/client/GwtTestCollections.java
```diff
@@ -0,0 +1,65 @@
...
+    public String getModuleName() {
+        return "TestModule";
+    }
+
+    public void testTuple() {
+        Tuple1<Integer> t = Tuple.of(1);
+        assertEquals((int) t._1(), 1);
+    }
+
+    private void applyCollection(Function1<char[], Traversable<Character>> factory) {
```
paplorinc: hmm, why `char`, because of `CharSeq`?

javaslang-gwt/src/test/java/client/GwtTestCollections.java
```diff
@@ -0,0 +1,65 @@
...
+    public void testCompileArray() {
+        applyCollection(Array::ofAll);
+    }
+
+    public void testCompileBitSet() {
+        applyCollection(BitSet::ofAll);
+    }
+
+    public void testCompileCharSeq() {
+        applyCollection(chars -> CharSeq.of(new String(chars)));
```
paplorinc: I think this can also be
```java
applyCollection(CharSeq::of);
```

javaslang-gwt/src/test/java/client/GwtTestCollections.java
```diff
@@ -0,0 +1,65 @@
...
+    public void testCompileLinkedHashSet() {
+        applyCollection(LinkedHashSet::ofAll);
+    }
+
+    public void testCompileList() {
+        applyCollection(List::ofAll);
+    }
+
+    public void testCompilePriorityQueue() {
+        applyCollection(chars -> PriorityQueue.ofAll(List.ofAll(chars)));
```
paplorinc: I think this should rather be
```java
applyCollection(chars -> PriorityQueue.ofAll(Iterator.ofAll(chars)));
```
otherwise you're iterating over the elements 3 times.
I know it's 'just' a test, but tests also serve as usage documentation.

javaslang-gwt/src/test/java/client/GwtTestCollections.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public void testCompileQueue() {
+        applyCollection(Queue::ofAll);
+    }
+
+    public void testCompileTreeSet() {
+        applyCollection(chars -> TreeSet.ofAll(List.ofAll(chars)));
```
paplorinc: same

javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -2792,6 +2792,7 @@ public int parseInt(int radix) {
      * @throws NumberFormatException If this {@code CharSeq} does not contain a parsable unsigned int.
      * @since 2.1.0
      */
+    @GwtIncompatible
     public int parseUnsignedInt() {
         return Integer.parseUnsignedInt(back);
```
paplorinc: we could make it GWT compatible, but I doubt it would be very useful from javascript:
```java
public int parseUnsignedInt() {
    int result = Integer.parseInt(back);
    if (result < 0) { throw new NumberFormatException(back + " is not unsigned!"); }
    return result;
}
```

javaslang-gwt/src/test/java/client/GwtTestCollections.java
ruslansennov: In this case test fails with `ClassCastException`. And I don't know what the problem. I changed this to `CharSeq.ofAll(Iterator.ofAll(chars))` (like other your comments)

javaslang-gwt/pom.xml
ruslansennov: GWT compiler requires this.

javaslang/src/main/java/javaslang/control/GwtIncompatible.java
ruslansennov: > ?

? :smile: 
This class is part of GWT, just copied to Javaslang to avoid additional dependency

javaslang-gwt/src/test/java/client/GwtTestCollections.java
ruslansennov: yep

javaslang/src/main/java/javaslang/collection/CharSeq.java
ruslansennov: I'm not sure about this. 
In fact I'm not sure that this method is useful in Java too :smile:

javaslang/src/main/java/javaslang/control/GwtIncompatible.java
paplorinc: But doesn't it state that you're not allowed to do that, i.e. copy right? :/

danieldietrich: This is what I found:

> All of GWT's code is available under the Apache 2.0 open source license.

Then mentioning Google's Copyright is the way to go, right?

paplorinc: I'm not sure, this is what the [doc](http://www.gwtproject.org/terms.html) states:

> 4. Redistribution
> 
> You may reproduce and distribute copies of the Work or Derivative Works thereof in any medium, with or without modifications, and in Source or Object form, provided that You meet the following conditions:
> 
> * You must give any other recipients of the Work or Derivative Works a copy of this License; and 
> 
> * You must cause any modified files to carry prominent notices stating that You changed the files; and 
> 
> * You must retain, in the Source form of any Derivative Works that You distribute, all copyright, patent, trademark, and attribution notices from the Source form of the Work, excluding those notices that do not pertain to any part of the Derivative Works; and 
> 
> * If the Work includes a "NOTICE" text file as part of its distribution, then any Derivative Works that You distribute must include a readable copy of the attribution notices contained within such NOTICE file, excluding those notices that do not pertain to any part of the Derivative Works, in at least one of the following places: within a NOTICE text file distributed as part of the Derivative Works; within the Source form or documentation, if provided along with the Derivative Works; or, within a display generated by the Derivative Works, if and wherever such third-party notices normally appear. The contents of the NOTICE file are for informational purposes only and do not modify the License. You may add Your own attribution notices within Derivative Works that You distribute, alongside or as an addendum to the NOTICE text from the Work, provided that such additional attribution notices cannot be construed as modifying the License.
>
> You may add Your own copyright statement to Your modifications and may provide additional or different license terms and conditions for use, reproduction, or distribution of Your modifications, or for any such Derivative Works as a whole, provided Your use, reproduction, and distribution of the Work otherwise complies with the conditions stated in this License.

javaslang/src/main/java/javaslang/collection/RedBlackTree.java
danieldietrich: Mmmh, why that? Does the following work?

```java
return ofAll((Comparator<T> & Serializable) T::compareTo, values);
```

Widening the type is safe in this case.

pom.xml
```diff
@@ -133,7 +134,7 @@ Note: The maven build currently needs to be started from the javaslang root dir
                     <executions>
                         <execution>
                             <id>attach-sources</id>
-                            <phase>verify</phase>
+                            <phase>generate-resources</phase>
```
danieldietrich: Is this important because GWT relies on sources? Just asking...

pom.xml
```diff
@@ -70,7 +70,7 @@ Note: The maven build currently needs to be started from the javaslang root dir
         <java.version>1.8</java.version>
         <jmh.version>1.13</jmh.version>
         <junit.version>4.12</junit.version>
-        <gwt.version>2.8.0-rc1</gwt.version>
+        <gwt.version>2.8.0-535f2fa9</gwt.version>
```
danieldietrich: Is this a specific snapshot or nightly build? :-)

javaslang/src/main/java/javaslang/control/GwtIncompatible.java
danieldietrich: Hi @tbroyer, do you have an advise on the best practice how to include the `@GwtIncompatible` annotation into a library that cannot depend on 3rd party libs.

Are we allowed to copy the source-code of the annotation, preserving the copyright file header _without_ including additional LICENSE files?

pom.xml
ruslansennov: yes

pom.xml
ruslansennov: This artifact must be build manually (see [README.md](https://github.com/ruslansennov/javaslang/blob/57db46ef641e22bd9168792e8a961f68bda957ff/javaslang-gwt/README.md))

javaslang-gwt/README.md
```diff
@@ -1,3 +1,10 @@
 # GWT support for Javaslang
 
-Highly Experimental, works and tested in IntelliJ Idea only
+Tested with 535f2fa973995f18d190a689127acbe134a42a15
+
+```
+git checkout 535f2fa9
+ant clean elemental dist-dev
+mvn install:install-file -Dfile=build/lib/gwt-user.jar -DgroupId=com.google.gwt -DartifactId=gwt-user -Dversion=2.8.0-535f2fa9 -Dpackaging=jar
```
tbroyer: Please use the `maven/push-gwt.sh` script instead; it will "unbundle" most dependencies and declare them as Maven dependencies.

javaslang-gwt/example/pom.xml
```diff
@@ -62,29 +62,15 @@
...
-                    <draftCompile>false</draftCompile>
-                    <disableCastChecking>true</disableCastChecking>
-                    <disableClassMetadata>true</disableClassMetadata>
-                    <disableRunAsync>true</disableRunAsync>
-                    <extraJvmArgs>-Xms2048m</extraJvmArgs>
-                    <extraJvmArgs>-Xmx2048m</extraJvmArgs>
-                    <inplace>false</inplace>
+                    <skipModule>true</skipModule>
+                    <moduleName>JavaslangExample</moduleName>
+                    <jvmArgs>-Xmx2048m</jvmArgs>
```
tbroyer: This should be:
```xml
<jvmArgs>
  <arg>-Xmx2048m</arg>
</jvmArgs>
```
I'm surprised (actually, not really) Maven doesn't complain.

javaslang-gwt/example/pom.xml
```diff
@@ -62,29 +62,15 @@
...
                 </configuration>
-                <executions>
```
tbroyer: Without `<executions>` and without a special `<packaging>`, the plugin won't do anything unless explicitly invoked from the command-line. Actually, I don't quite understand why this sample has `<packaging>jar</packaging>` (and includes the `maven-bundle-plugin`); I think this change is missing switching the packaging to `gwt-app`.

javaslang-gwt/pom.xml
```diff
@@ -47,8 +64,26 @@
...
+            </plugin>
+            <plugin>
+                <groupId>net.ltgt.gwt.maven</groupId>
+                <artifactId>gwt-maven-plugin</artifactId>
+                <version>${maven.gwt.plugin}</version>
+                <extensions>true</extensions>
+                <configuration>
+                    <skipModule>true</skipModule>
+                    <moduleName>Javaslang</moduleName>
+                    <jvmArgs>-Xmx2048m</jvmArgs>
```
tbroyer: Same here

javaslang-gwt/example/pom.xml
paplorinc: Do we really need it, if it works with defaults also?

javaslang/src/main/java/javaslang/control/GwtIncompatible.java
tbroyer: The Apache License allows you to do just that; but I believe you should then add the Google copyright to your `LICENSE` file, or at least add a note that parts of the code have a Google copyright.

But actually you don't need to copy the file: GWT expects an annotation with a simple of `GwtIncompatible`, in any package, will ignore any annotation attribute, and will only process the annotation on classes, fields, constructors and methods, and that's it. So just create such an annotation in your project that fulfills that contract, and put your own license (and copyright) on it.

paplorinc: Yup, that's my recommendation also, something like:
```java
@Retention(RetentionPolicy.CLASS)
@Target(ElementType.METHOD)
@interface GwtIncompatible {
}
```

tbroyer: Having it `@Documented` might be a good thing too.

pom.xml
danieldietrich: ok

javaslang/src/main/java/javaslang/control/GwtIncompatible.java
danieldietrich: @tbroyer thanks! @ruslansennov we should do that

pom.xml
tbroyer: Maybe better rely on the 2.8.0-SNAPSHOT for now, pending 2.8.0 final release that will include that change.

ruslansennov:
Thanks for comments, updated

javaslang/src/main/java/javaslang/control/GwtIncompatible.java
```diff
@@ -1,32 +1,21 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
 @Retention(RetentionPolicy.CLASS)
 @Target({
         ElementType.TYPE, ElementType.METHOD,
         ElementType.CONSTRUCTOR, ElementType.FIELD })
 @Documented
 @interface GwtIncompatible {
-    /**
```
paplorinc: hmm, why do you have several copies of the annotations now?

ruslansennov: To make them package visible

paplorinc: If I'm mistaken correctly, you only need these for the source code, right (i.e. Javaslang clients don't have to see the annotations)?
Would it make sense then to actually include the original `GWT` annotations as a build (provided?) dependency (i.e. used for build, but not packaged)?

danieldietrich: I'm not sure if the provided scope would work in all cases. For example I've found this of [StackOverflow](http://stackoverflow.com/a/24912101/1110815):

> If you're planning to generate a single JAR file with all of its dependencies (the typical xxxx-all.jar), then provided scope matters, because the classes inside this scope won't be package in the resulting JAR.

Also: What happens if an application inspects the annotations of a class/method by reflection and GwtIncompatible is not on the classpath? ClassNotFoundException?

I would use the `provided` scope only for dependencies which are known to be present at runtime.

paplorinc: http://stackoverflow.com/questions/3567413/why-doesnt-a-missing-annotation-cause-a-classnotfoundexception-at-runtime

Could we give it a try anyway?

danieldietrich: And then build a counter-example which uses reflection that fails?

tbroyer:
> Also: What happens if an application inspects the annotations of a class/method by reflection and GwtIncompatible is not on the classpath? ClassNotFoundException?

`getAnnotations` and `getDeclaredAnnotations` would simply not return it (because it has `@Retention(CLASS)`; it'd probably throw if it had `@Retention(RUNTIME)`).

> I would use the `provided` scope only for dependencies which are known to be present at runtime.

Ah, if only Maven had a _compile only_ scope… but it only has `<scope>provided</scope>` and/or `<optional>true</optional>` for dependencies that aren't transitively resolved. Either that or using a deploy POM different from the build POM.

danieldietrich:
@tbroyer thx!
@paplorinc let's give it a try then!

javaslang/src/main/java/javaslang/control/GwtIncompatible.java
tbroyer: > http://stackoverflow.com/questions/3567413/why-doesnt-a-missing-annotation-cause-a-classnotfoundexception-at-runtime

Ah yes, I remember better now: `@Retention(CLASS)` is only visible when looking at the class file itself (e.g. through ASM, or an annotation processor), `@Retention(RUNTIME)` makes it available through reflection (iff the class is available); so `@Retention(CLASS)` really is safe.

danieldietrich: @tbroyer Great, thanks for the explanation :)

danieldietrich:
@ruslansennov I prepare 2.1.0. What is your suggestion regarding the target version for the GWT module? I think it can be included in 2.1.0 but we should communicate the GWT support as _experimental_ to get first feedback. What do you think? Or are there still errors that prevent a valid GWT use-case?

ruslansennov:
First of all, I would keep `GwtIncompatible` annotations as is, i.e. included in `javaslang-core` as copy in every package. I believe that `provided` scope can be used in case if we will merge `javaslang-core` and `javaslang-gwt` modules in the future (it makes sense because all `javaslang-gwt` staff is quite small and can't broke main functionality)

> I prepare 2.1.0. What is your suggestion regarding the target version for the GWT module? I think it can be included in 2.1.0 but we should communicate the GWT support as experimental to get first feedback. What do you think? Or are there still errors that prevent a valid GWT use-case?

GWT support is not ready now. The Javaslang code is still not tested in GWT-scope and have a lot possibilities to fail. Also Javaslang is strongly Java8-oriented and we can find more bugs in GWT (see [fixed bug](https://github.com/gwtproject/gwt/issues/9406) and [probably bug](https://github.com/gwtproject/gwt/commit/468036ade51cd5094403ed2fb206e204c4f08fd3#commitcomment-18888927)). We should move this issue to 3.0.0

danieldietrich:
Good morning,

thank you Ruslan. Yes, 'provided'  dependencies will not be transitive. Would be interesting if it would work though...

However, I will move it to 3.0.0 then.

danieldietrich:
Great, thank you!! I followed your instructions/the javaslang-gwt/README.md and all tests succeeded. Will pull it in now.

Issue #1549, opened 2016-08-31
TreeMap.get throws NPE when the key is present and any value is null


ahammel:
TreeMap.get will throw a null pointer exception when trying to do a lookup on a map with nulls in the values. HashMap works as expected under the same circumstances.

``` java
void bug() {
    final HashMap<String, String> hashMap = HashMap.of(
            "Thisworks", "fine",
            "evenwith", null);

    hashMap.get("nokey");     // Option.none()
    hashMap.get("Thisworks"); // Option.some("fine")
    hashMap.get("evenwith");  // Option.some(null)

    final TreeMap<String, String> treeMap = TreeMap.of(
            "thisbreaks", "themap",
            "lolhax", null);

    treeMap.get("nokey");       // Option.none()
    treeMap.get("thisbreaks"); // NPE
    treeMap.get("lolhax");     // NPE
}
```

I'm using javaslang 2.0.3

Stack trace:

```
java.lang.NullPointerException
    at javaslang.Tuple2.compareTo(Tuple2.java:80)
    at javaslang.Tuple2.compareTo(Tuple2.java:96)
    at javaslang.Tuple2.compareTo(Tuple2.java:28)
    at javaslang.collection.RedBlackTreeModule$Node.find(RedBlackTree.java:418)
    at javaslang.collection.TreeMap.get(TreeMap.java:420)

```

ahammel:
Looks fixed on master :)

ahammel:
Although now I can't even get those tests to fail on 2.0.3, so maybe I'm crazy  ¯\_(ツ)_/¯ I'll have another look in the morning.

danieldietrich:
Hi @ahammel, thank you for reporting! I can reproduce it on 2.0.3 and confirm that it is fixed on the current SNAPSHOT. I will release Bugfix Release 2.0.4 neartime.

danieldietrich:
@ruslansennov do you know/remember with which PR we fixed it?

ruslansennov:
> do you know/remember with which PR we fixed it?

sorry :smile:

danieldietrich:
I think it is not fixed. Maybe the example now works on the current SNAPSHOT because the order of Map elements changed!? (A Map is unordered.)

The root error is thrown by the `Tuple2.compareTo` method, which calls this code:

```java
    @SuppressWarnings("unchecked")
[...]

PR #1550, opened 2016-09-01
Add tests for map with null keys

ahammel:
I ran into a bug in v2.0.3 which caused a TreeMap.get to throw an NPE
when _any_ of the keys were null. These tests show that the bug has been
cleared up as of this commit.

Original report: #1549


danieldietrich:
Great, many thanks!

PR #1554, opened 2016-09-02
Removed abstractmultimap

danieldietrich:
Hi @ruslansennov, @paplorinc,@zsolt-donca,

this is an ongoing (currently dirty) draft of the bugfix of #1326. I created an early PR so that you can track the changes.

This is one of several PRs. The checklist is updated [here](https://github.com/javaslang/javaslang/issues/1326#issuecomment-240726519).

**Goal:**

Remove all package-private super classes because method-references do not work for default implementations (JDK bug!).

**What I did so far:**

* Added Multimaps helper class similar to the new Maps class.
* Found another abstraction for a common Multimap Builder
* Started to move methods from AbstractMultimap to Multimaps. The generics and method arguments change.
* Changes LinkedHashMultimap to use Multimaps.

**TODO:**

* Move the remaining methods from AbstractMultimap to Multimaps
* Override missing methods and adjust return types in LinkedHashMap
* Do the same for HashMultimap and TreeMultimap
* **Fix the problem with the emptyInstance() when an underlying Comparator is present and the generic type changes. This blows up the Multimap when the original comparator is used on a new type.**
* Delete AbstractMultimap

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
+++ b/javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
@@ -26,79 +26,6 @@ import java.util.function.*;
  */
 abstract class AbstractMultimap<K, V, M extends Multimap<K, V>> implements Multimap<K, V> {
 
-    interface SerializableSupplier<T> extends Supplier<T>, Serializable {
-    }
-
-    private static final long serialVersionUID = 1L;
-
-    protected final Map<K, Traversable<V>> back;
-    protected final SerializableSupplier<Traversable<?>> emptyContainer;
-    private final ContainerType containerType;
-
-    AbstractMultimap(Map<K, Traversable<V>> back, ContainerType containerType, SerializableSupplier<Traversable<?>> emptyContainer) {
-        this.back = back;
-        this.containerType = containerType;
-        this.emptyContainer = emptyContainer;
-    }
-
-    protected abstract <K2, V2> Map<K2, V2> emptyMapSupplier();
-
-    protected abstract <K2, V2> Multimap<K2, V2> emptyInstance();
-
-    protected abstract <K2, V2> Multimap<K2, V2> createFromMap(Map<K2, Traversable<V2>> back);
-
-    @SuppressWarnings("unchecked")
-    private <K2, V2> Multimap<K2, V2> createFromEntries(Iterable<? extends Tuple2<? extends K2, ? extends V2>> entries) {
-        Map<K2, Traversable<V2>> back = emptyMapSupplier();
-        for (Tuple2<? extends K2, ? extends V2> entry : entries) {
-            if (back.containsKey(entry._1)) {
-                back = back.put(entry._1, containerType.add(back.get(entry._1).get(), entry._2));
-            } else {
-                back = back.put(entry._1, containerType.add(emptyContainer.get(), entry._2));
-            }
-        }
-        return createFromMap(back);
-    }
-
-    @Override
-    public <K2, V2> Multimap<K2, V2> bimap(Function<? super K, ? extends K2> keyMapper, Function<? super V, ? extends V2> valueMapper) {
-        Objects.requireNonNull(keyMapper, "keyMapper is null");
-        Objects.requireNonNull(valueMapper, "valueMapper is null");
-        final Iterator<Tuple2<K2, V2>> entries = iterator().map(entry -> Tuple.of(keyMapper.apply(entry._1), valueMapper.apply(entry._2)));
-        return createFromEntries(entries);
-    }
-
-    @Override
-    public boolean containsKey(K key) {
-        return back.containsKey(key);
-    }
-
-    @Override
-    public ContainerType getContainerType() {
-        return containerType;
-    }
-
-    @Override
-    public <K2, V2> Multimap<K2, V2> flatMap(BiFunction<? super K, ? super V, ? extends Iterable<Tuple2<K2, V2>>> mapper) {
-        Objects.requireNonNull(mapper, "mapper is null");
-        return foldLeft(this.emptyInstance(), (acc, entry) -> {
-            for (Tuple2<? extends K2, ? extends V2> mappedEntry : mapper.apply(entry._1, entry._2)) {
-                acc = acc.put(mappedEntry);
-            }
-            return acc;
-        });
-    }
-
-    @Override
-    public Option<Traversable<V>> get(K key) {
-        return back.get(key);
-    }
-
-    @Override
-    public Set<K> keySet() {
-        return back.keySet();
-    }
-
     @Override
     public <K2, V2> Multimap<K2, V2> map(BiFunction<? super K, ? super V, Tuple2<K2, V2>> mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
@@ -152,16 +79,6 @@ abstract class AbstractMultimap<K, V, M extends Multimap<K, V>> implements Multi
         return (M) (result == back ? this : createFromMap(result));
     }
 
-    @Override
-    public int size() {
-        return back.foldLeft(0, (s, t) -> s + t._2.size());
-    }
-
-    @Override
-    public Traversable<V> values() {
-        return Iterator.concat(back.values()).toStream();
-    }
-
     @SuppressWarnings("unchecked")
     @Override
     public M distinct() {
@@ -175,10 +92,11 @@ abstract class AbstractMultimap<K, V, M extends Multimap<K, V>> implements Multi
         return (M) createFromEntries(iterator().distinctBy(comparator));
     }
 
+    @SuppressWarnings("unchecked")
     @Override
-    public <U> Multimap<K, V> distinctBy(Function<? super Tuple2<K, V>, ? extends U> keyExtractor) {
+    public <U> M distinctBy(Function<? super Tuple2<K, V>, ? extends U> keyExtractor) {
         Objects.requireNonNull(keyExtractor, "keyExtractor is null");
-        return createFromEntries(iterator().distinctBy(keyExtractor));
+        return (M) createFromEntries(iterator().distinctBy(keyExtractor));
     }
 
     @Override
@@ -262,7 +180,7 @@ abstract class AbstractMultimap<K, V, M extends Multimap<K, V>> implements Multi
     }
 
     @Override
-    public <C> Map<C, Multimap<K, V>> groupBy(Function<? super Tuple2<K, V>, ? extends C> classifier) {
+    public <C> Map<C, M> groupBy(Function<? super Tuple2<K, V>, ? extends C> classifier) {
         return Collections.groupBy(this, classifier, this::createFromEntries);
     }
 
@@ -438,49 +356,4 @@ abstract class AbstractMultimap<K, V, M extends Multimap<K, V>> implements Multi
         return (M) (taken.length() == length() ? this : taken);
     }
 
-    @Override
-    public boolean equals(Object o) {
-        if (o == this) {
-            return true;
-        } else if (o != null && getClass().isAssignableFrom(o.getClass())) {
-            final AbstractMultimap<?, ?, ?> that = (AbstractMultimap<?, ?, ?>) o;
-            return this.back.equals(that.back);
-        } else {
-            return false;
-        }
-    }
-
-    @Override
-    public int hashCode() {
-        return back.hashCode();
-    }
-
-    @Override
-    public String stringPrefix() {
-        return getClass().getSimpleName() + "[" + emptyContainer.get().stringPrefix() + "]";
-    }
-
-    @Override
-    public String toString() {
-        return mkString(stringPrefix() + "(", ", ", ")");
-    }
-
-    @Override
-    public java.util.Map<K, Collection<V>> toJavaMap() {
-        final java.util.Map<K, Collection<V>> javaMap = new java.util.HashMap<>();
-        final Supplier<Collection<V>> javaContainerSupplier;
-        if (containerType == ContainerType.SEQ) {
-            javaContainerSupplier = java.util.ArrayList::new;
-        } else if (containerType == ContainerType.SET) {
-            javaContainerSupplier = java.util.HashSet::new;
-        } else if (containerType == ContainerType.SORTED_SET) {
-            javaContainerSupplier = java.util.TreeSet::new;
-        } else {
-            throw new IllegalStateException("Unknown ContainerType: " + containerType);
-        }
-        for (Tuple2<K, V> t : this) {
-            javaMap.computeIfAbsent(t._1, k -> javaContainerSupplier.get()).add(t._2);
-        }
-        return javaMap;
-    }
 }
--- a/javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
+++ b/javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
@@ -26,18 +26,6 @@ import java.util.function.*;
  */
 abstract class AbstractMultimap<K, V, M extends Multimap<K, V>> implements Multimap<K, V> {
 
-    @Override
-    public <K2, V2> Multimap<K2, V2> map(BiFunction<? super K, ? super V, Tuple2<K2, V2>> mapper) {
-        Objects.requireNonNull(mapper, "mapper is null");
-        return foldLeft(this.emptyInstance(), (acc, entry) -> acc.put(mapper.apply(entry._1, entry._2)));
-    }
-
-    @Override
-    public <V2> Multimap<K, V2> mapValues(Function<? super V, ? extends V2> valueMapper) {
-        Objects.requireNonNull(valueMapper, "valueMapper is null");
-        return map((k, v) -> Tuple.of(k, valueMapper.apply(v)));
-    }
-
     @SuppressWarnings("unchecked")
     @Override
     public M put(K key, V value) {
--- a/javaslang/src/main/java/javaslang/collection/HashMultimap.java
+++ b/javaslang/src/main/java/javaslang/collection/HashMultimap.java
@@ -6,15 +6,14 @@
 package javaslang.collection;
 
 import javaslang.Tuple2;
+import javaslang.control.Option;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Comparator;
 import java.util.Objects;
-import java.util.function.BiConsumer;
-import java.util.function.BinaryOperator;
-import java.util.function.Function;
-import java.util.function.Supplier;
+import java.util.function.*;
 import java.util.stream.Collector;
 
 /**
@@ -25,9 +24,21 @@ import java.util.stream.Collector;
  * @author Ruslan Sennov
  * @since 2.1.0
  */
-public final class HashMultimap<K, V> extends AbstractMultimap<K, V, HashMultimap<K, V>> implements Serializable {
+public final class HashMultimap<K, V> implements Multimap<K, V>, Serializable {
 
     private static final long serialVersionUID = 1L;
+
+    private final Map<K, Traversable<V>> back;
+    private final SerializableSupplier<Traversable<?>> emptyContainer;
+    private final ContainerType containerType;
+
+    private HashMultimap(Map<K, Traversable<V>> back, ContainerType containerType, SerializableSupplier<Traversable<?>> emptyContainer) {
+        this.back = back;
+        this.containerType = containerType;
+        this.emptyContainer = emptyContainer;
+    }
+
+    // -- static API
 
     public static <V> Builder<V> withSeq() {
         return new Builder<>(ContainerType.SEQ, List::empty);
@@ -136,7 +147,7 @@ public final class HashMultimap<K, V> extends AbstractMultimap<K, V, HashMultima
 
     /**
      * Narrows a widened {@code HashMultimap<? extends K, ? extends V>} to {@code HashMultimap<K, V>}
-     * by performing a type safe-cast. This is eligible because immutable/read-only
+     * by performing a type-safe cast. This is eligible because immutable/read-only
      * collections are covariant.
      *
      * @param map A {@code Map}.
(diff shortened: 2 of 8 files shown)


javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
```diff
@@ -175,10 +92,11 @@ public M distinctBy(Comparator<? super Tuple2<K, V>> comparator) {
         return (M) createFromEntries(iterator().distinctBy(comparator));
     }
 
+    @SuppressWarnings("unchecked")
     @Override
-    public <U> Multimap<K, V> distinctBy(Function<? super Tuple2<K, V>, ? extends U> keyExtractor) {
+    public <U> M distinctBy(Function<? super Tuple2<K, V>, ? extends U> keyExtractor) {
```
paplorinc: :+1:

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
```diff
@@ -175,10 +92,11 @@ public M distinctBy(Comparator<? super Tuple2<K, V>> comparator) {
...
         Objects.requireNonNull(keyExtractor, "keyExtractor is null");
-        return createFromEntries(iterator().distinctBy(keyExtractor));
+        return (M) createFromEntries(iterator().distinctBy(keyExtractor));
```
paplorinc: I think that if you change `createFromEntries` to `M` also, you won't nee these casts (multiple ones in this file)

javaslang/src/main/java/javaslang/collection/HashMultimap.java
```diff
@@ -25,10 +24,22 @@
  * @author Ruslan Sennov
  * @since 2.1.0
  */
-public final class HashMultimap<K, V> extends AbstractMultimap<K, V, HashMultimap<K, V>> implements Serializable {
+public final class HashMultimap<K, V> implements Multimap<K, V>, Serializable {
 
     private static final long serialVersionUID = 1L;
 
+    private final Map<K, Traversable<V>> back;
```
paplorinc: I find `back` to be a confusing name, is it an abbreviation of `backing` or is it meant as the opposite of `front` or does it mean that something is stored in reverse order?
In my opinion `delegate` might be more appropriate in this case.

javaslang/src/main/java/javaslang/collection/HashMultimap.java
```diff
@@ -149,24 +160,92 @@ private Builder(ContainerType containerType, SerializableSupplier<Traversable<?>
...
+            }
+            return acc;
+        });
     }
 
     @Override
-    protected <K2, V2> HashMultimap<K2, V2> createFromMap(Map<K2, Traversable<V2>> back) {
-        return new HashMultimap<>(back, getContainerType(), emptyContainer);
+    public int size() {
+        return back.foldLeft(0, (s, t) -> s + t._2.size());
```
paplorinc: hmm, linear `size` method?

javaslang/src/main/java/javaslang/collection/HashMultimap.java
```diff
@@ -25,10 +24,22 @@
  * @author Ruslan Sennov
  * @since 2.1.0
```
paplorinc: ?

javaslang/src/main/java/javaslang/collection/HashMultimap.java
```diff
@@ -149,24 +160,92 @@ private Builder(ContainerType containerType, SerializableSupplier<Traversable<?>
...
+        return Iterator.concat(back.values()).toStream();
+    }
+
+    // -- Object
+
+    @Override
+    public boolean equals(Object o) {
+        if (o == this) {
+            return true;
+        } else if (o instanceof HashMultimap) {
```
paplorinc: :+1: for removing reflection here

javaslang/src/main/java/javaslang/collection/HashMultimap.java
```diff
@@ -149,24 +160,92 @@ private Builder(ContainerType containerType, SerializableSupplier<Traversable<?>
...
+
+    @Override
+    public Traversable<V> values() {
+        return Iterator.concat(back.values()).toStream();
+    }
+
+    // -- Object
+
+    @Override
+    public boolean equals(Object o) {
```
paplorinc: could be a one-liner to increase signal-to-noise-ratio:
```java
return (that == this) || ((that instanceof HashMultimap) && this.delegate.equals(((HashMultimap<?, ?>) that).delegate);
```

javaslang/src/main/java/javaslang/collection/HashMultimap.java
```diff
@@ -149,24 +160,92 @@ private Builder(ContainerType containerType, SerializableSupplier<Traversable<?>
...
+    }
+
+    @Override
+    public int hashCode() {
+        return back.hashCode();
+    }
+
+    @Override
+    public String stringPrefix() {
+        return getClass().getSimpleName() + "[" + emptyContainer.get().stringPrefix() + "]";
```
paplorinc: I don't mind, but was this kept on purpose, i.e. no `"HashMultimap[" +`?

javaslang/src/main/java/javaslang/collection/HashMultimap.java
```diff
@@ -149,24 +160,92 @@ private Builder(ContainerType containerType, SerializableSupplier<Traversable<?>
...
+    }
+
+    @Override
+    public String toString() {
+        return mkString(stringPrefix() + "(", ", ", ")");
+    }
+
+    private <K2, V2> Map<K2, V2> emptyMapSupplier() {
```
paplorinc: I think it's confusing that this doesn't actually return a `Supplier`

javaslang/src/main/java/javaslang/collection/LinkedHashMultimap.java
```diff
@@ -132,11 +124,21 @@ private Builder(ContainerType containerType, SerializableSupplier<Traversable<?>
             };
             return Collector.of(supplier, accumulator, combiner, this::ofEntries);
         }
+
+        @Override
+        <K, V2 extends V> Map<K, Traversable<V2>> emptyMap() {
```
paplorinc: are these package private on purpose?

javaslang/src/main/java/javaslang/collection/LinkedHashMultimap.java
```diff
@@ -149,24 +151,107 @@ private Builder(ContainerType containerType, SerializableSupplier<Traversable<?>
...
 
     @Override
-    protected <K2, V2> Map<K2, V2> emptyMapSupplier() {
-        return LinkedHashMap.empty();
+    public boolean containsKey(K key) {
+        return back.containsKey(key);
     }
 
-    @SuppressWarnings("unchecked")
+    // TODO: Move this implementation to Multimaps and use Map Builder instead of Multimap.put
```
paplorinc: reminder: TODO

javaslang/src/main/java/javaslang/collection/LinkedHashMultimap.java
```diff
@@ -149,24 +151,107 @@ private Builder(ContainerType containerType, SerializableSupplier<Traversable<?>
...
+    }
+
+    @Override
+    public Set<K> keySet() {
+        return back.keySet();
+    }
+
+    @Override
+    public <K2, V2> LinkedHashMultimap<K2, V2> map(BiFunction<? super K, ? super V, Tuple2<K2, V2>> mapper) {
+        return null; // TODO ...
```
paplorinc: ah, ok, I see there are multiple ones :)

javaslang/src/main/java/javaslang/collection/LinkedHashMultimap.java
```diff
@@ -149,24 +151,107 @@ private Builder(ContainerType containerType, SerializableSupplier<Traversable<?>
...
+
+    @Override
+    public Traversable<V> values() {
+        return Iterator.concat(back.values()).toStream();
+    }
+
+    // -- Object
+
+    @Override
+    public boolean equals(Object o) {
```
paplorinc: same, can be a one-liner

javaslang/src/main/java/javaslang/collection/Multimaps.java
```diff
@@ -0,0 +1,171 @@
...
+ * @since 2.1.0
+ */
+final class Multimaps {
+
+    static <K, V, K2, V2, M extends Multimap<K2, V2>> M bimap(
+            Multimap<K, V> multimap,
+            OfEntries<K2, V2, M> ofEntries,
+            Function<? super K, ? extends K2> keyMapper,
+            Function<? super V, ? extends V2> valueMapper) {
+        Objects.requireNonNull(keyMapper, "keyMapper is null");
```
paplorinc: maybe put pre-conditions to the rest of the params also?

javaslang/src/main/java/javaslang/collection/Multimaps.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    static <K, V> int size(Map<K, Traversable<V>> back) {
+        return back.foldLeft(0, (s, t) -> s + t._2.size());
+    }
+
+    static <K, V> java.util.Map<K, Collection<V>> toJavaMap(Multimap<K, V> multimap) {
+        final java.util.Map<K, Collection<V>> javaMap = new java.util.HashMap<>();
+        final Supplier<Collection<V>> javaContainerSupplier;
+        final ContainerType containerType = multimap.getContainerType();
```
paplorinc: you could change this to an switch instead, as `ContainerType` is an enum and maybe extract to separate method, to get rid of `break` statements, e.g.:
```java
@Override
public java.util.Map<K, Collection<V>> toJavaMap() {
    final java.util.Map<K, Collection<V>> javaMap = new java.util.HashMap<>();
    final Supplier<Collection<V>> supplier = getJavaContainerSupplier();
    for (Tuple2<K, V> t : this) {
        javaMap.computeIfAbsent(t._1, k -> supplier.get()).add(t._2);
    }
    return javaMap;
}
private Supplier<Collection<V>> getJavaContainerSupplier() {
    switch (containerType) {
        case SEQ:
            return java.util.ArrayList::new;
        case SET:
            return java.util.HashSet::new;
        case SORTED_SET:
            return java.util.TreeSet::new;
        default:
            throw new IllegalStateException("Unknown ContainerType: " + containerType);
    }
}
```

javaslang/src/main/java/javaslang/collection/Multimaps.java
```diff
@@ -0,0 +1,171 @@
...
+        public abstract <K, V2 extends V> Multimap<K, V2> ofEntries(java.util.Map.Entry<? extends K, ? extends V2>... entries);
+
+        public abstract <K, V2 extends V> Multimap<K, V2> fill(int n, Supplier<Tuple2<? extends K, ? extends V2>> supplier);
+
+        public abstract <K, V2 extends V> Multimap<K, V2> tabulate(int n, Function<? super Integer, Tuple2<? extends K, ? extends V2>> f);
+
+        public abstract <K, V2 extends V> Collector<Tuple2<K, V2>, ArrayList<Tuple2<K, V2>>, ? extends Multimap<K, V2>> collector();
+
+        // Returns the empty Map used as back for Multimap
+        abstract <K, V2 extends V> Map<K, Traversable<V2>> emptyMap();
```
paplorinc: not public on purpose?

javaslang/src/main/java/javaslang/collection/Multimaps.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        static class Entries<K, V> implements Iterator<Object[]> {
+
+            int index = 0;
+            Object[] pairs;
+
+            Entries(Object[] pairs) {
+                Objects.requireNonNull(pairs, "pairs is null");
+                if ((pairs.length & 1) != 0) {
```
paplorinc: this should be extracted to an `isEven` method, like
```java
static boolean hasEvenSize(Object[] pairs) { return (pairs.length & 1) != 0; }
```

javaslang/src/main/java/javaslang/collection/TreeMultimap.java
```diff
@@ -194,7 +205,7 @@ private Builder(ContainerType containerType, SerializableSupplier<Traversable<?>
 
     /**
      * Narrows a widened {@code HashMultimap<? extends K, ? extends V>} to {@code HashMultimap<K, V>}
-     * by performing a type safe-cast. This is eligible because immutable/read-only
+     * by performing a type-safe cast. This is eligible because immutable/read-only
```
paplorinc: these will disappear after rebase, right?

javaslang/src/main/java/javaslang/collection/LinkedHashMultimap.java
```diff
@@ -192,12 +192,12 @@ public ContainerType getContainerType() {
 
     @Override
     public <K2, V2> LinkedHashMultimap<K2, V2> map(BiFunction<? super K, ? super V, Tuple2<K2, V2>> mapper) {
-        return null; // TODO ...
+        return Multimaps.map(this, emptyInstance(), mapper);
```
paplorinc: would it be possible to force push these, otherwise we're reviewing the same line multiple times (even though they were fixed in a later commit).

javaslang/src/main/java/javaslang/collection/Multimaps.java
```diff
@@ -161,10 +185,30 @@ V getValue(Object[] pair) {
         }
     }
 
+    /**
```
paplorinc: I envy your drive for writing Javadocs...

paplorinc:
@danieldietrich, good job so far, now go get some sleep :p

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
danieldietrich: The class AbstractMultimap is not relevant any more. You can ignore it. It is just a Clipboard for the ongoing work.

E.g. this method will be moved to `Multimaps` and get additional parameters, one of them is a function `OfEntries` that is type-safe. The cast will disappear.

javaslang/src/main/java/javaslang/collection/HashMultimap.java
danieldietrich: Will rename it.

Note: Currently only LinkedHashMultimap and Multimaps is worth to take a look at. HashMultimap and TreeMultimap are far behind...

javaslang/src/main/java/javaslang/collection/HashMultimap.java
danieldietrich: You are right, we should use the `size()` method of the delegate directly!

javaslang/src/main/java/javaslang/collection/HashMultimap.java
danieldietrich: This is all right, Ruslan did all the work and Multimap will be there in 2.1.0. I only copy-paste the impls and add additional generic parameters. Some fixes and a Builder-reimplementation on the way... It is a load of work, maybe I add my name...

javaslang/src/main/java/javaslang/collection/HashMultimap.java
danieldietrich: Yes, this could be applied to many equals methods of Javaslang. Fewer lines will increase code coverage automatically. Think we should do that in a separate PR for the whole repo.

javaslang/src/main/java/javaslang/collection/HashMultimap.java
danieldietrich: You are right - will change that.

javaslang/src/main/java/javaslang/collection/HashMultimap.java
danieldietrich: Yes. In LinkedHashMultimap it completely disappeared. Other Multimap impls will follow...

javaslang/src/main/java/javaslang/collection/LinkedHashMultimap.java
danieldietrich: Yes, used internally by the common builder impl. I will add an comment...

javaslang/src/main/java/javaslang/collection/LinkedHashMultimap.java
danieldietrich: will clean that up soon!

javaslang/src/main/java/javaslang/collection/Multimaps.java
danieldietrich: I didn't do it on purpose. There are parameters that are passed internally by the Multimap impls and there are user-parameters that are routed through. We only test the latter because we are sure that the former are correct.

javaslang/src/main/java/javaslang/collection/Multimaps.java
danieldietrich: Nice!

javaslang/src/main/java/javaslang/collection/Multimaps.java
danieldietrich: Yes, added a comment

javaslang/src/main/java/javaslang/collection/Multimaps.java
danieldietrich: will do

javaslang/src/main/java/javaslang/collection/TreeMultimap.java
danieldietrich: I'm not very used to rebasing. Did try it but the Atom editor had a problem and the interactive mode didn't work. Found a fix but using now a guy. There I squash/merge/cherry-pick and delete all commits manually.

From what I can see now, this will not disappear because I did it in a commit on this branch.

javaslang/src/main/java/javaslang/collection/LinkedHashMultimap.java
danieldietrich: Yes, what you see is a working darft. I will fore push later

javaslang/src/main/java/javaslang/collection/Multimaps.java
danieldietrich: This one is for developers. It may help understanding the code. But it could also a single-line comment...

danieldietrich:
zzZZzz

javaslang/src/main/java/javaslang/collection/HashMultimap.java
paplorinc: :) k, thanks

javaslang/src/main/java/javaslang/collection/HashMultimap.java
paplorinc: Fewer lines usually decrease it, if no new tests are added - exception being partial line matches, like this one.

javaslang/src/main/java/javaslang/collection/TreeMultimap.java
paplorinc: yes, that's what I meant :)

But other rebases can be done nicely from Idea also (`VCS` / `Git` / `Rebase`).

javaslang/src/main/java/javaslang/collection/HashMultimap.java
paplorinc: In other cases we store the length, e.g. `Tree`.
It bothers me, as a simple loop over the elements will end up being quadratic, not linear and I find that very counter-intuitive.

javaslang/src/main/java/javaslang/collection/HashMultimap.java
danieldietrich: Yes, you are right. then let's add some newlines ;-p

javaslang/src/main/java/javaslang/collection/TreeMultimap.java
danieldietrich: Thanks, will try that - still searching the right tool. Currently I use GitUp for Mac - nice and simple.

javaslang/src/main/java/javaslang/collection/HashMultimap.java
paplorinc: I think we should rather generate them ...

javaslang/src/main/java/javaslang/collection/HashMultimap.java
danieldietrich: I think it will not be quadratic because valid containers do store their size. It will be linear over keys.

However, storing the size might be a viable solution. We should do that in a separate PR when this re-design is finished. We currently have too many changes in one PR.

javaslang/src/main/java/javaslang/collection/HashMultimap.java
danieldietrich: Haha, we could generate veeery looong package private classes which do nothing. Then we will eventually reach 99.99% code coverage without covering code ;)

Is it already 1st April? :)

danieldietrich:
My approach looks too complicated/unintuitive to me.

Example:

```java
@Override
public LinkedHashMultimap<K, V> put(K key, V value) {
    return Multimaps.put(this, delegate, emptyContainer, containerType, builder()::ofMap, key, value);
}

static <K, V, M extends Multimap<K, V>> M put(
        M multimap,
        Map<K, Traversable<V>> delegate,
        Traversable<V> emptyContainer,
        ContainerType containerType,
        OfMap<K, V, M> ofMap,
        K key,
        V value) {
    final Traversable<V> values = delegate.get(key).getOrElse(emptyContainer);
    final Traversable<V> newValues = containerType.add(values, value);
    return (newValues == values) ? multimap : ofMap.apply(delegate.put(key, newValues));
}
```

E.g. we create a new `LinkedHashMultimap.Builder` instance to get an instance of the `OfMap` function, which wraps a delegate. We need an instance to get the instances `emptyContainer` and `containerType`. As an alternative we could duplicate the `ofMap` method - one for the Builder, one for the Multimap implementation.

You see, this starts to get cumbersome. Putting a value into a Map should be simple. I have the wrong abstraction here.

I like @ruslansennov approach to have an `AbstractMultimap` (btw - same applies for `AbstractMap` which I already removed and pushed the new `Maps` helper). But I don't want to make our abstract base impls public. They are not intended to be part of the user API. They just help us internally to factor out duplicate code.

Maybe there is another way to circumvent the JDK bug by just overriding all methods of the package private base class and delegating to them. I will test it.

danieldietrich:
What if... we just duplicate code. It would simplify things.

```java
@Override
public LinkedHashMultimap<K, V> takeUntil(Predicate<? super Tuple2<K, V>> predicate) {
    return takeWhile(predicate.negate());
}
```

We could also generate the files (all Multimaps and all Maps) by the code generator. Then we have the possibility to apply better optimizations and have fewer classes. On the other hand that would make it harder to get the hands on these classes for further changes.

danieldietrich:
This will still take some time. Maybe we should move this issue to the next (bugfix or minor) release in order to ship 2.0.4. In other words method references will not work for Maps/Multimaps. Workaround: use lambdas instead of method references. Good thing: It is a JDK bug, not a Javaslang bug, but should be fixed neartime.

I think we still have problems in TreeMap and/or TreeMultimap when the value type is changed by an operation (based on wrong Comparators). But I haven't investigated it yet.

Also operations of sorted collections that operate on multiple instances with different underlying Comparators make cause unexpected behavior. But this has to be checked and solved step-by-step. I'm alread on it by re-designing the RedBlackTree with some optimizations and preparation for the Navigable additions...

So much to do :) Known bugs are ok. The most important thing is to ship reliable, backward-compatible software.

danieldietrich:
We already removed AbstractMap. I wouldn't revert that change - but stop this change here.

I will move (the rest of) #1326 to 2.0.5 and go the generator-way.

paplorinc:
I'm trying to follow the changes, but I'm not sure I understand one thing: can't we make the abstractions public instead?

If I'm mistaken correctly the problem is that the returned value is a package-private superclass.
But can't we extract a public interface for the private superclass and return the interface instead?

(also, why isn't multimap simply a `Map<K, Seq<V>>`>)

danieldietrich:
> can't we make the abstractions public instead

No. They are only ballast and would pollute the user's view on the Javaslang API.

> If I'm mistaken correctly the problem is that the returned value is a package-private superclass.

That's not the problem. The returned value is a generic `M extends Multimap<K, V>`. The `M` is public (e.g. LinkedHashMultimap).

> But can't we extract a public interface for the private superclass and return the interface instead?

We have a public interface, it is `Multimap<K, V>`. But this would not solve the problem. It is a JDK bug finding the right method.

> (also, why isn't multimap simply a Map<K, Seq<V>>>)

Because Multimaps have methods that operate on `V` instead of `Traversable<V>`. (`Seq<V>` is not sufficient, we have also `Set<V>` and `SortedSet<V>`).

---

I like the idea of generated code. It solved the one and only problem that our AbstractMultimap also solves - factor out duplicate code into a single place.

But the generated code is better than the AbstractMultimap. We do not have the generic-hell, no type casts, fewer types within the class hierarchy, etc. There are many benefits.

There is only one drawback - the code generator is a little bit harder to write than non-generated code.

paplorinc:
> I like the idea of generated code.

Hmmm, wasn't that your counter-argument against collection specialization?

Or can we make the generator source a valid Java source? (i.e write once and forget that this is actually a base for other cases). Otherwise I think this would have a huge maintenance cost (non-standard way of programming, scaring many people away, providing more base for mistakes).

danieldietrich:
> > I like the idea of generated code.

> Hmmm, wasn't that your counter-argument against collection specialization?

I did not mean in the very general but only considering our options:

* Making the abstract super-classes public (- this is a NO-GO because it is not intended to be user API)
* Staying with package-private abstract super classes and override all methods in implementing classes. These just delegate to the super-class (- this leads to **maintenance hell**)
* Introducing package-private helper classes with static methods (- this leads to **unreadable, blown-up API, see current PR**)
* generate the code (- we get rid of much technical boilerplate, like casts, unnecessary types and method calls, etc.)

> Or can we make the generator source a valid Java source?

Interesting idea. I have already thought about self-expanding Java code. Xtend language does this with so-called Active Annotations. Scala has Macros. But these things actually _hide_ things like the annotation of many (in)famous Java frameworks.

Btw - the AbstractQueue does not have this problem. The queues do not need to be generated.

paplorinc:
As long as the source code is not Strings, I don't mind :)

danieldietrich:
> As long as the source code is not Strings, I don't mind :)

I don't understand. We check-in the generated source. It is like hand-written code. Unit tests run against it. The unit tests will not be generated (from what I can see now), we use the existing `Absrtact*MapTest` classes.

Also our CI build automatically checks if a user accidentally overwrites generated code with manual changes. This breaks the build.

**Generating the Maps and Multimaps should be safe regarding the Java compiler and manual user-changes!**

danieldietrich:
Obsolete, will be closed. We (most probably) will go the generator way.

paplorinc:
> I don't understand. 

I just meant that source code should be written in an IDE recognizable format.
In my opinion currently we have some source code in Strings that are a bit cumbersome to edit, i.e. no compiler or IDE help, only after regenerating everything can I tell if it's ok (like back in Borland C++ times).

danieldietrich:
> I just meant that source code should be written in an IDE recognizable format.

The quality of our source code and user experience are very important / priority 1. Javaslang library developer experience is also important but priority 2.

* The public API does not change, therefore **user experience** is not affected.
* A switch to generating Maps and Multimaps will (drastically?) increase the **quality of our source code**.
* Maintaining the generator templates is an additional effort, it affects the **developer experience**.

The right workflow will ease the pain of maintaining generator temples:

1. First directly edit/change the generated code. We will have full IDE support. Editing may involve the introduction of new types or the erasure of existing types. As always, write unit tests for the code. In this phase we do also refactoring until the code has a final state/seems perfect.
2. We use the IDE to test the changes. We have to take care not to run the generator (via maven build), it will overwrite our changes. It is recommended to create backups, e.g. in the form of local commits that will be squashed later.
3. If all works as expected, we adjust the generator templates in the way that they reflect the changes of the generated code we made in the first phase. We committed the direct changes of the generated code before. So we are able to verify if our generator adjustment is correct by looking at the `git diff`.

Phase 1. increases the developer experience during generator-driven development. Phase 2. is a pain in the ass, as always in stringly-driven development. But we need to do it to increase the quality of our library.

paplorinc:
Thanks for clarifying. I like the approach.
The alternative, of generating all cases from an annotated `.java` file would probably be more complex :)

PR #1557, opened 2016-09-04
Automatic, internal primitive storage support for `Vector`

paplorinc:
This PR modifies the optimized `Vector` to store its data internally in the format it was created from, i.e. 

``` java
Vector.ofAll(3, 1, 4, 1, 5); // stored internally as `int[]`
```

will create a `Vector` that has an `int[]` internally and

``` java
Vector.ofAll(Arrays.asList(3, 1, 4, 1, 5)); // boxed to `Integer[]` by `asList`
```

will have an `Object[]` inside.

The user doesn't have to know how it's stored internally (e.g. adding a `null` will automatically box everything in amortized constant time).

Internally many operations can work with primitive arrays without boxing (e.g. `slice`, `iterate`, `filter`, `distinct`), but others have to wrap them to make sense (e.g. `flatMap`).

Memory-wise storing primitives can be up to `~12×` more memory efficient (e.g. `53,968 bytes` for Vector<Byte> compared to `653,672 bytes` for `ArrayList<Byte>` (or `3,408,624 bytes` for Functional Java's `Seq<Byte>`)).
Interestingly many operations are faster also (e.g. `create` and `iterate`), sometimes even faster than the `Java` counterpart - e.g. `filter` or `map`.

The [benchmarks](https://github.com/paplorinc/javaslang/blob/Primitivization/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java) are as follows:

``` java
Operation Ratio                                                32     1024    32768
Create    slang_persistent/java_mutable                     1.51×    0.92×    1.17×
Create    slang_persistent/java_mutable_boxed              11.10×    8.64×    6.17×
Create    slang_persistent/java_mutable_boxed_stream       16.68×   10.77×    9.57×
Create    slang_persistent/fjava_persistent               229.91×  216.16×  192.30×
Create    slang_persistent/pcollections_persistent        126.90×  253.86×  375.34×
Create    slang_persistent/ecollections_persistent          2.05×    0.95×    1.18×
Create    slang_persistent/clojure_persistent               0.95×   15.40×   12.79×
Create    slang_persistent/scala_persistent                13.96×    9.32×    6.43×

Head      slang_persistent/java_mutable                     0.71×    0.53×    0.40×
Head      slang_persistent/fjava_persistent                 0.85×    0.66×    0.51×
Head      slang_persistent/pcollections_persistent          2.18×    3.75×    4.68×
Head      slang_persistent/ecollections_persistent          0.77×    0.61×    0.47×
Head      slang_persistent/clojure_persistent               0.70×    0.83×    0.90×
Head      slang_persistent/scala_persistent                 0.78×    0.63×    0.49×

Tail      slang_persistent/java_mutable                     0.71×    0.56×    0.39×
Tail      slang_persistent/fjava_persistent                52.84×   50.26×   52.36×
Tail      slang_persistent/pcollections_persistent          5.85×    9.15×   12.06×
Tail      slang_persistent/ecollections_persistent         16.88×  156.82× 4301.60×
Tail      slang_persistent/clojure_persistent               0.93×    0.69×    0.60×
Tail      slang_persistent/scala_persistent                 2.98×    4.87×    5.97×

Get       slang_persistent/java_mutable                     0.71×    0.12×    0.36×
Get       slang_persistent/fjava_persistent                55.92×   28.12×   79.81×
Get       slang_persistent/pcollections_persistent          8.32×    4.20×   17.46×
Get       slang_persistent/ecollections_persistent          0.71×    0.10×    0.34×
Get       slang_persistent/clojure_persistent               0.93×    0.37×    1.45×
Get       slang_persistent/scala_persistent                 1.13×    0.30×    0.95×

Update    slang_persistent/java_mutable                     0.11×    0.05×    0.03×
Update    slang_persistent/fjava_persistent               133.11×  262.66×  222.09×
Update    slang_persistent/pcollections_persistent          2.45×    4.30×    4.97×
Update    slang_persistent/ecollections_persistent         13.88×  162.43× 2589.05×
Update    slang_persistent/clojure_persistent               0.85×    1.23×    1.10×
Update    slang_persistent/scala_persistent                 1.03×    1.07×    1.14×

Map       slang_persistent/java_mutable                     2.19×    2.27×    2.00×
Map       slang_persistent/java_mutable_loop                0.53×    0.67×    0.54×
Map       slang_persistent/ecollections_persistent          1.16×    1.15×    1.06×
Map       slang_persistent/scala_persistent                 1.03×    1.43×    1.73×

Filter    slang_persistent/java_mutable                     1.97×    1.64×    1.15×
Filter    slang_persistent/ecollections_persistent          1.67×    1.14×    0.92×
Filter    slang_persistent/scala_persistent                 1.45×    1.44×    1.42×

Prepend   slang_persistent/java_mutable                     0.38×    0.71×   11.80×
Prepend   slang_persistent/fjava_persistent                 1.43×    1.62×    1.17×
Prepend   slang_persistent/pcollections_persistent          0.94×    2.47×    2.95×
Prepend   slang_persistent/ecollections_persistent          2.53×   27.04×  569.17×
Prepend   slang_persistent/clojure_persistent               5.90×  124.95× 3177.65×
Prepend   slang_persistent/scala_persistent                 0.22×    0.23×    0.15×

Append    slang_persistent/java_mutable                     0.25×    0.05×    0.03×
Append    slang_persistent/fjava_persistent                 5.18×    1.54×    1.02×
Append    slang_persistent/pcollections_persistent          2.71×    1.87×    2.22×
Append    slang_persistent/ecollections_persistent          7.93×   24.56×  560.25×
Append    slang_persistent/clojure_persistent               0.99×    0.23×    0.18×
Append    slang_persistent/scala_persistent                 0.81×    0.22×    0.16×

GroupBy   slang_persistent/java_mutable                     0.43×    0.65×    0.83×
GroupBy   slang_persistent/scala_persistent                 1.44×    1.05×    1.21×

Slice     slang_persistent/java_mutable                     0.54×    0.49×    0.45×
Slice     slang_persistent/pcollections_persistent          0.51×    0.42×    0.40×
Slice     slang_persistent/clojure_persistent               0.54×    0.45×    0.43×
Slice     slang_persistent/scala_persistent                 2.43×    5.61×    8.61×

Iterate   slang_persistent/java_mutable                     0.88×    0.45×    0.39×
Iterate   slang_persistent/fjava_persistent               368.25×  442.89×  289.54×
Iterate   slang_persistent/pcollections_persistent         14.61×   13.76×    9.42×
Iterate   slang_persistent/ecollections_persistent          2.35×    1.72×    1.52×
Iterate   slang_persistent/clojure_persistent               0.79×    1.03×    0.85×
Iterate   slang_persistent/scala_persistent                 1.53×    1.49×    0.90×
```

Note: Apparently `Java 9`  made some array copying optimizations (that made `append` slower than `prepend` in the first place), making `update`, `append` and `groupBy` slightly faster.

primitive:

``` java
Operation Ratio                                                32     1024    32768
Create    slang_persistent_int/java_mutable                 3.55×    1.71×    1.35×
Create    slang_persistent_int/java_mutable_boxed          26.16×   16.01×    7.09×
Create    slang_persistent_int/java_mutable_boxed_stream   39.31×   19.96×   11.01×
Create    slang_persistent_int/fjava_persistent           542.00×  400.65×  221.11×
Create    slang_persistent_int/pcollections_persistent    299.17×  470.54×  431.57×
Create    slang_persistent_int/ecollections_persistent      4.84×    1.76×    1.36×
Create    slang_persistent_int/clojure_persistent           2.23×   28.55×   14.71×
Create    slang_persistent_int/scala_persistent            32.91×   17.27×    7.39×

Tail      slang_persistent_int/java_mutable                 0.71×    0.56×    0.45×
Tail      slang_persistent_int/fjava_persistent            52.55×   50.20×   59.62×
Tail      slang_persistent_int/pcollections_persistent      5.82×    9.14×   13.73×
Tail      slang_persistent_int/ecollections_persistent     16.79×  156.62× 4898.01×
Tail      slang_persistent_int/clojure_persistent           0.92×    0.69×    0.68×
Tail      slang_persistent_int/scala_persistent             2.97×    4.87×    6.80×

Update    slang_persistent_int/java_mutable                 0.11×    0.05×    0.03×
Update    slang_persistent_int/fjava_persistent           124.86×  254.21×  215.66×
Update    slang_persistent_int/pcollections_persistent      2.30×    4.16×    4.83×
Update    slang_persistent_int/ecollections_persistent     13.02×  157.21× 2514.05×
Update    slang_persistent_int/clojure_persistent           0.80×    1.19×    1.07×
Update    slang_persistent_int/scala_persistent             0.97×    1.04×    1.11×

Map       slang_persistent_int/java_mutable                 2.01×    1.85×    1.58×
Map       slang_persistent_int/java_mutable_loop            0.48×    0.54×    0.43×
Map       slang_persistent_int/ecollections_persistent      1.07×    0.93×    0.84×
Map       slang_persistent_int/scala_persistent             0.94×    1.16×    1.37×

Filter    slang_persistent_int/java_mutable                 2.35×    1.63×    0.87×
Filter    slang_persistent_int/ecollections_persistent      2.00×    1.13×    0.70×
Filter    slang_persistent_int/scala_persistent             1.73×    1.43×    1.07×

Prepend   slang_persistent_int/java_mutable                 0.84×    0.78×   12.69×
Prepend   slang_persistent_int/fjava_persistent             3.12×    1.78×    1.25×
Prepend   slang_persistent_int/pcollections_persistent      2.05×    2.72×    3.18×
Prepend   slang_persistent_int/ecollections_persistent      5.50×   29.76×  612.04×
Prepend   slang_persistent_int/clojure_persistent          12.83×  137.53× 3416.94×
Prepend   slang_persistent_int/scala_persistent             0.49×    0.26×    0.17×

Append    slang_persistent_int/java_mutable                 0.23×    0.11×    0.03×
Append    slang_persistent_int/fjava_persistent             4.66×    3.00×    1.15×
Append    slang_persistent_int/pcollections_persistent      2.43×    3.64×    2.50×
Append    slang_persistent_int/ecollections_persistent      7.13×   47.83×  631.27×
Append    slang_persistent_int/clojure_persistent           0.89×    0.46×    0.20×
Append    slang_persistent_int/scala_persistent             0.73×    0.42×    0.18×

Slice     slang_persistent_int/java_mutable                 0.56×    0.47×    0.38×
Slice     slang_persistent_int/pcollections_persistent      0.52×    0.40×    0.34×
Slice     slang_persistent_int/clojure_persistent           0.55×    0.44×    0.37×
Slice     slang_persistent_int/scala_persistent             2.50×    5.42×    7.28×

Iterate   slang_persistent_int/java_mutable                 1.76×    0.88×    1.38×
Iterate   slang_persistent_int/fjava_persistent           737.64×  858.07× 1036.08×
Iterate   slang_persistent_int/pcollections_persistent     29.26×   26.65×   33.69×
Iterate   slang_persistent_int/ecollections_persistent      4.71×    3.33×    5.44×
Iterate   slang_persistent_int/clojure_persistent           1.58×    2.01×    3.05×
Iterate   slang_persistent_int/scala_persistent             3.06×    2.89×    3.23×
```

screenshot from 2016-09-10 14 20 50

and memory usages:

``` java
for `32` elements
  Java mutable @ ArrayList                            →   504 bytes
  Functional Java persistent @ Seq                    → 3,064 bytes
  PCollections persistent @ TreePVector               → 1,712 bytes
  Eclipse Collections persistent @ ImmutableArrayList →   496 bytes
  Clojure persistent @ PersistentVector               →   704 bytes
  Scala persistent @ Vector                           →   536 bytes
  Javaslang persistent @ Vector (Integer[])           →   544 bytes
  Javaslang persistent @ Vector (int[])               →   208 bytes
  Javaslang persistent @ Vector (byte[])              →   112 bytes

for `1024` elements
  Java mutable @ ArrayList                            →  19,064 bytes
  Functional Java persistent @ Seq                    → 104,760 bytes
  PCollections persistent @ TreePVector               →  55,984 bytes
  Eclipse Collections persistent @ ImmutableArrayList →  19,056 bytes
  Clojure persistent @ PersistentVector               →  20,504 bytes
  Scala persistent @ Vector                           →  19,736 bytes
  Javaslang persistent @ Vector (Integer[])           →  19,744 bytes
  Javaslang persistent @ Vector (int[])               →   4,816 bytes
  Javaslang persistent @ Vector (byte[])              →   1,744 bytes

for `32768` elements
  Java mutable @ ArrayList                            →   653,672 bytes
  Functional Java persistent @ Seq                    → 3,408,624 bytes
  PCollections persistent @ TreePVector               → 1,833,376 bytes
  Eclipse Collections persistent @ ImmutableArrayList →   653,664 bytes
  Clojure persistent @ PersistentVector               →   700,168 bytes
  Scala persistent @ Vector                           →   674,824 bytes
  Javaslang persistent @ Vector (Integer[])           →   674,832 bytes
  Javaslang persistent @ Vector (int[])               →   152,272 bytes
  Javaslang persistent @ Vector (byte[])              →    53,968 bytes
```

i.e. storing `byte`s in `functional java` would require `~63×` more storage.

### DIFF
--- a/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
@@ -5,7 +5,6 @@ import org.eclipse.collections.api.list.MutableList;
 import org.junit.Test;
 import org.openjdk.jmh.annotations.*;
 import org.openjdk.jmh.infra.Blackhole;
-import scala.Tuple2;
 import scala.collection.generic.CanBuildFrom;
 import scala.compat.java8.functionConverterImpls.FromJavaFunction;
 import scala.compat.java8.functionConverterImpls.FromJavaPredicate;
@@ -48,12 +47,12 @@ public class VectorBenchmark {
 
     public static void main(String... args) {
         JmhRunner.runDebugWithAsserts(CLASSES);
-        JmhRunner.runNormalNoAsserts(CLASSES, /*FUNCTIONAL_JAVA, PCOLLECTIONS, ECOLLECTIONS,*/ JAVA, CLOJURE, SCALA, JAVASLANG);
+        JmhRunner.runNormalNoAsserts(CLASSES, JAVA, FUNCTIONAL_JAVA, PCOLLECTIONS, ECOLLECTIONS, CLOJURE, SCALA, JAVASLANG);
     }
 
     @State(Scope.Benchmark)
     public static class Base {
-        @Param({ "100", "10000" })
+        @Param({ "10", "100", "1026" })
         public int CONTAINER_SIZE;
 
         int EXPECTED_AGGREGATE;
--- a/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
@@ -5,15 +5,17 @@ import org.eclipse.collections.api.list.MutableList;
 import org.junit.Test;
 import org.openjdk.jmh.annotations.*;
 import org.openjdk.jmh.infra.Blackhole;
+import scala.Tuple2;
 import scala.collection.generic.CanBuildFrom;
 import scala.compat.java8.functionConverterImpls.FromJavaFunction;
 import scala.compat.java8.functionConverterImpls.FromJavaPredicate;
 
 import java.util.Objects;
 import java.util.Random;
-import java.util.stream.Collectors;
 
 import static java.util.Arrays.asList;
+import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.toList;
 import static javaslang.JmhRunner.Includes.*;
 import static javaslang.JmhRunner.*;
 import static javaslang.collection.Collections.areEqual;
@@ -32,7 +34,10 @@ public class VectorBenchmark {
             Map.class,
             Filter.class,
             Prepend.class,
+            PrependAll.class,
             Append.class,
+            AppendAll.class,
+            Insert.class,
             GroupBy.class,
             Slice.class,
             Iterate.class
@@ -43,16 +48,17 @@ public class VectorBenchmark {
 
     public static void main(String... args) {
         JmhRunner.runDebugWithAsserts(CLASSES);
-        JmhRunner.runNormalNoAsserts(CLASSES, JAVA, FUNCTIONAL_JAVA, PCOLLECTIONS, ECOLLECTIONS, CLOJURE, SCALA, JAVASLANG);
+        JmhRunner.runNormalNoAsserts(CLASSES, /*FUNCTIONAL_JAVA, PCOLLECTIONS, ECOLLECTIONS,*/ JAVA, CLOJURE, SCALA, JAVASLANG);
     }
 
     @State(Scope.Benchmark)
     public static class Base {
-        @Param({ "32", "1024", "32768" /*, "1048576", "33554432", "1073741824" */ }) /* i.e. depth 1,2,3(,4,5,6) for a branching factor of 32 */
+        @Param({ "100", "10000" })
         public int CONTAINER_SIZE;
 
         int EXPECTED_AGGREGATE;
         Integer[] ELEMENTS;
+        int[] INT_ELEMENTS;
         int[] RANDOMIZED_INDICES;
 
         /* Only use this for non-mutating operations */
@@ -64,11 +70,14 @@ public class VectorBenchmark {
         clojure.lang.PersistentVector clojurePersistent;
         scala.collection.immutable.Vector<Integer> scalaPersistent;
         javaslang.collection.Vector<Integer> slangPersistent;
+        javaslang.collection.Vector<Integer> slangPersistentInt;
+        javaslang.collection.Vector<Byte> slangPersistentByte;
 
         @Setup
         public void setup() {
             final Random random = new Random(0);
             ELEMENTS = getRandomValues(CONTAINER_SIZE, false, random);
+            INT_ELEMENTS = ArrayType.asPrimitives(int.class, Array.of(ELEMENTS));
             RANDOMIZED_INDICES = shuffle(Array.range(0, CONTAINER_SIZE).toJavaStream().mapToInt(Integer::intValue).toArray(), random);
 
             EXPECTED_AGGREGATE = Array.of(ELEMENTS).reduce(JmhRunner::aggregate);
@@ -80,6 +89,11 @@ public class VectorBenchmark {
             clojurePersistent = create(clojure.lang.PersistentVector::create, javaMutable, v -> areEqual(v, javaMutable));
             scalaPersistent = create(v -> (scala.collection.immutable.Vector<Integer>) scala.collection.immutable.Vector$.MODULE$.apply(asScalaBuffer(v)), javaMutable, v -> areEqual(asJavaCollection(v), javaMutable));
             slangPersistent = create(javaslang.collection.Vector::ofAll, javaMutable, v -> areEqual(v, javaMutable));
+            slangPersistentInt = create(v -> javaslang.collection.Vector.ofAll(INT_ELEMENTS), javaMutable, v -> areEqual(v, javaMutable) && (v.trie.type.type() == int.class));
+
+            final byte[] BYTE_ELEMENTS = new byte[CONTAINER_SIZE];
+            random.nextBytes(BYTE_ELEMENTS);
+            slangPersistentByte = create(v -> javaslang.collection.Vector.ofAll(BYTE_ELEMENTS), javaMutable, v -> areEqual(v, Array.ofAll(BYTE_ELEMENTS)) && (v.trie.type.type() == byte.class));
         }
     }
 
@@ -88,6 +102,23 @@ public class VectorBenchmark {
         @Benchmark
         public Object java_mutable() {
             final java.util.List<Integer> values = new java.util.ArrayList<>(java.util.Arrays.asList(ELEMENTS));
+            assert areEqual(values, javaMutable);
+            return values;
+        }
+
+        @Benchmark
+        public Object java_mutable_boxed() {
+            final java.util.List<Integer> values = new java.util.ArrayList<>();
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values.add(INT_ELEMENTS[i]);
+            }
+            assert areEqual(values, javaMutable);
+            return values;
+        }
+
+        @Benchmark
+        public Object java_mutable_boxed_stream() {
+            final java.util.List<Integer> values = java.util.Arrays.stream(INT_ELEMENTS).boxed().collect(toList());
             assert areEqual(values, javaMutable);
             return values;
         }
@@ -131,6 +162,13 @@ public class VectorBenchmark {
         public Object slang_persistent() {
             final javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.ofAll(javaMutable);
             assert areEqual(values, javaMutable);
+            return values;
+        }
+
+        @Benchmark
+        public Object slang_persistent_int() {
+            final javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.ofAll(INT_ELEMENTS);
+            assert (values.trie.type.type() == int.class) && areEqual(values, javaMutable);
             return values;
         }
     }
@@ -198,9 +236,7 @@ public class VectorBenchmark {
             }
 
             @TearDown(Level.Invocation)
-            public void tearDown() {
-                javaMutable.clear();
-            }
+            public void tearDown() { javaMutable.clear(); }
         }
 
         @Benchmark
@@ -265,6 +301,16 @@ public class VectorBenchmark {
         @Benchmark
         public Object slang_persistent() {
             javaslang.collection.Vector<Integer> values = slangPersistent;
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                values = values.tail();
+            }
+            assert values.isEmpty();
+            return values;
+        }
+
+        @Benchmark
+        public Object slang_persistent_int() {
+            javaslang.collection.Vector<Integer> values = slangPersistentInt;
             for (int i = 0; i < CONTAINER_SIZE; i++) {
                 values = values.tail();
             }
@@ -359,9 +405,7 @@ public class VectorBenchmark {
             }
 
             @TearDown(Level.Invocation)
-            public void tearDown() {
-                javaMutable.clear();
-            }
+            public void tearDown() { javaMutable.clear(); }
         }
 
         @Benchmark
@@ -435,9 +479,20 @@ public class VectorBenchmark {
             assert values.forAll(e -> e == 0);
             return values;
         }
+
+        @Benchmark
+        public Object slang_persistent_int() {
+            javaslang.collection.Vector<Integer> values = slangPersistentInt;
+            for (int i : RANDOMIZED_INDICES) {
+                values = values.update(i, 0);
+            }
+            assert (values.trie.type.type() == int.class) && values.forAll(e -> e == 0);
+            return values;
+        }
     }
 
     public static class Map extends Base {
+        final CanBuildFrom canBuildFrom = scala.collection.immutable.Vector.canBuildFrom();
         private static int mapper(int i) { return i + 1; }
 
         @Benchmark
@@ -452,7 +507,7 @@ public class VectorBenchmark {
 
         @Benchmark
         public Object java_mutable() {
-            final java.util.List<Integer> values = javaMutable.stream().map(Map::mapper).collect(Collectors.toList());
+            final java.util.List<Integer> values = javaMutable.stream().map(Map::mapper).collect(toList());
             assert areEqual(values, Array.of(ELEMENTS).map(Map::mapper));
             return values;
         }
@@ -466,7 +521,6 @@ public class VectorBenchmark {
 
         @Benchmark
         public Object scala_persistent() {
-            final CanBuildFrom canBuildFrom = scala.collection.immutable.Vector.canBuildFrom();
             final scala.collection.immutable.Vector<Integer> values = (scala.collection.immutable.Vector<Integer>) scalaPersistent.map(new FromJavaFunction<>(Map::mapper), canBuildFrom);
             assert areEqual(asJavaCollection(values), Array.of(ELEMENTS).map(Map::mapper));
             return values;
@@ -478,6 +532,13 @@ public class VectorBenchmark {
             assert areEqual(values, Array.of(ELEMENTS).map(Map::mapper));
             return values;
         }
+
+        @Benchmark
+        public Object slang_persistent_int() {
+            final javaslang.collection.Vector<Integer> values = slangPersistentInt.map(Map::mapper);
+            assert areEqual(values, Array.of(ELEMENTS).map(Map::mapper));
+            return values;
+        }
     }
 
     public static class Filter extends Base {
@@ -485,7 +546,7 @@ public class VectorBenchmark {
 
         @Benchmark
         public Object java_mutable() {
-            final java.util.List<Integer> someValues = javaMutable.stream().filter(Filter::isOdd).collect(Collectors.toList());
+            final java.util.List<Integer> someValues = javaMutable.stream().filter(Filter::isOdd).collect(toList());
             assert areEqual(someValues, Array.of(ELEMENTS).filter(Filter::isOdd));
             return someValues;
         }
@@ -508,6 +569,13 @@ public class VectorBenchmark {
         public Object slang_persistent() {
             final javaslang.collection.Vector<Integer> someValues = slangPersistent.filter(Filter::isOdd);
             assert areEqual(someValues, Array.of(ELEMENTS).filter(Filter::isOdd));
+            return someValues;
+        }
+
+        @Benchmark
+        public Object slang_persistent_int() {
+            final javaslang.collection.Vector<Integer> someValues = slangPersistentInt.filter(Filter::isOdd);
+            assert (someValues.trie.type.type() == int.class) && areEqual(someValues, Array.of(ELEMENTS).filter(Filter::isOdd));
             return someValues;
         }
     }
@@ -588,6 +656,29 @@ public class VectorBenchmark {
             assert areEqual(values.reverse(), javaMutable);
             return values;
         }
+
+        @Benchmark
+        public Object slang_persistent_int() {
+            javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.ofAll(ELEMENTS[0]);
+            for (int i = 1; i < ELEMENTS.length; i++) {
+                values = values.prepend(ELEMENTS[i]);
+            }
+            assert (values.trie.type.type() == int.class) && areEqual(values.reverse(), javaMutable);
+            return values;
+        }
+    }
+
+    public static class PrependAll extends Base {
+        @Benchmark
+        public void slang_persistent(Blackhole bh) {
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                final java.util.List<Integer> front = javaMutable.subList(0, i);
+                final javaslang.collection.Vector<Integer> back = slangPersistent.slice(i, CONTAINER_SIZE);
+                final javaslang.collection.Vector<Integer> values = back.prependAll(front);
+                assert areEqual(values, javaMutable);
+                bh.consume(values);
+            }
+        }
     }
 
     /** Add all elements (one-by-one, as we're not testing bulk operations) */
@@ -663,23 +754,64 @@ public class VectorBenchmark {
             assert areEqual(values, javaMutable);
             return values;
         }
+
+        @Benchmark
+        public Object slang_persistent_int() {
+            javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.ofAll(INT_ELEMENTS[0]);
+            for (int i = 1; i < INT_ELEMENTS.length; i++) {
+                values = values.append(INT_ELEMENTS[i]);
+            }
+            assert (values.trie.type.type() == int.class) && areEqual(values, javaMutable);
+            return values;
+        }
+    }
+
+    public static class AppendAll extends Base {
+        final CanBuildFrom canBuildFrom = scala.collection.immutable.Vector.canBuildFrom();
+
+        @Benchmark
+        public void scala_persistent(Blackhole bh) {
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                final scala.collection.immutable.Vector<Integer> front = scalaPersistent.slice(0, i);
+                final java.util.List<Integer> back = javaMutable.subList(i, CONTAINER_SIZE);
+                final scala.collection.immutable.Vector<Integer> values = front.$plus$plus(asScalaBuffer(back), (CanBuildFrom<scala.collection.immutable.Vector<Integer>, Integer, scala.collection.immutable.Vector<Integer>>) canBuildFrom);
+                assert areEqual(asJavaCollection(values), javaMutable);
+                bh.consume(values);
+            }
+        }
+
+        @Benchmark
+        public void slang_persistent(Blackhole bh) {
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                final javaslang.collection.Vector<Integer> front = slangPersistent.slice(0, i);
+                final java.util.List<Integer> back = javaMutable.subList(i, CONTAINER_SIZE);
+                final javaslang.collection.Vector<Integer> values = front.appendAll(back);
+                assert areEqual(values, javaMutable);
+                bh.consume(values);
+            }
+        }
+    }
+
+    public static class Insert extends Base {
+        @Benchmark
+        public void slang_persistent(Blackhole bh) {
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                final Vector<Integer> values = slangPersistent.insert(i, 0);
+                assert values.size() == CONTAINER_SIZE + 1;
+                bh.consume(values);
+            }
+        }
     }
 
     public static class GroupBy extends Base {
         @Benchmark
-        public Object java_mutable() {
-            return javaMutable.stream().collect(Collectors.groupingBy(Integer::bitCount));
-        }
+        public Object java_mutable() { return javaMutable.stream().collect(groupingBy(Integer::bitCount)); }
 
         @Benchmark
-        public Object scala_persistent() {
-            return scalaPersistent.groupBy(func(Integer::bitCount));
-        }
+        public Object scala_persistent() { return scalaPersistent.groupBy(func(Integer::bitCount)); }
 
         @Benchmark
-        public Object slang_persistent() {
-            return slangPersistent.groupBy(Integer::bitCount);
-        }
+        public Object slang_persistent() { return slangPersistent.groupBy(Integer::bitCount); }
     }
 
     /** Consume the vector one-by-one, from the front and back */
@@ -733,30 +865,24 @@ public class VectorBenchmark {
                 bh.consume(values);
             }
         }
+
+        @Benchmark
+        public void slang_persistent_int(Blackhole bh) {
+            javaslang.collection.Vector<Integer> values = this.slangPersistentInt;
+            while (!values.isEmpty()) {
+                values = values.slice(1, values.size());
+                values = values.slice(0, values.size() - 1);
+                bh.consume(values);
+            }
+        }
     }
 
     /** Sequential access for all elements */
     public static class Iterate extends Base {
-        @State(Scope.Thread)
-        public static class Initialized {
-            final java.util.ArrayList<Integer> javaMutable = new java.util.ArrayList<>();
-
-            @Setup(Level.Invocation)
-            public void initializeMutable(Base state) {
-                java.util.Collections.addAll(javaMutable, state.ELEMENTS);
-                assert areEqual(javaMutable, asList(state.ELEMENTS));
-            }
-
-            @TearDown(Level.Invocation)
-            public void tearDown() {
-                javaMutable.clear();
-            }
-        }
-
         @Benchmark
-        public int java_mutable(Initialized state) {
+        public int java_mutable() {
             int aggregate = 0;
-            for (final java.util.Iterator<Integer> iterator = state.javaMutable.iterator(); iterator.hasNext(); ) {
+            for (final java.util.Iterator<Integer> iterator = javaMutable.iterator(); iterator.hasNext(); ) {
                 aggregate ^= iterator.next();
             }
             assert aggregate == EXPECTED_AGGREGATE;
@@ -821,6 +947,19 @@ public class VectorBenchmark {
             }
             assert aggregate == EXPECTED_AGGREGATE;
             return aggregate;
+        }
+
+        @Benchmark
+        public int slang_persistent_int() {
+            final int[] aggregate = { 0 };
+            slangPersistentInt.trie.<int[]> visit((ordinal, leaf, start, end) -> {
+                for (int i = start; i < end; i++) {
+                    aggregate[0] ^= leaf[i];
+                }
+                return -1;
+            });
+            assert aggregate[0] == EXPECTED_AGGREGATE;
+            return aggregate[0];
         }
     }
 }
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -34,6 +34,7 @@ def generateMainClasses(): Unit = {
   genAPI()
   genFunctions()
   genTuples()
+  genArrayTypes()
 
   /**
    * Generator of Match
@@ -2273,6 +2274,76 @@ def generateMainClasses(): Unit = {
 
             ${(1 to N).gen(genSeqMethod)("\n\n")}
 
+        }
+      """
+    }
+  }
+
+  /**
+    * Generator of javaslang.collection.*ArrayType
+    */
+  def genArrayTypes(): Unit = {
+    val types = Map(
+      "Object" -> "Object",
+      "boolean" -> "Boolean",
+      "byte" -> "Byte",
+      "char" -> "Character",
+      "double" -> "Double",
+      "float" -> "Float",
+      "int" -> "Integer",
+      "long" -> "Long",
+      "short" -> "Short"
+    ) // note: there is no void[] in Java
+
+    types.keys.foreach { arrayType =>
+      genJavaslangFile("javaslang.collection", s"${arrayType.capitalize}ArrayType")(genArrayType(arrayType))
+    }
+
+    /*
+     * Generates *ArrayType
+     */
+    def genArrayType(arrayType: String)(im: ImportManager, packageName: String, className: String): String = {
+      val wrapperType = types(arrayType)
+      val cast = if (wrapperType != "Object") s"($wrapperType)" else ""
+
+      xs"""
+         /**
+         * $arrayType[] helper to replace reflective array access.
+         * @author Pap Lőrinc
+         * @since 2.1.0
+         */
+        final class $className extends ArrayType<$wrapperType> {
+            private static final long serialVersionUID = 1L;
+            static final $className INSTANCE = new $className();
+            static final $arrayType[] EMPTY = new $arrayType[0];
+
+            private static $arrayType[] cast(Object array) { return ($arrayType[]) array; }
+
+            @Override
+            Class<$wrapperType> type() { return $arrayType.class; }
+
+            @Override
+            $arrayType[] empty() { return EMPTY; }
+
+            @Override
+            int lengthOf(Object array) { return (array == null) ? 0 : cast(array).length; }
+
+            @Override
+            $wrapperType getAt(Object array, int index) { return cast(array)[index]; }
+
+            @Override
+            void setAt(Object array, int index, Object value) { cast(array)[index] = $cast value; }
+
+            @Override
+            Object copy(Object array, int arraySize, int sourceFrom, int destinationFrom, int size) {
+                if (size == 0) {
+                    return new $arrayType[arraySize];
+                } else {
+                    final $arrayType[] result = new $arrayType[arraySize];
+                    System.arraycopy(array, sourceFrom, result, destinationFrom, size); /* has to be near the object allocation to avoid zeroing out the array */
+                    return result;
+                }
+            }
         }
       """
     }
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -2299,20 +2299,18 @@ def generateMainClasses(): Unit = {
       genJavaslangFile("javaslang.collection", s"${arrayType.capitalize}ArrayType")(genArrayType(arrayType))
     }
 
-    /*
-     * Generates *ArrayType
-     */
     def genArrayType(arrayType: String)(im: ImportManager, packageName: String, className: String): String = {
       val wrapperType = types(arrayType)
-      val cast = if (wrapperType != "Object") s"($wrapperType)" else ""
+      val cast = if (wrapperType != "Object") s" ($wrapperType)" else ""
 
       xs"""
-         /**
+        /**
          * $arrayType[] helper to replace reflective array access.
+         *
          * @author Pap Lőrinc
          * @since 2.1.0
          */
-        final class $className extends ArrayType<$wrapperType> {
+        final class $className implements ArrayType<$wrapperType>, ${im.getType("java.io.Serializable")} {
             private static final long serialVersionUID = 1L;
             static final $className INSTANCE = new $className();
             static final $arrayType[] EMPTY = new $arrayType[0];
@@ -2320,22 +2318,22 @@ def generateMainClasses(): Unit = {
             private static $arrayType[] cast(Object array) { return ($arrayType[]) array; }
 
             @Override
-            Class<$wrapperType> type() { return $arrayType.class; }
+            public Class<$wrapperType> type() { return $arrayType.class; }
 
             @Override
-            $arrayType[] empty() { return EMPTY; }
+            public $arrayType[] empty() { return EMPTY; }
 
             @Override
-            int lengthOf(Object array) { return (array == null) ? 0 : cast(array).length; }
+            public int lengthOf(Object array) { return (array == null) ? 0 : cast(array).length; }
 
             @Override
-            $wrapperType getAt(Object array, int index) { return cast(array)[index]; }
+            public $wrapperType getAt(Object array, int index) { return cast(array)[index]; }
 
             @Override
-            void setAt(Object array, int index, Object value) { cast(array)[index] = $cast value; }
+            public void setAt(Object array, int index, Object value) { cast(array)[index] =$cast value; }
 
             @Override
-            Object copy(Object array, int arraySize, int sourceFrom, int destinationFrom, int size) {
+            public Object copy(Object array, int arraySize, int sourceFrom, int destinationFrom, int size) {
                 if (size == 0) {
                     return new $arrayType[arraySize];
                 } else {
--- a/javaslang/src-gen/main/java/javaslang/collection/ObjectArrayType.java
+++ b/javaslang/src-gen/main/java/javaslang/collection/ObjectArrayType.java
@@ -1,0 +1,49 @@
+/*     / \____  _    _  ____   ______  / \ ____  __    _______
+ *    /  /    \/ \  / \/    \ /  /\__\/  //    \/  \  //  /\__\   JΛVΛSLΛNG
+ *  _/  /  /\  \  \/  /  /\  \\__\\  \  //  /\  \ /\\/ \ /__\ \   Copyright 2014-2016 Javaslang, http://javaslang.io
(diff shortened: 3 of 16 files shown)


danieldietrich:
Wow - this is a really interesting change - _The Javaslang_ way. Automatically switching to primitive representations, I think the Java world hasn't seen that to date, right?

Will review all open Vector PRs today...

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -30,18 +30,18 @@
...
+//            Tail.class,
+//            Get.class,
+//            Update.class,
+//            Map.class,
+//            Filter.class,
+//            Prepend.class,
+//            Append.class,
+            GroupBy.class
+//            Slice.class,
+//            Iterate.class
```
danieldietrich: Were these accidentally committed?

paplorinc: fixed, thanks

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -177,11 +177,18 @@ T get(int index) {
...
     @SuppressWarnings("unchecked")
     T[] getLeaf(int index) {
-        Object[] leaf = array;
-        for (int shift = depthShift; shift > 0; shift -= BRANCHING_BASE) {
-            leaf = getAt(leaf, digit(offset + index, shift));
+        if (depthShift == 0) {
+            return (T[]) array;
+        } else if (depthShift == BRANCHING_BASE) {
+            return getAt(array, firstDigit(offset + index, depthShift));
+        } else {
```
danieldietrich: The else branch covers `depthShift > 0 && (depthShift < BRANCHING_BASE || depthShift > BRANCHING_BASE)`, right?

Just asking to get sure that `depthShift` can be less than `BRANCHING_BASE`...

javaslang/src/main/java/javaslang/collection/LeafType.java
```diff
@@ -0,0 +1,258 @@
...
+    private static final long serialVersionUID = 1L;
+    final Class<T> type;
+
+    PrimitiveArray(Class<T> type) { this.type = type; }
+
+    @Override
+    Class<T> type() { return type; }
+
+    @Override
+    Object empty() { return java.lang.reflect.Array.newInstance(type, 0); }
```
ruslansennov: `java.lang.reflect.Array` is GWT-incompatible

paplorinc: This is a backup plan (i.e. we could spell out every primitive), though I'm not even sure whether this would work/is necessary in Javascipt.

@ruslansennov, what are your thoughts on the rest of the PRs?

ruslansennov: > This is a backup plan (i.e. we could spell out every primitive), though I'm not even sure whether this would work/is necessary in Javascipt.

It seems you should create a lot of implementations like `IntArray`?

> what are your thoughts on the rest of the PRs?

I'm very sorry, I can't find a few **consecutive** hours for review your previous PR :(
I still doubt in the effectiveness of tail and head arrays in the `Vector` (because of significantly code complication), but benchmarks are more than convincing :)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -914,7 +914,7 @@ public int lastIndexOf(T element, int end) {
             if (ordinal == 0) {
                 newLeading.val = copy;
             } else if (ordinal <= newMiddle.length) {
-                newMiddle[ordinal - 1] = copy;
+                newMiddle[ordinal - 1] = copy);
```
danieldietrich: Huh, where does this ')' come from? Does it really compile!?

danieldietrich:
@paplorinc 

I see your effort and the improvements. The implementation is very low level and got more complicated because of the heading and trailing arrays. But Vector will be very useful as general purpose collection.

Therefore I will merge this change. Please resolve the conflicts and I will pull it in.

Do I also have to pull in "Vector optimizations #1522" or does this PR also contain all changes of #1522?

danieldietrich:
What about the reflection stuff? Will Vector work with GWT?
If not, we should discuss how to proceed further.

paplorinc:
Will do my best to make it GWT compliant also, but I might need help from @ruslansennov, as currently the GWT tests are failing and I'm not sure how to decipher the error message :/.

danieldietrich:
@paplorinc it is important that GWT follows Javaslang core, not the other direction.
We write efficient code. If we can omit GWT incompatible API, we will to it. But we do not write slower/worse code in order to make GWT work. Just saying.
Pleade make the GWT-dpecific changes in a separate commit for better review-experience.

paplorinc:
@danieldietrich, will rebase, once #1522 is merged :)

paplorinc:
Pushed, but GWT complains again. I'm not sure how to fix those errors, any help?

paplorinc:
> Also I removed dead code: `LeafType.getAt(...)` and `LeafType.isPrimitive()` weren't used.

Not sure what you mean by dead code, every method is used and almost every line is tested (except for some redundant boundary checks in `BMT`, that were already checked in `Vector`):
```java
Element         Class       Method          Line
BitMappedTrie   100% (2/2)	100% (33/33)	97%  (152/156)
LeafType        100% (1/1)	100% (17/17)	100% (52/52)
Vector          100% (1/1)	100% (130/130)	100% (329/329)
```
* getAt usage: https://github.com/paplorinc/javaslang/blob/Primitivization/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java#L178
* isPrimitive usage: https://github.com/paplorinc/javaslang/blob/Primitivization/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java#L317

> I moved it to `VectorModule` because it is currently only used by `Vector`.

It's actually also used by `BitMappedTrie` and `VectorBenchmark`.
Should I really merge `Vector` with `BitMappedTrie` and `LeafType` (they're only meant to complement `Vector`)?
It would result in a huge, `~1900` line behemoth, that would contain multiple, separate responsibilities.

If you think this change is too complex, I don't mind separating it into more commits/files, but I need to know what is hard to understand. :)
I can also add more documentation or explain it here, just let me know where the difficulty is.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -41,32 +41,32 @@
     @SuppressWarnings("unchecked")
     static <T> BitMappedTrie<T> empty() { return (BitMappedTrie<T>) EMPTY; }
 
-    private final Object[] array;
```
paplorinc: intermediary commit, that removes all `Object[]` references, changing them to `Object`s and accessing them in a single place using `getAt`, `setAt` and `lengthOf`, so that these methods can be specialized in the next commit to every primitive.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -164,54 +164,53 @@ private boolean arePointingToSameLeaf(int i, int j) {
...
+            Object array = root;
             for (int shift = depthShift - BRANCHING_BASE; shift >= BRANCHING_BASE; shift -= BRANCHING_BASE) {
                 final int offset = digit(index, shift);
 
-                final Object previous = array[previousIndex];
-                final Object[] newNode = node.apply(previous, offset);
-                array[previousIndex] = newNode;
+                final Object previous = getAt(array, previousIndex);
+                final Object newNode = node.apply(previous, offset);
+                setAt(array, previousIndex, newNode);
```
paplorinc: since we have `Object`s everywhere, array notation cannot be used anymore, since the `Object[]` type cannot refer to primitive arrays, but `Object` can

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -133,6 +159,13 @@ public Object slang_persistent() {
             assert areEqual(values, javaMutable);
             return values;
         }
+
+        @Benchmark
+        public Object slang_persistent_int() {
+            final javaslang.collection.Vector<Integer> values = javaslang.collection.Vector.ofAll(INT_ELEMENTS);
+            assert (values.trie.type.type() == int.class) && areEqual(values, javaMutable);
```
paplorinc: these asserts also make sure that we actually have `int[]`  inside

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -61,59 +65,73 @@ private static int treeSize(int branchCount, int depthShift) {
...
-            return new BitMappedTrie<>(array, 0, size, shift);
+            return new BitMappedTrie<>(type, array, 0, size, shift);
         }
     }
 
     BitMappedTrie<T> prepend(T leading) {
-        final int newSize = length() + 1;
-        if (length() == 0) {
-            return new BitMappedTrie<>(asArray(leading), offset, newSize, depthShift);
+        if (cannotAdd(leading)) {
```
paplorinc: if we can't add an element to the `Vector` (i.e. `null` or `String` to a primitive collection), copy everything once, and add it to that instead.

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -29,6 +29,20 @@ public void shouldCreateAndGet() {
             for (int j = 0; j < actual.size(); j++) {
                 assertThat(expected.get(j)).isEqualTo(actual.get(j));
             }
+
+            final Vector<Integer> actualInt = Vector.ofAll(LeafType.<int[]> asPrimitives(int.class, expected));
```
paplorinc: create primitive int[] array and test that it contains the correct elements and that it's stored as `int[]`, not `Object[]`

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -29,6 +29,20 @@ public void shouldCreateAndGet() {
...
+            assert (i == 0) || (actualInt.trie.type.type() == int.class);
+            assertAreEqual(expected, actualInt);
+
+            final Seq<Character> expectedChar = expected.map(v -> (char) v.intValue());
```
paplorinc: same for `char[]`, which doesn't have an optimized `LeafType` yet, just a generic one via reflection (currently only `int[]` arrays are optimized speed-wize, the rest are all delegated to `java.lang.reflect.Array`).
Each primitive can be optimized on demand, as done in `IntArray`.
@danieldietrich, would you like me to add other primitive speed optimizations in this commit also, or is `int[]` enough for now (the rest are also stored as primitives, but their usage is slower because of the reflective call)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -269,20 +268,20 @@ private Vector(BitMappedTrie<T> trie) {
      */
     public static Vector<Short> ofAll(short... array) {
         Objects.requireNonNull(array, "array is null");
-        return ofAll(Iterator.ofAll(array));
+        return ofAll(BitMappedTrie.ofAll(array));
     }
 
     public static Vector<Character> range(char from, char toExclusive) {
-        return ofAll(Iterator.range(from, toExclusive));
+        return ofAll(LeafType.<char[]> asPrimitives(char.class, Iterator.range(from, toExclusive)));
```
paplorinc: ranges should be stored in space and iteration friendly primitive arrays <3

javaslang/src/main/java/javaslang/collection/LeafType.java
```diff
@@ -5,99 +5,238 @@
...
+            final int[] result = new int[arraySize];
+            System.arraycopy(array, sourceFrom, result, destinationFrom, size); // has to be near the object allocation to avoid zeroing out the array
+            return result;
+        }
+    }
+}
+
+@SuppressWarnings("unchecked")
+@GwtIncompatible
+final class PrimitiveArray<T> extends LeafType<T> {
```
paplorinc: generic converter for non-optimized array access (only needed if we don't privide converters for every primitive)

javaslang/src/main/java/javaslang/collection/LeafType.java
```diff
@@ -5,99 +5,238 @@
...
+}
 
-    /** array access, avoiding call-site casts */
-    @SuppressWarnings("unchecked")
-    static <T> T getAt(Object array, int index) { return (T) ((Object[]) array)[index]; }
-    @SuppressWarnings("unchecked")
-    static void setAt(Object array, int index, Object value) { ((Object[]) array)[index] = value; }
-    static int lengthOf(Object array) {
-        return ((Object[]) array).length;
+final class IntArray extends LeafType<Integer> {
```
paplorinc: `int`s are used very often, let's optimize their access (storage is the same in all cases)

javaslang/src/main/java/javaslang/collection/LeafType.java
```diff
@@ -5,99 +5,238 @@
...
+        for (int j = 0; j < WRAPPERS.length; j++) {
+            if (wrapper == WRAPPERS[j]) {
+                return j;
+            }
+        }
+        return -1;
+    }
+}
+
+final class ObjectArray extends LeafType<Object> {
```
paplorinc: general `Object[]` array access (including e.g. `Integer[]`)

paplorinc:
@danieldietrich, @zsolt-donca, added comments throughout the codebase to help in reviewing.

danieldietrich:
No, please do not change the implementation. IntelliJ showed a 'false positive'. The editor stated that these methods are not used.

I need some more time to review.

I see that from the performance perspective the code is fast. But on the abstraction level we did not find the right way yet I think. Future and Promise are a good example. I looked at the Scala code of Viktor Klang. There is a very small core where synchronization logic takes place (in Javaslang it is FutureImpl). It is like an onion. This core is very small. The outer layers don't know anything about the core other than it's public interface (take a look at Future). So it is ensured that no synchronization logic leaks to the outside. It is all very elegant and compositional. Some of the best code I've ever seen.

Our Vector works different. It is so highly optimized that the logic of the BMT leaks to the outside. I'm also not sure if Vector should know about the Leaf types. For reasons of stability and maintainability of the library this is really evil. The complexity increases (more than linear). I'm not sure if I'm able to maintain it or to add new functionality.

I need some additional time to look over the code. But for now I see no reason to change anything.

danieldietrich:
Btw - I intentionally did no special handling of primitives throughout Javaslang. It was a design decision. I did not want to fix the 'errors' of the Java language. With Project Valhalla we probably get primitive generics. No boxing will take place any more. Then the 'problem' will disappear automatically.

paplorinc:
> It is so highly optimized that the logic of the BMT leaks to the outside. 

I don't think it's fair to compare it against something conceptually as simple as a `Future`, but if you think the impl is still leaking, please advise on how to make it better, simpler!
`Vector` basically just delegates now, there is no BMT logic there.

> I did not want to fix the 'errors' of the Java language. 

I think this whole library's sole purpose is exactly that, to fix some of Java's shortcomings.

> With Project Valhalla we probably get primitive generics. 

`JDK 9` just got delayed again [to mid 2017](http://www.theregister.co.uk/2016/09/14/jdk_9_release_delay), version `10` probably won't be widespread enough for another `~5` years. And even when it will be, will `Javalang` simply deprecate `Java 8` & `9` support (it probably won't be fully backwards compatible)?
I think people need a memory efficient, fast, random access, persistent collection now.

javaslang/src/main/java/javaslang/collection/LeafType.java
danieldietrich: @paplorinc I've slept over the topic. What about generating the converters for every primitive type?
I would rename LeafType to PrimitiveArray and let IntArray, CharArray, ... extend it. All these types reside in a package-private final class Arrays.
I think we could reuse these types also in other collections, starting with Array :)

danieldietrich:
@paplorinc you are right. If I think there is sth leaking, I will provide a better solution. (Will be offline today most of the time...)

javaslang/src/main/java/javaslang/collection/LeafType.java
paplorinc: Interesting idea, I like it, I will try to come up with such a construct :)

paplorinc: @danieldietrich, should I put these in a separate module, i.e. `javaslang-primitive`?

danieldietrich: @paplorinc for now I see no reason to do that. the types should be package-private available for collection for now. Arrays are somehow collections. It is sufficient if our Javaslang collections use the primitive arrays. Our Array will abstract over all of the primitve + object arrays. It is a sell-point for our Array to not give them to the outside.

javaslang/generator/Generator.scala
```diff
@@ -2228,6 +2229,76 @@ def generateMainClasses(): Unit = {
...
+      "Object" -> "Object",
+      "boolean" -> "Boolean",
+      "byte" -> "Byte",
+      "char" -> "Character",
+      "double" -> "Double",
+      "float" -> "Float",
+      "int" -> "Integer",
+      "long" -> "Long",
+      "short" -> "Short"
+    ) // note: there is no void[] in Java
```
paplorinc: funny, I didn't know that :)

paplorinc:
@danieldietrich, pushed version with exhaustive primitive array helper generation - making it GWT compatible also, as it doesn't need reflection anymore.

danieldietrich:
Will check the code tomorrow!

paplorinc:
@danieldietrich, would you like me to change anything in this commit? :)

danieldietrich:
@paplorinc Will get my hands on the new version this evening. Sorry for the delay - so many things to do in parallel. The new Vector will be awesome!

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -151,67 +169,66 @@ private boolean arePointingToSameLeaf(int i, int j) {
...
-    private static <T> BitMappedTrie<T> collapsed(Object[] array, int offset, int length, int shift) {
+    private static <T> BitMappedTrie<T> collapsed(ArrayType<T> type, Object array, int offset, int length, int shift) {
         for (; shift > 0; shift -= BRANCHING_BASE) {
-            final int skippedElements = array.length - 1;
+            final int skippedElements = obj().lengthOf(array) - 1;
             if (skippedElements != digit(offset, shift)) {
                 break;
             }
-            array = getAt(array, skippedElements);
+            array = obj().getAt(array, skippedElements);
```
danieldietrich: We should not use (or even know about) `obj()` here (in BMT) at all. Instead use a static import of the ObjectArrayType methods `getAt` and `setAt` and use those here. We do not need the cast to `ArrayType<T>` because we deal with Object here.

(If we can't statically import them, we should use the qualified call ObjectArrayType.getAt...)

---

But it is still not clear to me why we are using ObjectArrayType directly. Can't the values be primitive?

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -151,67 +169,66 @@ private boolean arePointingToSameLeaf(int i, int j) {
...
             offset -= treeSize(skippedElements, shift);
         }
-        return new BitMappedTrie<>(array, offset, length, shift);
+        return new BitMappedTrie<>(type, array, offset, length, shift);
```
danieldietrich: Please rename `type` to `arrayType`. It was not obvious to me...

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -151,67 +169,66 @@ private boolean arePointingToSameLeaf(int i, int j) {
...
+                final Object previous = obj().getAt(array, previousIndex);
+                final Object newNode = node.apply(previous, offset);
+                obj().setAt(array, previousIndex, newNode);
 
                 previousIndex = offset;
                 array = newNode;
             }
 
-            array[previousIndex] = leaf.apply(array[previousIndex], lastDigit(index));
+            obj().setAt(array, previousIndex, leaf.apply(obj().getAt(array, previousIndex), lastDigit(index)));
```
danieldietrich:
```java
final Object previous = obj().getAt(array, previousIndex);
final Object newLeaf = leaf.apply(previous, offset);
/*ObjectArrayType.*/setAt(array, previousIndex, newLeaf);
```
<small>*) is the name _newLeaf_ correct?</small>

javaslang/src/main/java/javaslang/collection/ArrayType.java
```diff
@@ -0,0 +1,147 @@
...
+            type.setAt(results, i, array[i]);
+        }
+        return (T) results;
+    }
+
+    /* convert to primitive */
+    private static final Class<?>[] WRAPPERS = { Boolean.class, Byte.class, Character.class, Double.class, Float.class, Integer.class, Long.class, Short.class, Void.class };
+    private static final Class<?>[] PRIMITIVES = { boolean.class, byte.class, char.class, double.class, float.class, int.class, long.class, short.class, void.class };
+
+    static Class<?> asPrimitive(Class<?> wrapper) {
```
danieldietrich: Should be removed completely. Only use is in BMT.cannotAdd. See my comment there.

javaslang/src/main/java/javaslang/collection/ArrayType.java
```diff
@@ -0,0 +1,147 @@
...
+        final int i = primitiveIndex(wrapper);
+        return (i < 0) ? wrapper
+                       : PRIMITIVES[i];
+    }
+
+    private static int primitiveIndex(Class<?> wrapper) { /* linear search is faster than binary search here */
```
danieldietrich: Only use in asPrimitive(Class). Can be removed too.

javaslang/src/main/java/javaslang/collection/ArrayType.java
```diff
@@ -0,0 +1,147 @@
...
+        for (int i = 0; i < length; i++) {
+            array[i] = it.next();
+        }
+        return array;
+    }
+
+    @SuppressWarnings("unchecked")
+    static <T> T asPrimitives(Class<?> primitiveClass, Iterable<?> values) {
+        final Object[] array = Array.ofAll(values).toJavaArray();
+        assert (array.length == 0) || (primitiveClass == asPrimitive(array[0].getClass())) && !primitiveClass.isArray();
```
danieldietrich: we need to remove this assert please - it is the only remaing line that uses the code below that should be removed

javaslang/src/main/java/javaslang/collection/ArrayType.java
```diff
@@ -0,0 +1,147 @@
...
+        final Object results = type.newInstance(array.length);
+        for (int i = 0; i < array.length; i++) {
+            type.setAt(results, i, array[i]);
+        }
+        return (T) results;
+    }
+
+    /* convert to primitive */
+    private static final Class<?>[] WRAPPERS = { Boolean.class, Byte.class, Character.class, Double.class, Float.class, Integer.class, Long.class, Short.class, Void.class };
+    private static final Class<?>[] PRIMITIVES = { boolean.class, byte.class, char.class, double.class, float.class, int.class, long.class, short.class, void.class };
```
danieldietrich: both constants can be removed

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -266,46 +283,47 @@ private void setCurrentArray() {
...
             }
             return index;
         });
 
         return BitMappedTrie.ofAll(results);
     }
 
-    int length() { return length; }
+    int length()                         { return length; }
+    private boolean cannotAdd(T element) { return type != obj() && ((element == null) || (type.type() != asPrimitive(element.getClass()))); }
```
danieldietrich: We don't need to check the types. Because we already use `T` we only need to check if we want to insert `null` into a primitive array.

```java
private boolean addNullToPrimitiveArray(T element) { return !arrayType.isPrimitive() && element == null; }
```

I don't think we need an additional check if the element is an Object and we want to insert it into an primitive array. If the element is not null it should be unboxed, right?

If it is not unboxed we have to:

```java
private boolean addNullOrObjectToPrimitiveArray(T element) {
    return arrayType.isPrimitive() && (element == null || !arrayType.isPrimitive());
}
```

javaslang/src/main/java/javaslang/collection/ArrayType.java
```diff
@@ -0,0 +1,147 @@
+/*     / \____  _    _  ____   ______  / \ ____  __    _______
+ *    /  /    \/ \  / \/    \ /  /\__\/  //    \/  \  //  /\__\   JΛVΛSLΛNG
+ *  _/  /  /\  \  \/  /  /\  \\__\\  \  //  /\  \ /\\/ \ /__\ \   Copyright 2014-2016 Javaslang, http://javaslang.io
+ * /___/\_/  \_/\____/\_/  \_/\__\/__/\__\_/  \_//  \__/\_____/   Licensed under the Apache License, Version 2.0
+ */
+package javaslang.collection;
+
+import java.io.Serializable;
+
+abstract class ArrayType<T> implements Serializable {
```
danieldietrich: I would love to see one Array abstraction on the file system, e.g. ArrayType and have the implementations as auxiliary classes. The ArrayType then needs to be added to the generator.

But I suggested below to call ObjectArrayType.xxx directly instead of using obj(). Maybe we need to call obj() then. Or do you have another suggestion?

javaslang/generator/Generator.scala
```diff
@@ -2228,6 +2229,76 @@ def generateMainClasses(): Unit = {
       """
     }
   }
+
+  /**
+    * Generator of javaslang.collection.*ArrayType
+    */
```
paplorinc: I find indentations important also (though I aligned it this way on purpose).
Will change it.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: > We don't need to check the types. Because we already use T

you can add a `String` to a primitive array via `narrow`, e.g.
```java
@Test
public void shouldNarrowPrimitives() {
    final Vector<Object> object = Vector.narrow(range(0, 2));
    Vector<Object> actual = object.append("String");
    assertThat(actual).isEqualTo(of(0, 1, "String"));
}
```

(added this test explicitly now. `shouldBehaveLikeArray` already tested this implicitly)

> arrayType.isPrimitive() && (element == null || !arrayType.isPrimitive());

`A & (B | !A)` == `A && B`, since `!A` is always `false` (typo?)

---

let me know if you still prefer to change this

javaslang/src-gen/main/java/javaslang/collection/ObjectArrayType.java
```diff
@@ -0,0 +1,49 @@
...
+    Object[] empty() { return EMPTY; }
+
+    @Override
+    int lengthOf(Object array) { return (array == null) ? 0 : cast(array).length; }
+
+    @Override
+    Object getAt(Object array, int index) { return cast(array)[index]; }
+
+    @Override
+    void setAt(Object array, int index, Object value) { cast(array)[index] =  value; }
```
paplorinc: spacing ...

javaslang/src/main/java/javaslang/collection/ArrayType.java
paplorinc: Uff, I don't understand what you mean by `file system` in this context.
Also, I can't easily make the methods in `ObjectArrayType` static :/
Suggestions?

javaslang/src/main/java/javaslang/collection/ArrayType.java
```diff
@@ -0,0 +1,147 @@
...
+            } else if (int.class == type) {
+                return (ArrayType<T>) IntArrayType.INSTANCE;
+            } else if (long.class == type) {
+                return (ArrayType<T>) LongArrayType.INSTANCE;
+            } else if (short.class == type) {
+                return (ArrayType<T>) ShortArrayType.INSTANCE;
+            } else {
+                throw new IllegalArgumentException("Unknown type: " + type);
+            }
+        }
```
paplorinc: I wanted to separate the primitive types from the only non-primitive type.
Would you like me to merge the nested ifs instead?

javaslang/src/main/java/javaslang/collection/ArrayType.java
paplorinc: If you still think I should, I will, just let me know if it's still the case (see my other responses)

javaslang/src/main/java/javaslang/collection/ArrayType.java
paplorinc: `JMH` shows that the if/else version is ~5% faster, thanks :)

javaslang/src/main/java/javaslang/collection/ArrayType.java
danieldietrich: Please change the ArrayType like follows (example for int only). We make the following changes:

* also generate ArrayType
* make ArrayType an interface
* ArrayType should not extends Serializable. We've done this for other Javaslang types also - only impls implement Serializable (exception: some collections interfaces extends Function1 and are therefore Serializable)
* make the ArrayType impls static nested classes.

Also I suggest to rename type() to componentType() (see TODO below).

```java
/*     / \____  _    _  ____   ______  / \ ____  __    _______
 *    /  /    \/ \  / \/    \ /  /\__\/  //    \/  \  //  /\__\   JΛVΛSLΛNG
 *  _/  /  /\  \  \/  /  /\  \\__\\  \  //  /\  \ /\\/ \ /__\ \   Copyright 2014-2016 Javaslang, http://javaslang.io
 * /___/\_/  \_/\____/\_/  \_/\__\/__/\__\_/  \_//  \__/\_____/   Licensed under the Apache License, Version 2.0
 */
package javaslang.collection;

import javaslang.collection.*;

import java.io.Serializable;

interface ArrayType<T> {

    static <T> ArrayType<T> obj() { return (ArrayType<T>) ObjectArrayType.INSTANCE; }

    static <T> ArrayType<T> of(Object array) {
        final Class<?> type = array.getClass().getComponentType();
        if (!type.isPrimitive()) {
            return obj();
        } else if (boolean.class == type) {
            return (ArrayType<T>) BooleanArrayType.INSTANCE;
        } else if (byte.class == type) {
            return (ArrayType<T>) ByteArrayType.INSTANCE;
        } else if (char.class == type) {
            return (ArrayType<T>) CharArrayType.INSTANCE;
        } else if (double.class == type) {
            return (ArrayType<T>) DoubleArrayType.INSTANCE;
        } else if (float.class == type) {
            return (ArrayType<T>) FloatArrayType.INSTANCE;
        } else if (int.class == type) {
            return (ArrayType<T>) IntArrayType.INSTANCE;
        } else if (long.class == type) {
            return (ArrayType<T>) LongArrayType.INSTANCE;
        } else if (short.class == type) {
            return (ArrayType<T>) ShortArrayType.INSTANCE;
        } else {
            throw new IllegalArgumentException("Unknown type: " + type);
        }
    }

    Class<T> type(); // TODO: rename to componentType
    int lengthOf(Object array);
    T getAt(Object array, int index);

    Object empty();
    void setAt(Object array, int index, Object value);
    Object copy(Object array, int arraySize, int sourceFrom, int destinationFrom, int size);

    default Object newInstance(int length) { return copy(empty(), length); }

    /** System.arrayCopy with same source and destination */
    default Object copyRange(Object array, int from, int to) {
        final int length = to - from;
        return copy(array, length, from, 0, length);
    }

    /** Repeatedly group an array into equal sized sub-trees */
    default Object grouped(Object array, int groupSize) {
        final int arrayLength = lengthOf(array);
        assert arrayLength > groupSize;
        final Object results = obj().newInstance(1 + ((arrayLength - 1) / groupSize));
        obj().setAt(results, 0, copyRange(array, 0, groupSize));

        for (int start = groupSize, i = 1; start < arrayLength; i++) {
            final int nextLength = Math.min(groupSize, arrayLength - (i * groupSize));
            obj().setAt(results, i, copyRange(array, start, start + nextLength));
            start += nextLength;
        }

        return results;
    }

    /** clone the source and set the value at the given position */
    default Object copyUpdate(Object array, int index, T element) {
        final Object copy = copy(array, index + 1);
        setAt(copy, index, element);
        return copy;
    }

    default Object copy(Object array, int minLength) {
        final int arrayLength = lengthOf(array);
        final int length = Math.max(arrayLength, minLength);
        return copy(array, length, 0, 0, arrayLength);
    }

    /** clone the source and keep everything after the index (pre-padding the values with null) */
    default Object copyDrop(Object array, int index) {
        final int length = lengthOf(array);
        return copy(array, length, index, index, length - index);
    }

    /** clone the source and keep everything before and including the index */
    default Object copyTake(Object array, int lastIndex) {
        return copyRange(array, 0, lastIndex + 1);
    }

    /** Create a single element array */
    default Object asArray(T element) {
        final Object result = newInstance(1);
        setAt(result, 0, element);
        return result;
    }

    /** Store the content of an iterable in an array */
    static Object[] asArray(java.util.Iterator<?> it, int length) {
        final Object[] array = new Object[length];
        for (int i = 0; i < length; i++) {
            array[i] = it.next();
        }
        return array;
    }

    static <T> T asPrimitives(Class<?> primitiveClass, Iterable<?> values) {
        final Object[] array = Array.ofAll(values).toJavaArray();
        assert (array.length == 0) || (primitiveClass == primitiveType(array[0])) && !primitiveClass.isArray();
        final ArrayType<T> type = of((Class<T>) primitiveClass);
        final Object results = type.newInstance(array.length);
        for (int i = 0; i < array.length; i++) {
            type.setAt(results, i, array[i]);
        }
        return (T) results;
    }

    static <T> Class<?> primitiveType(T element) {
        final Class<?> wrapper = (element == null) ? Object.class : element.getClass();
        if (wrapper == Boolean.class) {
            return boolean.class;
        } else if (wrapper == Byte.class) {
            return byte.class;
        } else if (wrapper == Character.class) {
            return char.class;
        } else if (wrapper == Double.class) {
            return double.class;
        } else if (wrapper == Float.class) {
            return float.class;
        } else if (wrapper == Integer.class) {
            return int.class;
        } else if (wrapper == Long.class) {
            return long.class;
        } else if (wrapper == Short.class) {
            return short.class;
        } else {
            return wrapper;
        }
    }

    final class IntArrayType implements ArrayType<Integer>, Serializable {

        private static final IntArrayType INSTANCE = new IntArrayType();
        private static final int[] EMPTY = new int[0];

        private static int[] cast(Object array) { return (int[]) array; }

        @Override
        public Class<Integer> type() { return int.class; }

        @Override
        public int[] empty() { return EMPTY; }

        @Override
        public int lengthOf(Object array) { return (array == null) ? 0 : cast(array).length; }

        @Override
        public Integer getAt(Object array, int index) { return cast(array)[index]; }

        @Override
        public void setAt(Object array, int index, Object value) { cast(array)[index] = (Integer) value; }

        @Override
        public Object copy(Object array, int arraySize, int sourceFrom, int destinationFrom, int size) {
            if (size == 0) {
                return new int[arraySize];
            } else {
                final int[] result = new int[arraySize];
                System.arraycopy(array, sourceFrom, result, destinationFrom, size); /* has to be near the object allocation to avoid zeroing out the array */
                return result;
            }
        }
    }

}
```

javaslang/src/main/java/javaslang/collection/ArrayType.java
danieldietrich: See suggested code in the comment above.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -37,83 +39,101 @@
...
+        final int size = type.lengthOf(array);
         if (size == 0) {
             return empty();
         } else {
             int shift = 0;
-            for (; array.length > BRANCHING_FACTOR; shift += BRANCHING_BASE) {
-                array = grouped(array, BRANCHING_FACTOR);
+            for (ArrayType<T> t = type; t.lengthOf(array) > BRANCHING_FACTOR; shift += BRANCHING_BASE) {
+                array = t.grouped(array, BRANCHING_FACTOR);
+                t = obj();
```
danieldietrich: had to look a moment at this for-loop :)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -151,67 +169,66 @@ private boolean arePointingToSameLeaf(int i, int j) {
...
 
-            Object[] array = root;
+            Object array = root;
             for (int shift = depthShift - BRANCHING_BASE; shift >= BRANCHING_BASE; shift -= BRANCHING_BASE) {
                 final int offset = digit(index, shift);
 
-                final Object previous = array[previousIndex];
-                final Object[] newNode = node.apply(previous, offset);
-                array[previousIndex] = newNode;
+                final Object previous = obj().getAt(array, previousIndex);
```
danieldietrich: The usages of obj() are OK!

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -37,83 +39,101 @@
...
 
     private static int treeSize(int branchCount, int depthShift) {
         final int fullBranchSize = 1 << depthShift;
         return branchCount * fullBranchSize;
     }
 
-    static <T> BitMappedTrie<T> ofAll(Object[] array) {
-        final int size = array.length;
+    static <T> BitMappedTrie<T> ofAll(Object array) {
+        final ArrayType<T> type = ArrayType.of(array);
```
paplorinc: Renamed all to `type` for brevity :)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: yeah, trying to achieve a good signal-to-noise ratio in a language that doesn't want me to :)

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -43,16 +47,17 @@
...
     public static void main(String... args) {
         JmhRunner.runDebugWithAsserts(CLASSES);
-        JmhRunner.runNormalNoAsserts(CLASSES, JAVA, FUNCTIONAL_JAVA, PCOLLECTIONS, ECOLLECTIONS, CLOJURE, SCALA, JAVASLANG);
+        JmhRunner.runNormalNoAsserts(CLASSES, /*FUNCTIONAL_JAVA, PCOLLECTIONS, ECOLLECTIONS,*/ JAVA, CLOJURE, SCALA, JAVASLANG);
     }
 
     @State(Scope.Benchmark)
     public static class Base {
-        @Param({ "32", "1024", "32768" /*, "1048576", "33554432", "1073741824" */ }) /* i.e. depth 1,2,3(,4,5,6) for a branching factor of 32 */
+        @Param({ "100", "10000" })
```
paplorinc: 1.  I don't mind if the benchmark correctness tests (e.g. that `prependAll` concats to the start and not the end, i.e. that's I'm testing the speed of the correct thing) run only on the CI, an not locally.
2. The property based tests are incredibly useful. I have made them quicker now, they all run in 7 seconds :)
3. same as 1.

4. we could simply try parallel tests, as we're basically stateless, everything should work out of the box

danieldietrich: Looks all good to me. +1 for trying parallel tests - should be a flag in the maven surefire plugin.

javaslang/src/main/java/javaslang/collection/ArrayType.java
paplorinc: Mostly done - and pushed -, not sure about embedding them to the same file, though.
What's the advantage exactly, are we trying to simulate a sealed trait in Java this way?

Pushed most of the changes, if it's easier for you to merge-and-modify rather than explain further, I won't mind :)

danieldietrich:
@paplorinc 

> Mostly done - and pushed -, not sure about embedding them to the same file, though.
> What's the advantage exactly, are we trying to simulate a sealed trait in Java this way?

Yes, it is some kind of sealed trait behavior. It is a kind of coding style I applied in different places of Javaslang. I want to keep together things that belong together. It is some kind of local module. The implementation types are never used directly, like List.Cons & List.Nil. Let's apply this change and align to the other classes.

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
paplorinc: I could `BitMappedTrie<T> extends ObjectArrayType` to get rid of the `obj()`s, the end result would be cleaner, but it would be kinda' hacky :/

danieldietrich: No - obj() looks good to me - it was my fault.

paplorinc:
> Please just nest the *ArrayType into ArrayType as discussed in the original PR.

Is there a non-hacky way for this, I'm not proud of how this code generation part turned out :/

danieldietrich:
> Is there a non-hacky way for this, I'm not proud of how this code generation part turned out :/

Currently we need to do it the hacky way :-/

When we move to Gradle  I hope there is a better Scala integration. We will split the generator in several parts (one for ArrayType, one for Functions, one for Tuple, etc.)

paplorinc:
@danieldietrich, I tried, but I can't convince Scala.
Could you please do it after merging this :)?
Is there anything else you would like me to address?

danieldietrich:
@paplorinc

> Could you please do it after merging this :)?

Will do that, Sir. Compared to your great ideas that materialized in this PR this looks like a warm, mild summer breeze to me :)

danieldietrich:
Many thanks!

PR #1558, opened 2016-09-04
Benchmark framework updates

paplorinc:
* Added `Eclipse Collections`
* Added option to selectively disable benchmarking alternatives
* Added proper sorting of end-results
* Added more compact memory consumption display
* Ordered methods in VectorBenchmarks (and added `map` and `filter` methods)


danieldietrich:
Thanks!

Issue #1556, opened 2016-09-04
Tuple cleanup


danieldietrich:
# To be done before 2.1.0 release

We added static `Tuple*.sequence()` methods. These should be moved to `Tuple` instead.

Minor: We added Tuple*.append() and Tuple*.prepend(). These should be reordered alphabetically within the files.

ashrko619:
@danieldietrich Can I work on this? :)

danieldietrich:
Yes, that would be great! You are familiar with the code generator (Generator.scala), right?

ashrko619:
@danieldietrich Yes, Once I make the changes -  I just need to run the gen code right?

ashrko619:
@danieldietrich How do I run the code in Generator.scala ?

emmanueltouzery:
just build the app IIRC: `mvn install` and it'll regenerate the generated code.

ruslansennov:
`mvn install` also runs tests and benchmarks.
For your purposes `mvn generate-sources` is enough

danieldietrich:
Yes, I also do a `clean`. Choose one of these:

* `mvn clean generate-sources` for fast results
* `mvn clean compile` to see if your generate code does compile
* `mvn clean test` to see if the unit tests still run after applying changes

Before pushing to Github we should also do a

* `mvn javadoc:javadoc` to see, if the doclint complains

danieldietrich:
Fixed with #1559

Issue #1241, opened 2016-03-29
Replace `? extends TupleN` by `TupleN` for N = 0..8


danieldietrich:
Tuples are final. Therefore in most (all?) cases we need no bounds.

**Example:**

```java
public <K, V2 extends V> HashMultimap<K, V2> tabulate(int n, Function<? super Integer, ? extends Tuple2<? extends K, ? extends V2>> f) {
   ...
}
```

is substituted by

```java
public <K, V2 extends V> HashMultimap<K, V2> tabulate(int n, Function<? super Integer, Tuple2<? extends K, ? extends V2>> f) {
   ...
}
```

danieldietrich:
This is not possible everywhere. E.g. we need it in LinkedHashMap.Builder:

```java
@Override
public <K, V2 extends V> LinkedHashMultimap<K, V2> ofEntries(Iterable<? extends Tuple2<? extends K, ? extends V2>> entries) {
    return build(entries, Tuple2::_1, Tuple2::_2, (OfMap<K, V2, LinkedHashMultimap<K, V2>>) this::ofMap);
}
```

in order to use it as finisher for a Collector:

```java
@Override
public <K, V2 extends V> Collector<Tuple2<K, V2>, ArrayList<Tuple2<K, V2>>, LinkedHashMultimap<K, V2>> collector() {
    final Supplier<ArrayList<Tuple2<K, V2>>> supplier = ArrayList::new;
    final BiConsumer<ArrayList<Tuple2<K, V2>>, Tuple2<K, V2>> accumulator = ArrayList::add;
    final BinaryOperator<ArrayList<Tuple2<K, V2>>> combiner = (left, right) -> {
        left.addAll(right);
        return left;
    };
    return Collector.of(supplier, accumulator, combiner, this::ofEntries);
}
```

danieldietrich:
Code works fine, no hurry here. But we definitely should check all bounds. Will move it to 2.2.0.

danieldietrich:
Not possible as stated above.

PR #1559, opened 2016-09-08
Refactored sequence methods

ashrko619:
PR for #1556.

Here we have to name the methods as sequence* otherwise they have the same type erasure.

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -1278,42 +1278,6 @@ def generateMainClasses(): Unit = {
               }
             """)}
 
-            ${(i == 0).gen(xs"""
-              @Override
-              public <T> Tuple1<T> append(T value) {
-                 return new Tuple1<>(value);
-              }
-
-              @Override
-              public <T> Tuple1<T> prepend(T value) {
-                 return new Tuple1<>(value);
-              }
-            """)}
-
-            ${(i > 0 && i < N).gen(xs"""
-              @Override
-              public <T> Tuple${i + 1}<${(1 to i).gen(j => s"T$j")(", ")}, T> append(T value) {
-                  return Tuple.of(${(1 to i).gen(j => s"_$j")(", ")}, value);
-              }
-
-              @Override
-              public <T> Tuple${i + 1}<T, ${(1 to i).gen(j => s"T$j")(", ")}> prepend(T value) {
-                  return Tuple.of(value, ${(1 to i).gen(j => s"_$j")(", ")});
-              }
-            """)}
-
-            ${(i == N).gen(xs"""
-              @Override
-              public <T> Tuple append(T value) {
-                  throw new UnsupportedOperationException("Append to Tuple${N}");
-              }
-
-              @Override
-              public <T> Tuple prepend(T value) {
-                  throw new UnsupportedOperationException("Prepend to Tuple${N}");
-              }
-            """)}
-
         }
       """
     }
@@ -1361,9 +1325,8 @@ def generateMainClasses(): Unit = {
         val seqs = (1 to i).gen(j => s"Seq<T$j>")(", ")
         val Objects = im.getType("java.util.Objects")
         val Stream = im.getType("javaslang.collection.Stream");
-        val Iterator  = im.getType("javaslang.collection.Iterator");
         val emptyString = " "*38; // for indentation
-        val args  = (1 to i).gen(j => s"$Stream.narrow($Iterator.ofAll(tuples).map(Tuple$i::_$j).toStream())")(s",\n$emptyString")
+        val args  = (1 to i).gen(j => s"$Stream.ofAll(tuples).map(Tuple$i::_$j)")(s",\n$emptyString")
         xs"""
             static <$generics> Tuple$i<$seqs> sequence$i(Iterable<? extends Tuple$i<${(1 to i).gen(j => s"? extends T$j")(", ")}>> tuples) {
                 $Objects.requireNonNull(tuples, "tuples is null");
@@ -1402,24 +1365,6 @@ def generateMainClasses(): Unit = {
              * @return A new {@code Seq}.
              */
             $Seq<?> toSeq();
-
-            /**
-             * Appends the given {@code value} to the end of this Tuple and increases the arity by one.
-             *
-             * @param value the value that will be appended to the end of this Tuple
-             * @return a new Tuple that contains this values plus the new value
-             * @throws UnsupportedOperationException if {@code this.arity() == Tuple.MAX_ARITY}
-             */
-            <T> Tuple append(T value);
-
-            /**
-             * Prepends the given {@code value} before the start of this Tuple and increases the arity by one.
-             *
-             * @param value the value that will be prepended before the start of this Tuple
-             * @return a new Tuple that contains the new value plus this values
-             * @throws UnsupportedOperationException if {@code this.arity() == Tuple.MAX_ARITY}
-             */
-            <T> Tuple prepend(T value);
 
             // -- factory methods
 
@@ -2049,50 +1994,6 @@ def generateTestClasses(): Unit = {
                   }
                 """
               })("\n\n")}
-
-              ${(i == 0).gen(xs"""
-                @$test
-                public void shouldAppendTuple$i() {
-                    Tuple1<Integer> actual = Tuple0.instance().append(42);
-                    Tuple1<Integer> expected = new Tuple1<>(42);
-                    assertThat(actual).isEqualTo(expected);
-                }
-
-                @$test
-                public void shouldPrependTuple$i() {
-                    Tuple1<Integer> actual = Tuple0.instance().prepend(42);
-                    Tuple1<Integer> expected = new Tuple1<>(42);
-                    assertThat(actual).isEqualTo(expected);
-                }
-              """) }
-
-              ${(i > 0 && i < N).gen(xs"""
-                @$test
-                public void shouldAppendTuple$i() {
-                    Tuple${i + 1}<${(1 to i + 1).gen(_ => "Integer")(", ")}> actual = createIntTuple(${(1 to i).gen()(", ")}).append(42);
-                    Tuple${i + 1}<${(1 to i + 1).gen(_ => "Integer")(", ")}> expected = Tuple.of(${(1 to i).gen()(", ")}, 42);
-                    assertThat(actual).isEqualTo(expected);
-                }
-
-                @$test
-                public void shouldPrependTuple$i() {
-                    Tuple${i + 1}<${(1 to i + 1).gen(_ => "Integer")(", ")}> actual = createIntTuple(${(1 to i).gen()(", ")}).prepend(42);
-                    Tuple${i + 1}<${(1 to i + 1).gen(_ => "Integer")(", ")}> expected = Tuple.of(42, ${(1 to i).gen()(", ")});
-                    assertThat(actual).isEqualTo(expected);
-                }
-              """) }
-
-              ${(i == N).gen(xs"""
-                @$test(expected = UnsupportedOperationException.class)
-                public void shouldAppendTuple$i() {
-                    Tuple.of(${(1 to i).gen()(", ")}).append(42);
-                }
-
-                @$test(expected = UnsupportedOperationException.class)
-                public void shouldPrependTuple$i() {
-                    Tuple.of(${(1 to i).gen()(", ")}).prepend(42);
-                }
-              """) }
 
               @$test
               public void shouldApplyTuple() {
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -1358,12 +1358,16 @@ def generateMainClasses(): Unit = {
 
       def genSeqMethod(i: Int) = {
         val generics = (1 to i).gen(j => s"T$j")(", ")
-        val seqs = (1 to i).gen(j => s"Seq<? extends T$j>")(", ")
+        val seqs = (1 to i).gen(j => s"Seq<T$j>")(", ")
         val Objects = im.getType("java.util.Objects")
+        val Stream = im.getType("javaslang.collection.Stream");
+        val Iterator  = im.getType("javaslang.collection.Iterator");
+        val emptyString = " "*38; // for indentation
+        val args  = (1 to i).gen(j => s"$Stream.narrow($Iterator.ofAll(tuples).map(Tuple$i::_$j).toStream())")(s",\n$emptyString")
         xs"""
-            static <$generics> Tuple$i<$seqs> sequence$i(Iterable<Tuple$i<${(1 to i).gen(j => s"? extends T$j")(", ")}>> tuples) {
+            static <$generics> Tuple$i<$seqs> sequence$i(Iterable<? extends Tuple$i<${(1 to i).gen(j => s"? extends T$j")(", ")}>> tuples) {
                 $Objects.requireNonNull(tuples, "tuples is null");
-                return new Tuple$i<>(${(1 to i).gen(j => s"${im.getType("javaslang.collection.Iterator")}.ofAll(tuples).map(Tuple$i::_$j).toList()")(", ")});
+                return new Tuple$i<>($args);
             }
         """
       }
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -1324,13 +1324,13 @@ def generateMainClasses(): Unit = {
         val generics = (1 to i).gen(j => s"T$j")(", ")
         val seqs = (1 to i).gen(j => s"Seq<T$j>")(", ")
         val Objects = im.getType("java.util.Objects")
-        val Stream = im.getType("javaslang.collection.Stream");
-        val emptyString = " "*38; // for indentation
-        val args  = (1 to i).gen(j => s"$Stream.ofAll(tuples).map(Tuple$i::_$j)")(s",\n$emptyString")
+        val Stream = im.getType("javaslang.collection.Stream")
+        val widenedGenerics = (1 to i).gen(j => s"? extends T$j")(", ")
         xs"""
-            static <$generics> Tuple$i<$seqs> sequence$i(Iterable<? extends Tuple$i<${(1 to i).gen(j => s"? extends T$j")(", ")}>> tuples) {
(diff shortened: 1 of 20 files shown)


javaslang/src-gen/main/java/javaslang/Tuple.java
```diff
@@ -213,4 +215,45 @@ static Tuple0 empty() {
     static <T1, T2, T3, T4, T5, T6, T7, T8> Tuple8<T1, T2, T3, T4, T5, T6, T7, T8> of(T1 t1, T2 t2, T3 t3, T4 t4, T5 t5, T6 t6, T7 t7, T8 t8) {
         return new Tuple8<>(t1, t2, t3, t4, t5, t6, t7, t8);
     }
+
+    static <T1> Tuple1<Seq<? extends T1>> sequence1(Iterable<Tuple1<? extends T1>> tuples) {
+        Objects.requireNonNull(tuples, "tuples is null");
+        return new Tuple1<>(Iterator.ofAll(tuples).map(Tuple1::_1).toList());
+    }
+
+    static <T1, T2> Tuple2<Seq<? extends T1>, Seq<? extends T2>> sequence2(Iterable<Tuple2<? extends T1, ? extends T2>> tuples) {
```
ruslansennov: If we change signature to 
```java
static <T1, T2> Tuple2<Iterator<? extends T1>, Iterator<? extends T2>> sequence2(Iterable<Tuple2<? extends T1, ? extends T2>> tuples)
```
then these methods can be lazy. Also `toList()` will be omitted.
/cc @danieldietrich

javaslang/generator/Generator.scala
```diff
@@ -1265,46 +1265,40 @@ def generateMainClasses(): Unit = {
...
               }
-            """)}
 
-            ${(i > 0).gen(xs"""
-              public static $generics $className<${(1 to i).gen(j => s"Seq<? extends T$j>")(", ")}> sequence(Iterable<$className<${(1 to i).gen(j => s"? extends T$j")(", ")}>> tuples) {
-                Objects.requireNonNull(tuples, "tuples is null");
-                return new $className<>(${(1 to i).gen(j => s"${im.getType("javaslang.collection.Iterator")}.ofAll(tuples).map($className::_$j).toList()")(", ")});
+              @Override
+              public <T> Tuple prepend(T value) {
+                  throw new UnsupportedOperationException("Prepend to Tuple${N}");
```
paplorinc: could we return  a `Seq` here instead?
(maybe `Tuple` should implement `Seq` to make it possible)
@danieldietrich?

danieldietrich: I agree, that's not safe.

In an early stage of Javaslang I experimented with the integration of Tuple into the Value hierarchy. There were several problems, for details see #692 (e.g. [this](https://github.com/javaslang/javaslang/issues/692#issuecomment-151115513)). In the end I understood: 'A Tuple is a Tuple'. Not more, not less. It isn't a Seq, nor a Value.

We should not change the return type - too much confusion / not straight forward enough. Nor we should throw - that's unsafe.

I think these are our options:

* remove `append()`, `prepend()` from the base interface `Tuple` and remove them from `Tuple8`
* remove these methods completely

I don't know if these methods are used often in practice, haven't seen a use-case yet. So we should consider to remove them completely. Keep things simple! Less is more.

ruslansennov: > I don't know if these methods are used often in practice, haven't seen a use-case yet. So we should consider to remove them completely. Keep things simple! Less is more.

See [this](https://github.com/javaslang/javaslang-jackson/blob/master/src/main/java/javaslang/jackson/datatype/deserialize/TupleDeserializer.java#L70) example

danieldietrich: It looks like append() and prepend() are Tuple builders. In your example the elements are already known. Changing the example to the following would more concise but also inefficient regarding creation of Tuple instances:

```java
Tuple result = Tuple.empty();
for (int i = 0; i < list.size(); i++) {
    result = result.append(list.get(i));
}
```

Because append() and prepend() make most sense in a dynamic context (i.e. size of the List is unknown), the result type is just `Tuple`. In my opinion this kind of code is a corner case. It is comparable / has the same characteristics like reflection, which is a code smell and not type safe.

I'm not convinced that these methods are really useful, they do not encourage a good programming practice.

paplorinc: > I'm not convinced that these methods are really useful, they do not encourage a good programming practice.

I'm not convince either, but @ruslansennov's example seems like a case for a `list.toTuple` (and the opposite, `tuple.toList`), though the generics for those would probably be horrible :D

However, I think these discussions should be held before someone is finished implementing it already (sorry @ashrko619, I've been there numerous times also)...

danieldietrich: A `list.toTuple()` is not possible because the number of elements may exceed the max Tuple arity.

`tuple.toSeq()` exists already.

What we really need is a heterogenous list (`HList`). We have already an issue for that but the generics are not human-readable any more when the list grows. For now I would not implement an HList. It would start to make sense when we have local variable type inference in Java.

@ashrko619 Sorry for being torn back and forth. I think we need a more in-depth analysis for future features, including code examples, pros, cons etc.

Please give me some time to think about this feature...

javaslang/src-gen/main/java/javaslang/Tuple.java
danieldietrich: @ashrko619 @ruslansennov Please let us stay with `Seq` instead of `Iterator`. Then the static `sequence()` methods of all `Value` implementations have the same signature.

Also let's narrow the return type generics. I've learned that `Iterable<? extends Tuple....>` also comes handy in certain situations (see [this comment](https://github.com/javaslang/javaslang/issues/1241#issuecomment-242927975)).

I agree that it could make sense to have a lazy implementation underneath, so let's take `Stream`, like `Future.sequence` does.

The result looks like this:

```java
static <T1, T2> Tuple2<Seq<T1>, Seq<T2>> sequence2(Iterable<? extends Tuple2<? extends T1, ? extends T2>> tuples) {
    return new Tuple2<>(
            Iterator.ofAll(tuples).map(Tuple2::_1).toStream(),
            Iterator.ofAll(tuples).map(Tuple2::_2).toStream());
}
```

If possible, the generator should create the new lines like above for Tuple2, Tuple3, ...
For Tuple1 it can still be a one-liner:

```java
return new Tuple1<>(Iterator.ofAll(tuples).map(Tuple1::_1).toStream());
```

ashrko619: Here ```Iterator.ofAll(tuples).map(Tuple2::_1).toStream()``` returns a ```Stream<? extends T1>``` so the return type has to be ```Tuple2<Seq<? extends T1>, Seq<? extends T2>>```

danieldietrich: Please narrow the streams:

```java
Stream.narrow(Iterator.ofAll(...)...toStream());
```

danieldietrich: Double-checked it - should work!

```java
    static <T1, T2> Tuple2<Seq<T1>, Seq<T2>> sequence2(Iterable<? extends Tuple2<? extends T1, ? extends T2>> tuples) {
        return new Tuple2<>(
                Stream.narrow(Iterator.ofAll(tuples).map(Tuple2::_1).toStream()),
                Stream.narrow(Iterator.ofAll(tuples).map(Tuple2::_2).toStream()));
    }
```

paplorinc:
@ashrko619, please rebase your PR on `master` :)
([from Idea it's `VCS` -> `Git` -> `Rebase my github fork`](https://blog.jetbrains.com/idea/2011/02/advanced-github-integration-rebase-my-github-fork))

javaslang/src-gen/main/java/javaslang/Tuple.java
```diff
@@ -213,4 +216,73 @@ static Tuple0 empty() {
...
     }
+
+    static <T1> Tuple1<Seq<T1>> sequence1(Iterable<? extends Tuple1<? extends T1>> tuples) {
+        Objects.requireNonNull(tuples, "tuples is null");
+        return new Tuple1<>(Stream.narrow(Iterator.ofAll(tuples).map(Tuple1::_1).toStream()));
+    }
+
+    static <T1, T2> Tuple2<Seq<T1>, Seq<T2>> sequence2(Iterable<? extends Tuple2<? extends T1, ? extends T2>> tuples) {
+        Objects.requireNonNull(tuples, "tuples is null");
+        return new Tuple2<>(Stream.narrow(Iterator.ofAll(tuples).map(Tuple2::_1).toStream()),
```
paplorinc: I didn't have time to thoroughly investigate, but why `narrow` and `iterator` and `toStream`, why not simply:
```java
static <T1, T2> Tuple2<Seq<T1>, Seq<T2>> sequence2(Iterable<? extends Tuple2<? extends T1, ? extends T2>> tuples) {
    return new Tuple2<>(Stream.ofAll(tuples).map(Tuple2::_1),
            Stream.ofAll(tuples).map(Tuple2::_2));
}
```

javaslang/src-gen/main/java/javaslang/Tuple.java
```diff
@@ -213,4 +216,73 @@ static Tuple0 empty() {
     static <T1, T2, T3, T4, T5, T6, T7, T8> Tuple8<T1, T2, T3, T4, T5, T6, T7, T8> of(T1 t1, T2 t2, T3 t3, T4 t4, T5 t5, T6 t6, T7 t7, T8 t8) {
         return new Tuple8<>(t1, t2, t3, t4, t5, t6, t7, t8);
     }
+
+    static <T1> Tuple1<Seq<T1>> sequence1(Iterable<? extends Tuple1<? extends T1>> tuples) {
```
paplorinc: why is there a need for the number at the end, it should be distinguishable via parameters?

javaslang/src-gen/main/java/javaslang/Tuple0.java
```diff
@@ -127,12 +127,12 @@ private Object readResolve() {
     }
 
     @Override
-    public <T> Tuple1<T> prepend(T value) {
+    public <T> Tuple1<T> append(T value) {
```
paplorinc: what was the problem with the previous order?
`prepend` comes conceptually before `append`, i.e. `insert before` should be before `insert after`, I think

javaslang/src-gen/main/java/javaslang/Tuple8.java
```diff
@@ -571,17 +571,13 @@ public String toString() {
...
     public <T> Tuple append(T value) {
         throw new UnsupportedOperationException("Append to Tuple8");
     }
 
-    public static <T1, T2, T3, T4, T5, T6, T7, T8> Tuple8<Seq<? extends T1>, Seq<? extends T2>, Seq<? extends T3>, Seq<? extends T4>, Seq<? extends T5>, Seq<? extends T6>, Seq<? extends T7>, Seq<? extends T8>> sequence(Iterable<Tuple8<? extends T1, ? extends T2, ? extends T3, ? extends T4, ? extends T5, ? extends T6, ? extends T7, ? extends T8>> tuples) {
-      Objects.requireNonNull(tuples, "tuples is null");
-      return new Tuple8<>(Iterator.ofAll(tuples).map(Tuple8::_1).toList(), Iterator.ofAll(tuples).map(Tuple8::_2).toList(), Iterator.ofAll(tuples).map(Tuple8::_3).toList(), Iterator.ofAll(tuples).map(Tuple8::_4).toList(), Iterator.ofAll(tuples).map(Tuple8::_5).toList(), Iterator.ofAll(tuples).map(Tuple8::_6).toList(), Iterator.ofAll(tuples).map(Tuple8::_7).toList(), Iterator.ofAll(tuples).map(Tuple8::_8).toList());
+    @Override
+    public <T> Tuple prepend(T value) {
+        throw new UnsupportedOperationException("Prepend to Tuple8");
```
paplorinc: * you're modifying the code you have previously modified in the same PR. Please rebase it, it confuses reviewers.
* I still don't think it's a good idea to throw here :/

javaslang/src-gen/main/java/javaslang/Tuple5.java
```diff
@@ -409,17 +409,13 @@ public String toString() {
...
-    @Override
     public <T> Tuple6<T1, T2, T3, T4, T5, T> append(T value) {
         return Tuple.of(_1, _2, _3, _4, _5, value);
     }
 
-    public static <T1, T2, T3, T4, T5> Tuple5<Seq<? extends T1>, Seq<? extends T2>, Seq<? extends T3>, Seq<? extends T4>, Seq<? extends T5>> sequence(Iterable<Tuple5<? extends T1, ? extends T2, ? extends T3, ? extends T4, ? extends T5>> tuples) {
-      Objects.requireNonNull(tuples, "tuples is null");
-      return new Tuple5<>(Iterator.ofAll(tuples).map(Tuple5::_1).toList(), Iterator.ofAll(tuples).map(Tuple5::_2).toList(), Iterator.ofAll(tuples).map(Tuple5::_3).toList(), Iterator.ofAll(tuples).map(Tuple5::_4).toList(), Iterator.ofAll(tuples).map(Tuple5::_5).toList());
+    @Override
+    public <T> Tuple6<T, T1, T2, T3, T4, T5> prepend(T value) {
```
paplorinc: maybe name `T` as `T0` to conform to the numbering sequence

javaslang/src-gen/main/java/javaslang/Tuple.java
ashrko619: Here we need numbers at the end since they have the same type erasure
```Error:(227, 46) java: name clash: <T1,T2>sequence(java.lang.Iterable<? extends javaslang.Tuple2<? extends T1,? extends T2>>) and <T1>sequence(java.lang.Iterable<? extends javaslang.Tuple1<? extends T1>>) have the same erasure```

Comment (paplorinc) [inline javaslang/src-gen/main/java/javaslang/Tuple.java]: K, thanks :)

File: javaslang/src-gen/main/java/javaslang/Tuple.java (hunk 3 above)
Comment (danieldietrich) [inline javaslang/src-gen/main/java/javaslang/Tuple.java]: Yes! That is a better solution :)
@ashrko619 please do it like @paplorinc suggests

File: javaslang/src-gen/main/java/javaslang/Tuple0.java (hunk 5 above)
Comment (danieldietrich) [inline javaslang/src-gen/main/java/javaslang/Tuple0.java]: Preferences on the semantic ordering of methods may vary - depending of the person who is looking at the code.

I try to keep alphabetic ordering (which is, however, not continuously applied yet).

A (more or less) precise specification on which order we prefer you find in the file [CONTRIBUTION.md (section 'File structure')](https://github.com/javaslang/javaslang/blob/master/CONTRIBUTING.md#file-structure).

---

Side-note: Regarding  `append` and `prepend` I have come to a conclusion: we will remove them again. The builder should not be baked into the tuple itself. Also we block the evolution of Tuple (e.g. increasing the max number of tuples) by letting Tuple8.append/prepend throw.

@ashrko619 please don't feal 😢 about it (removing append/prepend again). It is hard to make correct API decisions. I will try to do deeper investigations before you create a PR next time!

Comment (danieldietrich) [inline javaslang/src-gen/main/java/javaslang/Tuple0.java]: Note: We remove append/prepend in another PR.

File: javaslang/src-gen/main/java/javaslang/Tuple5.java (hunk 7 above)
Comment (danieldietrich) [inline javaslang/src-gen/main/java/javaslang/Tuple5.java]: It does not matter any more - append/prepend will be removed.

File: javaslang/src-gen/main/java/javaslang/Tuple8.java (hunk 6 above)
Comment (danieldietrich) [inline javaslang/src-gen/main/java/javaslang/Tuple8.java]: > I still don't think it's a good idea to throw here :/

Right, we will remove it...

File: javaslang/src-gen/main/java/javaslang/Tuple0.java (hunk 5 above)
Comment (ashrko619) [inline javaslang/src-gen/main/java/javaslang/Tuple0.java]: oops, removed the methods in the PR itself 🐵. I'll revert and commit ?

Comment (danieldietrich) [inline javaslang/src-gen/main/java/javaslang/Tuple0.java]: This is ok!!!
Please don't revert...

File: javaslang/src-gen/main/java/javaslang/Tuple.java [hunk 8]
```diff
@@ -223,4 +207,73 @@ static Tuple0 empty() {
...
+        Objects.requireNonNull(tuples, "tuples is null");
+        return new Tuple2<>(Stream.ofAll(tuples).map(Tuple2::_1),
+                            Stream.ofAll(tuples).map(Tuple2::_2));
+    }
+
+    static <T1, T2, T3> Tuple3<Seq<T1>, Seq<T2>, Seq<T3>> sequence3(Iterable<? extends Tuple3<? extends T1, ? extends T2, ? extends T3>> tuples) {
+        Objects.requireNonNull(tuples, "tuples is null");
+        return new Tuple3<>(Stream.ofAll(tuples).map(Tuple3::_1),
+                            Stream.ofAll(tuples).map(Tuple3::_2),
+                            Stream.ofAll(tuples).map(Tuple3::_3));
```
danieldietrich: There is one more thing I saw by looking at the code. 🙈 Stream is persistent, so the tuples can be reused like this:

```java
final Stream<Tuple3<T1, T2, T3>> s = Stream.ofAll(tuples);
return new Tuple3<>(s.map(Tuple3::_1),
                    s.map(Tuple3::_2),
                    s.map(Tuple3::_3));
```

Could you please adjust the generator accordingly? 🙏🏽☺️

ashrko619: Shouldn't it be 
```java
 final Stream<Tuple3<? extends T1, ? extends T2, ? extends T3>> s = Stream.ofAll(tuples);
```

danieldietrich: Yes, you are right!

ashrko619: Since the args length is decreased now
```java
return new Tuple3<>(s.map(Tuple3::_1), s.map(Tuple3::_2), s.map(Tuple3::_3));
```
Can we have the entire statement in one line?

danieldietrich: Yup!

danieldietrich:
Perfect, this looks awesome!!

Issue #1465, opened 2016-08-01
Make `Javaslang` Java 9 compatible


paplorinc:
Related to https://github.com/javaslang/javaslang/issues/1339.

Steps:
- install `JDK 9` from e.g. https://jdk9.java.net/download
- add it to `Idea`
- change the project dependency to it
- try compiling it from `Idea`
- watch the compiler crash with `StackOverflowError` (maybe we should file a bug report for it)

The problems are mainly around `::compareTo`-like structures in `TreeMap` and `TreeSet`.
And with `Maven` in the `Scala` plugin for the generator.

danieldietrich:
Does it compile with Java 9 and Maven on the command line (without IntelliJ IDEA)?

If it is an IntelliJ IDEA related problem we should file a bug.

paplorinc:
It seems like a Java compiler bug again

paplorinc:
http://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8164080

danieldietrich:
I've created a bug report for the `T::compareTo` bug, internal review ID : 9044933.

See also [this Gist](https://gist.github.com/danieldietrich/9d91d3602a3f6b78cbea96326db8db82).

PR #1563, opened 2016-09-09
RFC: Migrate codebase to Java 9 compatibility

paplorinc:
WIP, feedback needed.
Addresses: #1465.

The code compiles but the `Scala` based code generation fails.
Running the tests without the generator reveals (unrelated?) bugs.

Still needs some love, but e.g. benchmarks are already functional.
Luckily some measurable speed improvements were implemented, i.e. `Vector`'s `append` is now almost as fast as it's `prepend` (#1557)

### DIFF
--- a/javaslang/src/main/java/javaslang/Value.java
+++ b/javaslang/src/main/java/javaslang/Value.java
@@ -860,7 +860,7 @@ public interface Value<T> extends Iterable<T> {
      */
     default <K extends Comparable<? super K>, V> SortedMap<K, V> toSortedMap(Function<? super T, ? extends Tuple2<? extends K, ? extends V>> f) {
         Objects.requireNonNull(f, "f is null");
-        return toSortedMap((Comparator<? super K> & Serializable) K::compareTo, f);
+        return toSortedMap(Comparator.naturalOrder(), f);
     }
 
     /**
@@ -992,8 +992,8 @@ public interface Value<T> extends Iterable<T> {
             return (PriorityQueue<T>) this;
         } else {
             final Comparator<T> comparator = (this instanceof Ordered<?>)
-                                             ? ((Ordered<T>) this).comparator()
-                                             : (Comparator<T> & Serializable) (o1, o2) -> ((Comparable<T>) o1).compareTo(o2);
+                    ? ((Ordered<T>) this).comparator()
+                    : (Comparator<T>) Comparator.naturalOrder();
             return toPriorityQueue(comparator);
         }
     }
@@ -1069,8 +1069,8 @@ public interface Value<T> extends Iterable<T> {
             return (TreeSet<T>) this;
         } else {
             final Comparator<T> comparator = (this instanceof Ordered<?>)
-                                             ? ((Ordered<T>) this).comparator()
-                                             : (Comparator<T> & Serializable) (o1, o2) -> ((Comparable<T>) o1).compareTo(o2);
+                    ? ((Ordered<T>) this).comparator()
+                    : (Comparator<T>) Comparator.naturalOrder();
             return toSortedSet(comparator);
         }
     }
--- a/javaslang/src/test/java/javaslang/collection/CharSeqTest.java
+++ b/javaslang/src/test/java/javaslang/collection/CharSeqTest.java
@@ -451,7 +451,7 @@ public class CharSeqTest {
 
     @Test
     public void shouldComputeDistinctByOfEmptyTraversableUsingComparator() {
-        final Comparator<Character> comparator = (i1, i2) -> i1 - i2;
+        final Comparator<Character> comparator = Comparator.comparingInt(i -> i);
         assertThat(empty().distinctBy(comparator)).isSameAs(empty());
     }
 
@@ -1004,7 +1004,7 @@ public class CharSeqTest {
 
     @Test
     public void shouldCalculateMaxByOfInts() {
-        assertThat(of('1', '2', '3').maxBy((i1, i2) -> i1 - i2)).isEqualTo(Option.some('3'));
+        assertThat(of('1', '2', '3').maxBy(Comparator.comparingInt(i -> i))).isEqualTo(Option.some('3'));
     }
 
     @Test
@@ -1060,7 +1060,7 @@ public class CharSeqTest {
 
     @Test
     public void shouldCalculateMinByOfInts() {
-        assertThat(of('1', '2', '3').minBy((i1, i2) -> i1 - i2)).isEqualTo(Option.some('1'));
+        assertThat(of('1', '2', '3').minBy(Comparator.comparingInt(i -> i))).isEqualTo(Option.some('1'));
     }
 
     @Test
@@ -3139,7 +3139,7 @@ public class CharSeqTest {
         assertThat(
                 CharSeq.unfoldRight('j', x -> x == 'a'
                                               ? Option.none()
-                                              : Option.of(new Tuple2<>(new Character(x), (char) (x - 1)))))
+                                              : Option.of(new Tuple2<>(x, (char) (x - 1)))))
                 .isEqualTo(of("jihgfedcb"));
     }
 
@@ -3153,7 +3153,7 @@ public class CharSeqTest {
         assertThat(
                 CharSeq.unfoldLeft('j', x -> x == 'a'
                                              ? Option.none()
-                                             : Option.of(new Tuple2<>((char) (x - 1), new Character(x)))))
+                                             : Option.of(new Tuple2<>((char) (x - 1), x))))
                 .isEqualTo(of("bcdefghij"));
     }
 
@@ -3167,7 +3167,7 @@ public class CharSeqTest {
         assertThat(
                 CharSeq.unfold('j', x -> x == 'a'
                                          ? Option.none()
-                                         : Option.of(new Tuple2<>((char) (x - 1), new Character(x)))))
+                                         : Option.of(new Tuple2<>((char) (x - 1), x))))
                 .isEqualTo(of("bcdefghij"));
     }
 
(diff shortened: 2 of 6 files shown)


javaslang/src/test/java/javaslang/control/TryTest.java
```diff
@@ -912,9 +911,9 @@ private RuntimeException error() {
 
     private static <T, X extends Throwable> Try<T> failure(Class<X> exceptionType) {
         try {
-            final X exception = exceptionType.newInstance();
+            final X exception = exceptionType.getConstructor().newInstance();
             return Try.failure(exception);
-        } catch (InstantiationException | IllegalAccessException x) {
+        } catch (InstantiationException | IllegalAccessException | NoSuchMethodException | InvocationTargetException ignored) {
```
talios: If you wanted to be cleaner, you could also catch `ReflectiveOperationException` (see [javadoc](https://docs.oracle.com/javase/7/docs/api/java/lang/ReflectiveOperationException.html)) which is the new superclass for all reflection based exceptions which came in with Java 1.7.

paplorinc: Thank you, I didn't know about that :).
Since this is part of the tests (i.e. it's not part of the public contract), we could simply catch a `Throwable`, since that's what we're testing.

javaslang/src/test/java/javaslang/collection/CharSeqTest.java
```diff
@@ -3129,7 +3129,7 @@ public void shouldUnfoldRightSimpleCharSeq() {
         assertThat(
                 CharSeq.unfoldRight('j', x -> x == 'a'
                                               ? Option.none()
-                                              : Option.of(new Tuple2<>(new Character(x), (char) (x - 1)))))
```
paplorinc: these constructors receive a warning now

javaslang/src/main/java/javaslang/Lazy.java
```diff
@@ -94,7 +94,7 @@ private Lazy(Supplier<? extends T> supplier) {
      */
     public static <T> Lazy<Seq<T>> sequence(Iterable<? extends Lazy<? extends T>> values) {
         Objects.requireNonNull(values, "values is null");
-        return Lazy.of(() -> List.ofAll(values).map(Lazy::get));
+        return Lazy.of(() -> List.ofAll(values).map(l -> l.get()));
```
paplorinc: this was weird, the compiler complained that `Lazy` was untyped here ...

javaslang/src/main/java/javaslang/Value.java
```diff
@@ -758,7 +758,7 @@ default CharSeq toCharSeq() {
      */
     default <K extends Comparable<? super K>, V> SortedMap<K, V> toSortedMap(Function<? super T, ? extends Tuple2<? extends K, ? extends V>> f) {
         Objects.requireNonNull(f, "f is null");
-        return toSortedMap((Comparator<? super K> & Serializable) K::compareTo, f);
+        return toSortedMap(Comparator.naturalOrder(), f);
```
paplorinc: these recursive types confuse the compiler

javaslang/src/test/java/javaslang/control/TryTest.java
```diff
@@ -912,10 +910,10 @@ private RuntimeException error() {
 
     private static <T, X extends Throwable> Try<T> failure(Class<X> exceptionType) {
         try {
-            final X exception = exceptionType.newInstance();
+            final X exception = exceptionType.getConstructor().newInstance();
```
paplorinc: direct `newInstance` call is also deprecated now

javaslang/src/main/java/javaslang/Lazy.java
danieldietrich: It think it may be the case because of the generic bounds of `values`...

Type-investigation seems to got worse in Java 9. These are the WTF!? moments during development - we know that the type is right but the compiler does complain.

Starting from JDK 1.8.0_40 the compiler got steadily better. It is sad to see that Java 9 is a step back :(

---

However, I'm not sure if I *want* to use a lambda here. Instead JDK 9 should be fixed :) That's also the case for other changes. Is it possible to defer some changes an let the build break? Maybe we should report these finding and hope that they get fixed until the release of JDK 9 (which may be [delayed again](http://mail.openjdk.java.net/pipermail/jdk9-dev/2016-September/004820.html)??).

If we let the JDK9 build break it should not affect the JDK8 build of our PRs, merges and so on...

javaslang/src/main/java/javaslang/Value.java
danieldietrich: This falls into the category of the above comment.

What if we remove `? super` in the cast?

```java
return toSortedMap((Comparator<K> & Serializable) K::compareTo, f);
```

danieldietrich:
Thanks for the changes!

We still have some time, the JDK9 compatibility will start to be interesting when Java 9 is released (which may take a while).

We should understand why existing things break and how we can fix it. The first approach should be to fix the JDK 9 (by issuing bugs to the tracker).

javaslang/src/main/java/javaslang/Lazy.java
paplorinc: I kinda' agree, the build isn't passing now either. I already opened a bug for a compiler crash, but they didn't fix it yet, I had to circumvent it.
When we do have an error though (please see the Travis log or pull it locally) or a simplification (e.g. the `Character` constructor), we should definitely take advantage of Java 9's wisdom.

javaslang/src/main/java/javaslang/Value.java
paplorinc:
```java
Error:(761, 59) java: incompatible types: invalid method reference
    method compareTo in interface java.lang.Comparable<T> cannot be applied to given types
      required: ? super K
      found: K,K
      reason: actual and formal argument lists differ in length
```

javaslang/src/main/java/javaslang/Lazy.java
danieldietrich: > When we do have an error though (please see the Travis log or pull it locally) or a simplification (e.g. the Character constructor), we should definitely take advantage of Java 9's wisdom.

true

javaslang/src/main/java/javaslang/Value.java
danieldietrich: After 2.0.4 and 2.1.0 (maybe 2.2.0) are through the door I will focus on Java 9. These hints are very helpful!

pom.xml
```diff
@@ -67,7 +67,7 @@ Note: The maven build currently needs to be started from the javaslang root dir
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <assertj.core.version>3.5.2</assertj.core.version>
         <eclipse.lifecycle.mapping.version>1.0.0</eclipse.lifecycle.mapping.version>
-        <java.version>1.8</java.version>
+        <java.version>1.9</java.version>
```
danieldietrich: This will compile Java 9 binaries, I think it should stay 1.8. Only the JDK used to compile has to change, right?

paplorinc: Maybe, this PR is not meant to be merged, it's mainly a Request For Comment, so that we know where we stand :)

paplorinc:
The latest compiler solves many issues, but it seems others were brought to our attention, e.g.:
```java
java.util.ConcurrentModificationException
	at java.util.HashMap.computeIfAbsent(java.base@9-ea/HashMap.java:1138)
	at javaslang.Function3.lambda$memoized$cc475b9f$1(Function3.java:200)
	at javaslang.collection.euler.Euler67Test.lambda$static$8440914d$1(Euler67Test.java:52)
	at javaslang.Function3.lambda$tupled$3fce4281$1(Function3.java:182)
	at java.util.HashMap.computeIfAbsent(java.base@9-ea/HashMap.java:1137)
	at javaslang.Function3.lambda$memoized$cc475b9f$1(Function3.java:200)
	at javaslang.collection.euler.Euler67Test.lambda$static$8440914d$1(Euler67Test.java:52)
	at javaslang.Function3.lambda$tupled$3fce4281$1(Function3.java:182)
	at java.util.HashMap.computeIfAbsent(java.base@9-ea/HashMap.java:1137)
	at javaslang.Function3.lambda$memoized$cc475b9f$1(Function3.java:200)
	at javaslang.collection.euler.Euler67Test.solve(Euler67Test.java:43)
	at javaslang.collection.euler.Euler67Test.shouldSolveProblem67(Euler67Test.java:38)
```

i.e. the `Function.memoize` might not be thread safe :/

paplorinc:
@danieldietrich, might be related to https://blog.jooq.org/2015/03/04/avoid-recursion-in-concurrenthashmap-computeifabsent

paplorinc:
Everything compiles with Java 9 now, and the tests are passing with Java 8. I think we can merge it, though the above memoization bug should be fixed on 9, evenually.

danieldietrich:
this sounds great! will check it tomorrow, need sleep :)
thx!

danieldietrich:
Wow, that's awesome! Thank you :)

I've experimented a bit with travis-ci env vars the last days. I think it is possible to start multiple builds in parallel with different configurations. Maybe we are possible to start a jdk8 and jdk9 build in parallel then. I think if one node fails, the complete build fails. We will see. This will be part of #1465

paplorinc:
Great! Have you investigated the memoization bug? Or how to make the build pass on Java 9 (e.g. the scala generation phase)?

danieldietrich:
@paplorinc Please file a ticket for the memoization bug. I will investigate it.

It is possible to use the env to spawn multiple parallel build nodes. We could use this technique to run multiple builds in parallel. But the question is for what purpose? It is important that a future Java 9 project (3rd party), that uses Javaslang, has no problems at runtime (and compile time). Until Javaslang moves to Java 9 only we need no Java 9 builds.

PR #1566, opened 2016-09-14
Get or else

ruslansennov:


danieldietrich:
Veeerryy nice :) Will review it in a minute...

javaslang/src/main/java/javaslang/collection/Multimap.java
```diff
@@ -233,6 +233,15 @@ default void forEach(BiConsumer<K, V> action) {
     Option<Traversable<V>> get(K key);
 
     /**
+     * Returns the value associated with a key, or a default value if the key is not contained in the map
+     *
+     * @param key the key
+     * @param defaultValue a default value
+     * @return the value associated with key if it exists, otherwise the result of the default value
+     */
+    Traversable<V> getOrElse(K key, Traversable<V> defaultValue);
```
danieldietrich: Can we widen the defaultValue type to `Traversable<? extends V>` without breaking anything?

javaslang/src/main/java/javaslang/collection/HashArrayMappedTrie.java
```diff
@@ -359,9 +373,18 @@ public boolean isEmpty() {
             this.value = value;
         }
 
+        private boolean itsMe(int keyHashCode, K key) {
```
danieldietrich: Two minor points:

* Let's name it `equals`, it is an optimized form of `equals(Object)` that first tests the `hash`.
* Let's also call the first param `keyHash` or rename the instance variable to `hashCode` to have consistent naming.

ruslansennov:
done

javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
```diff
@@ -95,6 +95,12 @@ public ContainerType getContainerType() {
     }
 
     @Override
+    @SuppressWarnings("unchecked")
+    public Traversable<V> getOrElse(K key, Traversable<? extends V> defaultValue) {
```
paplorinc: if you intend to return a `Traversable<V>`, shouldn't the default value be of that type also?

ruslansennov: Do you mean `Traversable<V> defaultValue`? No, Daniel is right.
See example:
```
public interface Multimap<K, V> {
    // ...
    Traversable<V> getOrElse(K key, Traversable<? extends V> defaultValue);
    Traversable<V> getOrElse2(K key, Traversable<V> defaultValue);
    // ...
}

public void example() {
    List<Integer> def = List.of(1);
    Multimap<Integer, Comparable<?>> map = HashMultimap.withSeq().empty();
    Traversable<Comparable<?>> r1 = map.getOrElse(1, def);   // OK
    Traversable<Comparable<?>> r2 = map.getOrElse2(1, def);  // Wrong 2nd argument type
    }
```

paplorinc: Good point, thanks (haven't noticed Daniel's review)

danieldietrich:
Thank you :)

PR #1570, opened 2016-09-16
Bugfix 1555 sets equality

### LINKED ISSUE #1555: Collection Equality (Sets, Seqs and Maps should be equal by content - modulo type)
One would typically expect implementations of a set interface to consider two sets equal regardless of subtype or ordering as long as they contained exactly the same elements.  
But in javaslang this:

``` java
Set<String> tree = TreeSet.of("a", "b", "c", "D");
Set<String> hash = HashSet.of("a", "b", "c", "D");
assertEquals(tree, hash);
```

Results in this:

```
java.lang.AssertionError: 
Expected :TreeSet(D, a, b, c)
Actual   :HashSet(a, b, c, D)
```

Is this the intended behavior or could it be improved?

pfijalki:
Willing to resolve #1555


javaslang/src/main/java/javaslang/Value.java
```diff
@@ -1021,6 +1013,26 @@ default CharSeq toCharSeq() {
...
+     * Tests equality of two objects assuming one, or both might be instances of Value.
+     *
+     * If one of a or b is Value instance congruence is tested.
+     * If neither is Objects{@link Objects#equals} is being invoked for two.
+     *
+     * @param a first object to test
+     * @param b fecond object to test
+     * @return true, if this equals o according to the rules defined above, otherwise false.
+     */
+    static boolean eq(Object a, Object b) {
```
danieldietrich: Extracting this method is possible but we do not need to introduce new public API.

On the 'level' of Value it is not intended to fall back to `equals`. The `eq(Object)` does only fall back for contained values.

javaslang/src/main/java/javaslang/collection/AbstractSet.java
```diff
@@ -0,0 +1,49 @@
...
+package javaslang.collection;
+
+import javaslang.Value;
+
+import java.util.function.Predicate;
+
+/**
+ * @since 2.0.5
+ */
+public abstract class AbstractSet<T> implements Set<T> {
```
danieldietrich: Our `abstract` classes (other than `AbstractIterator`) are not public.

javaslang/src/main/java/javaslang/collection/AbstractSet.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public boolean equals(Object obj) {
+        return trueIfSameOrWhenInstanceOfASet(obj, this::euqalsTypeUnsafe);
+    }
+
+    @Override
+    public boolean eq(Object obj) {
```
danieldietrich: Overwriting `eq` would break the contract of structural equality. E.g. this following must be true:

```java
TreeSet.of(1, 2).eq(Arrays.asList(1, 2)) == true

TreeSet.of(1, 2).eq(Arrays.asList(2, 1)) == false
```

javaslang/src/main/java/javaslang/collection/AbstractSet.java
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    private boolean euqalsTypeUnsafe(Set other) {
+        return this.forAll(other::contains);
+    }
+
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    private boolean eqTypeUnsafe(Set other) {
+        return this.forAll(thisElem -> other.existsUnique(otherElem -> Value.eq(thisElem, otherElem)));
```
danieldietrich: Side-note: because `other` is a `Set` `exists()` is equivalent to `exsistsUnique()`, am I right?

javaslang/src/main/java/javaslang/collection/HashSet.java
```diff
@@ -26,7 +26,7 @@
  * @author Ruslan Sennov, Patryk Najda, Daniel Dietrich
  * @since 2.0.0
  */
-public final class HashSet<T> implements Kind1<HashSet<?>, T>, Set<T>, Serializable {
+public final class HashSet<T> extends AbstractSet<T> implements Kind1<HashSet<?>, T>, Serializable {
```
danieldietrich: Inheriting from a package-private (see comment above) subclass causes trouble. Java 8 has a bug regarding method-references and inherited public impls of invisible super-classes...

A solution could be to create a helper class `Sets` (similar to `Maps` and `Arrays`) and add a methods

```
class Sets {
    static boolean equals(Set<?> set1, Object o) {
        if (set1 == o) {
            return true;
        } else if (o instanceof Set) {
            final Set<?> set2 = (Set<?>) o;
            ...
        } else {
            return false;
        }
    }
}
```

javaslang/src/main/java/javaslang/collection/HashSet.java
```diff
@@ -559,7 +559,7 @@ public boolean contains(T element) {
             return empty();
         } else {
             final HashArrayMappedTrie<U, U> that = foldLeft(HashArrayMappedTrie.empty(),
-                    (tree, t) -> addAll(tree, mapper.apply(t)));
+                                                            (tree, t) -> addAll(tree, mapper.apply(t)));
```
danieldietrich: Please don't reformat code. Do you use IntelliJ IDEA? Then please use our setting `<javaslang-root>/.ide/idea-settings.jar`

javaslang/src/main/java/javaslang/collection/HashSet.java
```diff
@@ -883,13 +883,11 @@ public int hashCode() {
...
-        if (o == this) {
-            return true;
-        } else if (o instanceof HashSet) {
+        if (o instanceof HashSet) {
             final HashSet<?> that = (HashSet<?>) o;
             return this.tree.equals(that.tree);
         } else {
-            return false;
+            return super.equals(o);
         }
```
danieldietrich: The implementation then looks like this:

```java
@Override
    return Sets.equals(this, o);
}
```

javaslang/src/test/java/javaslang/collection/SetsEqualityTest.java
```diff
@@ -0,0 +1,73 @@
...
+        // then
+        assertThat(first).isNotEmpty();
+        assertThat(first.get().eq(second.get())).describedAs("%s vs %s", firstClass, secondClass).isTrue();
+    }
+
+    @FunctionalInterface
+    private interface SetStringGenerator extends Function<String[], Set<String>> {
+        SetStringGenerator hashSet = HashSet::of;
+        SetStringGenerator linkedHashSet = LinkedHashSet::of;
+        SetStringGenerator treeSet = TreeSet::of;
```
danieldietrich: Please remove these three constants (but not the FunctionalInterface). The method references will be added above...

javaslang/src/test/java/javaslang/collection/SetsEqualityTest.java
```diff
@@ -0,0 +1,73 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class SetsEqualityTest {
+
+    private final String[] testStrings = { "a", "b", "c", "D" };
+
+    private final Map<Class<?>, SetStringGenerator> generators = HashMap.of(
+            HashSet.class, SetStringGenerator.hashSet,
+            LinkedHashSet.class, SetStringGenerator.linkedHashSet,
+            TreeSet.class, SetStringGenerator.treeSet);
```
danieldietrich: HashSet.class, HashSet::of,
LinkedHashSet.class, LinkedHashSet::of,
TreeSet.class, TreeSet::of

javaslang/src/test/java/javaslang/collection/SetsEqualityTest.java
```diff
@@ -0,0 +1,73 @@
...
+    public void shouldAllImplementationsBeCongruentForTheSameData() {
+        pairwiseDistinctSetClasses().forEach(pair -> this.congruenceTest(pair._1, pair._2));
+    }
+
+    private Set<Tuple2<? extends Class<?>, ? extends Class<?>>> pairwiseDistinctSetClasses() {
+        return generators.keySet()
+                .flatMap(firstClass -> generators.keySet()
+                        .filter(filteredKey -> filteredKey != firstClass)
+                        .map(inner -> Tuple.of(firstClass, inner))
+                        .collect(Collectors.toSet()));
```
danieldietrich:
```java
private Seq<Tuple2<Class<?>, Class<?>>> pairwiseDistinctSetClasses() {
    return generators.keySet().toList()
            .combinations(2)
            .map(seq -> Tuple.of(seq.get(0), seq.get(1)));
}
```

javaslang/src/test/java/javaslang/collection/SetsEqualityTest.java
```diff
@@ -0,0 +1,73 @@
...
+        final Option<Set<String>> first = generators.get(firstClass).map(sg -> sg.apply(testStrings));
+        final Option<Set<String>> second = generators.get(secondClass).map(sg -> sg.apply(testStrings));
+
+        // when
+        // then
+        assertThat(first).isNotEmpty();
+        assertThat(first).describedAs("%s vs %s", firstClass, secondClass).isEqualTo(second);
+    }
+
+    private void congruenceTest(Class<?> firstClass, Class<?> secondClass) {
```
danieldietrich: we don't need the congruenceTest

javaslang/src/test/java/javaslang/collection/SetsEqualityTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Test
+    public void shouldAllImplementationsBeEqualForTheSameData() {
+        pairwiseDistinctSetClasses().forEach(pair -> this.equalityTest(pair._1, pair._2));
+    }
+
+    @Test
+    public void shouldAllImplementationsBeCongruentForTheSameData() {
+        pairwiseDistinctSetClasses().forEach(pair -> this.congruenceTest(pair._1, pair._2));
+    }
```
danieldietrich: not needed any more

javaslang/src/test/java/javaslang/collection/SetsEqualityTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private final Map<Class<?>, SetStringGenerator> generators = HashMap.of(
+            HashSet.class, SetStringGenerator.hashSet,
+            LinkedHashSet.class, SetStringGenerator.linkedHashSet,
+            TreeSet.class, SetStringGenerator.treeSet);
+
+    @Test
+    public void shouldAllImplementationsBeEqualForTheSameData() {
+        pairwiseDistinctSetClasses().forEach(pair -> this.equalityTest(pair._1, pair._2));
+    }
```
danieldietrich: Please add tests for the negative case (sets are not equal). Also the second set should be initialized with test strings of different order (because of LinkedHashSet).

javaslang/src/main/java/javaslang/collection/HashSet.java
pfijalki: I used provided formatter, and doublechecked it before submitting. Just thought I'll add this byt the way (previous code was not compatible with formatter provided) - should I remove this change ?

danieldietrich: It is ok for me as-is. I recently did a few changes to the formatter regarding indentation. I will apply the new rules to the whole code-base soon, so this space actually does not matter.
Thank you!

javaslang/src/test/java/javaslang/collection/SetsEqualityTest.java
pfijalki: If I'd remove this I cannot add Hashset::of to my pretty generators map, because HashSet::of is not recognizable as a functional interface

javaslang/src/main/java/javaslang/Value.java
danieldietrich: Yes, I think for now we only need a package-private `Sets` class (similar to `Maps`), which contains one method:

```java
static boolean equals(Set<?> self, Object o) {
    if (o == self) {
        return true;
    } else if (o instanceof Set) {
        final Set<?> that = (Set<?>) that;
        return self.forAll(that::contains); // PROBLEM: SEE BELOW
    } else {
        return false;
    }
}
```

The `Set` interface needs to `@Override` the `equals` method. This should contain some javadoc about Set equality (all Set impls share the same equals method).

All Set implementations need to override the equals method like this:

```java
@Override
    return Sets.equals(this, o);
}
```

We only have one problem to solve in order to make the code above work: `contains()` currently takes a `T` instead of an `Object`. Maybe we need to change the signature of `Set#contains`...

javaslang/src/test/java/javaslang/collection/SetsEqualityTest.java
pfijalki: If I'd remove this I cannot add Hashset::of to my pretty generators map, because HashSet::of is not recognizable as a functional interface

danieldietrich: You are right, then please leave it for now. It will be possible when we add typed HashMap.of() methods. @ruslansennov is on it but it will still take some time before we can use them... (See #1608)

danieldietrich:
Thank you for your PR! The issue #1555 was recently solved with PR #1948 in a general way (for Seq, Set, Map, Multimap).

PR #1571, opened 2016-09-17
further improvement of Lazy.get()

danieldietrich:
thx @jbgi for the hint!

### DIFF
--- a/javaslang/src/main/java/javaslang/Lazy.java
+++ b/javaslang/src/main/java/javaslang/Lazy.java
@@ -44,7 +44,9 @@ public final class Lazy<T> implements Value<T>, Supplier<T>, Serializable {
 
     // read http://javarevisited.blogspot.de/2014/05/double-checked-locking-on-singleton-in-java.html
     private transient volatile Supplier<? extends T> supplier;
-    private volatile T value;
+
+    // does not need to be volatile, visibility piggy-backs on volatile read of `supplier`
+    private T value;
 
     // should not be called directly
     private Lazy(Supplier<? extends T> supplier) {
@@ -133,12 +135,13 @@ public final class Lazy<T> implements Value<T>, Supplier<T>, Serializable {
      */
     @Override
     public T get() {
+        // using a local var speeds up the double-check idiom by 25%, see Effective Java, Item 71
         Supplier<? extends T> tmp = supplier;
         if (tmp != null) {
             synchronized (this) {
                 tmp = supplier;
                 if (tmp != null) {
-                    value = supplier.get();
+                    value = tmp.get();
                     supplier = null; // free mem
                 }
             }


javaslang/src/main/java/javaslang/Lazy.java
```diff
@@ -133,12 +135,13 @@ private Lazy(Supplier<? extends T> supplier) {
      */
     @Override
     public T get() {
+        // using a local var speeds up the double-check idiom by 25%, see Effective Java, Item 71
         Supplier<? extends T> tmp = supplier;
```
paplorinc: I strongly think we should base speedups on JMH benchmarks, our intuition is most often wrong about these ...

javaslang/src/main/java/javaslang/Lazy.java
```diff
@@ -133,12 +135,13 @@ private Lazy(Supplier<? extends T> supplier) {
...
         if (tmp != null) {
             synchronized (this) {
                 tmp = supplier;
                 if (tmp != null) {
-                    value = supplier.get();
+                    value = tmp.get();
```
paplorinc: could you please rename `tmp` to something meaningful?

javaslang/src/main/java/javaslang/Lazy.java
```diff
@@ -133,12 +135,13 @@ private Lazy(Supplier<? extends T> supplier) {
      */
     @Override
     public T get() {
+        // using a local var speeds up the double-check idiom by 25%, see Effective Java, Item 71
```
paplorinc: this comment if for review only, I don't think it should stay in the code.
25%, based on what?

paplorinc: this comment if for review only, I don't think it should stay in the code

paplorinc:
?

danieldietrich:
Aargh, sorry @paplorinc. I've overseen the comments. Will apply your comments soon and send a new PR.

javaslang/src/main/java/javaslang/Lazy.java
danieldietrich: I agree that we should remove the speed-up comment.

However, it is hard to recognize design patterns/idioms when looking at code. Code is just a textual projection of abstract ideas (very much like 3D to 2D/Screen projection).

I want a comment for developers why we added a local var instead of using the well-known classical double-check idiom.

javaslang/src/main/java/javaslang/Lazy.java
danieldietrich: Will add a (local) test and post the results...

javaslang/src/main/java/javaslang/Lazy.java
danieldietrich: Reads of volatile variables need to access the main memory to ensure the value is actual. `tmp` could be named `cachedSupplier` or `localSupplier`.

Because we return `value` instead of `supplier`, we do not need to cache the `supplier` before entering the synchronized block.

Within the synchronized block we have two reads (1. not-null check, 2. get() call). So a changed supplier saves one main-memory read.

The optimized code then looks like this:

```java
public T get() {
    if (this.supplier != null) {
        synchronized (this) {
            // only one volatile read from main-memory
            Supplier<? extends T> supplier = this.supplier;
            if (supplier != null) {
                value = supplier.get();
                this.supplier = null; // free mem
            }
        }
    }
    return value;
}
```

Issue #1545, opened 2016-08-28
Add Tuple.fromEntry(java.util.Map.Entry) factory method and Tuple2.toEntry() conversion method


danieldietrich:
We recognize Javaslang <-> Java conversion methods by fromXxx, toXxx names

```java
java.util.Map.Entry<K, V> entry = ...;

// create Javaslang Tuple2 from Java Map.Entry
Tuple2<K, V> tuple = Tuple.fromEntry(entry);

// convert Javaslang Tuple2 to Java Map.Entry
java.util.Map.Entry<K, V> entry2 = tuple.toEntry();

assertThat(entry).isEqualTo(entry2);
```

ashrko619:
@danieldietrich I'd like to do this :)

danieldietrich:
@ashrko619 would love to see your PR!

emmanueltouzery:
should be closed I guess?

danieldietrich:
Right! Four eyes are better than two :)

PR #1572, opened 2016-09-18
Added tuple2 - Map.Entry conversion methods

ashrko619:
PR for #1545

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -987,6 +987,10 @@ def generateMainClasses(): Unit = {
       val Seq = im.getType("javaslang.collection.Seq")
       val List = im.getType("javaslang.collection.List")
       val Iterator = im.getType("javaslang.collection.Iterator")
+      if(i==2){
+        im.getType("java.util.Map")
+        im.getType("java.util.AbstractMap")
+      }
 
       xs"""
         /**
@@ -1119,6 +1123,17 @@ def generateMainClasses(): Unit = {
               public Tuple2<T2, T1> swap() {
                 return Tuple.of(_2, _1);
               }
+
+             /$javadoc
+             * Converts the tuple to java.util.Map.Entry {@code Tuple}.
+             *
+             * @return A  java.util.Map.Entry where the first element is the key and the second
+             * element is the value.
+             */
+              public Map.Entry<T1,T2> toEntry(){
+                  return new AbstractMap.SimpleEntry<>(_1, _2);
+              }
+
             """)}
 
             ${(i > 0).gen(xs"""
@@ -1315,6 +1330,7 @@ def generateMainClasses(): Unit = {
     def genBaseTuple(im: ImportManager, packageName: String, className: String): String = {
 
       val Seq = im.getType("javaslang.collection.Seq")
+      val Map = im.getType("java.util.Map")
 
       def genFactoryMethod(i: Int) = {
         val generics = (1 to i).gen(j => s"T$j")(", ")
@@ -1330,6 +1346,18 @@ def generateMainClasses(): Unit = {
           static <$generics> Tuple$i<$generics> of($paramsDecl) {
               return new Tuple$i<>($params);
           }
+          ${(i == 2).gen(xs"""
+
+           /**
+           * Creates a tuple of ${i.numerus("element")} from java.util.Map.Entry.
+           * @param entry A java.util.Map.Entry
+           * @return a tuple of ${i.numerus("element")}.
+           */
+            static <$generics> Tuple$i<$generics> fromEntry(Map.Entry<T1, T2> entry) {
+                return new Tuple$i<>(entry.getKey(), entry.getValue());
+            }
+
+              """)}
         """
       }
 
@@ -1894,6 +1922,11 @@ def generateTestClasses(): Unit = {
         val intGenerics = if (i == 0) "" else s"<${(1 to i).gen(j => s"Integer")(", ")}>"
         val functionArgs = if (i == 0) "()" else s"${(i > 1).gen("(") + (1 to i).gen(j => s"o$j")(", ") + (i > 1).gen(")")}"
         val nullArgs = (1 to i).gen(j => "null")(", ")
+        if(i==2){
+          im.getType("java.util.AbstractMap")
+          im.getType("java.util.Map")
+        }
+
 
         xs"""
           public class Tuple${i}Test {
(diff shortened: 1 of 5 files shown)


javaslang/generator/Generator.scala
```diff
@@ -1330,6 +1346,18 @@ def generateMainClasses(): Unit = {
...
               return new Tuple$i<>($params);
           }
+          ${(i == 2).gen(xs"""
+
+           /**
+           * Creates a tuple of ${i.numerus("element")} from java.util.Map.Entry.
+           * @param entry A java.util.Map.Entry
+           * @return a tuple of ${i.numerus("element")}.
+           */
+            static <$generics> Tuple$i<$generics> fromEntry(Map.Entry<T1, T2> entry) {
```
ruslansennov: Please use `$generics` instead of `T1, T2`

javaslang/generator/Generator.scala
```diff
@@ -1119,6 +1123,17 @@ def generateMainClasses(): Unit = {
...
               }
+
+             /$javadoc
+             * Converts the tuple to java.util.Map.Entry {@code Tuple}.
+             *
+             * @return A  java.util.Map.Entry where the first element is the key and the second
+             * element is the value.
+             */
+              public Map.Entry<T1,T2> toEntry(){
+                  return new AbstractMap.SimpleEntry<>(_1, _2);
```
ruslansennov: just question: why you use `SimpleEntry` but not `SimpleImmutableEntry`?

javaslang/generator/Generator.scala
```diff
@@ -1957,6 +1990,14 @@ def generateTestClasses(): Unit = {
                 public void shouldSwap() {
                     $assertThat(createIntTuple(1, 2).swap()).isEqualTo(createIntTuple(2, 1));
                 }
+
+                @$test
+                public void shouldConvertToEntry() {
+                    Tuple2<Integer, Integer> tuple= createIntTuple(1,2);
```
ruslansennov: `Tuple$i$generics`

javaslang/generator/Generator.scala
```diff
@@ -1957,6 +1990,14 @@ def generateTestClasses(): Unit = {
...
+                    Map.Entry<Integer,Integer> entry = new AbstractMap.SimpleEntry<>(1, 2);
```
ruslansennov: Map.Entry$generics

javaslang/generator/Generator.scala
ashrko619: I wasn't sure on what to use, so went with ```SimpleEntry``` - will change it to ```SimpleImmutableEntry``` :)

danieldietrich: @ashrko619 @ruslansennov Let's return a `SimpleEntry`.

Javaslang is immutable-by-design. The `java.util` package is a mutable-by-default parallel universe. The `Map.Entry` has a `setValue()` method. Returning an object that throws would be unsafe.

Thx for asking and giving an alternative. 👏

javaslang/src-gen/main/java/javaslang/Tuple.java
```diff
@@ -95,6 +96,15 @@ static Tuple0 empty() {
     }
 
     /**
+    * Creates a tuple of two elements from java.util.Map.Entry.
+    * @param entry A java.util.Map.Entry
+    * @return a tuple of two elements.
+    */
+     static <T1, T2> Tuple2<T1, T2> fromEntry(Map.Entry<T1, T2> entry) {
```
danieldietrich: I think we can widen the argument types:

```java
static <T1, T2> Tuple2<T1, T2> fromEntry(Map.Entry<? extends T1, ? extends T2> entry) { ...
```

Only immutable types are covariant in generic arguments, here is the classical counter example (taken from [StackOverflow](http://stackoverflow.com/a/18666878/1110815)):

```java
// Illegal code - because otherwise life would be Bad
List<Dog> dogs = new List<Dog>();
List<Animal> animals = dogs; // Awooga awooga
animals.add(new Cat());
Dog dog = dogs.get(0); // This should be safe, right?
```

Projected to our use-case, this can't happen:

```java
// Illegal code - because otherwise life would be Bad
Map.Entry<T, Dog> t_dog = new SimpleEntry<>(t, dog);
Tuple<T, Animal> t_animal = Tuple.fromEntry(entry);
Tuple<T, Animal> t_animal2 = t_animal.update2(new Cat());
Dog dog = t_dog.getValue(); // ok
```

**Side-note:** We can improve the generics of other types, too. We will do it in another PR.

```java
public interface Validation<E, T> extends Value<T> {
    // should be `Either<? extends E, ? extends T> either`
    static <E, T> Validation<E, T> fromEither(Either<E, T> either) { ... }
}
```

ashrko619: _Only immutable types are covariant in generic arguments_ - Nice 👍 
@danieldietrich will make the changes and thanks for the example 😄

javaslang/src-gen/test/java/javaslang/Tuple2Test.java
```diff
@@ -91,6 +93,13 @@ public void shouldSwap() {
     }
 
     @Test
+    public void shouldConvertToEntry() {
+        Tuple2<Integer, Integer> tuple= createIntTuple(1,2);
```
danieldietrich: minor (space): `tuple =` instead of `tuple=`

javaslang/src-gen/main/java/javaslang/Tuple2.java
```diff
@@ -133,15 +135,25 @@ public T2 _2() {
...
-      return Tuple.of(_2, _1);
-    }
+    * Converts the tuple to java.util.Map.Entry {@code Tuple}.
+    *
+    * @return A  java.util.Map.Entry where the first element is the key and the second
+    * element is the value.
+    */
+     public Map.Entry<T1, T2> toEntry(){
+         return new AbstractMap.SimpleEntry<>(_1, _2);
+     }
```
danieldietrich: The changes look good to me, thank you!
We are nearly there.
Before I merge, please align the lines right. The indentation needs to be fixed in the generator.
See also https://github.com/ashrko619/javaslang/blob/26257170f75eac814354e5f27bca6c48a7455a52/javaslang/src-gen/main/java/javaslang/Tuple2.java#L138-L156

danieldietrich:
Great, thank you!

Issue #1531, opened 2016-08-25
Align to Scala 2.12 Future API


danieldietrich:
Viktor Klang created great blog posts on the changes of Future in Scala 2.12:
- [x] [Missing canonical combinators: flatten](http://viktorklang.com/blog/Futures-in-Scala-2.12-part-1) (see #241) - _we use flatMap(t -&gt; t) instead_
- [ ] [Missing canonical combinators: zipWith](http://viktorklang.com/blog/Futures-in-Scala-2.12-part-2)
- [ ] [Missing canonical combinators: transform](http://viktorklang.com/blog/Futures-in-Scala-2.12-part-3)
- [ ] [Missing canonical combinators: transformWith](http://viktorklang.com/blog/Futures-in-Scala-2.12-part-4)
- [x] [Missing canonical combinators: Deprecations: onSuccess and onFailure](http://viktorklang.com/blog/Futures-in-Scala-2.12-part-5) - _in Vavr these methods return the Future instance, so they can be chained. In our case forEach is not a total version of onSuccess. We keep all methods._
- [ ] [Missing utilities: unit & never](http://viktorklang.com/blog/Futures-in-Scala-2.12-part-6)
- [x] [Deprecations and Refactors](http://viktorklang.com/blog/Futures-in-Scala-2.12-part-7) _We have refactors, see #2093, especially we removed Promise._
- [x] [Goodbye, sun.misc.Unsafe](http://viktorklang.com/blog/Futures-in-Scala-2.12-part-8) - _this does not apply to Vavr_
- [x] [Feature: Improved Promise Linking (I promise!)](http://viktorklang.com/blog/Futures-in-Scala-2.12-part-9) - _Currently out of scope for us. I think the risk in Java isn't that high for OOME because there is no tailrec. _

We should add additional combinators and see if the other blog posts (6-9) are relevant to Javaslang's Future.

See also #1529.

paplorinc:
FYI: http://www.scala-lang.org/news/2.12.0

hepin1989:
Alias to Scala 2.13

danieldietrich:
Our Future implementation will be dropped. There will be no Vavr 1.0 module `vavr-concurrent`.

Reason: Implementing our own concurrency library is too error-prone, concurrency is a hard problem. There are many(!) alternatives, first of all Java's native CompletableFuture

PR #1574, opened 2016-09-18
#1531: Implemented zipWith and transformResult

nicolaferraro:
I enjoyed playing with this library and I also implemented two of the methods listed in the linked issue.

I named `transformResult` the new transform function because of type erasure... this is still Java, even if you did a great work to improve it.

I wonder how you would implement `flatten` without implicits :)

See you!

### DIFF
--- a/javaslang/src/main/java/javaslang/concurrent/Future.java
+++ b/javaslang/src/main/java/javaslang/concurrent/Future.java
@@ -933,7 +933,7 @@ public interface Future<T> extends Value<T> {
     @Override
     default <U> Future<U> map(Function<? super T, ? extends U> mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
-        return mapTry(mapper::apply);
+        return transformResult(t -> t.map(mapper::apply));
     }
 
     default <U> Future<U> mapTry(CheckedFunction<? super T, ? extends U> mapper) {
--- a/javaslang/src/main/java/javaslang/concurrent/Future.java
+++ b/javaslang/src/main/java/javaslang/concurrent/Future.java
@@ -5,6 +5,7 @@
  */
 package javaslang.concurrent;
 
+import javaslang.Function2;
 import javaslang.Tuple;
 import javaslang.Tuple2;
 import javaslang.Value;
@@ -747,9 +748,7 @@ public interface Future<T> extends Value<T> {
      */
     default Future<T> recover(Function<? super Throwable, ? extends T> f) {
         Objects.requireNonNull(f, "f is null");
-        final Promise<T> promise = Promise.make(executorService());
-        onComplete(t -> promise.complete(t.recover(f)));
-        return promise.future();
+        return transformResult(t -> t.recover(f::apply));
     }
 
     /**
@@ -792,6 +791,23 @@ public interface Future<T> extends Value<T> {
     }
 
     /**
+     * Transforms the result of this {@code Future}, whether it is a success or a failure.
+     *
+     * @param f   A transformation
+     * @param <U> Generic type of transformation {@code Try} result
+     * @return A {@code Future} of type {@code U}
+     * @throws NullPointerException if {@code f} is null
+     */
+    default <U> Future<U> transformResult(Function<? super Try<T>, ? extends Try<U>> f) {
+        Objects.requireNonNull(f, "f is null");
+        final Promise<U> promise = Promise.make(executorService());
+        onComplete(t -> {
+            Try.run(() -> promise.complete(f.apply(t))).onFailure(promise::failure);
+        });
+        return promise.future();
+    }
+
+    /**
      * Returns a tuple of this and that Future result.
      * <p>
      * If this Future failed the result contains this failure. Otherwise the result contains that failure or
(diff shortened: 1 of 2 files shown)


nicolaferraro:
Related to https://github.com/javaslang/javaslang/issues/1531

javaslang/src/main/java/javaslang/concurrent/Future.java
```diff
@@ -792,6 +791,23 @@ default boolean isFailure() {
...
 
     /**
+     * Transforms the result of this {@code Future}, whether it is a success or a failure.
+     *
+     * @param f   A transformation
+     * @param <U> Generic type of transformation {@code Try} result
+     * @return A {@code Future} of type {@code U}
+     * @throws NullPointerException if {@code f} is null
+     */
+    default <U> Future<U> transformResult(Function<? super Try<T>, ? extends Try<U>> f) {
```
danieldietrich: The signature is OK and aligns to existing `transform()` methods that are already present.

I'm think need this instead (all over the lib, in another PR):

```java
<U> Future<U> transformResult(Function<? super Try<? super T>, ? extends Try<? extends U>> f)
```

See also [Sebastian's example](https://github.com/javaslang/javaslang/issues/352#issuecomment-125876580).

javaslang/src/main/java/javaslang/concurrent/Future.java
```diff
@@ -802,16 +818,35 @@ default boolean isFailure() {
...
+     *
+     * @param that Another Future
+     * @param combinator The combinator function
+     * @param <U>  Result type of {@code that}
+     * @param <R>  Result type of {@code f}
+     * @return A new Future that returns both Future results.
+     * @throws NullPointerException if {@code that} is null
+     */
+    @SuppressWarnings("unchecked")
+    default <U, R> Future<R> zipWith(Future<? extends U> that, Function2<T, U, R> combinator) {
```
danieldietrich: Please use a `BiFunction` instead of a `Function2` (which is a `BiFunction`) and add upper/lower type bounds:

```java
<U, R> Future<R> zipWith(Future<? extends U> that, BiFunction<? super T, ? super U, ? extends R> combinator)
```

javaslang/src/main/java/javaslang/concurrent/Future.java
```diff
@@ -802,16 +818,35 @@ default boolean isFailure() {
...
+        Objects.requireNonNull(combinator, "combinator is null");
+        final Promise<R> promise = Promise.make(executorService());
         onComplete(res1 -> {
             if (res1.isFailure()) {
-                promise.complete((Try.Failure<Tuple2<T, U>>) res1);
+                promise.complete((Try.Failure<R>) res1);
             } else {
                 that.onComplete(res2 -> {
-                    final Try<Tuple2<T, U>> result = res1.flatMap(t -> res2.map(u -> Tuple.of(t, u)));
+                    final Try<R> result = res1.flatMap(t -> res2.map(u -> combinator.apply(t, u)));
```
danieldietrich: Mmh, please add a unit test that let's the combinator fail, e.g. a `combinator` that throws a `RuntimeException`.

I expect that the `Future` that is returned by `zipWith` does never return.

**Note to myself:** We need also to check the other Future/FutureImpl methods. It must be ensured that in each case we get a defined result. We can do it by adding tests that use lambda arguments that just throw a RuntimeException. `map` delegated to `mapTry`, which should handle that case. The current optimization may change this behavior. #LetItCrash

danieldietrich: ...but I think the `Try.run(...).onFailure(promise::failure)` part of `transformResult` already handles this case.

Nevertheless, having a test helps to ensure it.

danieldietrich:
> I named transformResult the new transform function because of type erasure...

The current API names the underlying `Try<T>` value and has the following methods:

```java
get(): T
getCause(): Option<Throwable>
getValue(): Option<Try<T>>
```

So I think the best name choice for our transform method is `transformValue()` in order to align with the existing `getValue()`.

Please rename it along with my other review comments.

Many thanks!

nicolaferraro:
Yeah, `transformValue` is the best choice. I'm going to apply the changes.

danieldietrich:
Great, thanks :) Looking forward to pulling it!

nicolaferraro:
Only an issue with the `transformValue` function. I used this one:
```java
default <U> Future<U> transformValue(Function<? super Try<T>, ? extends Try<? extends U>> f)
```

Changing it to:
```java
default <U> Future<U> transformValue(Function<? super Try<? super T>, ? extends Try<? extends U>> f)
```
... is probably more correct, but forces me to add unchecked casts in every point I use the `transformValue` function.

danieldietrich:
@nicolaferraro Thank you, we can trust the compiler. Your version `? super Try<T>` is correct!

```java
import javaslang.control.Try;

import java.util.function.Function;

public class Test {

        Function<Try<String>, Try.Success<Integer>> f1 = v -> (Try.Success) Try.success(1);
        Function<Value<CharSequence>, Try.Success<Integer>> f2 = v -> (Try.Success) Try.success(1);
        Function<Value<String>, Try.Success<Integer>> f3 = v -> (Try.Success) Try.success(1);
        Function<Value<CharSequence>, Try.Success<Integer>> f4 = v -> (Try.Success) Try.success(1);
        Function<Try<String>, Try<Number>> f5 = v -> Try.success(1);
        Function<? super Try<? super String>, Try.Success<Integer>> f6 = v -> (Try.Success) Try.success(1);

        Future<String> future = null;

        Future<Number> result0_1 = future.transformValue0(f1); // OK
        Future<Number> result0_2 = future.transformValue0(f2); // Error
        Future<Number> result0_3 = future.transformValue0(f3); // Error
        Future<Number> result0_4 = future.transformValue0(f4); // Error
        Future<Number> result0_5 = future.transformValue0(f5); // OK
        Future<Number> result0_6 = future.transformValue0(f6); // Error

        Future<Number> result1_1 = future.transformValue1(f1); // OK
        Future<Number> result1_2 = future.transformValue1(f2); // Error
        Future<Number> result1_3 = future.transformValue1(f3); // OK
        Future<Number> result1_4 = future.transformValue1(f4); // Error
        Future<Number> result1_5 = future.transformValue1(f5); // OK
        Future<Number> result1_6 = future.transformValue1(f6); // OK

        Future<Number> result2_1 = future.transformValue2(f1); // Error
        Future<Number> result2_2 = future.transformValue2(f2); // Error
        Future<Number> result2_3 = future.transformValue2(f3); // Error
        Future<Number> result2_4 = future.transformValue2(f4); // Error
        Future<Number> result2_5 = future.transformValue2(f5); // Error
        Future<Number> result2_6 = future.transformValue2(f6); // OK

        Future<Number> result3_1 = future.transformValue3(f1); // Error
        Future<Number> result3_2 = future.transformValue3(f2); // Error
        Future<Number> result3_3 = future.transformValue3(f3); // Error
        Future<Number> result3_4 = future.transformValue3(f4); // Error
        Future<Number> result3_5 = future.transformValue3(f5); // Error
        Future<Number> result3_6 = future.transformValue3(f6); // Error

    }

    interface Future<T> extends Value<T> {

        <U> Future<U> transformValue0(Function<Try<T>, ? extends Try<? extends U>> f);

        <U> Future<U> transformValue1(Function<? super Try<T>, ? extends Try<? extends U>> f);

        <U> Future<U> transformValue2(Function<? super Try<? super T>, ? extends Try<? extends U>> f);

        <U> Future<U> transformValue3(Function<Try<? super T>, ? extends Try<? extends U>> f);

    }
}
```

danieldietrich:
Thank you, great addition!

PR #1576, opened 2016-09-18
Applied PR #1571 review comments of @paplorinc

danieldietrich:
Follow-up PR of #1571

### DIFF
--- a/javaslang/src/main/java/javaslang/Lazy.java
+++ b/javaslang/src/main/java/javaslang/Lazy.java
@@ -135,14 +135,13 @@ public final class Lazy<T> implements Value<T>, Supplier<T>, Serializable {
      */
     @Override
     public T get() {
-        // using a local var speeds up the double-check idiom by 25%, see Effective Java, Item 71
-        Supplier<? extends T> tmp = supplier;
-        if (tmp != null) {
+        if (this.supplier != null) {
             synchronized (this) {
-                tmp = supplier;
-                if (tmp != null) {
-                    value = tmp.get();
-                    supplier = null; // free mem
+                // only one volatile read from main-memory
+                final Supplier<? extends T> supplier = this.supplier;
+                if (supplier != null) {
+                    value = supplier.get();
+                    this.supplier = null; // free mem
                 }
             }
         }


danieldietrich:
@paplorinc please review

danieldietrich:
...still need to do the benchmarks...

danieldietrich:
@paplorinc I need your JMH expertise :)

In one benchmark I need an _uninitialized_ lazy value. We loop an amount of TIMES. Within the loop the lazy val needs to be not initialized on each call, to get sure we enter the `synchronized` block.

In fact we need also a test case with several threads, which is hard to benchmark because the system scheduler is not deterministic :-(

* how do I ensure that the testee is not initialized? Do I use `@Setup` with `Level.Invocation`? This is not recommended (see http://java-performance.info/jmh/)
* get() returns an int value `1`. Does the boxing influence the test in a negative way? I think not because it is a constant factor, regardless of the test case.


```java
        @Benchmark
        public int slang_localVar() {
            int aggregate = 0;
            for (int i = 1; i <= TIMES; i++) {
                // doubleCheckLocal needs to be uninitialized
                aggregate += doubleCheckLocal.get(); // get() returns 1
            }
            assert aggregate == TIMES;
            return aggregate;
        }
```

Thx in advance!

paplorinc:
Memoization (mutation / multi threading in general) performance is difficult measure properly.
I tried to come up with a good benchmarks, but I modified more things than I should quote here, so I created a separate PR for it (since it includes an alternative optimization also).
I hope you don't mind me hijacking your PR :)

The benchmark compares simple eager access (`java_eager`) to `Lazy`'s first access (`slang_lazy`) and subsequent, initialized access (`slang_inited_lazy`):

These are the measurements:
```java
Before:
    java_eager/slang_inited_lazy  1.17×
    java_eager/slang_lazy         6.89×

Current PR:
    java_eager/slang_inited_lazy  1.07×
    java_eager/slang_lazy         6.56×

Alternative:
    java_eager/slang_inited_lazy  1.02×
    java_eager/slang_lazy         6.73×
```

i.e. before is the worst, your solution is faster for first access, the alternative is faster after the first access.

See: #1577

danieldietrich:
Improved by @paplorinc in #1577 and therefore closed as duplicate.

PR #1577, opened 2016-09-19
`Lazy` access minor speedup

paplorinc:
Alternative to #1576
Memoization (mutation / multi threading in general) performance is difficult to measure properly.

The benchmark compares simple eager access (`java_eager`) to `Lazy`'s first access (`slang_lazy`) and subsequent, initialized access (`slang_inited_lazy`):

Since running the same measurement multiple times would measure different things (the first access would memoize, the rest would return the result of the calculation), I created a small array of `Lazy` values (small enough [to fit in the cache](https://github.com/emilk/ram_bench#measuring-it) but big enough to make the array access cost dissolve), and measure consuming it - recreating it before every measurement to assure uninitialized `Lazy`.

These are the results (please run them on your PC also via `javaslang.control.LazyBenchmark#main`):
```java
Before:
  java_eager/slang_inited_lazy 1.11×
  java_eager/slang_lazy        6.79×

After:
  java_eager/slang_inited_lazy 1.06×
  java_eager/slang_lazy        6.78×
```

Note: a proper test is still missing, thanks for the comments.

### DIFF
--- a/javaslang-benchmark/src/test/java/javaslang/control/LazyBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/control/LazyBenchmark.java
@@ -1,0 +1,83 @@
+package javaslang.control;
+
+import javaslang.JmhRunner;
+import javaslang.Lazy;
+import javaslang.collection.Array;
+import javaslang.collection.Iterator;
+import org.junit.Test;
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.infra.Blackhole;
+
+import static javaslang.JmhRunner.Includes.JAVA;
+import static javaslang.JmhRunner.Includes.JAVASLANG;
+
+public class LazyBenchmark {
+    static final Array<Class<?>> CLASSES = Array.of(
+            Get.class
+    );
+
+    @Test
+    public void testAsserts() { JmhRunner.runDebugWithAsserts(CLASSES); }
+
+    public static void main(String... args) {
+        JmhRunner.runDebugWithAsserts(CLASSES, JAVA, JAVASLANG);
+        JmhRunner.runSlowNoAsserts(CLASSES, JAVA, JAVASLANG);
+    }
+
+    @State(Scope.Benchmark)
+    public static class Base {
+        final int SIZE = 10;
+
+        Integer[] EAGERS;
+        javaslang.Lazy<Integer>[] INITED_LAZIES;
+
+        @Setup
+        @SuppressWarnings({ "unchecked", "rawtypes" })
+        public void setup() {
+            EAGERS = Iterator.range(0, SIZE).toJavaArray(Integer.class);
+            INITED_LAZIES = Iterator.of(EAGERS).map(i -> {
+                final Lazy<Integer> lazy = Lazy.of(() -> i);
+                lazy.get();
+                return lazy;
+            }).toJavaList().toArray(new Lazy[0]);
+        }
+    }
+
+    @Threads(4)
+    @SuppressWarnings({ "WeakerAccess", "rawtypes" })
+    public static class Get extends Base {
+        @State(Scope.Thread)
+        public static class Initialized {
+            javaslang.Lazy<Integer>[] LAZIES;
+
+            @Setup(Level.Invocation)
+            @SuppressWarnings("unchecked")
+            public void initializeMutable(Base state) {
+                LAZIES = Iterator.of(state.EAGERS).map(i -> Lazy.of(() -> i)).toJavaList().toArray(new Lazy[0]);
+            }
+        }
+
+        @Benchmark
+        public void java_eager(Blackhole bh) {
+            for (int i = 0; i < SIZE; i++) {
+                bh.consume(EAGERS[i]);
+            }
+        }
+
+        @Benchmark
+        public void slang_inited_lazy(Blackhole bh) {
+            for (int i = 0; i < SIZE; i++) {
+                assert INITED_LAZIES[i].isEvaluated();
+                bh.consume(INITED_LAZIES[i].get());
+            }
+        }
+
+        @Benchmark
+        public void slang_lazy(Initialized state, Blackhole bh) {
+            for (int i = 0; i < SIZE; i++) {
+                assert !state.LAZIES[i].isEvaluated();
+                bh.consume(state.LAZIES[i].get());
+            }
+        }
+    }
+}
--- a/javaslang/src/main/java/javaslang/Lazy.java
+++ b/javaslang/src/main/java/javaslang/Lazy.java
@@ -7,13 +7,20 @@ package javaslang;
 
 import javaslang.collection.Iterator;
 import javaslang.collection.List;
-import javaslang.collection.*;
+import javaslang.collection.Seq;
 import javaslang.control.Option;
 
-import java.io.*;
-import java.lang.reflect.*;
-import java.util.*;
-import java.util.function.*;
+import java.io.IOException;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Proxy;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
 
 /**
  * Represents a lazy evaluated value. Compared to a Supplier, Lazy is memoizing, i.e. it evaluates only once and
@@ -44,9 +51,7 @@ public final class Lazy<T> implements Value<T>, Supplier<T>, Serializable {
 
     // read http://javarevisited.blogspot.de/2014/05/double-checked-locking-on-singleton-in-java.html
     private transient volatile Supplier<? extends T> supplier;
-
-    // does not need to be volatile, visibility piggy-backs on volatile read of `supplier`
-    private T value;
+    private T value; // will behave as a volatile in reality, because a supplier volatile read will update all fields (see https://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html#volatile)
 
     // should not be called directly
     private Lazy(Supplier<? extends T> supplier) {
@@ -135,16 +140,13 @@ public final class Lazy<T> implements Value<T>, Supplier<T>, Serializable {
      */
     @Override
     public T get() {
-        // using a local var speeds up the double-check idiom by 25%, see Effective Java, Item 71
-        Supplier<? extends T> tmp = supplier;
-        if (tmp != null) {
-            synchronized (this) {
-                tmp = supplier;
-                if (tmp != null) {
-                    value = tmp.get();
-                    supplier = null; // free mem
-                }
-            }
+        return (supplier == null) ? value : computeValue();
+    }
+    private synchronized T computeValue() {
+        final Supplier<? extends T> s = supplier;
+        if (s != null) {
+            value = s.get();
+            supplier = null;
         }
         return value;
     }
--- a/javaslang/src/test/java/javaslang/LazyTest.java
+++ b/javaslang/src/test/java/javaslang/LazyTest.java
@@ -5,16 +5,19 @@
  */
 package javaslang;
 
-import javaslang.collection.List;
-import javaslang.collection.Seq;
+import javaslang.collection.*;
 import javaslang.control.Option;
 import javaslang.control.Try;
 import org.junit.Test;
 
 import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
 
+import static java.util.concurrent.CompletableFuture.runAsync;
 import static javaslang.Serializables.deserialize;
 import static javaslang.Serializables.serialize;
+import static javaslang.collection.Iterator.range;
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class LazyTest {
@@ -77,7 +80,7 @@ public class LazyTest {
         final Lazy<Integer> testee = Lazy.of(() -> 42);
         final Lazy<Integer> expected = Lazy.of(() -> 21);
 
-        assertThat(testee.map( i -> i / 2)).isEqualTo(expected);
+        assertThat(testee.map(i -> i / 2)).isEqualTo(expected);
     }
 
     @Test
@@ -86,8 +89,8 @@ public class LazyTest {
         final Option<Integer> expectedPositive = Option.some(42);
         final Option<Integer> expectedNegative = Option.none();
 
-        assertThat(testee.filter( i -> i % 2 == 0)).isEqualTo(expectedPositive);
-        assertThat(testee.filter( i -> i % 2 != 0)).isEqualTo(expectedNegative);
+        assertThat(testee.filter(i -> i % 2 == 0)).isEqualTo(expectedPositive);
+        assertThat(testee.filter(i -> i % 2 != 0)).isEqualTo(expectedNegative);
     }
 
     @Test
@@ -95,7 +98,7 @@ public class LazyTest {
         final Lazy<Integer> testee = Lazy.of(() -> 42);
         final Integer expected = 21;
 
-        final Integer actual = testee.transform( lazy -> lazy.get() / 2 );
+        final Integer actual = testee.transform(lazy -> lazy.get() / 2);
 
         assertThat(actual).isEqualTo(expected);
     }
@@ -173,9 +176,10 @@ public class LazyTest {
 
     @Test
     public void shouldSupportMultithreading() {
-        final boolean[] lock = new boolean[] { true };
+        final AtomicBoolean isEvaluated = new AtomicBoolean();
+        final AtomicBoolean lock = new AtomicBoolean();
         final Lazy<Integer> lazy = Lazy.of(() -> {
-            while (lock[0]) {
+            while (lock.get()) {
                 Try.run(() -> Thread.sleep(300));
             }
             return 1;
@@ -184,12 +188,38 @@ public class LazyTest {
             Try.run(() -> Thread.sleep(100));
             new Thread(() -> {
                 Try.run(() -> Thread.sleep(100));
-                lock[0] = false;
+                lock.set(false);
             }).start();
-            assertThat(lazy.isEvaluated()).isFalse();
+            isEvaluated.compareAndSet(false, lazy.isEvaluated());
             lazy.get();
         }).start();
+        assertThat(isEvaluated.get()).isFalse();
         assertThat(lazy.get()).isEqualTo(1);
+    }
+
+    @Test
+    @SuppressWarnings({ "StatementWithEmptyBody", "rawtypes" })
+    public void shouldBeConsistentFromMultipleThreads() throws Exception {
+        for (int i = 0; i < 100; i++) {
+            final AtomicBoolean canProceed = new AtomicBoolean(false);
+            final Vector<CompletableFuture<Void>> futures = Vector.range(0, 10).map(j -> {
+                final AtomicBoolean isEvaluated = new AtomicBoolean(false);
+                final Integer expected = ((j % 2) == 1) ? null : j;
+                Lazy<Integer> lazy = Lazy.of(() -> {
+                    assertThat(isEvaluated.getAndSet(true)).isFalse();
+                    return expected;
+                });
+                return Tuple.of(lazy, expected);
+            }).flatMap(t -> range(0, 5).map(j -> runAsync(() -> {
+                        while (!canProceed.get()) { /* busy wait */ }
+                        assertThat(t._1.get()).isEqualTo(t._2);
+                    }))
+            );
+
+            final CompletableFuture all = CompletableFuture.allOf(futures.toJavaList().toArray(new CompletableFuture<?>[0]));
+            canProceed.set(true);
+            all.join();
+        }
     }
 
     // -- equals
@@ -199,7 +229,7 @@ public class LazyTest {
         assertThat(Lazy.of(() -> 1).equals("")).isFalse();
         assertThat(Lazy.of(() -> 1).equals(Lazy.of(() -> 1))).isTrue();
         assertThat(Lazy.of(() -> 1).equals(Lazy.of(() -> 2))).isFalse();
-        Lazy<Integer> same = Lazy.of(() -> 1);
+        final Lazy<Integer> same = Lazy.of(() -> 1);
         assertThat(same.equals(same)).isTrue();
     }
 


javaslang/src/main/java/javaslang/Lazy.java
```diff
@@ -135,14 +133,11 @@ private Lazy(Supplier<? extends T> supplier) {
      */
     @Override
     public T get() {
-        // using a local var speeds up the double-check idiom by 25%, see Effective Java, Item 71
-        Supplier<? extends T> tmp = supplier;
-        if (tmp != null) {
+        if ((value == null) && (supplier != null)) {
```
danieldietrich: Wonderful idea! I have two remarks:

**1) supplier.get() may return null**

Our value has two possible null-states: null as undefined and null as initialized state. If it is initialized with null, the get() calls will be slower than in #1576.

**2) two reads from main memory within the synchronized block**

Because supplier is volatile, all occurrences will be read from main memory, in particular:

* if (supplier != null) { ... }
* value = supplier.get();

The following solution solves 1) and 2) by fusing the ideas of #1577  with those of #1576 and additionally adding an UNDEFINED object:

```java
    private static final Object UNDEFINED = new Object();

    private transient volatile Supplier<? extends T> supplier;

    private T value = (T) UNDEFINED;

    public T get() {
        if (value == UNDEFINED && this.supplier != null) {
            synchronized(this) {
                final Supplier<? extends T> supplier = this.supplier;
                if (supplier != null) {
                    value = supplier.get();
                    this.supplier = null;
                }
            }
        }
        return value;
    }
```

I ask myself if the isEvaluated method could now just use value like this:

```java
public boolean isEvaluated() {
    // fast read, memory synchronized with volatile write of supplier
    return value != UNDEFINED;
}
```

instead of

```java
public boolean isEvaluated() {
    // slow read from main memory
    return supplier == null;
}
```

javaslang/src/main/java/javaslang/Value.java
```diff
@@ -516,7 +516,7 @@ default CharSeq toCharSeq() {
      */
     @SuppressWarnings("unchecked")
     @GwtIncompatible("reflection is not supported")
-    default T[] toJavaArray(Class<T> componentType) {
+    default T[] toJavaArray(Class<? super T> componentType) {
```
danieldietrich: This is...... awesome :)

It seems to solve the unsolvable problem of getting the class of a generic type. I've played around with it and couldn't construct an unsafe example (yet).

Please do this change to a separate PR, this little change is really worth it. I will pull it before this PR then.

javaslang/src/main/java/javaslang/Lazy.java
```diff
@@ -44,8 +44,6 @@
 
     // read http://javarevisited.blogspot.de/2014/05/double-checked-locking-on-singleton-in-java.html
     private transient volatile Supplier<? extends T> supplier;
-
-    // does not need to be volatile, visibility piggy-backs on volatile read of `supplier`
```
danieldietrich: Maybe this comment is crap but some lines of Lazy (including this) definitely need to be commented.

I'm also no friend of blowing up concise code with _bla, bla_. But here is happening something important behind the curtain. We need to mention that non-volatile instance variables are synced to the main memory when volatile instance variables are updated. Our code strongly relies on this property.

> What does volatile do?
> (...)
> Writing to a volatile field has the same memory effect as a monitor release, and reading from a volatile field has the same memory effect as a monitor acquire. In effect, because the new memory model places stricter constraints on reordering of volatile field accesses with other field accesses, volatile or not, anything that was visible to thread A when it writes to volatile field f becomes visible to thread B when it reads f.

\- [JSR 133 FAQ](https://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html)

danieldietrich:
Wow, so many gems hidden in a small class like Lazy.

Would love to see updated benchmark results of the applied code review comments.

javaslang/src/main/java/javaslang/Lazy.java
danieldietrich: **IMPORTAT UPDATE**

When I interprete this right

> anything that was visible to thread A when it writes to volatile field f becomes visible to thread B when it reads f.

then the check `value == UNDEFINED` may not reflect the actual main memory state.

**Example:**

1. Thread A calls get(), acquires the lock and writes `value` and `supplier` (but is still in the synchronized block)
2. Thread B calls get() and checks `value == UNDEFINED`. Because Thread B did not read `supplier`, the initialized `value` gets not visible to Thread B. Then `supplier != null` is checked, which is `false`because Thread A set it to null. Because supplier was read, the initialized `value` is now visible to Thread B and the `return value` makes sense.

If the above cannot be proved to be wrong then, for the sake of correctness and safety, we should remove the first `value == UNDEFINED` check completely again. What remains is this:

```java
    private static final Object UNDEFINED = new Object();

    private transient volatile Supplier<? extends T> supplier;

    private T value = (T) UNDEFINED;

    public T get() {
        if (this.supplier != null) {
            synchronized(this) {
                final Supplier<? extends T> supplier = this.supplier;
                if (supplier != null) {
                    value = supplier.get();
                    this.supplier = null;
                }
            }
        }
        return value;
    }
```

This is exactly the solution proposed in #1576.

Also `isEvaluated()` stays as is.

danieldietrich:
Considering concurrent reads of `value` the optimization may lead to inconsistent reads (see comment above). Therefore we should implement a safe solution.
Thx for the great benchmarks!

paplorinc:
> Please do this change to a separate PR, this little change is really worth it.

Sure :)

> the get() calls will be slower than in #1576.

Indeed, but we're talking about `<5%` here, no other class is this optimal, I don't think this should be a criterion.

> then the check value == UNDEFINED may not reflect the actual main memory state

Yes, the same for `null` values, but that's what the subsequent synchronized block is for, which checks a volatile field, guaranteeing correctness. Once all the values are out of the cache, the non-volatile field check will suffice as it should only change once.

If I'm mistaken, could you please provide a test that fails?

danieldietrich:
@paplorinc 

> If I'm mistaken, could you please provide a test that fails?

Not sure if I'm able to do it. It requires multiple threads and one thread that reads an outdated `value` while checking at the same time that the `value` was updated.

I'm sure that I'm right here.

danieldietrich:
> If I'm mistaken, could you please provide a test that fails?

Will see what is possible...

javaslang/src/main/java/javaslang/Lazy.java
paplorinc: Now I got what the comment meant, I though it has historical significance only.
You're right, some comment IS needed here :)

javaslang/src/main/java/javaslang/Value.java
paplorinc: done in https://github.com/javaslang/javaslang/pull/1578

javaslang/src/main/java/javaslang/Lazy.java
jbgi: I would keep the volatile check before assuming `value` is correctly intialized (otherwise I think It is possible that it points to a "not finished" object).

I am also currently benchmarking lazy init schemes and I found that keeping the `get` method as short as possible is key to trigger inlining and the associated speedup. So the following may be prefered:
```java
    @Override
    public T get() {
        return supplier == null
            ? value
            : computeValue();
            //: computeValue_withLocalVar(); (is it faster ?)
    }
    
    private T computeValue() {
        synchronized (this) {
            if (supplier != null) {
                value = supplier.get();
                supplier = null; // free mem
            }
        }
        return value;
    }

    private T computeValue_withLocalVar() {
        T v;
        // using a local var speeds up the double-check idiom by 25% (maybe??), see Effective Java, Item 71
        synchronized (this) {
            Supplier<? extends T> tmp = supplier;
            if (tmp != null) {
                value = v = tmp.get();
                supplier = null; // free mem
            } else {
                v = value;
            }
        }
        return v;
    }
```
I could not get @paplorinc benchmark working so this still need to be validated...

danieldietrich: Hi @jbgi, inlining by keeping it small is a great idea!!
I also think that the read of `value` is dirty.
The local var `T v` most probably isn't needed because `value` isn't volatile and reads are cheap.
I will further play around with different solutions....

danieldietrich: Using a synchronized method is equivalent but more concise.

```java
private synchronized T computeValue() {
    final Supplier<? extends T> s = supplier;
    if (s != null) {
        value = s.get();
        supplier = null;
    }
    return value;
}
```

Will measure the variants later today.

paplorinc: I think we have a lot bigger problem, than optimizations: all the tests are passing with non-volatile fields also!

I tried creating a test that would fail, but haven't come up with one yet (only if I put 
`Try.run(() -> Thread.sleep(ThreadLocalRandom.current().nextInt(2)))` all over the production code):
```java
@Test
public void shouldBeConsistentFromMultipleThreads() throws Exception {
    final Vector<Tuple2<Lazy<Integer>, Integer>> values = Vector.range(0, 10000).map(i -> {
        final Integer expected = ((i % 2) == 1) ? null : i;
        final AtomicBoolean isEvaluated = new AtomicBoolean(false);
        final Lazy<Integer> lazy = Lazy.of(() -> {
            Try.run(() -> Thread.sleep(ThreadLocalRandom.current().nextInt(2)));
            assertThat(isEvaluated.getAndSet(true)).isFalse();
            return expected;
        });
        return Tuple.of(lazy, expected);
    });

    CompletableFuture.allOf(
            values.flatMap(t -> range(0, 10)
                    .map(j -> runAsync(() -> assertThat(t._1.get()).isEqualTo(t._2)))
            ).toJavaArray(CompletableFuture.class)
    ).join();
}
```

@danieldietrich, @jbgi, @zsolt-donca, can someone tell me why this isn't failing for non-volatile fields, i.e.:
```java
private Supplier<? extends T> supplier;
private T value;

@Override
public T get() {
    if (value == null) {
        synchronized (this) {
            if (supplier != null) {
                value = supplier.get();
                supplier = null;
            }
        }
    }
    return value;
}
```

(it IS failing if we do a non-synchronized, single check)

danieldietrich: When we use synchronized, this is always OK because the synch block ensures that supplier.get() (which sets the atomic boolean) is only entered by one thread at a time:

```java
assertThat(isEvaluated.getAndSet(true)).isFalse();
```

The second assert also can't fail because value has always the correct value. The worst case that can happen is, that supplier.get() is called twice (because the second thread did not see the initialized var).

The latter case (twice initialized) is what you see for non-volatile vars without the synchronized block. Because the synchronized block is missing, the first assertion fails (supplier.get() entered by two threads and atomic boolean assertion fails).

Right?

danieldietrich: @paplorinc we can do better ;) instead of

```java
CompletableFuture.allOf(
            values.flatMap(t -> range(0, 10)
                    .map(j -> runAsync(() -> assertThat(t._1.get()).isEqualTo(t._2)))
            ).toJavaArray(CompletableFuture.class)
    ).join();
```

we could use

```java
Future.sequence(
    values.flatMap(t -> range(0, 10)
                    .map(j -> runAsync(() -> assertThat(t._1.get()).isEqualTo(t._2)))
)/*.await()*/;
```

danieldietrich: or

```java
Future<Void> tasks = values.flatMap(t -> range(0, 10)
                    .map(j -> runAsync(() -> assertThat(t._1.get()).isEqualTo(t._2)));
boolean isOk = Future.fold(tasks, true, (result, task) -> result & task.isSuccess());
assertThat(isOk).isTrue();
```

(warning: untested code)

zsolt-donca: I've made some tests to verify whether it's actually required, in practice, to have a field marked as volatile when used in a double-checked lock. Theoretically, the mechanism can fail without the field being marked as volatile. 

According to my tests, it turns out that this is very much JVM-dependent. My test never failed when running against a Java 7 and 8 JVM. However, it did fail against Java 6.

My conclusion is that, at least since Java 7, there are some different runtime optimizations in place that either decrease the chance of the problem happening or make the _volatile_ modifier completely unneeded in this double-checked locking scenario. However, we should definitely not rely on this apparent behavior until there is an update to the Java Memory Model that clearly states that we can do so. 

I'll give you more details as soon as I got the time.

javaslang/src/test/java/javaslang/LazyTest.java
```diff
@@ -184,22 +190,39 @@ public void shouldSupportMultithreading() {
...
+            isEvaluated.compareAndSet(false, lazy.isEvaluated());
             lazy.get();
         }).start();
+        assertThat(isEvaluated.get()).isFalse();
         assertThat(lazy.get()).isEqualTo(1);
     }
 
+    @Test
+    public void shouldBeConsistentFromMultipleThreads() throws Exception {
+        CompletableFuture.allOf(
```
paplorinc: @danieldietrich, I couldn't manage to make it work with slang `Future` (that fails for non-double check), only with `CompletableFuture`

javaslang/src/main/java/javaslang/Lazy.java
paplorinc: It seems Java 7 introduced an implicit volatile when it detects double locking.
Should we ask help from @shipilev?

shipilev: No, there is no such thing as "implicit volatile". It's just your `volatile` `Supplier` is guarantees you always read the correct `value`.

From the JMM standpoint, you have to argue if there exists an execution that has this relation: read(supplier):null --PO/HB--> read(value):null? It does not, because read(supplier):null is always in SO (and therefore in SW) after write(supplier, null), and write(value, SOME_VALUE) --PO/HB--> write(supplier, null). If there is indeed an execution that does read(value):null, then it violates HB consistency. 

Some might think the whole case looks like the usual publishing piggybacking, regardless that we store the `null` into `supplier`:

Thread 1:
  store A;
  volatile store B; 

Thread 2:
  volatile read B // observes the value from store B
  read A; // guaranteed to observe store A

Glossary: read(field):val -- reading value "val" from "field"; write(field, val) -- writing value "val" to "field"; SO -- synchronization order; SW -- synchronizes-with order; PO -- program order; HB - happens-before order.

I would refer to https://shipilev.net/blog/2014/safe-public-construction/ for tests that actually fail without `volatile`.

zsolt-donca: I have also managed to write a test case that demonstrates the necessity of using `volatile` on the field of the double-checked locking mechanism. The test works on Java 6, 7 and 8. I managed to increase the chances of the concurrency problem being detected by introducing a _busy wait_ into the testing threads, acting as a live synchronization mechanism. The entire code is available [here](http://pastebin.com/P0hnVX4i).

It seems that Java 6 has different runtime behavior and it is easier to demonstrate the problem there.

danieldietrich: @shipilev Thanks for the great explanation!
One question. We saw that write(value, SOME_VALUE) is synchronized with write(supplier, null) because supplier is volatile. What does another Thread see that only performs read(value). Does an execution exist such that read(value) = null (uninitialized) and supplier == null?

shipilev: I am losing track what version we discuss. This one is incorrect:

```
private volatile Supplier<? extends T> supplier;  // dropping volatile here make it even worse
private T value;

@Override
public T get() {
    if (value == null) { // <--- non-volatile read, this is the data race
        synchronized (this) {
            if (supplier != null) {
                value = supplier.get();
                supplier = null;
            }
        }
    }
    return value;
}
```
You *have* to read supplier (and null-check it) before reading value to get happens-before through supplier write/read.

This one is correct:

```
private volatile Supplier<? extends T> supplier; // should be volatile!
private T value;

@Override
public T get() {
    if (supplier != null) { // <--- volatile read
        synchronized (this) {
            if (supplier != null) {
                value = supplier.get();
                supplier = null;
            }
        }
    }
    return value;
}
```

danieldietrich: @shipilev many thanks for the great explanation!! :)

@paplorinc @jbgi then we should use a straight forward solution and add two optimizations

1) enable inlining by keeping the get() method small
2) add a local var that holds the volatile instance var within the synchronized block to reduce the volatile reads from 2 to only 1.

**Solution:**

```java
private volatile Supplier<? extends T> supplier;
private T value;

@Override
public T get() {
    return (supplier == null) ? value : computeValue();
}

private synchronized T computeValue() {
    final Supplier<? extends T> lazyValue = supplier;
    if (lazyValue != null) {
        value = lazyValue.get();
        supplier = null;
    }
    return value;
}
```

@paplorinc Could you please change the PR accordingly if the benchmarks still look good?

Thanks all! This will be an awesome Lazy implementation :)

paplorinc: Thank you for your expertise @shipilev!

As my comments also stated, I needed help in finding out why the obviously wrong code (i.e. no volatile) passed all the tests, before merging this PR (which wasn't obvious to me, I though the volatile read after the non-volatile one will synchronize both. My mistake.).

Will try to come up with a test that fails for the unsafe version you pointed out, so that I can better understand it - and we need to base our decisions on failing and passing tests anyway!

Thanks!

danieldietrich: (Aksing myself if there is a chance that `computeValue()` is inlined by JIT. What happens then with the short `get()`? Will it be still inlined by Git?)

paplorinc: I will check those, once I have a failing test for all values that shouldn't work in theory, and pass for those, that should.

danieldietrich: Ok

paplorinc: @shipilev, I couldn't find the latest [`jcstress`](http://openjdk.java.net/projects/code-tools/jcstress) on a Maven repo (did find an older one for a [Gradle plugin](https://github.com/jerzykrlk/jcstress-gradle-plugin) though).
I built the project, but it needs [`Java 9`](http://hg.openjdk.java.net/code-tools/jcstress/file/default/pom.xml#l152), and even then I get a compiler warning because  [`JCStressTestProcessor`](http://hg.openjdk.java.net/code-tools/jcstress/file/default/jcstress-core/src/main/java/org/openjdk/jcstress/infra/processors/JCStressTestProcessor.java#l52) has a `@SupportedSourceVersion(SourceVersion.RELEASE_6)` (Javaslang needs `Java 8` and is built with `-werror` via `Maven`).

Do you recommend trying [thread-weaver](https://github.com/google/thread-weaver) instead - though only 18 commits worry me a bit?

ps. The code you commented on was experimental, i.e. trying to make the tests fail - without success, though. The original was (and is again):
```java
transient volatile Supplier<? extends T> supplier;
T value;

T get() {
    if ((value == null) && (supplier != null)) {
        synchronized (this) {
            final Supplier<? extends T> s = supplier;
            if (s != null) {
                value = s.get();
                supplier = null;
            }
        }
    }
    return value;
}
```
i.e. checking the non-volatile first seems to speed up the [benchmarks](https://github.com/javaslang/javaslang/pull/1577/commits/f956463ca578429ef03a999ea194ecc0499f0707#diff-179d7365a1b162ee88ba5aac7bd0c653R46).

danieldietrich: @paplorinc when I understood @shipilev right, the non-volatile read of `value`

```java
if (value == null && ...)
```

does not guarantee that (in the case of `value != null`) the second read does evaluate the same value:

```java
return value;
```

In other words your version is not deterministic and may produce unexpected results.

Am I right right?

shipilev: > In other words your version is not deterministic and may produce unexpected results.

Worse than that: reading a naked `value` is reading via the race. There is no guarantees that something that you have written **into the value fields themselves** is visible to the readers of value reference. Should read supplier first to get "acquire" part.

danieldietrich: @shipilev In our case that might be ok. We first read `value` because it is a fast read. We can have three cases then.

1. value was not written by any thread yet
2. read reflects the written value, i.e. it is visible to the thread.
3. written value is not visible to the current thread

In case 1. `value` is null and will be computed. After that it is visible to the current thread and will be returned by `get()`.

In the second case the write to `value` is visible. Then we are finished and can safely return it.

In the third case the write to `value` is not visible, it seems to be still null. Then `supplier` is read and it turns out that we already computed the value. Because of the volatile read of `supplier` our `value` is now ✨automagically✨ visible to the current thread and it can be safely returned by `get()`.

In other words we use `value == null` as a shortcut and we can detect not-visible writes by having the volatile 'afterburner' read of `supplier`.

Benchmarks say that is faster in most cases.

shipilev: > In the second case the write to value is visible. Then we are finished and can safely return it.

**WRONG**. Let me try to explain this again. You cannot safely return something you have read off the race, you have just constructed the unsafe publication. Even if you have observed the reference to `value`, you are not guaranteed to observe the transitive reads correctly, you have to piggyback on happens-before for this. And to piggy-back on all paths, you *have* to read `supplier` on all paths.

In other words, with "value == null" shortcut, you have just exposed your users to this (I am speculating about the syntax here, sorry, in hurry):

```
  class MyObject {
    int x; 
    MyObject(int x) { this.x = x; }
  }

  static final Lazy<MyObject> lazy = Lazy.with(() -> new MyObject(42));
  ...
  MyObject o = lazy.get(); // discover some other thread had already initalized
  assert (o.x == 42); // intermittently FAILS, because MyObject within Lazy was racily published
```

danieldietrich: @shipilev thanks for clarifying it again and for being so patient with us 🙏🏼
The missing part we did not consider was the one with the transitive dependencies.

@paplorinc please follow then [this solution](https://github.com/javaslang/javaslang/pull/1577#discussion_r80014687). Further improvements can be made later if necessary. I want this get merged to add it to the upcoming 2.0.4.

paplorinc: @danieldietrich, pushed a new version - though I'm not sure about speed or correctness anymore :p
FYI, `get` is small enough to be inlinable now , i.e. `javaslang.Lazy::get (19 bytes) inline (hot)`

javaslang-benchmark/src/test/java/javaslang/control/LazyBenchmark.java
```diff
@@ -0,0 +1,83 @@
...
+
+        @Setup
+        @SuppressWarnings({ "unchecked", "rawtypes" })
+        public void setup() {
+            EAGERS = Iterator.range(0, SIZE).toJavaArray(Integer.class);
+            INITED_LAZIES = Iterator.of(EAGERS).map(i -> {
+                final Lazy<Integer> lazy = Lazy.of(() -> i);
+                lazy.get();
+                return lazy;
+            }).toJavaList().toArray(new Lazy[0]);
```
paplorinc: @danieldietrich, this is ugly :/

danieldietrich: I agree. It will be a little less ugly if we add `T[] Value.toJavaArray(T[])` - the only safe solution :-/
See #1582

PR #1578, opened 2016-09-20
Generalized `Value#toJavaArray` to accept classes with generics

paplorinc:
Separated from https://github.com/javaslang/javaslang/pull/1577

### DIFF
--- a/javaslang/src/main/java/javaslang/Value.java
+++ b/javaslang/src/main/java/javaslang/Value.java
@@ -516,7 +516,7 @@ public interface Value<T> extends Iterable<T> {
      */
     @SuppressWarnings("unchecked")
     @GwtIncompatible("reflection is not supported")
-    default T[] toJavaArray(Class<T> componentType) {
+    default T[] toJavaArray(Class<? super T> componentType) {
         Objects.requireNonNull(componentType, "componentType is null");
         final java.util.List<T> list = toJavaList();
         return list.toArray((T[]) java.lang.reflect.Array.newInstance(componentType, list.size()));
@@ -1063,8 +1063,8 @@ interface ValueModule {
                 container.add(value.get());
             } else {
                 final int size = value instanceof Traversable && ((Traversable) value).isTraversableAgain()
-                        ? ((Traversable<V>) value).size()
-                        : 0;
+                                 ? ((Traversable<V>) value).size()
+                                 : 0;
                 container = containerSupplier.apply(size);
                 value.forEach(container::add);
             }
--- a/javaslang/src/test/java/javaslang/AbstractValueTest.java
+++ b/javaslang/src/test/java/javaslang/AbstractValueTest.java
@@ -28,12 +28,14 @@ import org.assertj.core.api.*;
 import org.junit.Test;
 
 import java.util.*;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.function.Function;
 
 import static java.lang.Integer.bitCount;
 import static javaslang.Serializables.deserialize;
 import static javaslang.Serializables.serialize;
+import static javaslang.control.Option.some;
 
 public abstract class AbstractValueTest {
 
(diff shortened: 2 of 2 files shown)


javaslang/src/main/java/javaslang/Value.java
```diff
@@ -516,7 +516,7 @@ default CharSeq toCharSeq() {
      */
     @SuppressWarnings("unchecked")
     @GwtIncompatible("reflection is not supported")
-    default T[] toJavaArray(Class<T> componentType) {
+    default T[] toJavaArray(Class<? super T> componentType) {
```
jbgi: note that with this change, the method is now unsafe:
```java
  Value<Option<Integer>> value = of(some(1), some(2));
  Option<Integer>[] options = value.toJavaArray(Object.class); // ClassCastException
```

danieldietrich: @jbgi Thanks - that is the missing counter-example I wasn't able to find. Then we have to revert this merge. /cc @paplorinc

paplorinc: Yes, but how DO we make it work for e.g. `Lazy<Integer>` otherwise?
I think we should only revert, if we can find a better solution...

jbgi: I think the only solution would be to provide an array wrapper class, alla https://github.com/functionaljava/functionaljava/blob/master/core/src/main/java/fj/data/Array.java

danieldietrich: > I think we should only revert, if we can find a better solution...

Definitely not. We will not introduce unsafe/breaking code. A better solution would be to deprectate (and remove in future) the whole method if it makes no sense.

danieldietrich: @jbgi could you please provide a little example (use-site perspective)?

danieldietrich: @jbgi [found it](https://github.com/functionaljava/functionaljava/blob/master/core/src/test/java/fj/data/ArrayTest.java#L14-L19) :)

```java
    @Test
    public void array_is_safe() {
        List<Integer> list = List.range(1, 2);

        assertThat(list.toArray().array(Integer[].class), instanceOf(Integer[].class));
    }
```

paplorinc: This (and the previous) also break for primitives, e.g.
```java
final Value<Integer> value = of(1, 2, 3);
final Integer[] ints = value.toJavaArray(int.class);
```

I guess this is why `Java` provides an actual array instead...

danieldietrich: @jbgi It does also not help for generic classes :-/

```java
final Value<Option<Integer>> value = of(Option.of(1), Option.of(2), Option.of(3));

// Compile error: Cannot select from parameterized type
final Option<Integer>[] ints = value.toJavaArray(Option<Integer>[].class);
```

@paplorinc It seems that `Class<? super T>` is the only solution (but unsafe).

I'm tempted to take it. What do you think? +1 or -1 :)

paplorinc: I think some unsafety is inherent. Arrays are really messy.
Since I find it unlikely to occur in real life (the primitive one should occur a lot more often), and since it will fail every time, signaling a real error (though not compile-time, but run-time), I think it's low-risk, and useful.

paplorinc: ... or we can deprecate it and do it Java's way, i.e. providing a preinitialized array as param.

danieldietrich: As @paplorinc said, these are our options:

```java
// unsafe
T[] toJavaArray(Class<? super T> componentType) // one size fits all

vs

// safe
T[] toJavaArray(Class<T> componentType) // variant for non-generic types
T[] toJavaArray(T[] array)
```

I think the first, unsafe variant is no viable solution.

Example:

```java
// before
Value<CharSequence> v = Option.of("");
CharSequence[] s = v.toJavaArray(CharSequence.class);

// after
Value<String> v = Option.of("");
String[] s = v.toJavaArray(CharSequence.class); // will eventually throw at runtime
```

The method `toJavaArray(Class<T>)` is still useful for non-generic types,
we do not need to deprecate it.

PR #1589, opened 2016-09-27
Added Value.toTree conversion methods

danieldietrich:
I prepare the next releases 2.0.4 and 2.1.0. To get sure that we stay backward compatible I write a tool that parses the output of the maven-bundle-plugin.

The 'parser' will be based on our collection API. First we will filter the console output, then convert the relevant lines to a tree (using this conversion methods) and then semantically enrich the parse tree and interpret the resulting tree of information.

We can define our own rules then what is backward compatible and what is not. E.g. these are backward compatible:
- widening of generic type arguments
- changing collection indexing from long to int (because java will convert ints to long at the call site)
- ...

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -195,6 +195,16 @@ release v2.1.0 - - - - - - + - - - X - - - + 2.1.1-SNAPSHOT
                           ...             ...
 ```
 
+### Backward compatibility
+
+We check backward compatibility by running the baseline goal of the maven-bundle-plugin:
+
+**Example:** Comparing API changes of current branch with version 2.0.0
+
+```bash
+mvn org.apache.felix:maven-bundle-plugin:baseline -DcomparisonVersion=2.0.0 -DskipTests
+```
+
 ### Major release
 
 #### Performing a release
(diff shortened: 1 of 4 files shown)


danieldietrich:
Note: This is a working change but I will probably `push --force` some optimizations of the algorithm. It is currently hacked from the brain to the keyboard...

danieldietrich:
Will fix the tests tomorrow - seems to be the isSingkeValued check and maybe unpredictable order of sets... Need to adjust the asserts.
Lazy me, checked it only for Lists. However, it does work like a charm - in iteration order...

danieldietrich:
There are still open issues regarding the conversion method:

* we need to go multiple levels down (but only one up) at a time
* we need to use a defaultValue instead of null
* Tree needs Tree.ofAll(Stream s, Comparator c, T defaultValue) (maybe also a default value Supplier variant?)

ronanM:
`ofAll(Iterable<? extends T> iterable, Comparator<? super T> comparator)` works for a **'level' comparator**.

So it can be a bit confusing. Maybe it's better to rename it with this 'level' notion ?

danieldietrich:
@ronanM Yes, will do it!

PR #1561, opened 2016-09-08
Changed memory usage display for benchmarks

paplorinc:
```java
for `32` elements
  Java mutable @ ArrayList                            →   504 bytes - 504 bytes
  Functional Java persistent @ Seq                    → 3,064 bytes -      3 KB
  PCollections persistent @ TreePVector               → 1,744 bytes -    1.7 KB
  Eclipse Collections persistent @ ImmutableArrayList →   496 bytes - 496 bytes
  Clojure persistent @ PersistentVector               →   704 bytes - 704 bytes
  Scala persistent @ Vector                           →   536 bytes - 536 bytes
  Javaslang persistent @ Vector                       →   528 bytes - 528 bytes

for `1024` elements
  Java mutable @ ArrayList                            →  19,064 bytes -  18.6 KB
  Functional Java persistent @ Seq                    → 104,760 bytes - 102.3 KB
  PCollections persistent @ TreePVector               →  56,016 bytes -  54.7 KB
  Eclipse Collections persistent @ ImmutableArrayList →  19,056 bytes -  18.6 KB
  Clojure persistent @ PersistentVector               →  20,504 bytes -    20 KB
  Scala persistent @ Vector                           →  19,736 bytes -  19.3 KB
  Javaslang persistent @ Vector                       →  19,728 bytes -  19.3 KB

for `32768` elements
  Java mutable @ ArrayList                            →   653,672 bytes - 638.4 KB
  Functional Java persistent @ Seq                    → 3,408,624 bytes -   3.3 MB
  PCollections persistent @ TreePVector               → 1,833,408 bytes -   1.7 MB
  Eclipse Collections persistent @ ImmutableArrayList →   653,664 bytes - 638.3 KB
  Clojure persistent @ PersistentVector               →   700,168 bytes - 683.8 KB
  Scala persistent @ Vector                           →   674,824 bytes -   659 KB
  Javaslang persistent @ Vector                       →   674,816 bytes -   659 KB
```

```java
for `10` elements
  Java mutable @ String                   →    64 bytes - 64 bytes
  Functional Java persistent @ LazyString → 2,480 bytes -   2.4 KB
  Javaslang persistent @ CharSeq          →    80 bytes - 80 bytes

for `100` elements
  Java mutable @ String                   →    240 bytes - 240 bytes
  Functional Java persistent @ LazyString → 23,536 bytes -     23 KB
  Javaslang persistent @ CharSeq          →    256 bytes - 256 bytes

for `1000` elements
[...]

PR #1590, opened 2016-09-28
Changed benchmark memory footprint computation to JOL

paplorinc:
All previous measurements were correctly done via `com.carrotsearch#java-sizeof` also, but JOL is more mainstream and part of the OpenJDK.

Note: Unlike `java-sizeof`, `JOL` gives a runtime warning (even with `jol.tryWithSudo=true` on Linux):
```
# WARNING: Unable to attach Serviceability Agent. You can try again with escalated privileges. Two options: a) use -Djol.tryWithSudo=true to try with sudo; b) echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope
```

Note2: you can find the memory measurements in: https://github.com/javaslang/javaslang/pull/1561

### DIFF
--- a/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
@@ -15,12 +15,9 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
-import static com.carrotsearch.sizeof.RamUsageEstimator.sizeOf;
 import static java.util.Arrays.asList;
 import static javaslang.JmhRunner.Includes.*;
 import static javaslang.JmhRunner.*;
-import static javaslang.collection.Arrays.copyRange;
-import static javaslang.collection.BitMappedTrie.branchingFactor;
 import static javaslang.collection.Collections.areEqual;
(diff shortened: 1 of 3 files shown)


javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -778,7 +775,6 @@ public void slang_persistent(Blackhole bh) {
             for (int i = 1; !values.isEmpty(); i++) {
                 values = values.slice(1, values.size());
                 values = values.slice(0, values.size() - 1);
-                assert sizeOf(values.trie) < (branchingFactor() * sizeOf(copyRange(ELEMENTS, i, javaMutable.size() - i))); /* detect memory leak */
```
paplorinc: slowed down tests too much (in debug mode only, didn't affect the actual benchmarks)

danieldietrich:
Thank you! It is a better choice to depend on an OpenJDK standard lib.
(I wonder if there is an impact of the warning. The tests still run. Do they have an appropriate accuracy regarding memory measurement?)

paplorinc:
yeah, but other functionality might not work from `JOL`, dunno yet.

danieldietrich:
Ok 👌

PR #1591, opened 2016-09-29
Future.of and Future.run waiting a duration of time before running

danieldietrich:
TODO:

* [ ] use nanoTime() instead of currentTimeMillis() or date-time API (see [this comment](https://github.com/vavr-io/vavr/pull/1591#discussion_r112820071))

___

JavaScript has

``` javascript
// waits 123 millis before calling the function
setTimeout(() => ..., 123);
```

This PR provides `Future.run` and `Future.of` methods that execute the given Supplier/Runnable after waiting a duration.


danieldietrich:
Ready for review!

javaslang/src/main/java/javaslang/concurrent/Future.java
```diff
@@ -309,10 +309,45 @@
...
+     * The thread supplied by the {@code DEFAULT_EXECUTOR_SERVICE} will be blocked at least {@code duration} milliseconds.
+     * If the {@code duration} &lt;= 0, the {@code computation} will be executed immediately.
+     *
+     * @param computation A computation.
+     * @param duration    The time in milliseconds to wait before executing the given {@code computation}.
+     * @param <T>         Type of the computation result.
+     * @return A new Future instance.
+     * @throws NullPointerException if computation is null.
+     */
+    static <T> Future<T> of(CheckedSupplier<? extends T> computation, long duration) {
```
ruslansennov: maybe `delay` is better than `duration`

danieldietrich: Yes, right! (I didn't like the name 'timeout' of JavaScript because it intends to cancel the computation after a specific amount of time.)

Delay is much better than duration.

danieldietrich:
Oh, javadoc formatting scrambled - will push --force a better version...

javaslang/src/main/java/javaslang/concurrent/Future.java
```diff
@@ -309,10 +309,45 @@
...
+     * @return A new Future instance.
+     * @throws NullPointerException if one of executorService of computation is null.
+     */
+    static <T> Future<T> of(ExecutorService executorService, CheckedSupplier<? extends T> computation, long delay) {
         Objects.requireNonNull(executorService, "executorService is null");
         Objects.requireNonNull(computation, "computation is null");
         final FutureImpl<T> future = new FutureImpl<>(executorService);
-        future.run(computation);
+        final CheckedSupplier<? extends T> task = (delay <= 0) ? computation : () -> {
+            Thread.sleep(delay);
```
hepin1989: sleep?

danieldietrich: @hepin1989 Thanks! I'm not used to HashWheelTimers. Our Future impl has no 'execution queue'. How could an impl look like?

Project Reactor's HashWheelTimer also supports Sleep:

https://projectreactor.io/old/api/reactor/fn/timer/HashWheelTimer.SleepWait.html

As an alternative there are Spin strategies that loop. These are costly (a whole core in a multicore environment). I think spinning a loop is no viable solution.

The sleep we use takes place on a different thread (if the executor-service isn't trivial). So it doesn't hurt much.

---

But I think there is something we didn't realize yet: We should take the time when the run method is entered. Because it may happen, that the wait is already over because the ExecutorService needed some time to spawn a new thread. Even if the time isn't over, we only need to wait the remaining delta since the run method was entered. This will be much more precise.

Thanks for the hint :)

hepin1989: you could take a look at monix Task.timeout and netty EventLoop,this kind of delay would cause issue ,javaslang is great,we are using it at taobao😃

danieldietrich:
Latest commit takes ExecutorService latency into account (including tests)

javaslang/src/main/java/javaslang/concurrent/Future.java
danieldietrich: @hepin1989 which features of Javaslang do you use mainly?

Are you interested in a guest blog-post about how you use Javaslang? I.e. on the [Javaslang Blog](http://blog.javaslang.io)

danieldietrich: **Update:** Netty also uses Java's ScheduledExecutorService under the hood: https://netty.io/4.0/api/io/netty/util/concurrent/EventExecutorGroup.html#schedule(java.util.concurrent.Callable,%20long,%20java.util.concurrent.TimeUnit)

That is the way to go.

hepin1989: @danieldietrich not like that,netty implement it via a dedicated delayed task queue
https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/concurrent/AbstractScheduledEventExecutor.java#L46

danieldietrich: Yes, but Netty does not implement the timeout/delay functionality at the core using delayed task queues.

It looks like to me that Netty uses a ScheduledExecutorService to produce a ScheduledFuture (like I stated above).

These ScheduledFutures are put on a PriorityQueue to ensure that the next delay that expires will be taken in to account. (The underlying java.util.PriorityQueue [does take the natural order of elements into account](https://docs.oracle.com/javase/7/docs/api/java/util/PriorityQueue.html). The java.util.concurrent.ScheduledFuture does extend `Comparable<Delayed>`, by which is ensured that the smaller delays are ordered first.)

**Question:** How uses Netty the PriorityQueue? What is it needed for? The ScheduledFuture will run after a specific delay, regardless if it is put on a queue or not.

Need to take a look at the source code of Netty. But my 'feeling' says, for now we don't need to put our futures on a PriorityQueue!

However, it could be useful to let Javaslang's Future implement `Comparable<java.uti.concurrent.Delayed>`, where the default delay is 0 (for non-delayed Futures).

javaslang/src/main/java/javaslang/concurrent/Future.java
```diff
@@ -309,10 +309,41 @@
...
+     * The thread supplied by the {@code DEFAULT_EXECUTOR_SERVICE} will be blocked at least {@code delay} milliseconds.
+     * If the {@code delay} &lt;= 0, the {@code computation} will be executed immediately.
+     *
+     * @param computation A computation.
+     * @param delay       The time in milliseconds to wait before executing the given {@code computation}.
+     * @param <T>         Type of the computation result.
+     * @return A new Future instance.
+     * @throws NullPointerException if computation is null.
+     */
+    static <T> Future<T> of(CheckedSupplier<? extends T> computation, long delay) {
```
paplorinc: please express the unit in the name also, people shouldn't have to look in the documentation to see whether `of (..., 1)` sleeps for `1 milllisecond` or `1 second`.
I recommend something like:
```java
static <T> Future<T> of(CheckedSupplier<? extends T> computation, long delayMillis)
```
instead, to mimic e.g.
```java
public static native long currentTimeMillis();
```

javaslang/src/main/java/javaslang/concurrent/Future.java
```diff
@@ -364,24 +395,53 @@
      * @throws NullPointerException if unit is null.
      */
     static Future<Void> run(CheckedRunnable unit) {
```
paplorinc: not sure why a runnable is called `unit` here

javaslang/src/main/java/javaslang/concurrent/Future.java
```diff
@@ -364,24 +395,53 @@
...
-        return run(DEFAULT_EXECUTOR_SERVICE, unit);
+        return run(DEFAULT_EXECUTOR_SERVICE, unit, 0);
     }
 
     /**
-     * Starts an asynchronous computation, backed by the given {@link ExecutorService}.
+     * Runs an asynchronous computation, backed by the given {@link ExecutorService}.
```
paplorinc: :+1:

javaslang/src/main/java/javaslang/concurrent/FutureImpl.java
```diff
@@ -193,7 +194,13 @@ void run(CheckedSupplier<? extends T> computation) {
                 // if the ExecutorService runs the computation
                 // - in a different thread, the lock ensures that the job is assigned before the computation completes
                 // - in the current thread, the job is already completed and the `job` variable remains null
-                final java.util.concurrent.Future<?> tmpJob = executorService.submit(() -> complete(Try.of(computation)));
+                final java.util.concurrent.Future<?> tmpJob = executorService.submit(() -> complete(Try.of(() -> {
+                    final long delta  = delay - System.currentTimeMillis() + start;
```
paplorinc: interesting ... 
Consider using Java 8 date-time-api's instead, like:
```java
Instant now = Instant.now();
...
Duration.between(now, Instant.now()).minusMillis(...)
```

javaslang/src/main/java/javaslang/concurrent/FutureImpl.java
```diff
@@ -193,7 +194,13 @@ void run(CheckedSupplier<? extends T> computation) {
...
+                    if (delta > 0) {
+                        Thread.sleep(delta);
```
paplorinc: `Math.max(0, delta)` ?

javaslang/src/test/java/javaslang/concurrent/ExecutorServices.java
```diff
@@ -37,47 +41,32 @@ static ExecutorService rejectingExecutorService() {
...
-
-    private static final class RejectingExecutorService extends AbstractExecutorService {
 
         @Override
         public <T> java.util.concurrent.Future<T> submit(Callable<T> task) {
-            throw new RejectedExecutionException();
+            try {
+                Thread.sleep(delay);
+                return new ImmediatelyDoneFuture<>(task.call());
+            } catch (Exception x) {
```
paplorinc: minor: `Exception e`?

javaslang/src/test/java/javaslang/concurrent/FutureTest.java
```diff
@@ -232,6 +231,42 @@ public void shouldCompleteWithFailureWhenExecutorServiceThrowsRejectedExecutionE
...
+        final long[] time = new long[] { System.currentTimeMillis() };
+        final Future<Void> future = Future.run(() -> time[0] = System.currentTimeMillis() - time[0], delay);
+        waitUntil(future::isCompleted);
+        assertThat(time[0]).isGreaterThanOrEqualTo(delay);
+    }
+
+    @Test
+    public void shouldWaitGivenDelayBeforeRunningUsingSpecificExecutorService() {
+        final long delay = 250;
+        final long[] time = new long[] { System.currentTimeMillis() };
```
paplorinc: you could use the built-in array initializer here (and in other places in the test):
```java
final long[] time = { System.currentTimeMillis() };
```

paplorinc:
@danieldietrich?

danieldietrich:
@paplorinc by accident during a late-night coding session. thank you!

javaslang/src/main/java/javaslang/concurrent/FutureImpl.java
hepin1989: shouldn't this be nano?

hepin1989:
https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/concurrent/EventExecutorGroup.java#L105

javaslang/src/main/java/javaslang/concurrent/FutureImpl.java
tedyoung: For elapsed time measurements, should use System.nanoTime() and not the date-time APIs nor currentTimeMillis(). See https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime-- for details -- mainly that nanoTime is guaranteed to be monotonic increasing (on most systems), whereas the date time APIs (which are based on currentTimeMillis) can move around during time changes and adjustments.

danieldietrich: @tedyoung thank you, that is helpful. I will use nanoTime().

javaslang/src/main/java/javaslang/concurrent/Future.java
chb0github: Actually, since it a unit of time, it should be something more like (or have an overload):
`Future<T> of(CheckedSupplier<? extends T> computation, long numUnits, TimeUnit unit)`

I had also though of Attaching a future submission to an event - in JS you can listen to events on arrays and such. So, having something like: `List.of(1).onRemove(thisList -> Future.of(...)`

Just some ideas.

danieldietrich:
Maybe we should use ScheduledExecutorService resp. ScheduledThreadPoolExecutor in order to decorate tasks...

http://download.java.net/java/jdk9/docs/api/java/util/concurrent/ScheduledThreadPoolExecutor.html

PR #1598, opened 2016-10-02
Fixed exception messages in Array

ashrko619:


danieldietrich:
Great thank you! Will include it in release 2.0.4 which is coming this weekend.

Issue #1584, opened 2016-09-25
Be a little more Lazy


danieldietrich:
We have methods that return `Xyz<Seq<...>, ...>`.

These should return a `Stream` instead of other `Seq` implementations in order to avoid unnecessay computations.

ashrko619:
I see some more opportunities where a ```Stream``` can be used instead of other implementations.
Ex : https://github.com/javaslang/javaslang/blob/master/javaslang/src/main/java/javaslang/collection/AbstractsQueue.java#L62
Can I refactor them here or should they go in another PR ? 😄

danieldietrich:
@ashrko619 please let's keep the scope of this PR focused on the methods that return wrapped Seq types.

> I see some more opportunities where a Stream can be used instead of other implementations.

We should discuss these by creating issues first. I would not 'refactor' the example you mentioned. A queue is backed on two lists: rear and front.

It makes most sense to switch to a Stream return type when having costly operations that can be deferred by making them intermediate/lazy. In your example of the Queue (link above) the List is only used to iterate elements. Also making the whole Queue lazy makes not much sense in my opinion because the elements are already known (they are enqueued and dequeued). Lazy evaluation makes most sense when the elements are not yet known.

danieldietrich:
Fixed with #1605

Issue #1493, opened 2016-08-15
Breaking changes in 3.0.0


danieldietrich:
# Breaking Changes

This is just a List of ideas. We will create separate PRs (and maybe also issues).

## Remove unnecessary ballast

* [x] Check if we can remove Either's Left/RightProjection now that Either is right-biased. See also Elm's Result type (similar to Either) - **yes, we can - issue created: #2071**
* [x] discontinue javaslang-pure~~
* [x] remove javaslang.Kind1, Kind2
* [x] remove Option.nothing()
* [x] consider making λ.Memoized package-private/internal. See #1905 
* [x] ~~remove Promise (see #1529) - (update: it will most probably stay for we will have new Future API for promises)~~
* [x] ~~rethink or remove run() for Match (see #1216, #1218)~~ - **I don't want to put too much additional effort into the Match API because Java will have native pattern matching in near future.**
* [x] Case API is ambiguous. We should allow only Function (not Supplier or value T) in order to calculate result
* [x] Value.toOption() and Value.getOption() are redundant. Remove one? (However, getOption needs to return this if `this instanceof Option`)
* [x] ~~Widen Validation's `List<E>` to `List<? extends E>` in many cases.~~ - **moved to #1565**
* [x] We should consider to change `Validation<List<E>, T>` to `Validation<Seq<E>, T>` and internally use `Vector` instead of `List`.
* [x] Make AbstractIterator package-private. ~~remove AbstractIterator and do hasNext()-checks manually in next(). (see [this comment](https://github.com/javaslang/javaslang/commit/d61f6a9cb6b2c2ae7d492c4f3a2d6b01d214724c#commitcomment-21225420)). Problem Iterator is an interface and we need to override toString(). Possible solution: make Iterator an abstract class. **Better solution:** Keep it as interface and do not provide toString() for Iterator - instead use hasNext() to check if it is empty.~~

## Other breaking changes

* [x] provide first char upper case static factory methods (see [#1431](https://github.com/javaslang/javaslang/issues/1431#issuecomment-244603613))
* [x] indexing collections with `int` instead of `long`
* [x] ~~consider renaming CheckedFunction0..8 to ThrowingFunction0..8 (one more think to not think about from the user perspective - which one throws, the checked or the unchecked function?)~~
* [x] ~~bring (creational method names of, ofAll) and (conversion names fromXxx, toXxx) into line, e.g.~~
[...]

PR #1605, opened 2016-10-05
Refactored to use Stream instead of other Seq impls to be lazy

ashrko619:
PR for #1584


javaslang/src/main/java/javaslang/control/Option.java
```diff
@@ -54,14 +59,14 @@
...
+        Stream<T> stream = Stream.empty();
         for (Option<? extends T> value : values) {
             if (value.isEmpty()) {
                 return Option.none();
             }
-            list = list.prepend(value.get());
+            stream = stream.append(value.get());
         }
-        return Option.some(list.reverse());
+        return Option.some(stream);
```
paplorinc: why is this an `Option` exactly, when it could simply return an empty stream instead?
@danieldietrich?

javaslang/src/test/java/javaslang/LazyTest.java
```diff
@@ -49,16 +49,16 @@ public void shouldMemoizeValues() {
 
     @Test
     public void shouldSequenceEmpty() {
-        final List<Lazy<Integer>> testee = List.empty();
+        final Seq<Lazy<Integer>> testee = Stream.empty();
```
paplorinc: I think this should remain a `List`

javaslang/src/test/java/javaslang/LazyTest.java
```diff
@@ -49,16 +49,16 @@ public void shouldMemoizeValues() {
...
         final Lazy<Seq<Integer>> sequence = Lazy.sequence(testee);
-        assertThat(sequence.get()).isEqualTo(List.empty());
+        assertThat(sequence.get()).isEqualTo(Stream.empty());
     }
 
     @Test
     public void shouldSequenceNonEmptyLazy() {
-        final List<Lazy<Integer>> testee = List.of(1, 2, 3).map(i -> Lazy.of(() -> i));
+        final Stream<Lazy<Integer>> testee = Stream.of(1, 2, 3).map(i -> Lazy.of(() -> i));
```
paplorinc: same

javaslang/src/main/java/javaslang/collection/Map.java
```diff
@@ -521,7 +519,7 @@ default int length() {
 
     default <U> Seq<U> traverse(BiFunction<K, V, ? extends U> mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
-        return foldLeft(List.empty(), (acc, entry) -> acc.append(mapper.apply(entry._1, entry._2)));
+        return foldLeft(Stream.empty(), (acc, entry) -> acc.append(mapper.apply(entry._1, entry._2)));
```
danieldietrich: `foldLeft` is not lazy, please implement it like this:

```java
    default <U> Seq<U> traverse(BiFunction<K, V, ? extends U> mapper) {
        return iterator().<U> map(entry -> mapper.apply(entry._1, entry._2)).toStream();
    }
```

javaslang/src/main/java/javaslang/collection/Multimap.java
```diff
@@ -297,8 +297,7 @@ default int length() {
     @Override
     default <U> Seq<U> map(Function<? super Tuple2<K, V>, ? extends U> mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
-        // don't remove cast, doesn't compile in Eclipse without it
-        return (Seq<U>) iterator().map(mapper).toStream();
+        return Stream.ofAll(iterator()).map(mapper);
```
danieldietrich: Nope :) The original expression `return (Seq<U>) iterator().map(mapper).toStream();` saves `n` Stream.Cons instances compared to the suggested solution.

javaslang/src/main/java/javaslang/collection/Multimap.java
```diff
@@ -551,8 +550,7 @@ default int length() {
     @Override
     default <U> Seq<U> flatMap(Function<? super Tuple2<K, V>, ? extends Iterable<? extends U>> mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
-        // don't remove cast, doesn't compile in Eclipse without it
-        return (Seq<U>) iterator().flatMap(mapper).toStream();
+        return Stream.ofAll(iterator()).flatMap(mapper);
```
danieldietrich: Also here: `return (Seq<U>) iterator().flatMap(mapper).toStream();` is already the best we can do!

javaslang/src/main/java/javaslang/collection/TreeMap.java
```diff
@@ -753,7 +753,7 @@ public int size() {
 
     @Override
     public Seq<V> values() {
-        return iterator().map(Tuple2::_2).toStream();
+        return  Stream.ofAll(iterator()).map(Tuple2::_2);
```
danieldietrich: Dito: stay with the original impl

javaslang/src/main/java/javaslang/control/Option.java
danieldietrich: @ashrko619 The sequence operation isn't lazy by nature. We have to evaluate all contained elements to decide whether the result is empty or not. Therefore we should choose an eager Seq implementation. Namely either we stay with `List` (because of good memory footprint) or choose `Vector` (because we safe the reverse operation).

I tend to use `Vector`.

---

@paplorinc It is the 'algebraic' definition of `sequence`:

```java
sequence(Seq[M[T]]) =::= M[Seq[T]]
```

The `sequence` operation on a Value (resp. Monad) reflects the properties of a Functor (first failing operation =&gt; empty result) in contrast to an Applicative (subsequent results are collected).

javaslang/src/test/java/javaslang/LazyTest.java
danieldietrich: Yes, please remain List here.

danieldietrich:
@ashrko619 Hi, could you please give me a status update? I'm currently consolidating all open issues/PRs. Thank you :)

ashrko619:
@danieldietrich I have been a bit caught up at work, will get the PR in by Wednesday 🙈

danieldietrich:
✨👍🏽😄👍🏽✨

javaslang/src/test/java/javaslang/LazyTest.java
paplorinc: I see, consider using `isEmpty` then :) (or leave `Stream.empty()`)

danieldietrich: @ashrko619 looks like it has to be `Stream.empty()`. Do you have a pointer to the test (File/Line number)?

javaslang/src/main/java/javaslang/Lazy.java
```diff
@@ -101,7 +101,7 @@ private Lazy(Supplier<? extends T> supplier) {
      */
     public static <T> Lazy<Seq<T>> sequence(Iterable<? extends Lazy<? extends T>> values) {
         Objects.requireNonNull(values, "values is null");
-        return Lazy.of(() -> List.ofAll(values).map(Lazy::get));
+        return Lazy.of(() -> Stream.ofAll(values).map(Lazy::get));
```
danieldietrich: Please use Vector.ofAll() instead of Stream.ofAll(). See an explanation below (at the Option.sequence() change).

javaslang/src/main/java/javaslang/control/Option.java
```diff
@@ -54,14 +60,14 @@
...
         for (Option<? extends T> value : values) {
             if (value.isEmpty()) {
                 return Option.none();
             }
-            list = list.prepend(value.get());
+            vector = vector.append(value.get());
         }
-        return Option.some(list.reverse());
+        return Option.some(vector);
     }
```
danieldietrich: I've double-checked all Javaslang types. Future, Lazy, Option, Validation and Try have a `sequence()` method. We should take care that all sequence() methods return the same. That is not always possible (as described below) but the others should use Vector instead of List now.

* [x] Future.sequence() can stay as-is. It uses Stream in order to defer the blocking call of Future.get()
* [x] Option.sequence() uses Vector now instead of List
* [ ] Lazy.sequence() needs to use Vector (instead of List or Stream)
* [ ] Validation.sequence() needs to use Vector instead of List
* [ ] Try.sequence() needs to use Vector instead of List

Please use Vector for the sequence() methods of Lazy, Validation and Try.

ashrko619: @danieldietrich ```Validation.sequence() ``` returns ```Validation<List<E>, Seq<T>> ```
If we change the return type to ```Validation<Seq<E>, Seq<T>> ``` then we can change the ```errors``` list to also be a Vector.

danieldietrich: @ashrko619 we can't do that right now. I targeted 3.0.0 for such compatibility breaking changes (see also #1493)

danieldietrich:
@ashrko619 this one is also nearly finished, right? Please merge to resolve conflicts, I'm looking forward to pull this one. Thank you.

javaslang/src/main/java/javaslang/Lazy.java
```diff
@@ -6,8 +6,9 @@
 package javaslang;
 
 import javaslang.collection.Iterator;
-import javaslang.collection.List;
 import javaslang.collection.Seq;
+import javaslang.collection.Stream;
```
sputnikci: \[Checkstyle\] ERROR: Unused import \- javaslang\.collection\.Stream\.

javaslang/src/main/java/javaslang/control/Option.java
```diff
@@ -7,12 +7,18 @@
 
 import javaslang.Value;
 import javaslang.collection.Iterator;
-import javaslang.collection.List;
-import javaslang.collection.*;
+import javaslang.collection.Seq;
+import javaslang.collection.Stream;
```
sputnikci: \[Checkstyle\] ERROR: Unused import \- javaslang\.collection\.Stream\.

javaslang/src/main/java/javaslang/control/Option.java
```diff
@@ -54,14 +59,14 @@
...
+        Vector<T> vector = Vector.empty();
         for (Option<? extends T> value : values) {
             if (value.isEmpty()) {
                 return Option.none();
             }
-            list = list.prepend(value.get());
+            vector = vector.append(value.get());
         }
-        return Option.some(list.reverse());
+        return Option.some(vector);
```
paplorinc: :+1: for using Vector here

javaslang/src/main/java/javaslang/collection/Map.java
```diff
@@ -503,13 +503,13 @@ default int length() {
     @Override
     default <U> Seq<U> scanLeft(U zero, BiFunction<? super U, ? super Tuple2<K, V>, ? extends U> operation) {
         Objects.requireNonNull(operation, "operation is null");
-        return Collections.scanLeft(this, zero, operation, List.empty(), List::prepend, List::reverse);
+        return iterator().scanLeft(zero, operation).toStream();
```
paplorinc: please explain :/

javaslang/src/main/java/javaslang/collection/Map.java
```diff
@@ -481,13 +479,13 @@ default int length() {
...
         Objects.requireNonNull(operation, "operation is null");
-        return Collections.scanLeft(this, zero, operation, List.empty(), List::prepend, List::reverse);
+        return Collections.scanLeft(this, zero, operation, Stream.empty(), Stream::append, Function.identity());
     }
 
     @Override
     default <U> Seq<U> scanRight(U zero, BiFunction<? super Tuple2<K, V>, ? super U, ? extends U> operation) {
         Objects.requireNonNull(operation, "operation is null");
-        return Collections.scanRight(this, zero, operation, List.empty(), List::prepend, Function.identity());
+        return Collections.scanRight(this, zero, operation, Stream.empty(), Stream::prepend, Function.identity());
```
danieldietrich: Stream makes most sense when doing it lazily (for maybe infinite amount of elements). The Stream::prepend currently evaluates the whole Seq eagerly. Therefore please use Vector here instead of Stream.

The benefit of Vector: It is also great for random access (approx. O(1)), Stream only has O(n) for random access

javaslang/src/main/java/javaslang/collection/Map.java
danieldietrich: Because Map.scanRight() returns a Vector (and Vector is our default Seq), Map.scanLeft should also return Vector

danieldietrich: It looks fine, no more changes required then

danieldietrich:
Many thanks!

PR #1609, opened 2016-10-06
Curried CheckedFunction2 and greater now returns the chain of Function1 followed by single CheckedFunction1 call - exactly what partial application would do.

SimY4:
Curried CheckedFunction2 and greater now returns the chain of Function1 followed by single CheckedFunction1 call - exactly what partial application would do.

This change was lost in branch 3.0.0 and since that branch become obsolete I'm creating a new pull requrest for master.

(cherry picked from commit d00af15)

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -1347,17 +1347,10 @@ def generateMainClasses(): Unit = {
         }
         val callApply = s"$refApply($params)"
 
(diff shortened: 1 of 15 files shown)


danieldietrich:
@SimY4 Oh, thank you! - and sorry. I have overseen that. #1473 was the *only* (!) PR we merged into another branch yet. We did merge all other PR's that target 3.0.0 into master. Therefore I removed the 3.0.0 branch.

PR #1612, opened 2016-10-07
API aliases: Unchecked

### LINKED ISSUE #1606: Add API#Unchecked methods that return unchecked Funtions
We should add these to `javaslang.API` _after_  pulling in #1500.

**Goal:**

```java
// before
list.map(CheckedFunction1.of(elem -> ...).unchecked())

// after
list.map(Unchecked(elem -> ...))
```

We do this for Function0..8:

```java
<R> Function0<R> Unchecked(CheckedFunction0<R> f) {
    return f.unchecked();
}

<T, R> Function1<T, R> Unchecked(CheckedFunction1<T, R> f) {
    return f.unchecked();
}

...
```

I don't think we need to widen the generic bounds of the function argument here.

valery1707:
Fix #1606

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -363,6 +363,22 @@ def generateMainClasses(): Unit = {
             """
           })("\n\n")}
 
+          ${(0 to N).gen(i => {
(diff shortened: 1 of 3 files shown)


danieldietrich:
Great, thank you! :)

PR #1614, opened 2016-10-09
Future#await return self after complete

### LINKED ISSUE #1607: Change Future#await to return the parent instance
As described in https://github.com/javaslang/javaslang/pull/1500#pullrequestreview-2890134, by modifying `Future#await` to return a `Future`, we could chain it, like:
```java
Future<?> future = Future.of(() -> 1).await();
```

To be investigated if it makes sense in reality, or just in the tests.

valery1707:
Fix #1607

### DIFF
--- a/javaslang/src-gen/test/java/javaslang/APITest.java
+++ b/javaslang/src-gen/test/java/javaslang/APITest.java
@@ -228,48 +228,42 @@ public class APITest {
 
     @Test
     public void shouldFutureWithSupplierReturnNotNull() {
-        final Future<?> future = Future(() -> 1);
-        future.await();
+        final Future<?> future = Future(() -> 1).await();
         assertThat(future).isNotNull();
         assertThat(future.isSuccess()).isTrue();
     }
 
     @Test
     public void shouldFutureWithinExecutorWithSupplierReturnNotNull() {
-        final Future<?> future = Future(Executors.newSingleThreadExecutor(), () -> 1);
-        future.await();
+        final Future<?> future = Future(Executors.newSingleThreadExecutor(), () -> 1).await();
         assertThat(future).isNotNull();
         assertThat(future.isSuccess()).isTrue();
     }
 
     @Test
     public void shouldFutureWithValueReturnNotNull() {
-        final Future<?> future = Future(1);
-        future.await();
+        final Future<?> future = Future(1).await();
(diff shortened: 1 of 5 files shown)


javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -228,48 +228,42 @@ public void shouldLeftReturnNotNull() {
 
     @Test
     public void shouldFutureWithSupplierReturnNotNull() {
-        final Future<?> future = Future(() -> 1);
-        future.await();
+        final Future<?> future = Future(() -> 1).await();
```
paplorinc: Thanks @valery1707 :)

In case @danieldietrich thinks this is battle-ready, could you please apply it to the usages in `FutureTest` also :)?

javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -228,48 +228,42 @@ public void shouldLeftReturnNotNull() {
...
+        final Future<?> future = Future(new Error()).await();
         assertThat(future).isNotNull();
         assertThat(future.isFailure()).isTrue();
     }
 
     @Test
     public void shouldFutureWithinExecutorWithErrorReturnNotNull() {
-        final Future<?> future = Future(Executors.newSingleThreadExecutor(), new Error());
-        future.await();
+        final Future<?> future = Future(Executors.newSingleThreadExecutor(), new Error()).await();
```
paplorinc: :+1: 
I hope this has real-world value also :)
Could we provide a non-trivial test for this?

javaslang/src/main/java/javaslang/concurrent/Future.java
```diff
@@ -539,7 +539,7 @@
     /**
      * Blocks the current Thread until this Future completed or returns immediately if this Future is already completed.
      */
```
danieldietrich: Please add a `@return this Future instance` to the Javadoc and test it with `mvn javadoc:javadoc`. We need to eliminate all doclint warnings, otherwise a release fails next time.

valery1707:
@danieldietrich JavaDoc fixed
@paplorinc I don't have any real-world application for this method yet. But I inline call to `await()` in `FutureTest`

danieldietrich:
Great, thanks :)

Issue #1615, opened 2016-10-13
GWT: Promise & Future support


ruslansennov:
@dharezlak says in Gitter chat:

> I played a little with the Future/Promise pair on the other side of the fence (I mean the browser side). This required a few modifications which were super-sourcing the `ExecutorService` interface and providing a simple implementation for it. Simple use cases worked just fine.

I believe this makes sense.

dharezlak:
@ruslansennov I also had to remove code using `Object#wait()|notify()` as these methods are not available in GWT. Also, some calls do not make sense on the client side such as `Future#get()|isEmpty()` as they use Java's notification mechanisms. One way to deal with this is to use GWT's replace mechanism which could use class implementations throwing exceptions for such cases (http://www.gwtproject.org/doc/latest/DevGuideCodingBasicsDeferred.html#replacement).

ruslansennov:
We can leave `<exclude name="concurrent/**"/>` entry in `Javaslang.gwt.xml`, but implement lightweight version of `Promise` and `Future` Javaslang's classes in GWT super-source folder. If you have played with it, could you try do this?

dharezlak:
I'll try...

dharezlak:
Where can I push the preliminary implementation?

PR #1616, opened 2016-10-13
GWT support for Promise and Future

dharezlak:
Addresses https://github.com/javaslang/javaslang/issues/1615.

- `javaslang.concurrent` package not ignored by GWT any more to minimize code duplicates,
- temporary use of guava to super-source the `java.util.concurrent ` package,
- `FutureImpl` overridden to get rid of `wait/notify` usage,
- a few simplistic tests added.


dharezlak:
@ruslansennov This is a preliminary implementation of what I had so far. I included the `javaslang.concurrent` package for the GWT compiller to minimize code duplicates for now and used guava as it provides super-source for a couple of classes from the `java.util.concurrent` package. Let me know how we should proceed with this. I am hoping to use it in 2.1.0 release :)

javaslang-gwt/src/test/java/client/GwtTestConcurrent.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class GwtTestConcurrent extends GWTTestCase {
+
+    @Override public String getModuleName()  {
+        return "TestModule";
+    }
+
+    public void testFutureSuccess() {
+        AtomicBoolean onCompleteCalled = new AtomicBoolean(false);
```
ruslansennov: There is no reason to use `AtomicBoolean` here in GWT, you can use any boolean holder, for example `boolean[]`.
Also this can help you to remove guava-dependency in `javaslang-gwt`module.
... and fix failed tests :)

ruslansennov:
@dharezlak still something wrong... I'll try to run tests on my computer later

danieldietrich:
Would it make sense to create a GWT ExecutorService impl that uses native JS to call ES6 Promise in order to perform async calls?

ruslansennov:
@dharezlak magic!
Could you explain why tests passed only with suite?

javaslang-gwt/pom.xml
```diff
@@ -50,6 +50,12 @@
             <version>${gwt.version}</version>
         </dependency>
         <dependency>
+            <groupId>com.google.guava</groupId>
```
danieldietrich: We should not rely on 3rd party libs.

dharezlak:
@danieldietrich I do not think that using JS native promise will gain us anything, we always operate in the browser's single-thread environment

danieldietrich:
@dharezlak yes, I think you are right. Wonder why Promise exists at all. Because of the error handling?

Function:

```javascript
function getUser(userId) {
  return new Promise((resolve, reject) => {
    try {
      resolve(db.getUserById(userId));
    } catch(error) {
      reject(`Error loading user ${userId}: ${error}`);
    }
  });
}
```

Usage:

```javascript
getUser('phantasy-id')
  .then(user => {
    // do something with loaded user
  })
  .catch(error => {
    // update UI with error message
  });
```

dharezlak:
@ruslansennov I did not manage to get to the bottom of the GWT test runs problem, just had a hunch for going with the recommended way of combining tests for the same module...

dharezlak:
@danieldietrich One of the reasons for JS promises to exist is stepping out of the so called callback hell (http://callbackhell.com/).

javaslang-gwt/src/test/java/client/ConcurrentTestGwt.java
```diff
@@ -0,0 +1,56 @@
...
+ * /___/\_/  \_/\____/\_/  \_/\__\/__/\__\_/  \_//  \__/\_____/   Licensed under the Apache License, Version 2.0
+ */
+package client;
+
+import com.google.gwt.junit.client.GWTTestCase;
+import javaslang.collection.List;
+import javaslang.concurrent.Future;
+import javaslang.concurrent.Promise;
+
+import java.util.concurrent.atomic.AtomicBoolean;
```
ruslansennov: unused import

javaslang/src/main/resources/super/java/util/concurrent/atomic/AtomicInteger.java
```diff
@@ -0,0 +1,325 @@
...
+ * {@code AtomicInteger} is used in applications such as atomically
+ * incremented counters, and cannot be used as a replacement for an
+ * {@link java.lang.Integer}. However, this class does extend
+ * {@code Number} to allow uniform access by tools and utilities that
+ * deal with numerically-based classes.
+ *
+ * @since 1.5
+ * @author Doug Lea
+ */
+public class AtomicInteger extends Number implements java.io.Serializable {
```
ruslansennov: Where this class is used?

javaslang-gwt/src/test/java/client/ConcurrentTestGwt.java
dharezlak: Fixed

javaslang/src/main/resources/super/java/util/concurrent/atomic/AtomicInteger.java
dharezlak: Got it. It's used in the `javaslang.concurrent.Future` type.

ruslansennov: thank you

ruslansennov:
@danieldietrich LGTM
As I see, nothing changed in javaslang core and GWT-compiler happy to see `javaslang.concurrent.Future`. After pulling this we should remove most of `GwtIncompatible` annotations from `API.*future*` methods /cc @valery1707

danieldietrich:
Hey guys, I looked through the changes and have some thoughts:

* Generally we should not include 3rd party framework specific stuff into the javaslang core module. Here we added super-sources. Why can't those reside within the javaslang-gwt module?
* Copying the Javaslang sources to a super-sources folder is no good idea. It leads to maintenance hell. The original sources and the super-sources will diverge over time.
* Including the JDK sources is not a good idea. Who is keeping them actual? It is bound to a specific JDK version. Javaslang will run on multiple JDK versions. We can't include every version of a JDK file at the same time.
* Including sources that are used by _current impls_ (like FutureImpl -> AtomicInteger) is also not a good idea. When changing FutureImpl, some dependencies might be obsolete. Zombies (like AtomicInteger in this example) will reside in our code-base.

The bottom line is that we can't add features to GWT client code like this...

What are your thoughts @dharezlak @ruslansennov ?

ruslansennov:
> Generally we should not include 3rd party framework specific stuff into the javaslang core module. Here we added super-sources. Why can't those reside within the javaslang-gwt module?

I believe this is kind of taste. Super-sources is just resource text files and I'm not sure that we should create artifact for this

> Copying the Javaslang sources to a super-sources folder is no good idea. It leads to maintenance hell. The original sources and the super-sources will diverge over time.

Our super-sources contains only JDK classes which not included in GWT emulation core.

> Including the JDK sources is not a good idea. Who is keeping them actual? It is bound to a specific JDK version. Javaslang will run on multiple JDK versions. We can't include every version of a JDK file at the same time.

It's only possibility to make Javaslang compilable. And as I know this is common solution for these cases. If `javaslang.concurrent.Future` depends from JDK executor, GWT-compiler trying to find in own emulation library and can not find it. We must give it. Our emulated executor in super-sources can be modified/simplified from JDK, but it should be

> Including sources that are used by current impls (like FutureImpl -> AtomicInteger) is also not a good idea. When changing FutureImpl, some dependencies might be obsolete. 

We will change out super-sources. Tests can tell us that something wrong

> Zombies (like AtomicInteger in this example) will reside in our code-base.

Oooops :)

dharezlak:
@danieldietrich In principle all your points are valid. We could address the first point by reorganizing the project by moving super-sources to the javaslang-gwt module and start deploying it to maven. Users who wish to use it with GWT would point to the javaslang-gwt artifact which would transitively include the javaslang core artifacts (sources included).
However, for the rest I don't see a clear approach at the moment. It's simply done this way for GWT libs.
Another issue worth mentioning here is super-source conflicts. When javaslang-gwt is used with another lib super-sourcing the same JDK classes (e.g. guava) they will be overwritten depending which lib is second in the GWT descriptor (*.gwt.xml files).

danieldietrich:
@dharezlak @ruslansennov Thank you for your suggestions.

I like the idea of only one include `javaslang-gwt` when using Javaslang with GWT. We should move all stuff to that module and deploy it.

This PR should not include any changes to Javaslang core.

dharezlak:
So, to summarize, we have the following todos:

* [ ] move everything GWT-related from javaslang core to javaslang-gwt (except the `GwtIncompatible` annotation),
* [ ] make the javaslang-gwt module dependent on the javaslang sources and make it maven-deployable,
* [ ] try to resolve potential guava conflicts for projects using both guava and javaslang-gwt (e.g. through providing different GWT descriptor files).

Will these give us a green light towards 2.1.0 release?

danieldietrich:
> Will these give us a green light towards 2.1.0 release?

from my point of view yes. @ruslansennov, does this make sense to you?

ruslansennov:
>  try to resolve potential guava conflicts for projects using both guava and javaslang-gwt (e.g. through providing different GWT descriptor files).

This is bonus task :)

> Will these give us a green light towards 2.1.0 release?

I think so

dharezlak:
@ruslansennov What should we do with the now removed JavaslangExample module?

ruslansennov:
> What should we do with the now removed JavaslangExample module?

I believe we should keep it. Maybe in own, non-deployable maven module "GWT-Example". Few commits ago I removed it :( But you can find it :)

Thank you so much for your contribution!

danieldietrich:
Hi @dharezlak, this looks great now! I will pull it in now.
Could you please re-add the Gwt-Example in its own non-deployable module?
Let's name it `javaslang-gwt-example` containing @ruslansennov's readme that describes how to setup and start the  example in the browser.
Many thanks!!

dharezlak:
Will do...

danieldietrich:
Awesome :)

Issue #1261, opened 2016-04-04
Add 'computeIfPresent' to Map


enelson:
I very well may be missing it, but I don't see a function on the `Map` type that is similar to Java 8's `computeIfPresent`, where it will either return the value for a given key, or compute a value and add it to the map. Am I missing it? If not, I'm happy to take this on if you'd like.

talios:
I don't believe there is one - I was looking for something similar the other day.  You'd want to add it to the new `Multimap` hierarchy as well.

enelson:
I would also add `computeIfAbsent` as well. Both of these are useful.

danieldietrich:
Yep, please compute, eh, implement them, they are absent. :)

paplorinc:
Hmmm, isn't `withDefaultValue` and `withDefaults` used exactly for this purpose?
```
* Turns this map from a partial function into a total function that
* returns a value computed by defaultFunction for all keys
* absent from the map.
```

danieldietrich:
Good point!

`computeIfAbsent` makes only sense for mutable collections. It comes handy when lazy evaluating values (e.g. for cache application). For example we use it in Function1:

```java
    @Override
    default Function1<T1, R> memoized() {
        if (isMemoized()) {
            return this;
        } else {
            final Lazy<R> forNull = Lazy.of(() -> apply(null));
            final Object lock = new Object();
            final java.util.Map<T1, R> cache = new java.util.HashMap<>();
            return (Function1<T1, R> & Memoized) t1 -> {
                if (t1 == null) {
                    return forNull.get();
                } else {
                    final R result;
                    synchronized (lock) {
                        result = cache.computeIfAbsent(t1, this::apply);
                    }
                    return result;
                }
            };
        }
    }
```

The mutable `cache` is automatically updated if the value is not present.

---

An immutable Map can't update its internal state. Therefore the updated collection has to be returned along with the value:

```java
Tuple2<V, ? extends Map<K, V>> computeIfAbsent(K key, Function<? super K, ? extends V> mappingFunction) {
    return getOption(key).map(value -> Tuple.of(value, this)).getOrElse(() -> {
        final V value = f.apply(key);
        return Tuple.of(value, put(key, value));
    });
}
```

The point of computeIfAbsent is to compute the value only once. Subsequent calls with the same key should return the 'cached' value. Function0..8 support this behavior using a `memoized()` version of the function.

---

Having this in mind we could achieve the same by using:

```java
[...]

Issue #1543, opened 2016-08-27
Add Map/Multimap.getOrElse(K key, V defaultValue)


danieldietrich:
Currently we have only `Option<V> get(K key)`. In order to prevent creation of unnecessary `Option` instances we generally first check `map.contains(key)` and call `get(key)` if a value is present, otherwise we provide a default value. A call `get(key).getOrElse(defaultValue)` creates an additional `Option` instance.

The `contains()`  and `get()` methods both do a HAMT lookup, which is relatively costly because the hash code of the key is computed. We could do better by providing a method

``` java
interface Map<K,V> {
    V getOrElse(K key, V defaultValue);
}
```

Same for Multimap.

See also #1261: Add 'computeIfPresent' / 'computeIfAbsent' to Map

---

When we have this method, the implementations of `Multimap.Builder.ofEntries` needs to be optimized accordingly. Maybe others too.

ruslansennov:
It seems we should return `Tuple2` as result of `computeIfPresent` / `computeIfAbsent` methods. Here is one of possible signature

```java
/**
 * If the specified key is not already associated with a value,
 * attempts to compute its value using the given mapping
 * function and enters it into this map.
 *
 * @param key key whose presence in this map is to be tested
 * @param mappingFunction mapping function
 * @return the {@link Tuple2} of current or modified map and existing or computed value associated with the specified key
 */
Tuple2<Map<K, V>, V> computeIfAbsent(K key, Function<? super K, ? extends V> mappingFunction);

/**
 * If the value for the specified key is present, attempts to
 * compute a new mapping given the key and its current mapped value.
 *
 * @param key key whose presence in this map is to be tested
 * @param remappingFunction remapping function
 * @return the {@link Tuple2} of current or modified map and the {@code Some} of the value associated 
 * with the specified key, or {@code None} if none
 */
Tuple2<Map<K, V>, Option<V>> computeIfPresent(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction);
```

danieldietrich:
This makes sense. Let's swap the generic params of the return type to be consistent to Queue.dequeue():

```java
class Queue<T> {
    Tuple2<T, Queue<T>> dequeue();
}
```

danieldietrich:
This is fixed.

If there is still sth missing regarding Multimap we will implement it with #1551

PR #1617, opened 2016-10-14
#1543 Map.computeIfPresent() & Map.computeIfAbsent

ruslansennov:


javaslang/src/main/java/javaslang/collection/Map.java
```diff
@@ -151,6 +151,28 @@ default boolean contains(Tuple2<K, V> element) {
     }
 
     /**
+     * If the specified key is not already associated with a value,
+     * attempts to compute its value using the given mapping
+     * function and enters it into this map.
```
valery1707: `this` map is unchanged anyway. Wouldn't the use of `this` reserved word be misleading?

danieldietrich:
Thank you!!

danieldietrich:
Fixes #1261

PR #1621, opened 2016-10-17
crossProduct of negative power return empty result

### LINKED ISSUE #1602: crossProduct of negative power should return empty result instead of throwing
Throwing exceptions makes the code unsafe. It is better to make the crossProduct function total for negative inputs.

valery1707:
Fix #1602 
Need I extend JavaDoc in `javaslang.collection.Seq#crossProduct(int)`?


valery1707:
@danieldietrich please check JavaDoc

danieldietrich:
Awesome, thank you!

PR #1622, opened 2016-10-17
Convert Value to Java Parallel Stream

### LINKED ISSUE #1592: Add Value.stream() and Value.parallelStream()
And deprecate Value.toJavaStream().

It seems odd but users might need to do it.

It is an integration-feature.

valery1707:
Fix #1592


danieldietrich:
That was fast, thank you!

Issue #1608, opened 2016-10-06
Make *Map factory methods safe


danieldietrich:
`Map(Object...)` is unsafe. Stuart Marks explained in his JavaOne 2016 Session [Collections Refueled](https://www.youtube.com/watch?v=LgR9ByD1dEw) how and why the Java language architects implemented [JEP 269](http://openjdk.java.net/jeps/269).

The essence regarding `Map.of()`is best described by this image:

![jep 269 api overview](https://cloud.githubusercontent.com/assets/743833/19150521/920f6fcc-8bc7-11e6-88c8-7448070e7a2b.png)

Javaslang needs to be safe. We will follow the decisions made in JEP 269 and therefore we will deprecate or `*Map.of(Object...)` methods and strongly discourage using them. Instead we will add methods
- `of(K k1, V v1)`
- `of(K k1, V v1, K k2, V v2)`
- ...

up to 10 key/value pairs. For more than 10 pairs we need to use tuples:
- `of(Tuple2<? extends K, ? extends V>...)`

Accordingly we will also deprecate `ofEntries(Tuple2<? extends K,? extends V>...)`.

danieldietrich:
@ruslansennov Is this s.th. for you? No hurry here...

danieldietrich:
We haven't defined a standard for deprecation yet. Currently I use the following:

* annotate methods with `@Deprecated`
* add javadoc `@deprecated <description>`

Descriptions are one of or a combination of these:

* `Should not be used any more because ...` (e.g. `it is unsafe`)
* `Will be removed in a future version.`
* ...

ruslansennov:
I'll try to implement this next week

danieldietrich:
Great :)

ruslansennov:
Let's take this code:
```java
final TreeMap<String, Integer> tm = TreeMap.of(Tuple.of("one", 1), Tuple.of("two", 2));
```
Compilation fails with
```
Error: reference to of is ambiguous
  both method 
<K,V>of(javaslang.Tuple2<? extends K,? extends V>,javaslang.Tuple2<? extends K,? extends V>) 
in javaslang.collection.TreeMap and method 
<K,V>of(K,V) in javaslang.collection.TreeMap match
```
The problem can be solved something like this:
```java
final TreeMap<String, Integer> tm = 
    TreeMap.<String, Integer>of(Tuple.of("one", 1), Tuple.of("two", 2));
```
but it is inconvenient.

[...]

PR #1624, opened 2016-10-17
Map.of(...) #1608

ruslansennov:


javaslang/src/main/java/javaslang/collection/TreeMap.java
```diff
@@ -133,24 +133,21 @@ private TreeMap(RedBlackTree<Tuple2<K, V>> entries) {
...
      * @return A new Map containing the given entries
      */
+    @Deprecated
     @SuppressWarnings("unchecked")
     public static <K extends Comparable<? super K>, V> TreeMap<K, V> of(Object... pairs) {
-        return of((Comparator<? super K> & Serializable) K::compareTo, pairs);
-    }
-
-    @SuppressWarnings("unchecked")
-    public static <K, V> TreeMap<K, V> of(Comparator<? super K> keyComparator, Object... pairs) {
```
ruslansennov: @danieldietrich please check this code.
This method removed instead of make them deprecated. Compiler says "possible ambiguous with of(keyComparator, k1, v1, ..., k10, v10)". The problem observed only with TreeMap and only in this line of code. I have no idea whats wrong

javaslang/generator/Generator.scala
```diff
@@ -309,17 +310,20 @@ def generateMainClasses(): Unit = {
...
 
-          /$javadoc
-           * Alias for {@link $mapType#of(Object...)}
-           *
-           * @param <K>   The key type.
-           * @param <V>   The value type.
-           * @param pairs A list of key-value pairs.
-           * @return A new {@link $mapType} instance containing the given entries
-           */
-          public static <K, V> $mapType<K, V> $name(Object... pairs) {
```
ruslansennov: `API.*Map(Objects...)` removed because fortunately still not released

javaslang/generator/Generator.scala
```diff
@@ -309,17 +310,20 @@ def generateMainClasses(): Unit = {
...
-              return $mapType.of(pairs);
-          }
+          ${(2 to VARARGS).gen(i => {
+            xs"""
+              /$javadoc
+               * Alias for {@link $mapType#of(${(1 to i).gen(j => "K, V")(", ")})}
```
valery1707: Javadoc linter in stage `javadoc:javadoc` will be generate warnings in link to `HashMap#of(K, V, K, V)`. I think JavaDoc linter does not known about generics.
Link `HashMap#of(Object, Object, Object, Object)` will be correctly parsed by JavaDoc linter and Java IDE.

javaslang/generator/Generator.scala
```diff
@@ -764,17 +768,20 @@ def generateMainClasses(): Unit = {
...
-           * @param pairs A list of key-value pairs.
-           * @return A new {@link $TreeMapType} instance containing the given entries
-           */
-          public static <K extends Comparable<? super K>, V> $TreeMapType<K, V> SortedMap(Object... pairs) {
-              return $TreeMapType.of(pairs);
-          }
+          ${(2 to VARARGS).gen(i => {
+            xs"""
+              /$javadoc
+               * Alias for {@link $TreeMapType#of(${(1 to i).gen(j => "K, V")(", ")})}
```
valery1707: Warning from JavaDoc linter as above

javaslang/generator/Generator.scala
ruslansennov: thank you, will fix this

javaslang/src/main/java/javaslang/collection/TreeMap.java
danieldietrich: No errors here... Which version do you use? I have `Java 1.8.0_101`.

The following code does work:

```java
public class Test {

        // TreeMap((1, a), (2, b), (3, c), (4, d), (5, e), (6, f), (7, g), (8, h), (9, i), (10, j))
        final Map<Integer, String> map = TreeMap.of(Integer::compareTo,
                1, "a",
                2, "b",
                3, "c",
                4, "d",
                5, "e",
                6, "f",
                7, "g",
                8, "h",
                9, "i",
                10, "j"
                );

        // TreeMap((1, a), (2, b), (3, c), (4, d), (5, e), (6, f), (7, g), (8, h), (9, i), (10, j))
        System.out.println(map);

    }
}
```

I tend to keep the method and add a unit test in TreeMapTest. If the CI build is ok, we can keep it.

javaslang/src/test/java/javaslang/collection/euler/Euler17Test.java
```diff
@@ -99,7 +99,7 @@ default int letterCount(Seq<Integer> range) {
             1_000, "thousand",
             1_000_000, "million",
             1_000_000_000, "billion"
-    );
+    ).grouped(2).toSortedMap(pair -> Tuple.of((Integer) pair.get(0), (String) pair.get(1)));
```
danieldietrich: you fox, nice trick :)

javaslang/src/main/java/javaslang/collection/TreeMap.java
ruslansennov: Please check this situation: 
class `TreeMap` has **both** methods `of(Comparator, Objects...)` and `of(Comparator, k1, v1, ... , k10, v10)`.
My Java (it seems `101u` too) fails at compilation phase.

ruslansennov:
reproduced: 
```
[WARNING] /home/travis/build/javaslang/javaslang/javaslang/src/main/java/javaslang/collection/TreeMap.java:[151,40] 
<K,V>of(java.util.Comparator<? super K>,java.lang.Object...) in javaslang.collection.TreeMap is potentially ambiguous with 
<K,V>of(java.util.Comparator<? super K>,K,V,K,V,K,V,K,V,K,V,K,V,K,V,K,V,K,V,K,V) in javaslang.collection.TreeMap
```

danieldietrich:
I see. I tested it only in IntelliJ, which does not break the build on warnings. There are also other ambiguities:

ambiguous

<small>**Note:** In this example my impl of the deprecated methods throws in order to double-check at runtime which method was really called.</small>

Now we have three choices:

* disable the build error on compiler warnings in the pom --> No.
* don't ship this change with a minor release (e.g. 2.1.0) --> Don't want to do that because of safety.
* remove the deprecated method(s) and break compatibility --> Maybe this :-/

ruslansennov:
> remove the deprecated method(s) and break compatibility --> Maybe this :-/

maybe there is only one possibility

danieldietrich:
Please remove all `of(Object...)` and `of(Comparator, Object...)` methods. And I mean *all*!

If we break, then the right way. These methods are unsafe. Having them, even if deprecated, is a risk for our users.

ruslansennov:
> If we break, then the right way.

OK, now all is clear for me.
I'll do the same for `of(Entry...)`, `of(Tuple2)` and for all methods in all multimaps

danieldietrich:
@ruslansennov don't do that. what would be the minimal-invasive solution? just removing the ambiguous TreeMap.of(Comparator, Object...) method, right?

ruslansennov:
Yes, removing all factory methods with `Object...` argument and replacing with `KVKV`

danieldietrich:
> Yes, removing all factory methods with `Object...` argument and replacing with `KVKV`

@ruslansennov I'm torn back and forth - but I think you are right. These methods are a pain in the ass. Keeping them as `@Deprecated` would not prevent from using them. We need to remove the methods containing Object... args also from the Multimap*.Builder classes.

javaslang/src/main/java/javaslang/collection/HashMap.java
```diff
@@ -121,11 +121,113 @@ private HashMap(HashArrayMappedTrie<K, V> trie) {
...
+     * @return A new Map containing the given entries
+     */
+    public static <K, V> HashMap<K, V> of(K k1, V v1, K k2, V v2, K k3, V v3, K k4, V v4, K k5, V v5, K k6, V v6, K k7, V v7, K k8, V v8, K k9, V v9, K k10, V v10) {
+        return new HashMap<>(HashArrayMappedTrie.<K, V> empty().put(k1, v1).put(k2, v2).put(k3, v3).put(k4, v4).put(k5, v5).put(k6, v6).put(k7, v7).put(k8, v8).put(k9, v9).put(k10, v10));
+    }
+
+    /**
+     * Creates a HashMap of the given list of key-value pairs.
+     *
+     * @deprecated Should not be used any more because unsafe
```
paplorinc: suggestion: 
`@deprecated Should not be used any more because it's unsafe. Use the (K, V) vararg alternative or the Tuple alternative instead`

javaslang/src/main/java/javaslang/collection/LinkedHashMap.java
```diff
@@ -94,11 +94,14 @@ private LinkedHashMap(Queue<Tuple2<K, V>> list, HashMap<K, V> map) {
     /**
      * Creates a LinkedHashMap of the given list of key-value pairs.
      *
+     * @deprecated Should not be used any more because unsafe
```
paplorinc: same

javaslang/generator/Generator.scala
```diff
@@ -309,17 +297,20 @@ def generateMainClasses(): Unit = {
...
-           */
-          public static <K, V> $mapType<K, V> $name(Object... pairs) {
-              return $mapType.of(pairs);
-          }
+          ${(1 to VARARGS).gen(i => {
+            xs"""
+              /$javadoc
+               * Alias for {@link $mapType#of(${(1 to i).gen(j => "Object, Object")(", ")})}
+               *
+               * @param <K>   The key type.
```
paplorinc: minor: formatting, too many spaces before the params

javaslang/generator/Generator.scala
```diff
@@ -309,17 +297,20 @@ def generateMainClasses(): Unit = {
...
-           * @param pairs A list of key-value pairs.
-           * @return A new {@link $mapType} instance containing the given entries
-           */
-          public static <K, V> $mapType<K, V> $name(Object... pairs) {
-              return $mapType.of(pairs);
-          }
+          ${(1 to VARARGS).gen(i => {
+            xs"""
+              /$javadoc
+               * Alias for {@link $mapType#of(${(1 to i).gen(j => "Object, Object")(", ")})}
```
paplorinc: Since we know the type of the params, we may want to show it in the javadoc also, i.e.
```"K, V")(", ")```

File: javaslang/src/main/java/javaslang/collection/TreeMap.java [hunk 10]
```diff
@@ -203,6 +207,105 @@ private TreeMap(RedBlackTree<Tuple2<K, V>> entries) {
...
 
     /**
+     * Creates a HashMap of the given list of key-value pairs.
+     *
+     * @param <K> The key type
+     * @param <V> The value type
+     * @return A new Map containing the given entries
+     */
+    public static <K extends Comparable<? super K>, V> TreeMap<K, V> of(K k1, V v1, K k2, V v2) {
+        return of((Comparator<? super K> & Serializable) K::compareTo, k1, v1, k2, v2);
```
paplorinc: I think this cast (`(Comparator<? super K> & Serializable`) is not Java 9 safe :/

javaslang/src/main/java/javaslang/collection/TreeMap.java
```diff
@@ -218,6 +321,114 @@ private TreeMap(RedBlackTree<Tuple2<K, V>> entries) {
     }
 
     /**
+     * Creates a HashMap of the given list of key-value pairs.
+     *
+     * @param keyComparator key comparator
```
paplorinc: `* @param keyComparator The comparator used to sort the entries by their key.`

javaslang/src/main/java/javaslang/collection/TreeMap.java
```diff
@@ -218,6 +321,114 @@ private TreeMap(RedBlackTree<Tuple2<K, V>> entries) {
...
+     * @param <K> The key type
+     * @param <V> The value type
+     * @return A new Map containing the given entries
```
paplorinc: Were the parameter docs omitted on purpose here?

javaslang/src/test/java/javaslang/collection/euler/Euler17Test.java
paplorinc: For type safety and better documentation via tests, maybe we should use `.merge()` instead, i.e.
```java
static final Map<Integer, String> LENGTHS = TreeMap.of(
        1, "one",
        2, "two",
        3, "three",
        4, "four",
        5, "five",
        6, "six",
        7, "seven",
        8, "eight",
        9, "nine",
        10, "ten"
).merge(TreeMap.of(
        11, "eleven",
        12, "twelve",
        13, "thirteen",
        14, "fourteen",
        15, "fifteen",
        16, "sixteen",
        17, "seventeen",
        18, "eighteen",
        19, "nineteen",
        20, "twenty"
)).merge(TreeMap.of(
        30, "thirty",
        40, "forty",
        50, "fifty",
        60, "sixty",
        70, "seventy",
        80, "eighty",
        90, "ninety",
        100, "hundred",
        1_000, "thousand",
        1_000_000, "million",
        1_000_000_000, "billion"
));
```
(though this doesn't seem to compile ... @danieldietrich, why is this working if I extract the parts and merge them in a method, but not inline?)

Maybe we could extend the `merge` to vararg also, so that larger maps can be created in a safe way also :)

javaslang/src/main/java/javaslang/collection/TreeMap.java
```diff
@@ -218,6 +321,114 @@ private TreeMap(RedBlackTree<Tuple2<K, V>> entries) {
...
+
+    /**
+     * Creates a HashMap of the given list of key-value pairs.
+     *
+     * @param keyComparator key comparator
+     * @param <K> The key type
+     * @param <V> The value type
+     * @return A new Map containing the given entries
+     */
+    public static <K extends Comparable<? super K>, V> TreeMap<K, V> of(Comparator<? super K> keyComparator, K k1, V v1, K k2, V v2, K k3, V v3, K k4, V v4, K k5, V v5, K k6, V v6, K k7, V v7, K k8, V v8, K k9, V v9, K k10, V v10) {
```
paplorinc: we could reuse the previous ones here, i.e.
intead of
```java
return TreeMap.<K, V> empty(keyComparator).put(k1, v1).put(k2, v2).put(k3, v3).put(k4, v4).put(k5, v5).put(k6, v6).put(k7, v7).put(k8, v8).put(k9, v9).put(k10, v10);
```
```java
return of(keyComparator, k1, v1, k2, v2, k3, v3, k4, v4, k5, v5, k6, v6, k7, v7, k8, v8, k9, v9).put(k10, v10);
```

javaslang/src/main/java/javaslang/collection/LinkedHashMap.java
```diff
@@ -149,6 +152,123 @@ private LinkedHashMap(Queue<Tuple2<K, V>> list, HashMap<K, V> map) {
...
     /**
+     * Creates a LinkedHashMap of the given list of key-value pairs.
+     *
+     * @param <K> The key type
+     * @param <V> The value type
+     * @return A new Map containing the given entries
+     */
+    public static <K, V> LinkedHashMap<K, V> of(K k1, V v1, K k2, V v2) {
+        final HashMap<K, V> map = HashMap.of(k1, v1, k2, v2);
+        final Queue<Tuple2<K, V>> list = Queue.of(Tuple.of(k1, v1), Tuple.of(k2, v2));
```
paplorinc: we could use the newly created aliases here
```java
final Queue<Tuple2<K, V>> list = Queue(Tuple(k1, v1), Tuple(k2, v2));
```

paplorinc:
@ruslansennov, @danieldietrich, @valery1707, we should add these possibilities to the aliases also and provide a convenient way of extending it to more than 10 elements.
I suggest making `merge` a vararg, to allow comfortable concatenation :)

danieldietrich:
@paplorinc 

> I suggest making merge a vararg, to allow comfortable concatenation :)

It would blow up the API too much. For more than 10 args the ofEntries method is used (e.g. in conjunction with the Tuple(K, V) alias). It is evil enough that we now have all these *Map.of(KVKVKVKV...) methods.

Let's not throw in too much new API when we do not know how frequently it is used at all...

/cc @ruslansennov @valery1707

ruslansennov:
@danieldietrich please leave this PR unmerged for a while

javaslang/src/test/java/javaslang/collection/euler/Euler17Test.java
danieldietrich: @paplorinc is also a fox - just saying that we have Java-trickery experts here :)

danieldietrich:
> please leave this PR unmerged for a while

Ok, will do!

javaslang/src/main/java/javaslang/collection/LinkedHashMap.java
```diff
@@ -92,28 +92,198 @@ private LinkedHashMap(Queue<Tuple2<K, V>> list, HashMap<K, V> map) {
...
+     * @param e5  The 5th entry
+     * @param e6  The 6th entry
+     * @param e7  The 7th entry
+     * @param e8  The 8th entry
+     * @return A new Map containing the given entries
+     */
+    @SuppressWarnings("unchecked")
+    public static <K, V> LinkedHashMap<K, V> ofEntries(Tuple2<? extends K, ? extends V> e1, Tuple2<? extends K, ? extends V> e2, Tuple2<? extends K, ? extends V> e3, Tuple2<? extends K, ? extends V> e4, Tuple2<? extends K, ? extends V> e5, Tuple2<? extends K, ? extends V> e6, Tuple2<? extends K, ? extends V> e7, Tuple2<? extends K, ? extends V> e8) {
+        final HashMap<K, V> map = HashMap.ofEntries(e1, e2, e3, e4, e5, e6, e7, e8);
+        final Queue<Tuple2<K, V>> list = Queue.of((Tuple2<K, V>) e1, (Tuple2<K, V>) e2, (Tuple2<K, V>) e3, (Tuple2<K, V>) e4, (Tuple2<K, V>) e5, (Tuple2<K, V>) e6, (Tuple2<K, V>) e7, (Tuple2<K, V>) e8);
```
paplorinc: hmm, these casts are ugly, but I could't come up with a better solution (except for chaining the methods, as I suggested earlier ... @ruslansennov, could you please react to that?)

Edit: I was still composing this review, not sure why you were notified, @ruslansennov :p

javaslang/src/main/java/javaslang/collection/LinkedHashMap.java
```diff
@@ -181,39 +468,203 @@ private LinkedHashMap(Queue<Tuple2<K, V>> list, HashMap<K, V> map) {
...
-     * @param <K>     The key type
-     * @param <V>     The value type
+     * @param <K> The key type
+     * @param <V> The value type
+     * @param e1  The 1st entry
+     * @param e2  The 2nd entry
+     * @return A new Map containing the given entries
+     */
+    @SuppressWarnings("unchecked")
+    public static <K, V> LinkedHashMap<K, V> ofEntries(java.util.Map.Entry<? extends K, ? extends V> e1, java.util.Map.Entry<? extends K, ? extends V> e2) {
```
danieldietrich: Hi @ruslansennov,

I think we do not need the same KVKVKV variants for ofEntries. We not use just varags:

```java
ofEntries(j.u.Map.Entry<? extends K, ? extends V>... entries)
ofEntries(Tuple2<? extends K, ? extends V>... entries)
```

(also with Comparator).

javaslang/src/main/java/javaslang/collection/LinkedHashMap.java
paplorinc: @ruslansennov?

ruslansennov:
>     public static <K, V> LinkedHashMap<K, V> ofEntries(Tuple2<? extends K, ? extends V> e1, Tuple2<? extends K, ? extends V> e2, Tuple2<? extends K, ? extends V> e3, Tuple2<? extends K, ? extends V> e4, Tuple2<? extends K, ? extends V> e5, Tuple2<? extends K, ? extends V> e6, Tuple2<? extends K, ? extends V> e7, Tuple2<? extends K, ? extends V> e8) {
>         final HashMap<K, V> map = HashMap.ofEntries(e1, e2, e3, e4, e5, e6, e7, e8);
>         final Queue<Tuple2<K, V>> list = Queue.of((Tuple2<K, V>) e1, (Tuple2<K, V>) e2, (Tuple2<K, V>) e3, (Tuple2<K, V>) e4, (Tuple2<K, V>) e5, (Tuple2<K, V>) e6, (Tuple2<K, V>) e7, (Tuple2<K, V>) e8);
> hmm, these casts are ugly, but I could't come up with a better solution (except for chaining the methods, as I suggested earlier ... @ruslansennov, could you please react to that?)

Can't find your comment directly (do you removed it?)
@danieldietrich proposed to remove this methods, let's wait his comment

javaslang/src/main/java/javaslang/collection/LinkedHashMap.java
paplorinc: this is not a vararg to avoid https://docs.oracle.com/javase/tutorial/java/generics/nonReifiableVarargsType.html#heap_pollution, right?
We do have many of these, though, maybe it's ok in our case :)

javaslang/src/main/java/javaslang/collection/LinkedHashMap.java
paplorinc: in case you will keep these, consider creating a method `t` here also, just for the casts :)

javaslang/src/main/java/javaslang/collection/TreeMap.java
```diff
@@ -131,75 +131,682 @@ private TreeMap(RedBlackTree<Tuple2<K, V>> entries) {
...
+     */
+    public static <K extends Comparable<? super K>, V> TreeMap<K, V> ofEntries(Tuple2<? extends K, ? extends V> e1, Tuple2<? extends K, ? extends V> e2, Tuple2<? extends K, ? extends V> e3, Tuple2<? extends K, ? extends V> e4, Tuple2<? extends K, ? extends V> e5, Tuple2<? extends K, ? extends V> e6, Tuple2<? extends K, ? extends V> e7, Tuple2<? extends K, ? extends V> e8) {
+        return ofEntries((Comparator<? super K> & Serializable) K::compareTo, e1, e2, e3, e4, e5, e6, e7, e8);
+    }
+
+    /**
+     * Creates a {@code TreeMap} of the given entries.
+     * The underlying key comparator is the natural comparator of K.
+     *
+     * @param <K>   The key type
```
paplorinc: minor: indentation

javaslang/src/main/java/javaslang/collection/TreeMap.java
```diff
@@ -288,84 +1003,200 @@ private TreeMap(RedBlackTree<Tuple2<K, V>> entries) {
...
+     * @param e2  The 2nd entry
+     * @param e3  The 3rd entry
+     * @param e4  The 4th entry
+     * @param e5  The 5th entry
+     * @param keyComparator The comparator used to sort the entries by their key.
+     * @return A new TreeMap containing the given entry.
+     */
+    public static <K, V> TreeMap<K, V> ofEntries(Comparator<? super K> keyComparator, java.util.Map.Entry<? extends K, ? extends V> e1, java.util.Map.Entry<? extends K, ? extends V> e2, java.util.Map.Entry<? extends K, ? extends V> e3, java.util.Map.Entry<? extends K, ? extends V> e4, java.util.Map.Entry<? extends K, ? extends V> e5) {
+        Objects.requireNonNull(e5, "The 5th entry is null");
+        return TreeMap.<K, V> ofEntries(keyComparator, e1, e2, e3, e4).put(e5.getKey(), e5.getValue());
```
paplorinc: I think you could create a `t` here also, that creates a `Tuple` from an `Entry`

javaslang/src/test/java/javaslang/collection/AbstractMapTest.java
```diff
@@ -891,26 +886,6 @@ public void shouldFillTheSeqWith0ElementsWhenNIsNegative() {
         assertThat(mapFill(-1, () -> new Tuple2<>(1, 1))).isEqualTo(empty());
     }
 
-    @SuppressWarnings("unchecked")
-    @Test
-    public void mapOfTuplesShouldReturnTheSingletonEmpty() {
```
paplorinc: no empty check anymore, or were these duplicates?

javaslang/src/test/java/javaslang/collection/euler/Euler17Test.java
paplorinc: @ruslansennov, @danieldietrich, I think this test shows us that we may need a more composable solution to be real-world-ready. @ruslansennov, could we solve this without trickery :)?

javaslang/src/main/java/javaslang/collection/LinkedHashMap.java
ruslansennov: isn't `ofEntries(j.u.Map.Entry<? extends K, ? extends V>... entries)` unsafe?

javaslang/src-gen/test/java/javaslang/APITest.java
```diff
@@ -709,38 +709,328 @@ public void shouldEmptyLinkedMapReturnNotNull() {
...
+    public void shouldLinkedMapFromMapReturnNotNull() {
+        assertThat(LinkedMap(Collections.singletonMap(1, '1'))).isNotNull();
     }
 
     @Test
-    public void shouldLinkedMapFromTuplesReturnNotNull() {
-        assertThat(LinkedMap(Tuple(1, '1'), Tuple(2, '2'), Tuple(3, '3'))).isNotNull();
+    public void shouldCreateLinkedMapFrom1Pairs() {
+      Map<Integer, Integer> map = LinkedMap(1, 2);
+      assertThat(map.getOrElse(1, 0)).isEqualTo(2);
```
paplorinc: consider using `map.apply(1)` instead of `map.getOrElse(1, 0)`

javaslang/src/main/java/javaslang/collection/LinkedHashMap.java
danieldietrich: Immutable objects are covariant in their generic arguments. However, we will not restrict Map.Entry to be immutable here. Internally we do not store the entries, instead we create immutable Tuples.

We do not provide special methods ofEntries(Map.EntryKVKV) and ofEntries(Tuple2KVKV).

javaslang/src/main/java/javaslang/collection/LinkedHashMap.java
danieldietrich: No additional method `t` here please. We will use `Tuple2.narrow()` which does not exist. I will create an issue and we will change these casts in a later PR.

ruslansennov:
Sorry guys
I'm totally confused. I have to close this PR and open new #1664 with `Map.of(KVKVKV)` only.

paplorinc:
@ruslansennov, what' the source of confusion?

ruslansennov:
> @ruslansennov, what' the source of confusion?

Huge PR, a lot of comments, different opinions about `ofEntries(Tuple/Entry/KV)`...
I prefer small PRs which is more observable (this is only my personally sight)

PR #1631, opened 2016-10-19
Overrides numeric Map operations.

danieldietrich:
It was necessary for sum(), product() and average(), because a Map consists of Tuples, which are not numeric. We now operate on Map values.

(Minor fixes in Traversable: Avoids unnecessary object creation in empty case.)


javaslang/src/main/java/javaslang/collection/Map.java
```diff
@@ -387,6 +408,26 @@ default int length() {
...
+     *
+     * @return {@code Some(maximum)} of this elements or {@code None} if this is empty or no {@code Comparator} is applicable
+     */
+    @SuppressWarnings("unchecked")
+    @Override
+    default Option<Tuple2<K, V>> max() {
+        if (isEmpty()) {
+            return Option.none();
+        } else {
+            if (head()._2 instanceof Comparator) {
```
ruslansennov: `Comparable`?

javaslang/src/main/java/javaslang/collection/Map.java
```diff
@@ -413,6 +454,37 @@ default int length() {
...
+     *
+     * @return {@code Some(minimum)} of this elements or {@code None} if this is empty or no {@code Comparator} is applicable
+     */
+    @SuppressWarnings("unchecked")
+    @Override
+    default Option<Tuple2<K, V>> min() {
+        if (isEmpty()) {
+            return Option.none();
+        } else {
+            if (head()._2 instanceof Comparator) {
```
ruslansennov: dito

javaslang/src/main/java/javaslang/collection/Map.java
danieldietrich: Yes, good catch! Thought there are already test - have to check them again...

danieldietrich: Multimap does need similar overrides, I will add them, too.
Also the coverage decreased. I will take care of it.

javaslang/src/main/java/javaslang/collection/Map.java
```diff
@@ -413,6 +454,37 @@ default int length() {
...
+
+    /**
+     * Calculates the <em>product</em> of this {@code Map} values, assuming that the values are numeric.
+     *
+     * @return a {@code Number} representing the product of this {@code Map} values
+     * @throws UnsupportedOperationException if the values are not numeric
+     */
+    @Override
+    default Number product() {
+        return iterator().map(t -> t._2).product();
```
valery1707: In `average` you use `Tuple2::_2` for mapping function

javaslang/src/main/java/javaslang/collection/Map.java
```diff
@@ -521,6 +593,17 @@ default int length() {
...
 
     /**
+     * Calculates the <em>sum</em> of this {@code Map} values, assuming that the values are numeric.
+     *
+     * @return a {@code Number} representing the sum of this {@code Map} values
+     * @throws UnsupportedOperationException if the values are not numeric
+     */
+    @Override
+    default Number sum() {
+        return iterator().map(t -> t._2).sum();
```
valery1707: dito

javaslang/src/main/java/javaslang/collection/Map.java
danieldietrich: will change it

javaslang/src/main/java/javaslang/collection/Map.java
danieldietrich: > Comparable?

The tests use `IntMap<T>` which extends `Traversable<T>`. We need to extends `Map` in order to test the overrides.

However, currently we need IntMap because our AbstractMapTest extends the AbstractTraversableTest and otherwise the method overrides do not compile.

I will search for another solution.

This PR: simple change, big impact o_O

danieldietrich:
It is wrong by me to try to fix it by overriding the numeric methods in Maps. It increases the overall complexity of the application. It is a sign that the design is wrong. Also it will be a reason to move to 3.0.0 to get it right from the user-perspective.

This should be solved straight-forward by letting Map extends `Traversable<V>` instead of `Traversable<Tuple2<K, V>>`.

**Workaround:** `map.values().sum();`

PR #1632, opened 2016-10-19
Added API shortcuts (TODO and print*)

danieldietrich:
This is a nice one :) Some might think it is unnecessary because these are only side-effecting methods - but I think it will be useful for the development phase:

As usual one import (**side-note: IntelliJ support / template would be awsome**):

```java
import static javaslang.API.*;
```

Then

```java
// instead of System.out.println("test")
println("test");

// instead of System.out.printf("Hello %s", name);
printf("Hello %s", name);
```

Also rapid prototyping in a functional style is possible. Just create APIs without using `null` as placeholder. Our `TODO` is equivalent to Scala's `???`:

```java
public HttpResponse getResponse(HttpRequest request) {
    return TODO("need to create http response");
}

// we create the request later
final HttpResponse response = getResponse(TODO());
```


danieldietrich:
@ruslansennov Help needed! GWT complains `The method printf(String, Object[]) is undefined for the type PrintStream`. May it be a problem with varargs calls? We have `API.printf(String, Object...)` that calls `PrintStream.printf(String, Object...)`:

```java
public static void printf(String format, Object... args) {
    System.out.printf(format, args);
}
```

This is the build error:

```bash
[INFO] --- gwt-maven-plugin:1.0-rc-6:compile (default-compile) @ javaslang-gwt-example ---
[INFO] Compiling module JavaslangExample
[INFO]    Ignored 10 units with compilation errors in first pass.
[INFO] Compile with -strict or with -logLevel set to TRACE or DEBUG to see all errors.
[INFO]    [ERROR] Errors in 'jar:file:/home/travis/build/javaslang/javaslang/javaslang/target/javaslang-3.0.0-SNAPSHOT-sources.jar!/javaslang/API.java'
[INFO]       [ERROR] Line 161: The method printf(String, Object[]) is undefined for the type PrintStream
[INFO]    Tracing compile failure path for type 'javaslang.API'
[INFO]       [ERROR] Errors in 'jar:file:/home/travis/build/javaslang/javaslang/javaslang/target/javaslang-3.0.0-SNAPSHOT-sources.jar!/javaslang/API.java'
[INFO]          [ERROR] Line 161: The method printf(String, Object[]) is undefined for the type PrintStream
[INFO]    [ERROR] Hint: Your source appears not to live underneath a subpackage called 'client'; no problem, but you'll need to use the <source> directive in your module to make it accessible
```

javaslang/src/main/java/javaslang/NotImplementedError.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public class NotImplementedError extends Error {
+
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Creates a {@code NotImplementedError} containing the message "an implementation is missing".
+     */
+    public NotImplementedError() {
+        super("an implementation is missing");
```
valery1707: Maybe start message with upper case char? As in [java.text.DateFormatSymbols#getInstance(Locale)](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/text/DateFormatSymbols.java?av=f#339)

ruslansennov:
> The method printf(String, Object[]) is undefined for the type PrintStream

This is true. [Here](https://github.com/gwtproject/gwt/blob/master/user/super/com/google/gwt/emul/java/io/PrintStream.java) is GWT super-source of `PrintStream`. No `printf*` methods emulation

danieldietrich:
@ruslansennov Thank you for the link!

PR #1645, opened 2016-10-26
Add Try.of method accepting void supplier.

whirlwin:
It would be cool if `Try` accepted a void method. E.g.

```java
Try.of(() -> Thread.sleep(1))
    .flatMap(nothing -> doSomethingElse());
```

Not sure if "void supplier" is the best name though, as it seems to be kind of self contradicting.

### DIFF
--- a/javaslang/src/main/java/javaslang/control/Try.java
+++ b/javaslang/src/main/java/javaslang/control/Try.java
@@ -45,6 +45,23 @@ public interface Try<T> extends Value<T> {
     }
 
     /**
+     * Creates a Try of a CheckedVoidSupplier.
+     *
+     * @param supplier A checked supplier
+     * @param <T>      Component type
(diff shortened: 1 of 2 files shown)


danieldietrich:
I think this will raise ambiguities. Why not use `Try.run`?

danieldietrich:
@whirlwin I thought about your suggestion. I also like to have same factory method names, it is better than learning new API (of, run, ...).

Also I verified that `of(CheckedSupplier)` does not clash with `of(CheckedVoidSupplier)`.

However, with Project Valhalla `Supplier<void>` will be the same as `Runnable` (as I understand Brian Goetz' [mission statement](http://mail.openjdk.java.net/pipermail/valhalla-spec-experts/2016-October/000175.html)). So having both methods named `of` is no good idea from the strategic viewpoint.

I hope you can live with `Try.run()`. I will close this PR.

Many thanks for participating! I hope I will see more PR's from you in future :-)

PR #1646, opened 2016-10-26
Fixes truncation bug of dropWhile and null handling of takeWhile and …

### LINKED ISSUE #1641: Incorrect Span behaviour
java> List.of(1,1,2,2,1,1).span(x -> x == 1)
(List(1, 1), List(2, 2))
// It should (List(1, 1), List(2, 2, 1, 1)) The list is simply truncated here
java> List.of(1,1,2,2,4,4).span(x -> x == 1)
(List(1, 1), List(2, 2, 4, 4))

It seems it is using map behind the scene. 
I am using version 2.0.2

danieldietrich:
…dropWhile

Fixes #1641

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Iterator.java
+++ b/javaslang/src/main/java/javaslang/collection/Iterator.java
@@ -1392,24 +1392,37 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
             final Iterator<T> that = this;
             return new AbstractIterator<T>() {
 
-                private T next = null;
+                private T next;
+                private boolean cached = false;
+                private boolean first = true;
 
                 @Override
                 public boolean hasNext() {
-                    while (next == null && that.hasNext()) {
-                        final T value = that.next();
-                        if (!predicate.test(value)) {
-                            next = value;
+                    if (cached) {
+                        return true;
+                    } else if (first) {
+                        first = false;
+                        while (that.hasNext()) {
+                            next = that.next();
+                            if (!predicate.test(next)) {
+                                cached = true;
+                                return true;
+                            }
                         }
+                        return false;
+                    } else if (that.hasNext()) {
+                        next = that.next();
+                        cached = true;
+                        return true;
+                    } else {
+                        return false;
                     }
-                    return next != null;
                 }
 
                 @Override
                 public T getNext() {
-                    final T result = next;
-                    next = null;
-                    return result;
+                    cached = false;
+                    return next;
                 }
             };
         }
@@ -1920,27 +1933,31 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
             final Iterator<T> that = this;
             return new AbstractIterator<T>() {
 
-                private T next = null;
+                private T next;
+                private boolean cached = false;
                 private boolean finished = false;
 
                 @Override
                 public boolean hasNext() {
-                    while (!finished && next == null && that.hasNext()) {
-                        final T value = that.next();
-                        if (predicate.test(value)) {
-                            next = value;
-                        } else {
-                            finished = true;
+                    if (cached) {
+                        return true;
+                    } else if (finished) {
+                        return false;
+                    } else if (that.hasNext()) {
+                        next = that.next();
+                        if (predicate.test(next)) {
+                            cached = true;
+                            return true;
                         }
                     }
-                    return next != null;
+                    finished = true;
+                    return false;
                 }
 
                 @Override
                 public T getNext() {
-                    final T result = next;
-                    next = null;
-                    return result;
+                    cached = false;
+                    return next;
                 }
             };
         }
--- a/javaslang/src/test/java/javaslang/collection/AbstractTraversableTest.java
+++ b/javaslang/src/test/java/javaslang/collection/AbstractTraversableTest.java
@@ -489,8 +489,13 @@ public abstract class AbstractTraversableTest extends AbstractValueTest {
     }
 
     @Test
-    public void shouldDropWhileCorrect() {
+    public void shouldDropWhileAccordingToPredicate() {
         assertThat(of(1, 2, 3).dropWhile(i -> i < 2)).isEqualTo(of(2, 3));
+    }
+
+    @Test
+    public void shouldDropWhileAndNotTruncate() {
+        assertThat(of(1, 2, 3).dropWhile(i -> i % 2 == 1)).isEqualTo(of(2, 3));
     }
 
     // -- existsUnique
@@ -1536,7 +1541,7 @@ public abstract class AbstractTraversableTest extends AbstractValueTest {
         assertThat(actual).isEqualTo(expected);
     }
 
-    @Test // #201
+    @Test
     public void shouldSlideNonNilBySize2() {
         final List<Traversable<Integer>> actual = of(1, 2, 3, 4, 5).sliding(2).toList().map(Vector::ofAll);
         final List<Traversable<Integer>> expected = List.of(Vector.of(1, 2), Vector.of(2, 3), Vector.of(3, 4), Vector.of(4, 5));
@@ -1603,6 +1608,12 @@ public abstract class AbstractTraversableTest extends AbstractValueTest {
     @Test
     public void shouldSpanNonNil() {
         assertThat(of(0, 1, 2, 3).span(i -> i < 2)).isEqualTo(Tuple.of(of(0, 1), of(2, 3)));
+    }
+
+    @Test
+    public void shouldSpanAndNotTruncate() {
+        assertThat(of(1, 1, 2, 2, 3, 3).span(x -> x % 2 == 1)).isEqualTo(Tuple.of(of(1, 1), of(2, 2, 3, 3)));
+        assertThat(of(1, 1, 2, 2, 4, 4).span(x -> x == 1)).isEqualTo(Tuple.of(of(1, 1), of(2, 2, 4, 4)));
     }
 
     // -- spliterator
--- a/javaslang/src/test/java/javaslang/collection/CharSeqTest.java
+++ b/javaslang/src/test/java/javaslang/collection/CharSeqTest.java
@@ -1238,7 +1238,7 @@ public class CharSeqTest {
                 .isEqualTo(List.of(of('1'), of('2'), of('3')));
     }
 
-    @Test // #201
+    @Test
     public void shouldSlideNonNilBySize2() {
         assertThat(of('1', '2', '3', '4', '5').sliding(2).toList())
                 .isEqualTo(List.of(of('1', '2'), of('2', '3'), of('3', '4'), of('4', '5')));


javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1392,24 +1392,37 @@ public T getNext() {
             final Iterator<T> that = this;
             return new AbstractIterator<T>() {
 
-                private T next = null;
+                private T next;
+                private boolean cached = false;
+                private boolean test = true;
```
paplorinc: please rename `test` to reflect the intent better ...
`isFirstValue` or something?

javaslang/src/test/java/javaslang/collection/AbstractTraversableTest.java
```diff
@@ -489,10 +489,16 @@ public void shouldDropWhileAllIfPredicateIsTrue() {
     }
 
     @Test
-    public void shouldDropWhileCorrect() {
+    public void shouldDropWhileAccordingToPredicate() {
         assertThat(of(1, 2, 3).dropWhile(i -> i < 2)).isEqualTo(of(2, 3));
     }
 
+    // fixes #1641
```
paplorinc: Should we add these kinds of comments for new tests from now on?

danieldietrich: No, I will remove it. What counts is the current state of the application, it reflects the truth. The less organizational overhead, the better. Commits to a file are browsable. Also the full-text search function for issues and PRs is great. Thank you for asking!

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1392,24 +1392,37 @@ public T getNext() {
             final Iterator<T> that = this;
             return new AbstractIterator<T>() {
 
-                private T next = null;
+                private T next;
```
paplorinc: I think these can be simplified
```java
@Override
default Iterator<T> dropUntil(Predicate<? super T> predicate) {
    return find(predicate)
            .map(t -> of(t).concat(this))
            .getOrElse(empty());
}

@Override
default Iterator<T> dropWhile(Predicate<? super T> predicate) {
    return dropUntil(predicate.negate());
}
```

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1920,27 +1933,31 @@ public T getNext() {
...
```
paplorinc: all tests seem to be passing without these modifications also (`takeWhile`/`takeUntil`) :/

paplorinc:
@danieldietrich, the difficulty of implementing the `takeWhile` originates from the impotence of the `Iterator` interface.

It would be a trivial implementation, if the collections would implement an abstract `toStream` instead (where we shouldn't cache `next()` in `hasNext()`, as `head()` won't consume anything), and the `iterator()` would be implemented in `Value`, calling `toStream().iterator()`.

danieldietrich:
@paplorinc

> the difficulty of implementing the takeWhile originates from the impotence of the Iterator interface.

I know. The Iterator interface is stateful and it has to be. we need to take care of i_de_mpotence of hasNext

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: Will check it. had problems with `(1,1,2,2,1,1).span(i -> i == 1)`. unordered and sorted collections tests failed...
So I used `(1,1,2,2,3,3).span(i -> i % 2 == 1)` instead.
These tests should fail with unmodified Iterator class!

paplorinc:
> we need to take care of i _de_ mpotence of hasNext

I think we could avoid it by working with `Stream`s instead

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: From what I can see the `concat(this)` will not work. We need to concat the remaining elements _after_ t.

danieldietrich:
> I think we could avoid it by working with Streams instead

Not possible. Like in Scala, Iterator is the only TraversableOnce implementation. It is more memory efficient that Stream because it does not need additional objects per node (like the 'lazy' Supplier of Stream). A stateful Iterator has its use-case and its place in a mature collection library.

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: Right - forgot that we are mutably stateful. However, the other is more efficient.

danieldietrich: But having a simple solution would be also nice. I will think about it.

paplorinc: would that `==` the `empty()` instance? I think we're checking for reference equality in case of non-match

danieldietrich: > Feel free to benchmark it :)

Generally yes :) But I think I need to be the pragmatic developer in this case. Otherwise we will never see 2.1.0. My plan is to resolve 68 issues (tendency increasing because of new tickets). Every day an issue => two months.
Every 3 days an issue => half a year...

But yes, quality is very important!!

danieldietrich: > would that == the empty() instance?

Mmhh, you're right. We need to `empty()`!

danieldietrich: @paplorinc For now we follow the stateful design and low-level implementation of Scala's Iterator. See https://github.com/scala/scala/blob/2.12.x/src/library/scala/collection/Iterator.scala#L794-L823

paplorinc: > My plan is to resolve 68 issues (tendency increasing because of new tickets). 

I think our issues originate from more complicated code.

I have benchmarked the two alternatives like follows:
```java
public static class DropWhile extends Base {
    final Predicate<Integer> predicate = e -> e != ELEMENTS[CONTAINER_SIZE / 2];

    @Benchmark
    public int slang_persistent() {
        int length = Iterator.of(ELEMENTS).dropWhile(predicate).length();
        assert length == Array.of(ELEMENTS).dropWhile(predicate).length();
        return length;
    }
}
```
i.e. let's `drop` the iterator until the middle, and consume all resulting elements via `length`.

The results for `10`, `1_000` and `100_000` elements:
The explicit, long solution:
```java
Params          Score   Unit
    10  10,827,509.92  ops/s
  1000     144,264.98  ops/s
100000       1,333.39  ops/s
```

The reusing, short solution:
```java
Params          Score   Unit
    10   1,688,373.91  ops/s
  1000     191,893.42  ops/s
100000       2,065.07  ops/s
```

i.e. for very few elements, your solution is `10x` faster, after which the other one becomes faster.

danieldietrich: I think my solution can be made faster - I will try... (but currently busy with the Gradle online course)

danieldietrich:
I will merge the fix now. If we need to optimize s.th. we will do it in another PR.

zsolt-donca:
> It is more memory efficient than Stream because it does not need additional objects per node (like the 'lazy' Supplier of Stream).

It's not the only way to create a stream associated to a collection. Having a stream as a pair of head and a lazy tail is indeed useful when the stream is the result of a computation (e.g. generating fibonacci numbers or whatnot), where the best part is that the stream can be infinite.

However, if what you have is a data structure (e.g. a `List` or a `Vector`) with a well-defined structure and with a finite size, you can create much more optimal Stream implementations. For example, you can have a Stream implementation that directly references the underlying collection. In this case, implementing most (if not all) Stream-operations (e.g. `head`, `tail`, `map`, `filter`, `forEach` etc.) can be done in terms of executing the corresponding operation on the data structure and wrapping the result as a new Stream. In case of `Vector`, the resulting Stream would probably have the same performance characteristics (time complexity) as the current Stream (implemented in `ConsImpl`).

I'm not trying to question the justification for having low-level, mutable classes such as Iterator. They are indeed needed for performance-critical cases, or just for compatibility with Java. However, I don't think it's a good idea to have so much logic implemented in terms of `Iterator`, such as `dropWhile`; iterators are mutable, aren't referentially transparent: basically, they have all the problems that functional programming is trying to solve. 

I believe that implementing `Stream` based on `Iterator` for all collection types (as it is now in Javaslang) is the wrong way of doing it. Also, I don't believe that it's fundamentally impossible to write high-performance code with just using FP techniques. You just need to use the correct abstractions and get the design right.

As a general rule, one should try to keep the amount of mutable, side-effectful code minimal. I've always found it difficult to properly implement `Iterator` because it's defined in terms of `hasNext` and `next`; to check whether it has a next element, you usually need to fetch it, which you'll then need to store To return the next element, you might be able to use the element that is already fetched by `hasNext`; but it is also perfectly valid to call `next` multiple times without calling `hasNext` in between, so you also need to prepare to fetch the element directly in `next`. I've always found this a nightmare; since I know FP, I can also name it why (lack of referential transparency). 

Now, I find it pretty surprising that `javaslang.collection.Iterator` exists with such rich semantics, and that it actually extends `javaslang.collection.Traversable`. Since `Iterator` has _traversable once_ semantics, it breaks referential transparency in regards to the `Traversable` interface, which is at the core of Javaslang's collection hierarchy. This means that any code using the  `Traversable` type might not be referentially transparent, given it received an `Iterator` instance. What is the reason for `Iterator` extending `Traversable`?

danieldietrich:
Hi @zsolt-donca,

you are right, Iterator is only traversable once. We are aware of this issue (see #1295) but it will be a breaking change that is targeted for 3.0.0.

Algebraic types are beautiful, because they are simple. Stream is such a sum-type. Basically it has Cons and Empty. The algebraic properties let us use it in a for-comprehension, because it is a sum-type we are able to pattern-match it. We will not add internal magic like "if it is this or that then we will internally act like this or that". Algebraic data types do not work like that. What you describe is a universal data structure that is optimized for the underlying characteristics (e.g. finite, could also be indexed vs linear, etc). But that is a completely different approach than ADTs. It does not work for our collections.

Purely functional data-structures use foldRight to iterate over the elements - Erik Meyer's Haskell Course FP101 gives great insight. But Java is not purely functional, it allows us to mutate elements. Iterator is mutable by definition. When we get next() we cannot go back again. Its use-case is best described by the (future) type-hierachy: TraversableOnce. Stream highly depends on it (as you already described) because Iterator is used as element generator.

Again, we align to Scala. Javaslang brings the Scala goodness to Java, and that is on a purpose. It is not a good idea to ignore all the research and effort that was necessary to create Scala's collection library. It is a pool of wisdom. Scala 3.0.0 is in the make, there are already efforts to give the collections a refresh. We need to follow the discussion and to understand the main ideas behind the design. We should not make the mistake to try to be more clever than others and invent collections that are mostly 'engineered' from the low-level performance view-point.

I hope you understand that Iterator will stay part of Javaslang. Because of its mutable nature, most implementations will be mutable.

\- Daniel

danieldietrich:
> iterators are mutable, aren't referentially transparent: basically, they have all the problems that functional programming is trying to solve.

Btw - don't confuse it. Mutable state is not evil, _shared_ mutable state is evil!

PR #1650, opened 2016-10-30
Optimized `prependAll`, `appendAll` and `insertAll` for `Vector`

paplorinc:
Since @danieldietrich requested the original `Vector` to be simplified (good call, the code is a lot simpler now!), I've been thinking how to keep the performance AND maintainable code.

The solution was the optimization of the bulk operations instead, i.e. `appendAll` and `prependAll`.
Though they are difficult to benchmarks, I ended up measuring all reasonable combinations of recreating the original array by splitting it and recombining via `prependAll` and `appendAll`, therefore including all corner cases that could differ in speed.

I couldn't really compare it against Java (I can't recreate everything after a modification) or Scala (doesn't have bulk `prepend`, just `append`), so I compared it against the previous implementation:

``` java
Operation  Ratio                   10      100    1026
PrependAll slang_new/slang_old   3.63×  12.93×  17.64×
AppendAll  slang_new/slang_old   1.28×   8.86×  16.82×
Insert     slang_new/slang_old   0.48×   6.65×  18.87×
```
Note: I optimized `insertAll` also, by always iterating over the smaller collection.

---
The benchmarks are as follows:
boxed:
```java
Operation  Ratio                                                10      100     1026 
Create     slang_persistent/java_mutable                     1.68×    1.31×    1.28×
Create     slang_persistent/java_mutable_boxed               3.93×    5.28×    7.80×
Create     slang_persistent/java_mutable_boxed_stream        7.84×    7.27×    8.90×
Create     slang_persistent/fjava_persistent                67.37×  165.32×  197.55×
Create     slang_persistent/pcollections_persistent         27.73×  107.59×  202.16×
Create     slang_persistent/ecollections_persistent          2.24×    1.12×    0.88×
Create     slang_persistent/clojure_persistent               1.00×   11.33×   10.11×
Create     slang_persistent/scala_persistent                 7.13×    7.92×    7.70×

Head       slang_persistent/java_mutable                     0.66×    0.49×    0.39×
Head       slang_persistent/fjava_persistent                 0.83×    0.64×    0.51×
Head       slang_persistent/pcollections_persistent          1.70×    2.31×    3.16×
Head       slang_persistent/ecollections_persistent          0.62×    0.60×    0.49×
Head       slang_persistent/clojure_persistent               0.71×    0.83×    0.69×
Head       slang_persistent/scala_persistent                 0.79×    0.59×    0.49×

Tail       slang_persistent/java_mutable                     0.88×    0.48×    0.49×
Tail       slang_persistent/fjava_persistent                30.80×   44.14×   40.51×
Tail       slang_persistent/pcollections_persistent          4.49×    5.65×    7.78×
Tail       slang_persistent/ecollections_persistent         19.18×   19.57×  128.28×
Tail       slang_persistent/clojure_persistent               0.94×    0.66×    0.63×
Tail       slang_persistent/scala_persistent                 2.64×    3.49×    4.55×

Get        slang_persistent/java_mutable                     0.80×    0.41×    0.27×
Get        slang_persistent/fjava_persistent                28.58×   51.36×   60.27×
Get        slang_persistent/pcollections_persistent          4.56×    6.10×    8.81×
Get        slang_persistent/ecollections_persistent          1.30×    0.39×    0.22×
Get        slang_persistent/clojure_persistent               0.88×    1.48×    0.74×
Get        slang_persistent/scala_persistent                 1.09×    1.03×    0.71×

Update     slang_persistent/java_mutable                     0.18×    0.06×    0.04×
Update     slang_persistent/fjava_persistent                62.71×  147.10×  175.57×
Update     slang_persistent/pcollections_persistent          1.62×    2.21×    2.82×
Update     slang_persistent/ecollections_persistent          4.97×   22.35×  112.51×
Update     slang_persistent/clojure_persistent               0.75×    1.41×    0.89×
Update     slang_persistent/scala_persistent                 1.21×    1.07×    0.76×

Map        slang_persistent/java_mutable                     2.16×    2.37×    2.30×
Map        slang_persistent/java_mutable_loop                0.48×    0.51×    0.72×
Map        slang_persistent/ecollections_persistent          1.39×    0.96×    1.13×
Map        slang_persistent/scala_persistent                 1.21×    1.18×    1.43×

Filter     slang_persistent/java_mutable                     1.88×    1.94×    1.53×
Filter     slang_persistent/ecollections_persistent          1.04×    1.49×    1.06×
Filter     slang_persistent/scala_persistent                 1.47×    1.55×    1.38×

Prepend    slang_persistent/java_mutable                     0.11×    0.12×    0.23×
Prepend    slang_persistent/fjava_persistent                 0.39×    0.56×    0.51×
Prepend    slang_persistent/pcollections_persistent          0.22×    0.49×    0.84×
Prepend    slang_persistent/ecollections_persistent          0.38×    1.50×    9.03×
Prepend    slang_persistent/clojure_persistent               0.91×    4.70×   42.84×
Prepend    slang_persistent/scala_persistent                 0.09×    0.08×    0.08×

Append     slang_persistent/java_mutable                     0.04×    0.03×    0.02×
Append     slang_persistent/fjava_persistent                 0.65×    0.66×    0.57×
Append     slang_persistent/pcollections_persistent          0.29×    0.44×    0.70×
Append     slang_persistent/ecollections_persistent          0.47×    1.57×    9.45×
Append     slang_persistent/clojure_persistent               0.15×    0.10×    0.10×
Append     slang_persistent/scala_persistent                 0.15×    0.10×    0.08×

AppendAll  slang_persistent/scala_persistent                 2.79×    1.52×    1.67×

GroupBy    slang_persistent/java_mutable                     0.35×    0.53×    0.89×
GroupBy    slang_persistent/scala_persistent                 1.25×    1.17×    0.92×

Slice      slang_persistent/java_mutable                     0.63×    0.43×    0.45×
Slice      slang_persistent/pcollections_persistent          0.56×    0.40×    0.37×
Slice      slang_persistent/clojure_persistent               0.60×    0.43×    0.41×
Slice      slang_persistent/scala_persistent                 2.45×    3.13×    4.87×

Iterate    slang_persistent/java_mutable                     0.75×    0.30×    0.40×
Iterate    slang_persistent/fjava_persistent               191.76×  259.16×  396.44×
Iterate    slang_persistent/pcollections_persistent         13.86×    6.54×   13.70×
Iterate    slang_persistent/ecollections_persistent          1.14×    1.21×    1.68×
Iterate    slang_persistent/clojure_persistent               0.95×    0.78×    1.33×
Iterate    slang_persistent/scala_persistent                 2.03×    0.88×    1.43×
```

primitives
```java
Create     slang_persistent_int/java_mutable                 2.61×    1.29×    1.22×
Create     slang_persistent_int/java_mutable_boxed           6.12×    5.20×    7.41×
Create     slang_persistent_int/java_mutable_boxed_stream   12.19×    7.16×    8.46×
Create     slang_persistent_int/fjava_persistent           104.79×  162.83×  187.87×
Create     slang_persistent_int/pcollections_persistent     43.12×  105.97×  192.25×
Create     slang_persistent_int/ecollections_persistent      3.49×    1.10×    0.83×
Create     slang_persistent_int/clojure_persistent           1.55×   11.16×    9.61×
Create     slang_persistent_int/scala_persistent            11.09×    7.80×    7.32×

Tail       slang_persistent_int/java_mutable                 0.83×    0.48×    0.50×
Tail       slang_persistent_int/fjava_persistent            29.10×   44.56×   40.75×
Tail       slang_persistent_int/pcollections_persistent      4.24×    5.71×    7.82×
Tail       slang_persistent_int/ecollections_persistent     18.13×   19.76×  129.03×
Tail       slang_persistent_int/clojure_persistent           0.89×    0.67×    0.64×
Tail       slang_persistent_int/scala_persistent             2.49×    3.52×    4.58×

Update     slang_persistent_int/java_mutable                 0.20×    0.06×    0.04×
Update     slang_persistent_int/fjava_persistent            68.52×  150.03×  178.25×
Update     slang_persistent_int/pcollections_persistent      1.77×    2.26×    2.86×
Update     slang_persistent_int/ecollections_persistent      5.43×   22.80×  114.22×
Update     slang_persistent_int/clojure_persistent           0.82×    1.44×    0.91×
Update     slang_persistent_int/scala_persistent             1.32×    1.09×    0.77×

Map        slang_persistent_int/java_mutable                 2.11×    2.08×    1.79×
Map        slang_persistent_int/java_mutable_loop            0.47×    0.45×    0.56×
Map        slang_persistent_int/ecollections_persistent      1.36×    0.85×    0.88×
Map        slang_persistent_int/scala_persistent             1.18×    1.04×    1.11×

Filter     slang_persistent_int/java_mutable                 2.09×    2.28×    1.13×
Filter     slang_persistent_int/ecollections_persistent      1.15×    1.76×    0.78×
Filter     slang_persistent_int/scala_persistent             1.63×    1.82×    1.02×

Prepend    slang_persistent_int/java_mutable                 0.13×    0.15×    0.24×
Prepend    slang_persistent_int/fjava_persistent             0.46×    0.69×    0.53×
Prepend    slang_persistent_int/pcollections_persistent      0.25×    0.59×    0.87×
Prepend    slang_persistent_int/ecollections_persistent      0.45×    1.84×    9.33×
Prepend    slang_persistent_int/clojure_persistent           1.08×    5.74×   44.25×
Prepend    slang_persistent_int/scala_persistent             0.10×    0.10×    0.08×

Append     slang_persistent_int/java_mutable                 0.04×    0.03×    0.02×
Append     slang_persistent_int/fjava_persistent             0.57×    0.70×    0.64×
Append     slang_persistent_int/pcollections_persistent      0.25×    0.46×    0.79×
Append     slang_persistent_int/ecollections_persistent      0.42×    1.66×   10.68×
Append     slang_persistent_int/clojure_persistent           0.14×    0.10×    0.11×
Append     slang_persistent_int/scala_persistent             0.14×    0.11×    0.10×

Slice      slang_persistent_int/java_mutable                 0.62×    0.41×    0.48×
Slice      slang_persistent_int/pcollections_persistent      0.55×    0.39×    0.40×
Slice      slang_persistent_int/clojure_persistent           0.58×    0.42×    0.44×
Slice      slang_persistent_int/scala_persistent             2.40×    3.05×    5.18×

Iterate    slang_persistent_int/java_mutable                 1.01×    0.89×    0.79×
Iterate    slang_persistent_int/fjava_persistent           256.18×  777.01×  796.19×
Iterate    slang_persistent_int/pcollections_persistent     18.52×   19.60×   27.52×
Iterate    slang_persistent_int/ecollections_persistent      1.52×    3.63×    3.37×
Iterate    slang_persistent_int/clojure_persistent           1.27×    2.33×    2.66×
Iterate    slang_persistent_int/scala_persistent             2.71×    2.64×    2.87×
```

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -8,6 +8,7 @@
 import Generator._
 import JavaGenerator._
 
+import collection.immutable.ListMap
 import scala.language.implicitConversions
 
 val N = 8
@@ -2267,8 +2268,7 @@ def generateMainClasses(): Unit = {
     */
   def genArrayTypes(): Unit = {
 
-    val types = Map(
-      "Object" -> "Object",
+    val types = ListMap(
       "boolean" -> "Boolean",
       "byte" -> "Byte",
       "char" -> "Character",
@@ -2276,10 +2276,14 @@ def generateMainClasses(): Unit = {
       "float" -> "Float",
       "int" -> "Integer",
       "long" -> "Long",
-      "short" -> "Short"
+      "short" -> "Short",
+      "Object" -> "Object" // fallback
     ) // note: there is no void[] in Java
 
     genJavaslangFile("javaslang.collection", "ArrayType")((im: ImportManager, packageName: String, className: String) => xs"""
+      import javaslang.Tuple2;
+      import java.util.Collection;
+
       /**
        * Helper to replace reflective array access.
        *
@@ -2287,7 +2291,6 @@ def generateMainClasses(): Unit = {
        * @since 2.1.0
        */
       interface ArrayType<T> {
-
           @SuppressWarnings("unchecked")
           static <T> ArrayType<T> obj() { return (ArrayType<T>) ObjectArrayType.INSTANCE; }
 
@@ -2296,33 +2299,16 @@ def generateMainClasses(): Unit = {
           T getAt(Object array, int index);
 
           Object empty();
-          void setAt(Object array, int index, Object value);
+          void setAt(Object array, int index, T value) throws ClassCastException;
           Object copy(Object array, int arraySize, int sourceFrom, int destinationFrom, int size);
 
           @SuppressWarnings("unchecked")
           static <T> ArrayType<T> of(Object array) { return of((Class<T>) array.getClass().getComponentType()); }
           @SuppressWarnings("unchecked")
           static <T> ArrayType<T> of(Class<T> type) {
-              if (!type.isPrimitive()) {
-                  return (ArrayType<T>) obj();
-              } else if (boolean.class == type) {
-                  return (ArrayType<T>) BooleanArrayType.INSTANCE;
-              } else if (byte.class == type) {
-                  return (ArrayType<T>) ByteArrayType.INSTANCE;
-              } else if (char.class == type) {
-                  return (ArrayType<T>) CharArrayType.INSTANCE;
-              } else if (double.class == type) {
-                  return (ArrayType<T>) DoubleArrayType.INSTANCE;
-              } else if (float.class == type) {
-                  return (ArrayType<T>) FloatArrayType.INSTANCE;
-              } else if (int.class == type) {
-                  return (ArrayType<T>) IntArrayType.INSTANCE;
-              } else if (long.class == type) {
-                  return (ArrayType<T>) LongArrayType.INSTANCE;
-              } else if (short.class == type) {
-                  return (ArrayType<T>) ShortArrayType.INSTANCE;
-              } else {
-                  throw new IllegalArgumentException("Unknown type: " + type);
+              ${types.keys.toSeq.gen(arrayType =>
+                  s"""${if (arrayType == types.keys.last) "} else" else s"${if (arrayType == types.keys.head) "if" else "} else if"} ($arrayType.class == type)"} {
+                     |    return (ArrayType<T>) ${arrayType.capitalize + className}.INSTANCE;""".stripMargin)("\n")}
               }
           }
 
@@ -2391,41 +2377,29 @@ def generateMainClasses(): Unit = {
           }
 
           @SuppressWarnings("unchecked")
+          static <T> Object[] asArray(Iterable<? extends T> iterable) {
+              if (iterable instanceof Collection<?>) {
+                  final Collection<? extends T> collection = (Collection<? extends T>) iterable;
+                  return collection.toArray();
+              } else {
+                  final Tuple2<Iterable<? extends T>, Integer> iterableAndSize = Collections.withSize(iterable);
+                  return asArray(iterableAndSize._1.iterator(), iterableAndSize._2);
+              }
+          }
+
+          @SuppressWarnings("unchecked")
           static <T> T asPrimitives(Class<?> primitiveClass, Iterable<?> values) {
               final Object[] array = Array.ofAll(values).toJavaArray();
-              assert (array.length == 0) || (primitiveClass == primitiveType(array[0])) && !primitiveClass.isArray();
+              assert (array.length == 0) || !primitiveClass.isArray();
               final ArrayType<T> type = of((Class<T>) primitiveClass);
               final Object results = type.newInstance(array.length);
               for (int i = 0; i < array.length; i++) {
-                  type.setAt(results, i, array[i]);
+                  type.setAt(results, i, (T) array[i]);
               }
               return (T) results;
           }
 
-          static <T> Class<?> primitiveType(T element) {
-              final Class<?> wrapper = (element == null) ? Object.class : element.getClass();
-              if (wrapper == Boolean.class) {
-                  return boolean.class;
-              } else if (wrapper == Byte.class) {
-                  return byte.class;
-              } else if (wrapper == Character.class) {
-                  return char.class;
-              } else if (wrapper == Double.class) {
-                  return double.class;
-              } else if (wrapper == Float.class) {
-                  return float.class;
-              } else if (wrapper == Integer.class) {
-                  return int.class;
-              } else if (wrapper == Long.class) {
-                  return long.class;
-              } else if (wrapper == Short.class) {
-                  return short.class;
-              } else {
-                  return wrapper;
-              }
-          }
-
-          ${types.keys.toSeq.sorted.gen(arrayType =>
+          ${types.keys.toSeq.gen(arrayType =>
             genArrayType(arrayType)(im, packageName, arrayType.capitalize + className)
           )("\n\n")}
       }
@@ -2433,7 +2407,7 @@ def generateMainClasses(): Unit = {
 
     def genArrayType(arrayType: String)(im: ImportManager, packageName: String, className: String): String = {
       val wrapperType = types(arrayType)
-      val cast = if (wrapperType != "Object") s" ($wrapperType)" else ""
+      val isPrimitive = arrayType != "Object"
 
       xs"""
         final class $className implements ArrayType<$wrapperType>, ${im.getType("java.io.Serializable")} {
@@ -2456,7 +2430,15 @@ def generateMainClasses(): Unit = {
             public $wrapperType getAt(Object array, int index) { return cast(array)[index]; }
 
             @Override
-            public void setAt(Object array, int index, Object value) { cast(array)[index] =$cast value; }
+            public void setAt(Object array, int index, $wrapperType value) ${if (isPrimitive) "throws ClassCastException " else ""}{
+                ${if (isPrimitive)
+                """if (value == null) {
+                  |    throw new ClassCastException();
+                  |} else {
+                  |    cast(array)[index] = value;
+                  |}""".stripMargin
+              else "cast(array)[index] = value;" }
+            }
 
             @Override
             public Object copy(Object array, int arraySize, int sourceFrom, int destinationFrom, int size) {
--- a/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
+++ b/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
@@ -11,9 +11,12 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 
 import static java.util.function.Function.identity;
+import static javaslang.API.Vector;
 import static javaslang.collection.ArrayType.obj;
-import static javaslang.collection.ArrayType.primitiveType;
-import static javaslang.collection.NodeModifier.*;
+import static javaslang.collection.Collections.isTraversableAgain;
+import static javaslang.collection.Collections.reverseIterator;
+import static javaslang.collection.NodeModifier.COPY_NODE;
+import static javaslang.collection.NodeModifier.IDENTITY;
 
 /**
  * A `bit-mapped trie` is a very wide and shallow tree (for integer indices the depth will be `≤6`).
@@ -55,6 +58,7 @@ final class BitMappedTrie<T> implements Serializable {
         this.length = length;
         this.depthShift = depthShift;
 
+        assert offset >= 0;
         assert length <= treeSize(BRANCHING_FACTOR, depthShift);
         assert (EMPTY == null) || ((length > 0) && (ArrayType.of(array).lengthOf(array) > 0));
     }
@@ -81,83 +85,118 @@ final class BitMappedTrie<T> implements Serializable {
 
     private BitMappedTrie<T> boxed() { return map(identity()); }
 
-    BitMappedTrie<T> prepend(T leading) {
-        if (cannotAdd(leading)) {
-            return boxed().prepend(leading);
-        } else {
-            final int newSize = length() + 1;
-            if (length() == 0) {
-                return new BitMappedTrie<>(type, type.asArray(leading), offset, newSize, depthShift);
-            } else {
-                Object array = this.array;
-                int shift = depthShift, offset = this.offset;
-                if (isFullLeft()) {
-                    array = obj().copyUpdate(obj().empty(), BRANCHING_FACTOR - 1, array);
-                    shift += BRANCHING_BASE;
-                    offset = treeSize(BRANCHING_FACTOR - 1, shift);
-                }
-
-                offset -= 1;
-                array = modifyLeaf(array, shift, offset, COPY_NODE, updateLeafWith(type, leading));
-                return new BitMappedTrie<>(type, array, offset, newSize, shift);
-            }
+    BitMappedTrie<T> prependAll(Iterable<? extends T> iterable) {
+        if (!isTraversableAgain(iterable)) {
+            iterable = Vector(iterable);
         }
+
+        try {
+            return prepend(reverseIterator(iterable));
+        } catch (ClassCastException ignored) {
+            return boxed().prepend(reverseIterator(iterable));
+        }
+    }
+    private BitMappedTrie<T> prepend(java.util.Iterator<? extends T> iterator) {
+        BitMappedTrie<T> result = this;
+        while (iterator.hasNext()) {
+            Object array = result.array;
+            int shift = result.depthShift, offset = result.offset;
+            if (result.isFullLeft()) {
+                array = obj().copyUpdate(obj().empty(), BRANCHING_FACTOR - 1, array);
+                shift += BRANCHING_BASE;
+                offset = treeSize(BRANCHING_FACTOR - 1, shift);
+            }
+
+            final MutableInt delta = new MutableInt(0);
+            array = result.modifyLeaf(array, shift, offset - 1, COPY_NODE, prependToLeaf(iterator, delta));
+            result = new BitMappedTrie<>(type, array, offset - delta.val, result.length + delta.val, shift);
+        }
+        return result;
     }
     private boolean isFullLeft() { return offset == 0; }
-
-    BitMappedTrie<T> append(T trailing) {
-        if (cannotAdd(trailing)) {
-            return boxed().append(trailing);
-        } else {
-            final int newSize = length() + 1;
-            if (length() == 0) {
-                return new BitMappedTrie<>(type, type.asArray(trailing), offset, newSize, depthShift);
-            } else {
-                Object array = this.array;
-                int shift = depthShift;
-                if (isFullRight(newSize)) {
-                    array = obj().asArray(array);
-                    shift += BRANCHING_BASE;
-                }
-
-                array = modifyLeaf(array, shift, offset + length(), COPY_NODE, updateLeafWith(type, trailing));
-                return new BitMappedTrie<>(type, array, offset, newSize, shift);
+    private NodeModifier prependToLeaf(java.util.Iterator<? extends T> iterator, MutableInt delta) {
+        return (array, index) -> {
+            final Object copy = type.copy(array, BRANCHING_FACTOR);
+            for (; iterator.hasNext() && index >= 0; index--) {
+                type.setAt(copy, index, iterator.next());
+                delta.val++;
             }
+            return copy;
+        };
+    }
+
+    BitMappedTrie<T> appendAll(Iterable<? extends T> iterable) {
+        if (!isTraversableAgain(iterable)) {
+            iterable = Vector(iterable);
+        }
+
+        try {
+            return append(iterable.iterator());
+        } catch (ClassCastException ignored) {
+            return boxed().append(iterable.iterator());
         }
     }
-    private boolean isFullRight(int newSize) { return (offset + newSize) > treeSize(BRANCHING_FACTOR, depthShift); }
+    private BitMappedTrie<T> append(java.util.Iterator<? extends T> iterator) {
+        BitMappedTrie<T> result = this;
+        while (iterator.hasNext()) {
+            Object array = result.array;
+            int shift = result.depthShift;
+            if (result.isFullRight()) {
+                array = obj().asArray(array);
+                shift += BRANCHING_BASE;
+            }
+
+            final MutableInt delta = new MutableInt(0);
+            array = result.modifyLeaf(array, shift, result.offset + result.length, COPY_NODE, appendToLeaf(iterator, delta));
+            result = new BitMappedTrie<>(type, array, result.offset, result.length + delta.val, shift);
+        }
+        return result;
+
+    }
+    private boolean isFullRight() { return (offset + length + 1) > treeSize(BRANCHING_FACTOR, depthShift); }
+    private NodeModifier appendToLeaf(java.util.Iterator<? extends T> iterator, MutableInt delta) {
+        return (array, index) -> {
+            final Object copy = type.copy(array, BRANCHING_FACTOR);
+            for (int length = type.lengthOf(copy); iterator.hasNext() && index < length; index++) {
+                type.setAt(copy, index, iterator.next());
+                delta.val++;
+            }
+            return copy;
+        };
+    }
 
     BitMappedTrie<T> update(int index, T element) {
-        if (cannotAdd(element)) {
-            return boxed().update(index, element);
-        } else {
+        try {
             final Object root = modifyLeaf(array, depthShift, offset + index, COPY_NODE, updateLeafWith(type, element));
-            return new BitMappedTrie<>(type, root, offset, length(), depthShift);
+            return new BitMappedTrie<>(type, root, offset, length, depthShift);
+        } catch (ClassCastException ignored) {
+            return boxed().update(index, element);
         }
     }
+    private NodeModifier updateLeafWith(ArrayType<T> type, T element) { return (a, i) -> type.copyUpdate(a, i, element); }
 
     BitMappedTrie<T> drop(int n) {
         if (n <= 0) {
             return this;
-        } else if (n >= length()) {
+        } else if (n >= length) {
             return empty();
         } else {
             final int index = offset + n;
             final Object root = arePointingToSameLeaf(0, n)
                                 ? array
                                 : modifyLeaf(array, depthShift, index, obj()::copyDrop, IDENTITY);
-            return collapsed(type, root, index, length() - n, depthShift);
+            return collapsed(type, root, index, length - n, depthShift);
         }
     }
 
     BitMappedTrie<T> take(int n) {
-        if (n >= length()) {
+        if (n >= length) {
             return this;
         } else if (n <= 0) {
             return empty();
         } else {
             final int index = n - 1;
-            final Object root = arePointingToSameLeaf(index, length() - 1)
+            final Object root = arePointingToSameLeaf(index, length - 1)
                                 ? array
                                 : modifyLeaf(array, depthShift, offset + index, obj()::copyTake, IDENTITY);
             return collapsed(type, root, offset, n, depthShift);
@@ -203,7 +242,8 @@ final class BitMappedTrie<T> implements Serializable {
                 array = newNode;
             }
 
-            obj().setAt(array, previousIndex, leaf.apply(obj().getAt(array, previousIndex), lastDigit(index)));
+            final Object newLeaf = leaf.apply(obj().getAt(array, previousIndex), lastDigit(index));
+            obj().setAt(array, previousIndex, newLeaf);
             return root;
         }
     }
@@ -236,7 +276,7 @@ final class BitMappedTrie<T> implements Serializable {
 
     Iterator<T> iterator() {
         return new Iterator<T>() {
-            private final int globalLength = BitMappedTrie.this.length();
+            private final int globalLength = BitMappedTrie.this.length;
             private int globalIndex = 0;
 
             private int index = lastDigit(offset);
@@ -296,7 +336,7 @@ final class BitMappedTrie<T> implements Serializable {
             return index;
         });
 
-        return (length() == length)
+        return (this.length == length)
                ? this
                : BitMappedTrie.ofAll(type.copyRange(results, 0, length));
     }
@@ -315,15 +355,13 @@ final class BitMappedTrie<T> implements Serializable {
         return BitMappedTrie.ofAll(results);
     }
 
-    int length()                         { return length; }
-    private boolean cannotAdd(T element) { return (type != obj()) && (type.type() != primitiveType(element)); }
+    int length() { return length; }
 }
 
 @FunctionalInterface
 interface NodeModifier {
     Object apply(Object array, int index);
 
-    static <T> NodeModifier updateLeafWith(ArrayType<T> type, T element) { return (a, i) -> type.copyUpdate(a, i, element); }
     NodeModifier COPY_NODE = (o, i) -> obj().copy(o, i + 1);
     NodeModifier IDENTITY = (o, i) -> o;
 }
@@ -331,4 +369,9 @@ interface NodeModifier {
 @FunctionalInterface
 interface LeafVisitor<T> {
     int visit(int index, T leaf, int start, int end);
+}
+
+class MutableInt {
+    int val;
+    MutableInt(int val) { this.val = val; }
 }
--- a/javaslang/src/main/java/javaslang/collection/Collections.java
+++ b/javaslang/src/main/java/javaslang/collection/Collections.java
@@ -5,6 +5,7 @@
  */
 package javaslang.collection;
 
+import javaslang.Tuple2;
 import javaslang.control.Option;
 
 import java.util.ArrayList;
@@ -14,6 +15,9 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
+
+import static javaslang.API.List;
+import static javaslang.API.Tuple;
 
 /**
  * Internal class, containing helpers.
@@ -123,7 +127,7 @@ final class Collections {
                                                                                U zero, BiFunction<? super T, ? super U, ? extends U> operation,
                                                                                C cumulativeResult, BiFunction<C, U, C> combiner,
                                                                                Function<C, R> finisher) {
-        final Iterator<? extends T> reversedElements = seq(elements).reverseIterator(); // TODO a List will be reversed too many times this way
+        final Iterator<? extends T> reversedElements = reverseIterator(elements);
         return scanLeft(reversedElements, zero, (u, t) -> operation.apply(t, u), cumulativeResult, combiner, finisher);
     }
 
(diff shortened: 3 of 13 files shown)


javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -43,12 +48,12 @@
 
     public static void main(String... args) {
         JmhRunner.runDebugWithAsserts(CLASSES);
-        JmhRunner.runNormalNoAsserts(CLASSES, JAVA, FUNCTIONAL_JAVA, PCOLLECTIONS, ECOLLECTIONS, CLOJURE, SCALA, JAVASLANG);
+        JmhRunner.runNormalNoAsserts(CLASSES, /*FUNCTIONAL_JAVA, PCOLLECTIONS, ECOLLECTIONS,*/ JAVA, CLOJURE, SCALA, JAVASLANG);
```
paplorinc: disabled FUNCTIONAL_JAVA, PCOLLECTIONS and ECOLLECTIONS by default

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -43,12 +48,12 @@
...
     }
 
     @State(Scope.Benchmark)
     public static class Base {
-        @Param({ "32", "1024", "32768" /*, "1048576", "33554432", "1073741824" */ }) /* i.e. depth 1,2,3(,4,5,6) for a branching factor of 32 */
+        @Param({ "100", "10000" })
```
paplorinc: reduced the benchmarks parameters to two to speed up builds

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -590,6 +591,19 @@ public Object slang_persistent() {
         }
     }
 
+    public static class PrependAll extends Base {
```
paplorinc: can't compare it against anything :/

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -665,21 +679,86 @@ public Object slang_persistent() {
...
+        final CanBuildFrom canBuildFrom = scala.collection.immutable.Vector.canBuildFrom();
+
         @Benchmark
-        public Object java_mutable() {
-            return javaMutable.stream().collect(Collectors.groupingBy(Integer::bitCount));
+        public void scala_persistent(Blackhole bh) {
+            for (int i = 0; i < CONTAINER_SIZE; i++) {
+                final scala.collection.immutable.Vector<Integer> front = scalaPersistent.slice(0, i);
+                final java.util.List<Integer> back = javaMutable.subList(i, CONTAINER_SIZE);
+                final scala.collection.immutable.Vector<Integer> values = front.$plus$plus(asScalaBuffer(back), (CanBuildFrom<scala.collection.immutable.Vector<Integer>, Integer, scala.collection.immutable.Vector<Integer>>) canBuildFrom);
```
paplorinc: ... Scala is really not meant to be called from Java ...

javaslang-benchmark/src/test/java/javaslang/collection/VectorBenchmark.java
```diff
@@ -737,26 +816,10 @@ public void slang_persistent(Blackhole bh) {
 
     /** Sequential access for all elements */
     public static class Iterate extends Base {
-        @State(Scope.Thread)
```
paplorinc: no need to recreate it in `Iterate` as we don't change the `javaMutable` in any way :)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -73,71 +76,91 @@ private static int treeSize(int branchCount, int depthShift) {
...
 
-    BitMappedTrie<T> prepend(T leading) {
-        final int newSize = length() + 1;
-        if (length() == 0) {
-            return new BitMappedTrie<>(asArray(leading), offset, newSize, depthShift);
-        } else {
-            Object[] array = this.array;
-            int shift = depthShift, offset = this.offset;
-            if (isFullLeft()) {
+    BitMappedTrie<T> prependAll(Iterable<? extends T> iterable) {
```
paplorinc: generalized `prepend` to reuse all newly inserted arrays (minimizing the copies)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -73,71 +76,91 @@ private static int treeSize(int branchCount, int depthShift) {
...
+            if (result.isFullLeft()) {
                 array = copyUpdate(Arrays.empty(), BRANCHING_FACTOR - 1, array);
                 shift += BRANCHING_BASE;
                 offset = treeSize(BRANCHING_FACTOR - 1, shift);
             }
 
-            offset -= 1;
-            array = modifyLeaf(array, shift, offset, COPY_NODE, updateLeafWith(leading));
-            return new BitMappedTrie<>(array, offset, newSize, shift);
+            final MutableInt delta = new MutableInt(0);
```
paplorinc: `MutableInt` is more elegant than a single element `int[]`.
We need this for multiple return (without using `Tuples` and boxing)

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -73,71 +76,91 @@ private static int treeSize(int branchCount, int depthShift) {
...
 
-    BitMappedTrie<T> append(T trailing) {
-        final int newSize = length() + 1;
-        if (length() == 0) {
-            return new BitMappedTrie<>(asArray(trailing), offset, newSize, depthShift);
-        } else {
-            Object[] array = this.array;
-            int shift = depthShift;
-            if (isFullRight(newSize)) {
+    BitMappedTrie<T> appendAll(Iterable<? extends T> iterable) {
```
paplorinc: same with `append`

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -73,71 +76,91 @@ private static int treeSize(int branchCount, int depthShift) {
...
+            }
+            return copy;
+        };
     }
-    private boolean isFullRight(int newSize) { return (offset + newSize) > treeSize(BRANCHING_FACTOR, depthShift); }
 
     BitMappedTrie<T> update(int index, T element) {
         final Object[] root = modifyLeaf(array, depthShift, offset + index, COPY_NODE, updateLeafWith(element));
-        return new BitMappedTrie<>(root, offset, length(), depthShift);
+        return new BitMappedTrie<>(root, offset, length, depthShift);
```
paplorinc: @danieldietrich requested previously that I call `length` via the field, not the method.
Now this class is encapsulated enough that I applied it :)

javaslang/src/main/java/javaslang/collection/Collections.java
```diff
@@ -123,7 +123,7 @@ static int hash(Iterable<?> iterable) {
                                                                                U zero, BiFunction<? super T, ? super U, ? extends U> operation,
                                                                                C cumulativeResult, BiFunction<C, U, C> combiner,
                                                                                Function<C, R> finisher) {
-        final Iterator<? extends T> reversedElements = seq(elements).reverseIterator(); // TODO a List will be reversed too many times this way
+        final Iterator<? extends T> reversedElements = reverseIterator(elements);
```
paplorinc: this change should make `scanRight` faster also :)

javaslang/src/main/java/javaslang/collection/Collections.java
```diff
@@ -191,7 +191,25 @@ protected T getNext() {
         if (iterable instanceof Seq) {
             return (Seq<T>) iterable;
         } else {
-            return List.ofAll(iterable);
+            return Vector.ofAll(iterable);
```
paplorinc: changed `List` to `Vector` for efficient `reverseIterator`

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -165,8 +164,8 @@ private Vector(BitMappedTrie<T> trie) {
             final Object[] array = ((Collection<? extends T>) iterable).toArray();
             return ofAll(BitMappedTrie.ofAll(array));
         } else {
-            final Seq<? extends T> elems = seq(iterable);
-            final Object[] array = asArray(elems.iterator(), elems.size());
+            final Seq<? extends T> result = (iterable instanceof Seq) ? (Seq<T>) iterable : List.ofAll(iterable);
+            final Object[] array = asArray(result.iterator(), result.size());
```
danieldietrich: We create a Seq to get the size. That's a waste (of mem and cpu/gc) because several new objects are created. Instead I would create two times an iterator(), the first for counting, the second for getting the elements:

```java
int size = 0;
for (Iterator<?> iter = iterable.iterator(); iter.hasNext(); iter.next()) { size += 1; }
final Object[] array = asArray(iterable.iterator(), size);
```

Additionally we could also do an extra check for `Traversable`:

```java
} else if (iterable instanceof Traversable<?>) {
    final Traverable<T> traversable = (Traversable<T>) iterable;
    final Object[] array = asArray(traversable.iterator(), traversable.size());
    return ofAll(BitMappedTrie.ofAll(array));
} ...
```

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -569,22 +568,14 @@ private Vector(BitMappedTrie<T> trie) {
...
-            for (T element : elements) {
-                result = result.append(element);
-            }
-            return result.isEmpty() ? empty()
-                                    : result;
-        }
+    public Vector<T> appendAll(Iterable<? extends T> iterable) {
+        Objects.requireNonNull(iterable, "iterable is null");
+        return isEmpty() ? ofAll(iterable)
+                         : wrap(trie.appendAll(iterable));
```
danieldietrich: good!

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -736,9 +727,11 @@ public int indexOf(T element, int from) {
         if ((index < 0) || (index > length())) {
             throw new IndexOutOfBoundsException("insert(" + index + ", e) on Vector of length " + length());
         } else {
-            final Vector<T> begin = take(index);
+            final Vector<T> begin = take(index).appendAll(elements);
             final Vector<T> end = drop(index);
-            return begin.appendAll(elements).appendAll(end);
+            return (begin.size() > end.size())
+                   ? begin.appendAll(end)
+                   : end.prependAll(begin);
```
danieldietrich: good!

javaslang/src/main/java/javaslang/collection/Collections.java
```diff
@@ -191,7 +191,25 @@ protected T getNext() {
...
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    static <T> Iterator<? extends T> reverseIterator(Iterable<? extends T> iterable) {
+        if (iterable instanceof java.util.List) {
+            java.util.List<T> list = (java.util.List<T>) iterable;
+            return new Iterator<T>() {
+                java.util.ListIterator<T> delegate = list.listIterator(list.size());
```
danieldietrich: good!

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -277,7 +300,7 @@ private void setCurrentArray() {
             return index;
         });
 
-        return (length() == length)
+        return (this.length == length)
```
danieldietrich: +1

javaslang/src/main/java/javaslang/collection/Vector.java
paplorinc: I thought about it, but if the `iterable` is `TraversableOnce`, I would consume it for the `size`, right?

danieldietrich: Right - but do it. The problem of consumption also occurs in other places, e.g. when an TraversableOnce is used in a For loop etc. It is a design-error of the type hierarchy that will be fixed with 3.0.

paplorinc: Fixed it in a separate commit, thanks :)

javaslang/src/main/java/javaslang/collection/Arrays.java
```diff
@@ -35,6 +39,19 @@
         return array;
     }
 
+    @SuppressWarnings("unchecked")
+    static <T> Object[] asArray(Iterable<? extends T> iterable) {
+        if (iterable instanceof Collection<?>) {
+            final Collection<? extends T> collection = (Collection<? extends T>) iterable;
+            return collection.toArray();
+        } else if (iterable instanceof Traversable) {
+            final Traversable<T> traversable = asTraversableAgain(iterable);
```
danieldietrich: please do a one-liner like in reverseIterator() and delete the asTraversableAgain method completely:

```java
} else if (iterable instanceof Traversable && !(iterable instanceof Iterator)) { // TODO: remove second check with 3.0.0
    final Traversable<T> traversable = (Trabersable<T>) iterable;
```

javaslang/src/main/java/javaslang/collection/Collections.java
```diff
@@ -187,12 +189,14 @@ protected T getNext() {
     }
 
     @SuppressWarnings("unchecked")
-    static <T> Seq<T> seq(Iterable<? extends T> iterable) {
-        if (iterable instanceof Seq) {
-            return (Seq<T>) iterable;
-        } else {
-            return Vector.ofAll(iterable);
+    static <T> Traversable<T> asTraversableAgain(Iterable<? extends T> iterable) {
```
danieldietrich: method to be deleted (see comment above)

javaslang/src/test/java/javaslang/collection/euler/Euler67Test.java
```diff
@@ -58,7 +59,7 @@ private static int solve(String fileName) {
         return Vector.ofAll(
                 Utils.readLines(Utils.file(fileName)).map(line ->
                         Arrays.stream(line.split("\\s")).map(Integer::parseInt)
-                ).map(s -> Vector.ofAll(s::iterator))
+                ).map(s -> List.ofAll(s::iterator).toVector())
```
danieldietrich: Can be reverted again.

javaslang/src/main/java/javaslang/collection/Arrays.java
paplorinc: There are still tests that are traversableonce ... Are you suggesting that I simply ignore it (and delete the tests that rely on it?)

javaslang/src/test/java/javaslang/collection/euler/Euler67Test.java
paplorinc: can it? :/

javaslang/src/main/java/javaslang/collection/Arrays.java
danieldietrich: Please point me to (one of) the tests. Iterator should be the only TraversableOnce (in our code-base).

paplorinc: It is, but where do we convert it to a `TraversableAgain`, if you want to delete the `asTraversableAgain` method? If I just igore those, I need to fix the tests, as I've done for `Euler67Test`

javaslang/src/test/java/javaslang/collection/euler/Euler67Test.java
danieldietrich: @paploric you are right - please keep the asTraversableAgain method then but please write a TODO that we remove it with 3.0.0. I think the test can be reverted then, right?

javaslang/src/main/java/javaslang/collection/Arrays.java
paplorinc: The `iterable` that's returned by the test is not an instance of `Iterator`, it's some anonymous lambda.
I don't think we can consume it twice, I recommend reverting it to
```java
if (iterable instanceof Collection<?>) {
    final Collection<? extends T> collection = (Collection<? extends T>) iterable;
    return collection.toArray();
} else {
    final Traversable<T> traversable = asTraversableAgain(iterable);
    return asArray(traversable.iterator(), traversable.size());
}
```

javaslang/src/test/java/javaslang/collection/euler/Euler67Test.java
paplorinc: I have reverted the test, but I needed to wrap the general `Iterable` to a `List`, otherwise I might not be able to go over it multiple times (e.g. it might be a `Stream`, that won't return the same `iterator` multiple times)

danieldietrich:
Right decision - use List then!

paplorinc:
@danieldietrich, rebased :)

danieldietrich:
Thank you! Need some sleep - will come back to this issue tomorrow!

javaslang/generator/Generator.scala
```diff
@@ -2284,27 +2285,29 @@ def generateMainClasses(): Unit = {
     */
   def genArrayTypes(): Unit = {
 
-    val types = Map(
-      "Object" -> "Object",
+    val types = ListMap(
```
paplorinc: the order matters, i.e. made "Object" the fallback

javaslang/generator/Generator.scala
```diff
@@ -2313,33 +2316,16 @@ def generateMainClasses(): Unit = {
           T getAt(Object array, int index);
 
           Object empty();
-          void setAt(Object array, int index, Object value);
+          void setAt(Object array, int index, T value) throws ClassCastException;
```
paplorinc: made it more typesafe

javaslang/generator/Generator.scala
```diff
@@ -2313,33 +2316,16 @@ def generateMainClasses(): Unit = {
...
-                  return (ArrayType<T>) FloatArrayType.INSTANCE;
-              } else if (int.class == type) {
-                  return (ArrayType<T>) IntArrayType.INSTANCE;
-              } else if (long.class == type) {
-                  return (ArrayType<T>) LongArrayType.INSTANCE;
-              } else if (short.class == type) {
-                  return (ArrayType<T>) ShortArrayType.INSTANCE;
-              } else {
-                  throw new IllegalArgumentException("Unknown type: " + type);
+              ${types.keys.toSeq.gen(arrayType =>
```
paplorinc: let's generate these also

javaslang/generator/Generator.scala
```diff
@@ -2408,49 +2394,37 @@ def generateMainClasses(): Unit = {
...
-                  return long.class;
-              } else if (wrapper == Short.class) {
-                  return short.class;
-              } else {
-                  return wrapper;
-              }
-          }
-
-          ${types.keys.toSeq.sorted.gen(arrayType =>
+          ${types.keys.toSeq.gen(arrayType =>
```
paplorinc: managed to get rid of this by relying on the runtime to throw a `ClassCastException`, instead of checking every insertion for correctness (i.e. a `null` or `String` value in a primitive array)

javaslang/generator/Generator.scala
```diff
@@ -2473,7 +2447,15 @@ def generateMainClasses(): Unit = {
             public $wrapperType getAt(Object array, int index) { return cast(array)[index]; }
 
             @Override
-            public void setAt(Object array, int index, Object value) { cast(array)[index] =$cast value; }
+            public void setAt(Object array, int index, $wrapperType value) ${if (isPrimitive) "throws ClassCastException " else ""}{
+                ${if (isPrimitive)
+                """if (value == null) {
+                  |    throw new ClassCastException();
```
paplorinc: this would result in a `NPE` instead ...

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -81,83 +85,118 @@ private static int treeSize(int branchCount, int depthShift) {
...
-                Object array = this.array;
-                int shift = depthShift, offset = this.offset;
-                if (isFullLeft()) {
-                    array = obj().copyUpdate(obj().empty(), BRANCHING_FACTOR - 1, array);
-                    shift += BRANCHING_BASE;
-                    offset = treeSize(BRANCHING_FACTOR - 1, shift);
-                }
+    BitMappedTrie<T> prependAll(Iterable<? extends T> iterable) {
+        if (!isTraversableAgain(iterable)) {
+            iterable = Vector(iterable);
```
paplorinc: I tried my best, but if we want optimistic modification, this is our best shot, I think.
Made it traversable-again so that we can start from scratch if we can't add the element (will get a `ClassCastException`); box it and try again.

javaslang/src/main/java/javaslang/collection/Collections.java
```diff
@@ -186,12 +190,45 @@ protected T getNext() {
         return tabulate(n, ignored -> supplier.get());
     }
 
+    static <T> Tuple2<Iterable<? extends T>, Integer> withSize(Iterable<? extends T> iterable) {
```
paplorinc: created all these helpers to bridge the gap between the Java and Javaslang collections.

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -569,23 +563,13 @@ private Vector(BitMappedTrie<T> trie) {
     }
 
     @Override
-    public Vector<T> append(T element) {
-        return wrap(trie.append(element));
-    }
+    public Vector<T> append(T element) { return appendAll(Iterator.of(element)); }
```
paplorinc: a single `append` is almost `3x` slower than before, but the bulk operations (`*all`) are `~17x` faster than before, making it faster than `Scala`

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -616,13 +600,7 @@ private Vector(BitMappedTrie<T> trie) {
...
     @Override
     public Vector<T> drop(int n) {
-        if (n <= 0) {
-            return this;
-        } else if (n >= length()) {
-            return empty();
-        } else {
-            return wrap(trie.drop(n));
-        }
+        return wrap(trie.drop(n));
```
paplorinc: could finally get rid of these, increasing coverage :)

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -30,52 +30,62 @@ public void shouldCreateAndGet() {
...
             }
 
+            assert (i == 0) || !actual.trie.type.type().isPrimitive();
+
             /* boolean */
             final Seq<Boolean> expectedBoolean = expected.map(v -> v > 0);
             final Vector<Boolean> actualBoolean = Vector.ofAll(ArrayType.<boolean[]> asPrimitives(boolean.class, expectedBoolean));
             assert (i == 0) || (actualBoolean.trie.type.type() == boolean.class);
             assertAreEqual(expectedBoolean, actualBoolean);
+            assertAreEqual(expectedBoolean.append(null), actualBoolean.append(null));
```
paplorinc: needed for 100% coverage :)

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -311,8 +342,16 @@ public void shouldBehaveLikeArray() {
...
     }
 
-    private int percent(Random random) { return random.nextInt(100); }
-
+    private int percent(Random random) { return random.nextInt(101); }
+    private Iterable<Object> randomValues(Random random, int count) {
+        final Vector<Object> values = Vector.range(0, count).map(v -> randomValue(random));
+        final int percent = percent(random);
+        if (percent < 30) {
+            return values.toJavaList();  /* not Traversable */
```
paplorinc: test adding from Java collections also

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -219,55 +229,76 @@ public void shouldBehaveLikeArray() {
...
-                    history = history.append(Tuple.of(expected, actual));
+                    history = history.append(Tuple(expected, actual));
+                }
+
+                if (percent(random) < 10) {
+                    final int index = random.nextInt(expected.size() + 1);
+                    Iterable<Object> values = randomValues(random, random.nextInt(2 * BRANCHING_FACTOR));
+                    expected = expected.insertAll(index, values);
+
+                    values = (percent(random) < 50) ? Iterator.ofAll(values) : values;  /* not traversable again */
```
paplorinc: test adding from non-traversable-again aso

javaslang/src/test/java/javaslang/collection/VectorPropertyTest.java
```diff
@@ -219,55 +229,76 @@ public void shouldBehaveLikeArray() {
...
                 if (percent(random) < 20) {
-                    final ArrayList<Object> values = new ArrayList<>();
-                    for (int k = 0; k < random.nextInt(j + 1); k++) {
-                        values.add(random.nextInt());
-                    }
-                    expected = Array.ofAll(values);
-                    final boolean isPrimitive = percent(random) < 30;
-                    actual = isPrimitive ? Vector.narrow(Vector.ofAll(ArrayType.<int[]> asPrimitives(int.class, values))) : Vector.ofAll(values);
+                    expected = Array.ofAll(Vector(randomValues(random, 100)).filter(v -> v instanceof Integer));
+                    actual = (percent(random) < 30) ? Vector.narrow(Vector.ofAll(ArrayType.<int[]> asPrimitives(int.class, expected))) : Vector.ofAll(expected);
```
paplorinc: convert everything to a primitive int with 30% likelihood

javaslang/generator/Generator.scala
danieldietrich: My `xs"""..."""` implementation does _automatic_ indentation. No more `.stripMargin` necessary. But we leave it this way, it is ok!

danieldietrich:
We should not modify the existing `prepend()` and `append()` both in Vector and BitMappedTrie. Especially a single prepend() or append() should not create an additional Iterator on their way.

It is good that you made appendAll() and prependAll() faster but I think that append() for example is a much more commonly used operation. We should not make it that much slower!

I'm not sure if the ClassCastException related change is the right way to go. There are other options than running into an exception, for example adding some kind of introspection to the ArrayType, e.g. `ArrayType.isPrimitive(Object)` that checks sth. like `((Object[]) obj).getComponentType().isPrimitive()`.

However, I'm not sure about this extra layer of automatic boxing and unboxing of underlying arrays. It adds additional _magic_ that is not transparent to a user. The Vector currently has several orthogonal layers of this quality. For changes to Vector the developer needs to know all the internals of these additional _magic_ layers in order to keep the Vector stable. That is really evil. As I mentioned in the [recent change to CONTRIBUTION.md](https://github.com/javaslang/javaslang/commit/79a489ec027c1fd9636fb583615a63ac4713080e):

> Rule 3. Fancy algorithms are slow when n is small, and n is usually small.
> Rule 4. Fancy algorithms are buggier than simple ones, and they're much harder to implement.

What we do here is _way too fancy_. Do we know if users really use prependAll() and appendAll() frequently? And even if so - are these methods _really_ bottlenecks in their specific applications? I think we add way too much fancy optimizations. Also the other stuff recently added to Vector. I love implementations that are dead-simple and where the performance directly comes from the most basic data structures, like Tries. I do not care so much about being _m_ times faster for a specific method than other implementations.

**There are days where I'm tempted to revert the whole Vector implementation to its most basic form of BitMappedTrie without all the fancy stuff. Yes, it will be slower, but also it will be just beautiful in its natural form and so much more easy to maintain and -most important- less error prone.**

/cc @ruslansennov

javaslang/generator/Generator.scala
paplorinc: will change this, thanks :)

paplorinc:
> We should not modify the existing prepend() and append() both in Vector and BitMappedTrie.

I appreciate your feedback, @danieldietrich, it made previous implementations a lot more readable.

To be on the same page, I urge you to spend some quality IDE time with this commit or maybe merge-and-change it, so that you familiarize yourself with it.

I would like to stress that this is inherently a very complicated system, I've spent months simplifying it.
If you could try to refactor it in your IDE, I'm sure you would appreciate all the optimizations I have't done! :)

> I think that append() for example is a much more commonly used operation. We should not make it that much slower!

I agree, though we need to give a possibility for people to optimize.
Removing the leading and trailing arrays made the code a lot simpler, but append got ~100x slower than the mutable counterpart. This commit fixes this discrepancy in a lot more elegant way.

We cannot convince people to go functional, when they see that their application could experience two orders of magnitude slowdown. Tackling this is inherently complex.

> I'm not sure if the ClassCastException related change is the right way to go. 

The problem is that I need to change the underlying structure during iteration, whenever an illegal value is found. If you can recommend a simpler way (via an Iterator, that already consumed the illegal value), I would really appreciate it.
I could iterate over everything to check for inconsistencies and again to add the values. That however would slow down the happy path (i.e. you can add the new element without problem).

> It adds additional magic that is not transparent to a user. 

The users don't have to know about this, but should be able to take advantage of it (e.g. by using the alternative iteration to avoid boxing). If in the future we decide to remove it, the users won't really be affected (only memory- and speedwise).

> That is really evil.

Please advise an alternative way, without making it an order of magnitude slower. Our users expect us to sacrifice some of our comfort for theirs, i.e. if it's more difficult for us, but a lot simpler for them, we should go for it!

> are these methods really bottlenecks in their specific applications? 

You need to create a vector to be able to use it. This commit is about creating vectors an order of magnitude faster! People need to have alternatives.

> love implementations that are dead-simple and where the performance directly comes from the most basic data structures, like Tries. 

Daniel, tries are not simple! Please take a look at any other implementation (I recommend [Scala](http://github.com/scala/scala/blob/2.12.x/src/library/scala/collection/immutable/Vector.scala) or [Clojure](https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/PersistentVector.java)). Our methods are usually far simpler - though we provide a lot more functionality than any of them.

Please spend some IDE time with this so that you can get convinced - or to convince me!
The interface of Vector hasn't changed, we can always revert in the future, if some parts turn out to be unmaintainable.
Until then there is no point in being afraid, I think this is rather exciting :)

zsolt-donca:
Deciding on which direction to optimize is a tough decision. Both maintainability and performance are important factors, and comparing the two is tough as well, as they affect different people: maintainability affects the contributors and performance affects the end-users.

Reverting all important Vector optimizations based on a possible maintenance overhead seems like a premature optimization to me. Do we really know **now** if it will turn out as a maintenance problem? Maybe there won't be any need to do major touches on Vector for years. In all that time, our users could benefit from the added performance (which could be a good selling point, but that's another discussion). If there turns out to be something that's difficult or impossible to do with certain optimizations in place, we can decide **then** what needs to be taken out or simplified. Maybe it wouldn't be **every** Vector optimization, just some of them. We can't possibly foresee the future and shouldn't try to do so. While I know that the same argument could be made against making the performance optimizations now, there are some important differences: we know that runtime performance (both CPU and memory) is generally important to our users. We also know that there are potentially more people using and benefiting of the additional performance than people maintaining and contributing to Javaslang. That shifts the balance to me.

I see the point of having additional optimization layers as a possible maintenance cost. IMO this could be reasonably addressed by **documenting** them someplace. Just as [javaslang.io](http://www.javaslang.io/) is meant to explain Javaslang's functionality to the end-users, there could be a document somewhere explaining Javaslang's _architecture_ to the maintainers/contributors (maybe similarly to [CONTRIBUTING.md](https://github.com/javaslang/javaslang/blob/79a489ec027c1fd9636fb583615a63ac4713080e/CONTRIBUTING.md)), detailing the important **design decisions**; I can imagine a section explaining the Vector's basics and its internal layers of optimizations.

danieldietrich:
@paplorinc, @zsolt-donca thank you for your constructive feedback, I really appreciate it. For now you convinced me to give it a try. The code lives, we can adjust it later if necessary. However, Vector turned to the most complex collection of Javaslang.
Nevertheless, many thanks for the ongoing efforts!

PR #1653, opened 2016-10-30
Made ArrayType implementations nested classes.

danieldietrich:


danieldietrich:
This is just a refactoring of class location.

PR #1654, opened 2016-10-31
Makes Javaslang types directly pattern-matchable

### LINKED ISSUE #1649: Integrate pattern matching Patterns tightly into Javaslang in form of unapply()
## Steps to make types directly pattern-matchable

* [ ] **Introduce Pattern.unapply()** - Currently Pattern extends Function, i.e. it has apply(). This clashes with existing apply() methods of Value implementations. Therefor Pattern should not extend Function. Instead of apply() it should offer unapply() (which has to be unique for all patterns).
* [ ] **Let all Javaslang Values and Tuples extends some Pattern0..8** - Then the types can be passed to Match Cases. Pattern is a class. Java classes can have only one super-class. That should be no problem because our types are a tree of _interfaces_ with final _class_ leaves. However, we need to solve that the static _of(...)_ methods need to return a specific impl type (like List.Cons, which is a Pattern1, and List.Nil, which is a Pattern0). The aliases API.* return also specific types. List() returns List.Nil (by calling List.empty()) and List(...) return List.Cons (by calling List.of(...)). What return aliases for interfaces like Seq() and Map()? Maybe they need to return impls, i.e. Vector and HashMap. **This still needs to be investigated/solved.**
* [ ] **Add deprecation-support to javaslang-match** - The annotation processor needs to recognize deprecated types and deprecated methods. The generated source needs to reflect deprecated types and members.
* [ ] **Deprecate javaslang.$ (and all of its methods)** - With the point above this will automatically deprecate the generated javaslang.Patterns type and all of its methods. (javaslang.$ will be removed in 3.0.0)

---

Driver for this issue was:
- we have Patterns containing static Pattern factory methods `Some(T)`, `None`, ...
- now we have API containing static object factory methods `Some(T)`, `None`

These ~~clash~~ are redundant.

**Solution:** We fuse Javaslang types with Patterns by adding unapply() methods to the types.

**Goals:**
- deprecate javaslang.$ and javaslang.Patterns classes
- let ADT implementations implement unapply(), which returns a Pattern (needed for pattern matching)

**Impl details:**
- maybe the instance method unapply() will require us to make most of our collection classes interfaces (like HashMap, Vector, ...). Cons will have head, tail, Empty.unapply will return Pattern0. The existing classes cannot implement both.
- The static API factory methods need to return sum types, e.g.

``` java
static <T, Z extends Option<T> & Pattern0<T>> Z Some(T t) { return Option.some(t); }
```
- enhance javaslang-match annotation processor to recognize @Deprecated classes and methods and use these also in the generated code

---

Scala has tightly integrated ADTs, pattern matching etc. into the language. Sealed traits and case classes are used for these purposes:

``` scala
sealed trait List[+A]
case object Nil extends List[Nothing]
case class Cons[+A](head: A, tail: List[A]) extends List[A]
```

Scala has apply/unapply methods that are needed to construct objects from parts and deconstruct objects into parts (informally):

``` scala
def apply(head: A, tail: List[A]) = new List(head, tail)
def unapply(list: List[A]) = (list.head, list.tail)
```

---

In Java these features are missing on the language level. Javaslang implements the companion functionality (apply/unapply) in form of static helper methods:

``` java
interface List<T> {

    static List<T> empty() { return Nil.instance(); }
    static List<T> of(T...) { ... };

    T head();
    List<T> tail();

    static final class Cons<T> implements List<T> {
         ...
        Pattern2 unapply(); // TODO: add generic types here
    }

    static final class Nil<T> implements List<T> {
        ...
        Pattern0 unapply(); // TODO: add generic types here
    }
}
```

Please note that the static factories return the interface type, not the implementation types.

danieldietrich:
Early PR (ongoing commits...)

Fixes #1649

### DIFF
--- a/javaslang/src-gen/main/java/javaslang/API.java
+++ b/javaslang/src-gen/main/java/javaslang/API.java
@@ -3351,15 +3351,23 @@ public final class API {
 
     // 1) Atomic patterns $(), $(value), $(predicate)
 
+    private static final Pattern0<Object> ANY = new Pattern0<Object>() {
+        @Override
+        public Option<Object> unapply(Object o) {
+            return Option.some(o);
+        }
+    };
+
     /**
      * Wildcard pattern, matches any value.
      *
      * @param <T> injected type of the underlying value
      * @return a new {@code Pattern0} instance
      */
+    @SuppressWarnings("unchecked")
     @GwtIncompatible
     public static <T> Pattern0<T> $() {
-        return Pattern0.any();
+        return (Pattern0<T>) ANY;
     }
 
     /**
@@ -3373,7 +3381,7 @@ public final class API {
     public static <T> Pattern0<T> $(T prototype) {
         return new Pattern0<T>() {
             @Override
-            public Option<T> apply(T obj) {
+            public Option<T> unapply(T obj) {
                 return Objects.equals(obj, prototype) ? Option.some(obj) : Option.none();
             }
         };
@@ -3440,7 +3448,7 @@ public final class API {
         Objects.requireNonNull(predicate, "predicate is null");
         return new Pattern0<T>() {
             @Override
-            public Option<T> apply(T obj) {
+            public Option<T> unapply(T obj) {
                 return predicate.test(obj) ? Option.some(obj) : Option.none();
             }
         };
@@ -3482,11 +3490,10 @@ public final class API {
 
         // -- CASES
 
-        // javac needs fqn's here
         public interface Case<T, R> extends java.util.function.Function<T, javaslang.control.Option<R>> {
         }
 
-        public static final class Case0<T, R> implements Case<T, R> {
+        static final class Case0<T, R> implements Case<T, R> {
 
             private final Pattern0<T> pattern;
             private final Function<? super T, ? extends R> f;
@@ -3498,11 +3505,11 @@ public final class API {
 
             @Override
             public Option<R> apply(T o) {
-                return pattern.apply(o).map(f);
+                return pattern.unapply(o).map(f);
             }
         }
 
-        public static final class Case1<T, T1, R> implements Case<T, R> {
+        static final class Case1<T, T1, R> implements Case<T, R> {
 
             private final Pattern1<T, T1> pattern;
             private final Function<? super T1, ? extends R> f;
@@ -3514,11 +3521,11 @@ public final class API {
 
             @Override
             public Option<R> apply(T obj) {
-                return pattern.apply(obj).map(f);
+                return pattern.unapply(obj).map(f);
             }
         }
 
-        public static final class Case2<T, T1, T2, R> implements Case<T, R> {
+        static final class Case2<T, T1, T2, R> implements Case<T, R> {
 
             private final Pattern2<T, T1, T2> pattern;
             private final BiFunction<? super T1, ? super T2, ? extends R> f;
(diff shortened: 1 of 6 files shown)


javaslang/src-gen/main/java/javaslang/API.java
```diff
@@ -2905,149 +2906,149 @@ private Match(T value) {
 
         // javac needs fqn's here
         public interface Case<T, R> extends java.util.function.Function<T, javaslang.control.Option<R>> {
-        }
 
-        public static final class Case0<T, R> implements Case<T, R> {
+            final class Case0<T, R> implements Case<T, R> {
```
ruslansennov: fuuuuu.... ))

danieldietrich:
**Note to myself:** Test both widened and non-widened type T

1) Widened

```java
public interface Option<T> extends Value<T> {

    final class Some<T> extends Pattern1<Some<? extends T>, T>
                        implements Option<T>, Serializable {
        @Override
        public Option<T> unapply(Some<? extends T> t) {
            return ...;
        }
    }

    final class None<T> extends Pattern0<None<? extends T>>
                        implements Option<T>, Serializable {
        @Override
        public Option<None<? extends T>> unapply(None<? extends T> ts) {
            return ...;
        }
    }
}
```

2) Non-Widened

```java
public interface Option<T> extends Value<T> {

    final class Some<T> extends Pattern1<Some<T>, T>
                        implements Option<T>, Serializable {
        @Override
        public Option<T> unapply(Some<T> t) {
            return ...;
        }
    }

    final class None<T> extends Pattern0<None<T>>
                        implements Option<T>, Serializable {
        @Override
        public Option<None<T>> unapply(None<T> ts) {
            return ...;
        }
    }
}
```

**Note:** It seems odd to me that Pattern0 'extracts itself' - it should unapply zero parts instead, i.e. unapply() should return Option&lt;Void&gt;. Added it to the [list of breaking changes in 3.0.0](#1493).

**Note 2:** As far as I know are Scala's case classes flat, i.e. they cannot extend another class (but implementing traits works). We experience the same with Javaslang's pattern matching because Pattern0..8 have to be abstract classes.

In Javaslang interfaces like Option are like sealed traits. The implementing final classes are like case classes. Because they extends Pattern they technically cannot extend another class.

---

If the above works in they way I expect it then we will have a **huge fusion of Value and Pattern**. For example the following will be possible:

```java
Tuple2<Option<String>, Option<Integer>> obj = ...;

Match(obj).of(
    Case(Tuple(Some("Java"), None()), (some, none) -> ...),
    Case(Tuple(None(), Some(8)), (none, some) -> ...),
    Case(Tuple($(), $()), (o1, o2) -> ...)
);
```

With other words when using our new API aliases, we need $ only as wildcard and for predicates but not for values any more. **But without having tested it this cannot be confirmed.**

danieldietrich:
While playing around with the code and the API I got a feeling for the domain. This change will not be backward-compatible - it has too much impact on the existing API.

The main idea will be to make Value impls directly pattern matchable. This can be achieved by extending Pattern0, which is a class. That is a problem, because each class can inherit from exactly one super class. Pattern0 has to be a class because an interface will introduce ambiguous APIs. The whole solution I currently have in mind looks fragile and not well thought through.

Today I realized that the simplest solution would be to leave things as they are and move the generated Patterns from javaslang.Patterns to javaslang.API. Done.

Therefore I will close this PR and move this the issue #1649 to 3.0.0 (for further investigation).

PR #1655, opened 2016-10-31
Slightly improved toArray(Class)

danieldietrich:


javaslang/src/main/java/javaslang/Value.java
```diff
@@ -546,11 +546,10 @@ default CharSeq toCharSeq() {
...
                     componentType == short.class ? Short.class :
                     componentType == void.class ? Void.class : null;
-            return toJavaArray((Class<T>) boxedType);
-        } else {
-            final java.util.List<T> list = toJavaList();
-            return list.toArray((T[]) java.lang.reflect.Array.newInstance(componentType, list.size()));
+            componentType = (Class<T>) boxedType;
         }
+        final java.util.List<T> list = toJavaList();
+        return list.toArray((T[]) java.lang.reflect.Array.newInstance(componentType, list.size()));
```
paplorinc: now that we have `ArrayType`, we can create this without reflection (making it `GWT` compatible also).
Would be neet if you could JMH both, as it may also be faster than reflection :)

danieldietrich: Thanks for the hint!

danieldietrich: @paplorinc I don't see how we could use ArrayType. E.g. `asArray(java.util.Iterator<?> it, int length)` returns an `Object[]`, which cannot be cast to a `T[]`.

Do you have a hint for me?

---

Example:

```java
java.util.List<Integer> list = Arrays.asList(1, 2, 3, 4);
Integer[] ints = (Integer[]) ArrayType.asArray(list.iterator(), list.size());
for (int i = 0; i < ints.length; i++) {
    int j = ints[i];
    System.out.println(j);
}
```

Output:

```
Exception in thread "main" java.lang.ClassCastException: [Ljava.lang.Object; cannot be cast to [Ljava.lang.Integer;
```

paplorinc: Meh, you're right, it's not applicable in this particular case :/

danieldietrich: Ok, thank you!

Issue #1656, opened 2016-11-02
LinkedHashMap values method does not return values in order


thekaratekid05:
It was my expectation that calling LinkedHashMap.values() would return a Seq that had the same ordering as the map.

ruslansennov:
this is an obvious error
```java
@Override
public Seq<V> values() {
    return map.values();
}
```

PR #1657, opened 2016-11-02
Fixes LinkedHashMap::values() #1656

ruslansennov:


javaslang/src/test/java/javaslang/collection/LinkedHashMapTest.java
```diff
@@ -69,8 +69,14 @@ protected String className() {
 
     @Test
     public void shouldKeepOrder() {
-        CharSeq actual = LinkedHashMap.<Integer, Character> empty().put(3, 'a').put(2, 'b').put(1, 'c').foldLeft(CharSeq.empty(), (s, t) -> s.append(t._2));
-        assertThat(actual).isEqualTo(CharSeq.of("abc"));
+        List<Character> actual = LinkedHashMap.<Integer, Character> empty().put(3, 'a').put(2, 'b').put(1, 'c').foldLeft(List.empty(), (s, t) -> s.append(t._2));
+        Assertions.assertThat(actual).isEqualTo(List.of('a', 'b', 'c'));
```
ruslansennov: It was a big mistake to override `assers*` methods. They don't check ordering of iterables. It is sufficient for `Map`, but very dangerous for other cases.

danieldietrich: Yes - also IntMap and IntMultimap hide methods overridden by Map/Multimap because IntMap only extends Traversable. It is not sufficient that they are backed by Map/Multimap. This is the reason we did not recognize a bug...

With 3.0.0 we will change the collection type hierarchy 'a bit'

* Iterator will not be Iterable/Traversable anymore, instead it will be TraversableOnce
* Map/Multimap will extend Traversable&lt;V&gt; instead of Traversable&lt;K, V&gt;

This will make IntMap/IntMultimap superfluous. But for the overridden assert* methods I have currently no better solution...

danieldietrich: Would it be simpler to have String instead of List?

```java
String actual = LinkedHashMap....mkString();
Assertions.assertThat(actual).isEqualTo("abc");
```

danieldietrich:
Thank you! The test is ok as-is.

javaslang/src/test/java/javaslang/collection/LinkedHashMapTest.java
ruslansennov: > This is the reason we did not recognize a bug...

No, this bug was not recognized because of there was not test for it :) But we could not notice the bug in `fold()`, if it were

PR #1678, opened 2016-11-14
noticed that the ifPresent test sections duplicates the forEach tests

emmanueltouzery:
it seems javaslang used to have a `ifPresent` method on `Option` and it was removed or renamed to `forEach`.

Now the test section for `ifPresent` duplicates exactly the one for `forEach` which is a little lower in the same file:
https://github.com/javaslang/javaslang/blob/master/javaslang/src/test/java/javaslang/control/OptionTest.java#L350-L364

so this PR removes the now duplicate tests.


danieldietrich:
Thank you!

PR #1682, opened 2016-11-16
Add fold(ifEmpty, f) on Option

mathieuancelin:
I'd love to be able to use 'fold' the Option type, so I created this pull request.
I hope it's fine like that, all tests are green on my machine.

Let me know if there is any issue

### DIFF
--- a/javaslang/src/main/java/javaslang/control/Option.java
+++ b/javaslang/src/main/java/javaslang/control/Option.java
@@ -330,6 +330,22 @@ public interface Option<T> extends Value<T> {
         return f.apply(this);
     }
 
+    /**
+     * Returns the result of applying {@code f} to this Option's value if the Option is nonempty.
+     * Otherwise, evaluates expression {@code ifEmpty}.
+     *
+     * @param ifEmpty the expression to evaluate if Option is empty (None)
+     * @param f       the function to apply if Option is nonempty (Some)
+     * @param <U>     Type of folding result
+     * @return An instance of type {@code U}
+     * @throws NullPointerException if {@code ifEmpty} or {@code f} is null
+     */
+    default <U> U fold(Supplier<? extends U> ifEmpty, Function<? super T, ? extends U> f) {
+        Objects.requireNonNull(ifEmpty, "ifEmpty is null");
+        Objects.requireNonNull(f, "f is null");
+        return this.isEmpty() ? ifEmpty.get() : f.apply(this.get());
+    }
+
     @Override
     default Iterator<T> iterator() {
         return isEmpty() ? Iterator.empty() : Iterator.of(get());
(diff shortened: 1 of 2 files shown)


danieldietrich:
Thank you Mathieu, well done (one feedback: we omit `this.`).

This is a tough decision. When I saw the email notification about this PR my first thoughts were:

* "Hell, why not"
* "Having just Option.fold does not scale very well"
* "Ok, I think the Foldable interface was once part of Value, our mother of of monadic types. Maybe we can move it back from Traversable to Value. But wait - aren't there also foldLeft, foldRight, ... do these make sense for single-valued types?"

In the end I thought we should just add the fold to Option (without moving the Foldable interface) because it is/might be well known from Scala (or Haskell).

But then I read the following article and I agree: [Option.fold() considered unreadable](http://www.nurkiewicz.com/2014/06/optionfold-considered-unreadable.html)

I got feedback that there are already (too) many ways to achieve the same. We should take care keeping the API simple and comprehensible (but powerful). This is no dogma, there are examples where we added aliases to the existing API, e.g. Map.zip() as shortcut for Map.zip(e -> e). However, the user does not need to change his _thinking_, he wants a zip and he gets a zip.

Adding a fold(sup, f) as shortcut for map(f).getOrElse(sup) is different compared to adding a zip alias. The consumer of the API has to think in different ways of solving the same problem.

In the article I linked above **Martin Odersky** is cited: "I personally find methods (...) that take two closures as arguments are often overdoing it."

I think the Option.fold method does not fit well, therefore we should close this PR, for the following reasons:

* Traversable is Foldable, but Value is not. At a first glance it seems inconsistent that Values like Option (Try, Either, ...) have fold but do not implement Foldable. This confuses our users.
* There are too many ways to achieve the same. Keeping things simple is important. We also target an audience that is not familiar with Haskell or Scala.
* There are already several ways to express fold(_function_, _function_)

These are the same:

```java
option.fold(sup, f);

option.map(f).getOrElse(sup);

option.isDefined() ? f.apply(option.get()) : sup.get();

Match(option).of(
    Case(Some($()), f::apply),
    Case(None(), sup::get)
);
```

My suggestions are:

1. Use `option.map(f).getOrElse(sup)`
2. If it is crucial not to create an intermediate object (like map does) then use `option.isDefined() ? f.apply(option.get()) : sup.get()`

I hope you understand my decision to steer Javaslang in this particular direction. Looking at all your popular Github organizations I think you do.

I also hope to see future PRs from you :)

Greets

Daniel

danieldietrich:
Closing it now. Don't hesitate to share your thoughts.

mathieuancelin:
Hi Daniel,

It's okay, I understand your point and I can live without `fold` on `Option` :-) 

It's just that I'm used to it in Scala, so it would have been nice to be able to use it on Javaslang's `Option`.

Is there a place were I can discuss future pull requests with the team before writing them ? 

Thanks for your answer

Mathieu

danieldietrich:
Hi Mathieu,

> Is there a place were I can discuss future pull requests with the team before writing them ?

the best way is to create an issue. First I will label it as question and we can brainstorm how it integrate into the existing design etc.

Greets

\- Daniel

mathieuancelin:
Ok, thanks

PR #1686, opened 2016-11-17
fixes #1680 rename leftMap to mapError & deprecate leftMap

### LINKED ISSUE #1680: leftMap vs mapLeft
We have `Validation.leftMap` and `Either.mapLeft`.
Probably we should unify the naming for 3.0?
We could add Validation.mapLeft or Either.leftMap, deprecate the one we don't like and remove it in 3.0.

Except if it's like that on purpose, for some reason?

emmanueltouzery:
i can also make a second commit to remove leftMap meant for 3.0+ and that you would not cherry-pick in 2.x, let me know if you'd like that.

### DIFF
--- a/javaslang/src/main/java/javaslang/Value.java
+++ b/javaslang/src/main/java/javaslang/Value.java
@@ -907,7 +907,7 @@ public interface Value<T> extends Iterable<T> {
      */
     default <L> Validation<L, T> toValidation(L invalid) {
         if (this instanceof Validation) {
-            return ((Validation<?, T>) this).leftMap(ignored -> invalid);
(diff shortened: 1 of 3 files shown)


danieldietrich:
Hi Emmanuel,
thank you - this looks great to me.

> let me know if you'd like that

let's wait with that for now. If possible I want to avoid cherry-picking for the next release. There will be more than 200 PRs, each having multiple commit in general. That would be overkill. Instead I will analyze and drill down API changes and double-check (binary) compatibility.

After 2.1.0 is released I want to start to work on 3.0.0. There will be (one?) commit then that removes all deprecated API.

Many thanks!

PR #1688, opened 2016-11-18
fix a little annoyance, ignore emacs interlock files

emmanueltouzery:
i'm actually using emacs to develop, this turns up from time to time.
https://www.emacswiki.org/emacs/AutoSave#toc3


danieldietrich:
Cool, Emacs! Not heard that often, yet :)

PR #1689, opened 2016-11-18
fix a bunch of javadoc warnings and errors

emmanueltouzery:
some of the errors i'm fixing were from my PRs :-)
listing all of the key-values for the map builders looks a bit ugly even in the generated javadoc actually. But probably it's as it should be...

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -430,7 +430,7 @@ def generateMainClasses(): Unit = {
           }
 
           /$javadoc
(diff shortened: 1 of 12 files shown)


danieldietrich:
Many thanks! I did not know that we can use

```java
<pre>{@code
   ...
}</pre>
```

instead of

```java
<pre><code>
   ...
</code></pre>
```

emmanueltouzery:
yes turns out the {@code is much better because > and < are allowed inside. they're NOT allowed inside <code>, and that was a problem with these examples which I migrated because of the lambda arrow, "->".

PR #1691, opened 2016-11-18
Fixed toCharSeq()

### LINKED ISSUE #1685: Iterator.toCharSeq() error
The following code fails 
```java
@Test
public void fail(){
    CharSeq seq = Iterator.of('a','b','c').toCharSeq();
    assertThat(seq.toString()).isEqualTo("abc");
}
```
with the following error
```
org.junit.ComparisonFailure: 
Expected :"abc"
Actual   :"Iterator(?)"
````
The problem is in this method
```java
//Value.java
default CharSeq toCharSeq() {
    return CharSeq.of(toString());
}
```
Here ```toString()``` calls ```AbstractIterator.toString()```

ashrko619:
Fixes #1685

### DIFF
--- a/javaslang/src/main/java/javaslang/Value.java
+++ b/javaslang/src/main/java/javaslang/Value.java
@@ -493,7 +493,13 @@ public interface Value<T> extends Iterable<T> {
      * @return A new {@link CharSeq}.
      */
     default CharSeq toCharSeq() {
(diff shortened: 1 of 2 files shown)


danieldietrich:
Awesome, thank you for revealing this bug!!

PR #1693, opened 2016-11-19
make javadoc alignment at least locally coherent

emmanueltouzery:
@paplorinc was commenting about the javadoc alignment being inconsistent.
actually sometimes we have the descriptions aligned between themselves, sometimes they are only one space character away from the keys. Maybe we could harmonize that through the project, maybe not, but yes, at least we should be coherent in a single javadoc comment, that's what this commit does.

see https://github.com/javaslang/javaslang/commit/b97080b199adc2abe51b8ebb05d646acecfac8fe#commitcomment-19891096

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Map.java
+++ b/javaslang/src/main/java/javaslang/collection/Map.java
@@ -280,7 +280,7 @@ public interface Map<K, V> extends Traversable<Tuple2<K, V>>, Function1<K, V> {
     /**
      * Returns the value associated with a key, or a default value if the key is not contained in the map.
      *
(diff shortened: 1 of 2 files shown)


danieldietrich:
Thank you!

I think there are still several formatting-inconsistencies throughout the codebase - but on the other hand we take care of formatting. The repo is in a relatively good shape.

I plan to re-format the whole repo (*.java) soon. The Checkstyle configuration needs to be accordingly adjusted to be consistent with our IntelliJ formatter. For now we will use Sputnik review comments as hints if there is s.th. wrong beyond formatting.

PR #1700, opened 2016-11-22
optimized random value generators

danieldietrich:


javaslang-test/src/main/java/javaslang/test/Gen.java
```diff
@@ -424,3 +413,27 @@
         return f.apply(this);
     }
 }
+
+final class GenModule {
```
sputnikci: \[Checkstyle\] ERROR: Utility classes should not have a public or default constructor\.

javaslang-test/src/main/java/javaslang/test/Gen.java
```diff
@@ -424,3 +413,27 @@
...
+     * @return A value generator, choosen according to the given frequencies and the underlying n
+     */
+    static <T> Gen<T> frequency(int n, java.util.Iterator<Tuple2<Integer, Gen<T>>> iter) {
+        do {
+            final Tuple2<Integer, Gen<T>> freqGen = iter.next();
+            final int k = freqGen._1;
+            if (n <= k) {
+                return freqGen._2;
+            } else {
+                n = n - k;
```
sputnikci: \[Checkstyle\] ERROR: Assignment of parameter 'n' is not allowed\.

danieldietrich: Ok for me - it is performed on a copy on the stack. Btw - when dealing with immutable objects (like in Javaslang), then nothing evil can happen at all!

javaslang-test/src/main/java/javaslang/test/Gen.java
danieldietrich: Will fix that

javaslang-test/src/main/java/javaslang/test/Gen.java
```diff
@@ -424,3 +413,27 @@
...
```
sputnikci: \[Checkstyle\] ERROR: Assignment of parameter 'n' is not allowed\.

Issue #1699, opened 2016-11-22
Gen choose fails with stream of instances of generic interface


gtrefs:
# Problem statement
```Java
Stream<Supplier<String>> strings = Stream.of(() -> "test", () -> "test2");
Property.def("Test")
         .forAll(Gen.choose(strings).arbitrary())
         .suchThat(s -> true)
         .check()
         .assertIsSatisfied();
}
```
fails with following stack trace:
```Java
Nov 22, 2016 1:46:12 AM org.junit.platform.launcher.core.ServiceLoaderTestEngineRegistry loadTestEngines
INFORMATION: Discovered TestEngines with IDs: [junit-jupiter]

java.lang.ArrayStoreException
	at java.lang.System.arraycopy(Native Method)
	at java.util.ArrayList.toArray(ArrayList.java:408)
	at javaslang.Value.toJavaArray(Value.java:489)
	at javaslang.test.Gen.choose(Gen.java:205)
...
```
# Possible cause
This exception is thrown `if the runtime type of the specified array is not a supertype of the runtime type of every element in this list` (see [JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#toArray-T:A-)). I assume the `choose` method of `Gen` to be the culprit
```Java
final T[] array = stream.toJavaArray((Class<T>) stream.head().getClass());
```
Although `Supplier<String>` is a supertype of `stream.head()` the actual type of `stream.head()` is something like `test.Clazz$$Lambda$120/90205195`. In method `toJavaArray` of interface `Value`            
this type is used to indicate the component type of the resulting array:
```
final java.util.List<T> list = toJavaList();
return list.toArray((T[]) java.lang.reflect.Array.newInstance(componentType, list.size()));`
```
The type is correct for the first element, but not for the second. If the second `Supplier` is removed from the Stream, the test succeeds.
```Java
Stream<Supplier<String>> strings = Stream.of(() -> "test");
Property.def("Test")
         .forAll(Gen.choose(strings).arbitrary())
         .suchThat(s -> true)
         .check()
         .assertIsSatisfied(); // works fine
}
```
# Solution proposals
1. Use [TypeTokens](https://github.com/google/guava/wiki/ReflectionExplained) if appropiate
2. Add new choose method with additional class parameter
```Java
Gen.choose(strings, Supplier.class) // example
```

danieldietrich:
Hi @gtrefs,

[...]

PR #1702, opened 2016-11-22
Fix Gen choose

gtrefs:
For instances of generic interfaces, choose was throwing an ArrayStoreException because of different runtime types of the elements within the iterable. As suggested by Daniel this commit uses Iterator instead of Stream. This resolves the issue #1699.

### DIFF
--- a/javaslang-test/src/test/java/javaslang/test/GenTest.java
+++ b/javaslang-test/src/test/java/javaslang/test/GenTest.java
@@ -179,6 +179,14 @@ public class GenTest {
         assertForAll(() -> Gen.choose(i).apply(RANDOM), c -> c == 1);
     }
 
+    @Test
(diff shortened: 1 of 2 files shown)


javaslang-test/src/test/java/javaslang/test/GenTest.java
```diff
@@ -179,6 +179,14 @@ public void shouldChooseFromIterable() throws Exception {
         assertForAll(() -> Gen.choose(i).apply(RANDOM), c -> c == 1);
     }
 
+    @Test
+    public void shouldChooseFromIterableWithInstancesOfGenericInterface(){
```
sputnikci: \[Checkstyle\] ERROR: '\{' is not preceded with whitespace\.

gtrefs:
:+1: nice to know

danieldietrich:
Fixes #1699

Issue #1683, opened 2016-11-16
applicative lifting for Option, Try, Future...


emmanueltouzery:
javaslang has one implementation of an applicative: `Validation`.
But for instance if I have two `Option` and want to combine them, or two `Try`, or two `Future`... today we have `sequence`. And it's really quite good already!

I do wonder, would it be good to add also `Option.lift`, `Try.lift`, `Future.lift` and so on. These would be:

    Function<Option<T>,Option<U>,Option<V>> Option.lift(Function<T,U,V>)
    Function2<Option<T>,Option<U>,Option<V>,Option<W>> Option.lift(Function2<T,U,V,W>)
    ...
    up to Function7, and same for Try, Future and probably others.

Then you can do for instance:

    Future.lift(Clazz::handleResults).apply(future1, future2)

compared to today

    Future.sequence(List.of(future1,future2)).map(l -> handleResults(l.head(), l.get(1))

it's a little untractable however typing the types for local variables, until we'll get `var`.

Anyway, wrote all in pseudo-code, because I'm thinking about this, didn't try any code. I saw there is something about `Applicative` in `javaslang-pure`, but it doesn't seem finished.

Anyway, any opinion about that `lift` idea? Except if there already something else today besides `sequence` that I missed? Also, that's a little of "bloat" I guess, another like 7 functions per type we want to cover, of which I listed at least three for now...
Also I didn't actually have a need for that yet, was just thinking about applicatives in general the other day :-)

danieldietrich:
Hi Emmanuel,

thank you for sharing your thoughts! In fact the javaslang-**pure** module is exactly the place where functionality like lifting functions to operate on an algebraic layer should be located.

It is meant to be a _purely functional_ add-on to the javaslang-core.

But yes, it is unfinished. Towards Javaslang 2.0.0 I factored-out functionality from the core to the _pure_ module but it is still in an experimental state.

We should take this opportunity and add the functionality you suggested. We did the lifting already for [Monads](https://github.com/javaslang/javaslang/blob/master/javaslang-pure/src-gen/main/java/javaslang/algebra/Monad.java). Applicative needs to be extended in a similar way.

A first step would be to move Application into Generator.scala and add the lift functions. We should keep the operations minimal - Monad currently only has map and flatMap, Applicativ has map. Looking at Haskell is always a good source of useful information.

emmanueltouzery:
[...]

PR #1704, opened 2016-11-24
Make Kind1 and Kind2 extends derive4j-hkt interfaces.

jbgi:
so that javaslang can:
- enjoy enhanced compile-time verification (including generated safe cast)
- be compatible at the kind level with a wider ecosystem (highj/cyclops-react).

this commit also fix some errors that have been catched by the hkt processor.

More info (eg. how to customize (or even disable) code generation) at https://github.com/derive4j/hkt.

cc: @johnmcclean-aol

### DIFF
--- a/javaslang/pom.xml
+++ b/javaslang/pom.xml
@@ -29,6 +29,11 @@
             <scope>compile</scope>
         </dependency>
         <dependency>
+            <groupId>org.derive4j.hkt</groupId>
+            <artifactId>hkt</artifactId>
+            <version>${hkt.version}</version>
+        </dependency>
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <scope>test</scope>
--- a/javaslang/src/main/java/javaslang/Kind1.java
+++ b/javaslang/src/main/java/javaslang/Kind1.java
@@ -5,6 +5,8 @@
  */
 package javaslang;
 
+import org.derive4j.hkt.__;
+
 /**
  * Recursive self type representing {@code TYPE<T>}, which allows similar behavior to higher-kinded types.
  *
@@ -13,5 +15,5 @@ package javaslang;
  * @author Daniel Dietrich
  * @since 2.0.0
  */
-public interface Kind1<TYPE extends Kind1<TYPE, ?>, T> {
+public interface Kind1<TYPE extends Kind1<TYPE, ?>, T> extends __<TYPE, T> {
 }
--- a/javaslang/src/main/java/javaslang/Kind2.java
+++ b/javaslang/src/main/java/javaslang/Kind2.java
@@ -5,6 +5,8 @@
  */
 package javaslang;
 
+import org.derive4j.hkt.__2;
+
 /**
  * Recursive self type representing {@code TYPE<T1, T2>}, which allows similar behavior to higher-kinded types.
  *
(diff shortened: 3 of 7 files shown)


danieldietrich:
Hi Jean-Baptiste,

thank you for the PR! I'm afraid - one of the few constraints is that Javaslang cannot depend on any 3rd party libraries, no matter if it makes sense to include a feature or how many/few types are used.

From CONTRIBUTION.md:

screen shot 2016-11-24 at 11 49 43 pm

Beside that the Javaslang core makes no use of Kind1/Kind2 any more. Currently only the algebraic Monad.lift() and narrow() API makes use of Kind1. But the Monad interface seems artificial. There is a [discussion](#1683) to remove the algebraic module and the Kind1/Kind2 types in the next major version and in favor of adding the lift methods directly to the classes they belong to.

Therefor I will close this PR. However, the bugs you found should be fixed.

(Did you thought of bytecode enhancement? You could provide modules for libraries and manipulate the bytecode (at runtime) in order to add your interfaces.)

--

One question beside that - you once asked for returning the interface instead of implementations when calling factory methods, e.g. Option.some() and Option.none() should return Option instead of Some and None. This was important for your code generation w.r.t. the visitor pattern, right?

jbgi:
Thanks for the response.
I was expecting it. The reason why the hkt library could be an exception to the rule is that is intended to be an extension to the java language it-self.
The thing is that even if javaslang it-self does not make use of higher-kinded polymorphism, users could have benefited from the type being compatible in order to easily create type-classes instances on their own.
But I understand if you prefer javaslang to use inheritance polymorphism instead, for the sake of not confusing users not familiar with higher-kinded polymorphism.

as for Option.some()/none() I think derive4j should be fine if they return the concrete class. Although, you may loose type inference in some situations by doing this.

johnmcclean-aol:
If you are interested I've created Dervice4j HKT encodings and type class instances (Unit, Functor, Applicative, Monad, MonadZero, MonadPlus, Comonad, Traversable & Foldable) for Javaslang types inside the cyclops-javaslang project. Because they are using Derive4j they will also work with HighJ

*[Javaslang HKT 'type constructors'](https://github.com/aol/cyclops/tree/master/cyclops-javaslang/src/main/java/com/aol/cyclops/javaslang/hkt)
*[Javaslang Type classes](https://github.com/aol/cyclops/tree/master/cyclops-javaslang/src/main/java/com/aol/cyclops/javaslang/hkt/typeclasses/instances)

danieldietrich:
@johnmcclean-aol Oh my god! ;-) Do you know anyone who uses these (in the way they are intended to be used)?

**User:** I need to encode a computation based on a List of elements.

**Answer:** Choose one...

* List
* ListType
* ListInstances
  * Applicative<ListType.µ>
  * Foldable<ListType.µ>
  * Higher<ListType.µ,R>
  * MonadPlus<ListType.µ>
  * MonadZero<ListType.µ>
  * Traverse<ListType.µ>
  * Unit<ListType.µ>

Please watch [this video](https://www.youtube.com/watch?v=iKyIKozv8a8). Erik Osheim talks about "Principles for Modular, Functional, Approachable Libraries" - somewhere within the first 15-20 minutes he mentions that users are scared by type-classes.

johnmcclean-aol:
@danieldietrich If you need a List, use a List :)

User: I need to encode a computation based on a List of elements.

Answer: Use a list

User: I would like to encode an abstract computation over monads or applicatives so it is reusable across all such types

Answer: **In the absence of HKT and type classes** whistles...

danieldietrich:
We experiment with pulling this abstraction into static lift methods that are directly located in the monadic types, e.g. Option.lift(f) -> liftedFunction

No need for hkt's and narrowing anymore.

Issue #1495, opened 2016-08-15
Add Traversable isOrdered() indicator and Ordered interface


danieldietrich:
This comes handy when converting a specific Traversable to an ordered collection like SortedSet or PriorityQueue. We could get rid of using `instanceof`.

``` java
// old code
if (this instanceof SortedSet) {
    Comparator<T> comparator = ((SortedSet<T>) this).comparator();
    ...
} else if (this instanceof PriorityQueue) {
    Comparator<T> comparator = ((PriorityQueue <T>) this).comparator();
    ...
} else if (this instanceof ...) {
    Comparator<T> comparator = ((...<T>) this).comparator();
    ...
} else {
    ...
}

// new code
if (isOrdered()) {
    Comparator<T> comparator = ((Ordered<T>) this).comparator();
    ...
} else {
   ...
}
```

The new `Ordered` interface looks like this:

``` java
interface Ordered<T> {
    Comparator<T> comparator();
}
```

It will be implemented by SortedSet, PriorityQueue, ... The existing `comparator()` methods need an `@Override` annotation then.

danieldietrich:
This issue goes adhere with #1488

danieldietrich:
See also #1635

ruslansennov:
Closed by #1711

PR #1711, opened 2016-11-26
introduces Ordered interface #1495

ruslansennov:


javaslang/src/main/java/javaslang/Value.java
```diff
@@ -1032,8 +1035,8 @@ default CharSeq toCharSeq() {
         final Comparator<T> comparator;
         if (this instanceof TreeSet<?>) {
             return (TreeSet<T>) this;
-        } else if (this instanceof SortedSet<?>) {
-            comparator = ((SortedSet<T>) this).comparator();
+        } else if (this instanceof Ordered<?>) {
```
ruslansennov: here we can't use this code:
```java
if (isOrdered()) {
    Comparator<T> comparator = ((Ordered<T>) this).comparator();
    ...
} else {
   ...
}
```
because `isOrdered()` belongs to `Traversable` but not to `Value`. But `Ordered` combines `PriorityQueue`, `SortedSet` and `SortedMap` and should be useful

danieldietrich:
will take a look this evening!

javaslang/src/main/java/javaslang/collection/Ordered.java
```diff
@@ -0,0 +1,25 @@
...
+ * @param <T> Component type
+ * @author Ruslan Sennov, Daniel Dietrich
+ * @since 2.1.0
+ */
+public interface Ordered<T> {
+
+    /**
+     * Returns the comparator which defines the order of the elements contained in this collection.
+     *
+     * @return This set's comparator.
```
danieldietrich: This must not necessarily be a _set_. I would write

```java
@return The comparator that defines the order of this collection's elements.
```

javaslang/src/main/java/javaslang/Value.java
danieldietrich: makes sense!

javaslang/src/main/java/javaslang/Value.java
```diff
@@ -954,12 +954,15 @@ default CharSeq toCharSeq() {
...
         if (this instanceof PriorityQueue<?>) {
             return (PriorityQueue<T>) this;
+        } else if (this instanceof Ordered<?>) {
+            comparator = ((Ordered<T>) this).comparator();
         } else {
-            final Comparator<T> comparator = (Comparator<T> & Serializable) (o1, o2) -> ((Comparable<T>) o1).compareTo(o2);
-            return toPriorityQueue(comparator);
+            comparator = (Comparator<T> & Serializable) (o1, o2) -> ((Comparable<T>) o1).compareTo(o2);
         }
+        return toPriorityQueue(comparator);
```
danieldietrich: The control-flow is somehow 'overlapping' because the if-else-block cannot be substituted with a function call.

Please use this more 'straight-forward' version:

```java
if (this instanceof PriorityQueue) {
    return (PriorityQueue<T>) this;
} else {
    final Comparator<T> comparator = (this instanceof Comparator)
        ? ((Comparator<T>) this)
        : (Comparator<T> & Serializable) (o1, o2) -> ((Comparable<T>) o1).compareTo(o2);
    return toPriorityQueue(comparator);
}
```

javaslang/src/main/java/javaslang/Value.java
danieldietrich: Don't know why we used the `<?>` everywhere in instanceof checks - they are not really needed and add no value here. Nevermind, you don't need to change all occurrences in this PR.

danieldietrich:
Great, thank you!!

PR #1720, opened 2016-11-29
add Value.toCompletableFuture and Future.fromCompletableFuture (#1494)

mduesterhoeft:
addresses #1494

PR #1734, opened 2016-12-01
remove redundant type parameters

mduesterhoeft:
Follow up on PR #1720 - remove unnecessary type parameters


danieldietrich:
Thank you!

PR #1742, opened 2016-12-02
Reformatted code with new settings

danieldietrich:
We already talked about reformatting the code-base.

Now the time has come ...


danieldietrich:
**Just one note:** Style, especially formatting, is one of the most controversial topics discussed by developers. I've played around hours with different code-formats. My goals were:

* not to diverge too much from our current code format
* increase code readability
* not add too much whitespace (especially when aligning code to the right)

As a result I switched only a few settings to 'Align when multiline':

* Ternary operation (especially needed when having multiple branches: `cond1 ? expr1 : cond2 ? expr2 : alt`)
* try-with-resources
* for-statement

Other categories could be aligned by inserting an additional line-break:

```java
// array initializer
String[] arr = new String[] {
        "a",
        "b",
        "c"
};

// chained method calls
T result = object
        .map()
        .filter()
        .reduce();
```

PR #1736, opened 2016-12-01
Added wrapping `toJava` methods to the collections

paplorinc:
Addresses https://github.com/javaslang/javaslang/issues/1728

This PR is mainly meant to receive feedback.
~~More thorough tests should be added, and~~ the rest of the interfaces should also be covered (`Map`, `Set`, etc)

Issue #1728, opened 2016-11-30
View as immutable java.util collections without copying


seanf:
For efficient interop with general Java code, it would be very handy if the collections could be viewed as immutable `java.util` collections, without having to populate (eg) a new `java.util.HashMap`.

Fixed with #1952:

* [x] `List.asJavaList()`

Addressed in #1933:

* [ ] `Map.asJavaMap()`
* [ ] `MultiMap.asJavaMap()`
* [ ] `Set.asJavaSet()`

danieldietrich:
Great idea!!

seanf:
There's some similar code in Dexx Collections which may or may not be useful: https://github.com/andrewoma/dexx/tree/master/collection/src/main/java/com/github/andrewoma/dexx/collection/internal/adapter

paplorinc:
What should happen with the illegal method, e.g. `remove` when they're backed by an immutable collection?

danieldietrich:
#ThrowAllTheThings&trade;

paplorinc:
Theoretically we could, but what wouldn't it be very counter-intuitive, e.g.
```java
final java.util.List<T> list = source.toJavaList();
java.util.Collections.shuffle(list);
return ofAll.apply(list);
```
would either `throw new UnsupportedOperationException` or would shuffle a linked list in `O(N^2)` time

seanf:
I was just thinking the standard `throw new UnsupportedOperationException` approach, as taken by the immutable wrappers in the JDK[1], Guava immutable collections[2] and Scala[3]. It's not pretty, but it is documented behaviour for the Java collections - "Throws: UnsupportedOperationException - if the put operation is not supported by this map"[4].

If someone wants to get a mutable Java Map out of a Javaslang Map, that's what `toJavaMap()` is for - in most cases mutations would probably far more efficient with a standard java.util.HashMap than a tricksy delegating implementation. Except in cases where the initial copy is expensive and there are few mutations afterwards. But I think that would be a less common use case.

[1] https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#unmodifiableMap-java.util.Map-
[2] https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/ImmutableList.html#add-int-E-
[3] http://docs.scala-lang.org/overviews/collections/conversions-between-java-and-scala-collections.html
[4] https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#put-K-V-

[...]

PR #1751, opened 2016-12-08
Add mutable/immutable java.util.List wrappers

seanf:
Based on the work in PR https://github.com/javaslang/javaslang/pull/1736 for (part of) issue https://github.com/javaslang/javaslang/issues/1728.

I suggest reviewing this PR via reviewable (it will add comments here in github): https://reviewable.io/reviews/javaslang/javaslang/1751. I generally find it makes it easier to keep track of individual concerns in the code.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/JavaConverters.java
+++ b/javaslang/src/main/java/javaslang/collection/JavaConverters.java
@@ -7,9 +7,11 @@ package javaslang.collection;
 
 import java.io.Serializable;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.ListIterator;
 import java.util.function.BiFunction;
+import java.util.function.UnaryOperator;
 
 import static javaslang.API.TODO;
 import static javaslang.collection.Collections.areEqual;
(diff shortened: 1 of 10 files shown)


javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,739 @@
...
+        }
+
+        public Seq<Object> narrowed() {
+            return Seq.narrow(delegate);
+        }
+
+        // -- Object.*
+
+        @Override
+        public boolean equals(Object o) {
```
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

seanf: Can we turn this off in PMD? Or should I just call super.hashCode()? Because that's the implementation I want to use. I considered inheriting super.equals too, but this implementation is probably more efficient.

danieldietrich:
@seanf @paplorinc I'm thankful for all your effort but this is too much for me :-) We now have two PRs open for the same issue and they diverge somehow.

The naming changed here, I created code that was not parameterized with an interface and so on. It is too much effort for me to review on that detail level on two diverging PRs.

Beside that I've created already an own implementation in order to be able to give review code examples. This is a third impl for that issue now.

Long story short: I want to finish my impl with your suggestions and close the existing two PRs in order to reduce the review noise and time effort.

That's the first time I do that and I hope not to piss off anyone...

---

Side-note: For now we focus on using plain Github for reviews.

seanf:
Closed in favour of https://github.com/javaslang/javaslang/pull/1752

paplorinc:
Thanks @seanf :)

PR #1752, opened 2016-12-09
Adds java.util Collection Views

danieldietrich:
Addresses #1728

This is in an early state. The following things are to be done:

* [x] Name methods asJavaMutable() and asJavaImmutable() (instead of asJava())
* [ ] Write failing unit tests that cover all(!) details of the java.util API (especially the exceptional cases)
* [ ] ~~Optimize the *.ofAll(Iterable) methods in the way that we check for instanceof View and return delegate if possible~~ We will not do that in a first step. It would pull the concept of collection views into the persistent collections. But both concerns should be separated.
* [x] Solve remaining TODOs
* [ ] **Re-enable Sputnik in .travis.yml**
---
**Note:** We already have two PRs (#1736 by @paplorinc , #1751 by @seanf) that target the issue. But there are too many details to mention / refactor. It is most economic to do it myself. Please understand me.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Array.java
+++ b/javaslang/src/main/java/javaslang/collection/Array.java
@@ -16,6 +16,8 @@ import java.util.stream.Collector;
 
 import static java.util.Arrays.copyOf;
 import static java.util.Arrays.sort;
+import static javaslang.collection.JavaConverters.ChangePolicy.IMMUTABLE;
+import static javaslang.collection.JavaConverters.ChangePolicy.MUTABLE;
 
 /**
  * Array is a Traversable wrapper for {@code Object[]} containing elements of type {@code T}.
@@ -600,6 +602,26 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
         return ofAll(iterator().<R> collect(partialFunction));
     }
     
+    @Override
+    public java.util.List<T> asJavaImmutable() {
+        return JavaConverters.asJava(this, IMMUTABLE);
+    }
+
+    @Override
+    public Array<T> asJavaImmutable(Consumer<? super java.util.List<T>> action) {
+        return Collections.asJava(this, action, IMMUTABLE);
+    }
+
+    @Override
+    public java.util.List<T> asJavaMutable() {
+        return JavaConverters.asJava(this, MUTABLE);
+    }
+
+    @Override
+    public Array<T> asJavaMutable(Consumer<? super java.util.List<T>> action) {
+        return Collections.asJava(this, action, MUTABLE);
+    }
+
     @Override
     public boolean hasDefiniteSize() {
         return true;
--- a/javaslang/src/main/java/javaslang/collection/Collections.java
+++ b/javaslang/src/main/java/javaslang/collection/Collections.java
@@ -5,6 +5,8 @@
  */
 package javaslang.collection;
 
+import javaslang.collection.JavaConverters.ChangePolicy;
+import javaslang.collection.JavaConverters.ListView;
 import javaslang.control.Option;
 
 import java.util.ArrayList;
@@ -13,6 +15,7 @@ import java.util.Objects;
 import java.util.function.*;
 
 import static javaslang.collection.ArrayType.asArray;
+import static javaslang.collection.JavaConverters.ChangePolicy.MUTABLE;
 
 /**
  * Internal class, containing helpers.
@@ -32,6 +35,13 @@ final class Collections {
             }
         }
         return iter1.hasNext() == iter2.hasNext();
+    }
+
+    static <T, C extends Seq<T>> C asJava(C source, Consumer<? super java.util.List<T>> action, ChangePolicy changePolicy) {
+        Objects.requireNonNull(action, "action is null");
+        final ListView<T, C> view = JavaConverters.asJava(source, changePolicy);
+        action.accept(view);
+        return view.getDelegate();
     }
 
     @SuppressWarnings("unchecked")
--- a/javaslang/src/main/java/javaslang/collection/JavaConverters.java
+++ b/javaslang/src/main/java/javaslang/collection/JavaConverters.java
@@ -1,0 +1,449 @@
+/*     / \____  _    _  ____   ______  / \ ____  __    _______
+ *    /  /    \/ \  / \/    \ /  /\__\/  //    \/  \  //  /\__\   JΛVΛSLΛNG
+ *  _/  /  /\  \  \/  /  /\  \\__\\  \  //  /\  \ /\\/ \ /__\ \   Copyright 2014-2017 Javaslang, http://javaslang.io
+ * /___/\_/  \_/\____/\_/  \_/\__\/__/\__\_/  \_//  \__/\_____/   Licensed under the Apache License, Version 2.0
+ */
+package javaslang.collection;
+
+import java.io.Serializable;
+import java.util.*;
+import java.util.Iterator;
+import java.util.List;
+import java.util.function.Supplier;
+
+import static javaslang.API.TODO;
+
+/**
+ * THIS CLASS IS INTENDED TO BE USED INTERNALLY ONLY!
+ * <p>
+ * This helper class provides methods that return views on Java collections.
+ * The view creation and back conversion take O(1).
+ *
+ * @author Daniel Dietrich
+ * @since 2.1.0
+ */
+class JavaConverters {
+
+    private JavaConverters() {
+    }
+
+    static <T, C extends Seq<T>> ListView<T, C> asJava(C seq, ChangePolicy changePolicy) {
+        return new ListView<>(seq, changePolicy.isMutable());
+    }
+
+    static <T, C extends Set<T>> SetView<T> asJava(C set, ChangePolicy changePolicy) {
+        return TODO("new SetView<>(set, changePolicy.isMutable());");
+    }
+
+    static <T, C extends SortedSet<T>> NavigableSetView<T> asJava(C sortedSet, ChangePolicy changePolicy) {
+        return TODO("new NavigableSetView<>(set, changePolicy.isMutable());");
+    }
+
+    static <K, V, C extends Map<K, V>> MapView<K, V> asJava(C map, ChangePolicy changePolicy) {
+        return TODO("new MapView<>(map, changePolicy.isMutable());");
+    }
+
+    static <K, V, C extends SortedMap<K, V>> NavigableMapView<K, V> asJava(C sortedMap, ChangePolicy changePolicy) {
+        return TODO("new NavigableMapView<>(map, changePolicy.isMutable());");
+    }
+
+    /*
+    static <K, V, C extends Multimap<K, V>> MapView<K, java.util.Collection<V>> asJava(C multimap, ChangePolicy changePolicy) {
+        return TODO("new MapView<>(map, changePolicy.isMutable());");
+    }
+
+    TODO: create interface javaslang.collection.SortedMultimap
+    static <K, V, C extends SortedMultimap<K, V>> NavigableMapView<K, java.util.Collection<V>> asJava(C sortedMultimap, ChangePolicy changePolicy) {
+        return TODO("new NavigableMapView<>(map, changePolicy.isMutable());");
+    }
+    */
+
+    enum ChangePolicy {
+
+        IMMUTABLE, MUTABLE;
+
+        boolean isMutable() {
+            return this == MUTABLE;
+        }
+    }
+
+    // -- private view implementations
+
+    /**
+     * Encapsulates the access to delegate and performs mutability checks.
+     *
+     * @param <C> The Javaslang collection type
+     */
+    private static abstract class HasDelegate<C extends Traversable<?>> implements Serializable {
+
+        private static final long serialVersionUID = 1L;
+
+        private C delegate;
+        private final boolean mutable;
+
+        HasDelegate(C delegate, boolean mutable) {
+            this.delegate = delegate;
+            this.mutable = mutable;
+        }
+
+        protected boolean isMutable() {
+            return mutable;
+        }
+
+        C getDelegate() {
+            return delegate;
+        }
+
+        protected boolean setDelegateAndCheckChanged(Supplier<C> delegate) {
+            ensureMutable();
+            final C previousDelegate = this.delegate;
+            final C newDelegate = delegate.get();
+            final boolean changed = newDelegate != previousDelegate;
+            if (changed) {
+                this.delegate = newDelegate;
+            }
+            return changed;
+        }
+
+        protected void setDelegate(Supplier<C> newDelegate) {
+            ensureMutable();
+            this.delegate = newDelegate.get();
+        }
+
+        protected void ensureMutable() {
+            if (!mutable) {
+                throw new UnsupportedOperationException();
+            }
+        }
+    }
+
+    static class ListView<T, C extends Seq<T>> extends HasDelegate<C> implements java.util.List<T> {
+
+        private static final long serialVersionUID = 1L;
+
+        ListView(C delegate, boolean mutable) {
+            super(delegate, mutable);
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public boolean add(T element) {
+            return setDelegateAndCheckChanged(() -> (C) getDelegate().append(element));
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public void add(int index, T element) {
+            setDelegate(() -> (C) getDelegate().insert(index, element));
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public boolean addAll(Collection<? extends T> collection) {
+            Objects.requireNonNull(collection, "collection is null");
+            return setDelegateAndCheckChanged(() -> (C) getDelegate().appendAll(collection));
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public boolean addAll(int index, Collection<? extends T> collection) {
+            Objects.requireNonNull(collection, "collection is null");
+            return setDelegateAndCheckChanged(() -> (C) getDelegate().insertAll(index, collection));
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public void clear() {
+            setDelegate(() -> (C) getDelegate().take(0));
+        }
+
+        @Override
+        public boolean contains(Object obj) {
+            @SuppressWarnings("unchecked") final T that = (T) obj;
+            return getDelegate().contains(that);
+        }
+
+        @Override
+        public boolean containsAll(Collection<?> collection) {
+            Objects.requireNonNull(collection, "collection is null");
+            @SuppressWarnings("unchecked") final Collection<T> that = (Collection<T>) collection;
+            return getDelegate().containsAll(that);
+        }
+
+        @Override
+        public T get(int index) {
+            return getDelegate().get(index);
+        }
+
+        @Override
+        public int indexOf(Object obj) {
+            @SuppressWarnings("unchecked") final T that = (T) obj;
+            return getDelegate().indexOf(that);
+        }
+
+        @Override
+        public boolean isEmpty() {
+            return getDelegate().isEmpty();
+        }
+
+        @Override
+        public Iterator<T> iterator() {
+            return getDelegate().iterator();
+        }
+
+        @Override
+        public int lastIndexOf(Object obj) {
+            @SuppressWarnings("unchecked") final T that = (T) obj;
+            return getDelegate().lastIndexOf(that);
+        }
+
+        @Override
+        public java.util.ListIterator<T> listIterator() {
+            return new ListIterator<>(this, 0);
+        }
+
+        @Override
+        public java.util.ListIterator<T> listIterator(int index) {
+            return new ListIterator<>(this, index);
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public T remove(int index) {
+            return setDelegateAndGetPreviousElement(index, () -> (C) getDelegate().removeAt(index));
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public boolean remove(Object obj) {
+            @SuppressWarnings("unchecked") final T that = (T) obj;
+            return setDelegateAndCheckChanged(() -> (C) getDelegate().remove(that));
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public boolean removeAll(Collection<?> collection) {
+            Objects.requireNonNull(collection, "collection is null");
+            @SuppressWarnings("unchecked") final Collection<T> that = (Collection<T>) collection;
+            return setDelegateAndCheckChanged(() -> (C) getDelegate().removeAll(that));
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public boolean retainAll(Collection<?> collection) {
+            Objects.requireNonNull(collection, "collection is null");
+            @SuppressWarnings("unchecked") final Collection<T> that = (Collection<T>) collection;
+            return setDelegateAndCheckChanged(() -> (C) getDelegate().retainAll(that));
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public T set(int index, T element) {
+            return setDelegateAndGetPreviousElement(index, () -> (C) getDelegate().update(index, element));
+        }
+
+        @Override
+        public int size() {
+            return getDelegate().size();
+        }
+
+        @Override
+        public List<T> subList(int fromIndex, int toIndex) {
+            return new ListView<>(getDelegate().subSequence(fromIndex, toIndex), isMutable());
+        }
+
+        @Override
+        public Object[] toArray() {
+            return getDelegate().toJavaArray();
+        }
+
+        @GwtIncompatible("reflection is not supported")
+        @SuppressWarnings("unchecked")
+        @Override
+        public <U> U[] toArray(U[] array) {
+            Objects.requireNonNull(array, "array is null");
+            final U[] target;
+            final int length = getDelegate().length();
+            if (array.length < length) {
+                final Class<? extends Object[]> newType = array.getClass();
+                target = (newType == Object[].class)
+                         ? (U[]) new Object[length]
+                         : (U[]) java.lang.reflect.Array.newInstance(newType.getComponentType(), length);
+            } else {
+                if (array.length > length) {
+                    array[length] = null;
+                }
+                target = array;
+            }
+            final Iterator<T> iter = iterator();
+            for (int i = 0; i < length; i++) {
+                target[i] = (U) iter.next();
+            }
+            return target;
+        }
+
+        // -- Object.*
+
+        @Override
+        public boolean equals(Object o) {
+            if (o == this) {
+                return true;
+            } else if (o instanceof java.util.List) {
+                return Collections.areEqual(getDelegate(), (java.util.List<?>) o);
+            } else {
+                return false;
+            }
+        }
+
+        @Override
+        public int hashCode() {
+            // DEV-NOTE: Ensures that hashCode calculation is stable, regardless of delegate.hashCode()
+            return Collections.hash(getDelegate());
+        }
+
+        @Override
+        public String toString() {
+            return getDelegate().mkString("[", ", ", "]");
+        }
+
+        // -- private helpers
+
+        private T setDelegateAndGetPreviousElement(int index, Supplier<C> delegate) {
+            ensureMutable();
+            final T previousElement = getDelegate().get(index);
+            setDelegate(delegate);
+            return previousElement;
+        }
+
+        // DEV-NOTE: ListIterator is intentionally not Serializable
+        private static class ListIterator<T, C extends Seq<T>> implements java.util.ListIterator<T> {
+
+            private ListView<T, C> list;
+            private int index;
+            private boolean dirty = true;
+
+            ListIterator(ListView<T, C> list, int index) {
+                if (index < 0 || index > list.size()) {
+                    throw new IndexOutOfBoundsException("Index: " + index + ", Size: " + list.size());
+                }
+                this.list = list;
+                this.index = index;
+            }
+
+            @Override
+            public boolean hasNext() {
+                return index < list.size();
+            }
+
+            @Override
+            public T next() {
+                if (!hasNext()) {
+                    throw new NoSuchElementException();
+                }
+                dirty = false;
+                return list.get(index++);
+            }
+
+            @Override
+            public int nextIndex() {
+                return index;
+            }
+
+            @Override
+            public boolean hasPrevious() {
+                return index > 0;
+            }
+
+            @Override
+            public T previous() {
+                if (!hasPrevious()) {
+                    throw new NoSuchElementException();
+                }
+                dirty = false;
+                return list.get(--index);
+            }
+
+            @Override
+            public int previousIndex() {
+                return index - 1;
+            }
+
+            @Override
+            public void remove() {
+                checkDirty();
+                list.remove(index);
+                dirty = true;
+            }
+
+            @Override
+            public void set(T value) {
+                checkDirty();
+                list.set(index, value);
+            }
+
+            @Override
+            public void add(T value) {
+                /* TODO:
+                 * The new element is inserted before the implicit
+                 * cursor: a subsequent call to {@code next} would be unaffected, and a
+                 * subsequent call to {@code previous} would return the new element.
+                 * (This call increases by one the value that would be returned by a
+                 * call to {@code nextIndex} or {@code previousIndex}.)
+                 */
+                // may throw a ClassCastException accordingly to j.u.ListIterator.add(T)
+                list.add(index++, value);
+                dirty = true;
+            }
+
+            private void checkDirty() {
+                if (dirty) {
+                    throw new IllegalStateException();
+                }
+            }
+        }
+    }
+
+    static abstract class SetView<T> extends HasDelegate<Set<T>> implements java.util.Set<T> {
+
+        private static final long serialVersionUID = 1L;
+
+        SetView(Set<T> delegate, boolean mutable) {
+            super(delegate, mutable);
+        }
+
+        // TODO
+    }
+
+    static abstract class NavigableSetView<T> extends HasDelegate<SortedSet<T>> implements java.util.NavigableSet<T> {
+
+        private static final long serialVersionUID = 1L;
+
+        NavigableSetView(SortedSet<T> delegate, boolean mutable) {
+            super(delegate, mutable);
+        }
+
+        // TODO
+    }
+
+    static abstract class MapView<K, V> extends HasDelegate<Map<K, V>> implements java.util.Map<K, V> {
+
+        private static final long serialVersionUID = 1L;
+
+        MapView(Map<K, V> delegate, boolean mutable) {
+            super(delegate, mutable);
+        }
+
+        // TODO
+    }
+
+    static abstract class NavigableMapView<K, V> extends HasDelegate<SortedMap<K, V>> implements java.util.NavigableMap<K, V> {
+
+        private static final long serialVersionUID = 1L;
+
+        NavigableMapView(SortedMap<K, V> delegate, boolean mutable) {
+            super(delegate, mutable);
+        }
+
+        // TODO
+    }
+}
--- a/javaslang/src/main/java/javaslang/collection/JavaConverters.java
+++ b/javaslang/src/main/java/javaslang/collection/JavaConverters.java
@@ -7,11 +7,8 @@ package javaslang.collection;
 
 import java.io.Serializable;
 import java.util.*;
-import java.util.Iterator;
-import java.util.List;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
-
-import static javaslang.API.TODO;
 
 /**
  * THIS CLASS IS INTENDED TO BE USED INTERNALLY ONLY!
@@ -30,33 +27,6 @@ class JavaConverters {
     static <T, C extends Seq<T>> ListView<T, C> asJava(C seq, ChangePolicy changePolicy) {
         return new ListView<>(seq, changePolicy.isMutable());
     }
-
-    static <T, C extends Set<T>> SetView<T> asJava(C set, ChangePolicy changePolicy) {
-        return TODO("new SetView<>(set, changePolicy.isMutable());");
-    }
-
-    static <T, C extends SortedSet<T>> NavigableSetView<T> asJava(C sortedSet, ChangePolicy changePolicy) {
-        return TODO("new NavigableSetView<>(set, changePolicy.isMutable());");
-    }
-
-    static <K, V, C extends Map<K, V>> MapView<K, V> asJava(C map, ChangePolicy changePolicy) {
-        return TODO("new MapView<>(map, changePolicy.isMutable());");
-    }
-
-    static <K, V, C extends SortedMap<K, V>> NavigableMapView<K, V> asJava(C sortedMap, ChangePolicy changePolicy) {
-        return TODO("new NavigableMapView<>(map, changePolicy.isMutable());");
-    }
-
-    /*
-    static <K, V, C extends Multimap<K, V>> MapView<K, java.util.Collection<V>> asJava(C multimap, ChangePolicy changePolicy) {
-        return TODO("new MapView<>(map, changePolicy.isMutable());");
-    }
-
-    TODO: create interface javaslang.collection.SortedMultimap
-    static <K, V, C extends SortedMultimap<K, V>> NavigableMapView<K, java.util.Collection<V>> asJava(C sortedMultimap, ChangePolicy changePolicy) {
-        return TODO("new NavigableMapView<>(map, changePolicy.isMutable());");
-    }
-    */
 
     enum ChangePolicy {
 
@@ -187,8 +157,8 @@ class JavaConverters {
         }
 
         @Override
-        public Iterator<T> iterator() {
-            return getDelegate().iterator();
+        public java.util.Iterator<T> iterator() {
+            return new Iterator<>(this);
         }
 
         @Override
@@ -248,7 +218,7 @@ class JavaConverters {
         }
 
         @Override
-        public List<T> subList(int fromIndex, int toIndex) {
+        public java.util.List<T> subList(int fromIndex, int toIndex) {
             return new ListView<>(getDelegate().subSequence(fromIndex, toIndex), isMutable());
         }
 
@@ -263,7 +233,8 @@ class JavaConverters {
         public <U> U[] toArray(U[] array) {
             Objects.requireNonNull(array, "array is null");
             final U[] target;
-            final int length = getDelegate().length();
+            final C delegate = getDelegate();
+            final int length = delegate.length();
             if (array.length < length) {
                 final Class<? extends Object[]> newType = array.getClass();
                 target = (newType == Object[].class)
@@ -275,7 +246,7 @@ class JavaConverters {
                 }
                 target = array;
             }
-            final Iterator<T> iter = iterator();
+            final java.util.Iterator<T> iter = delegate.iterator();
             for (int i = 0; i < length; i++) {
                 target[i] = (U) iter.next();
             }
@@ -286,13 +257,7 @@ class JavaConverters {
 
         @Override
         public boolean equals(Object o) {
-            if (o == this) {
-                return true;
-            } else if (o instanceof java.util.List) {
-                return Collections.areEqual(getDelegate(), (java.util.List<?>) o);
-            } else {
-                return false;
-            }
+            return o == this || o instanceof java.util.List && Collections.areEqual(getDelegate(), (java.util.List<?>) o);
         }
 
         @Override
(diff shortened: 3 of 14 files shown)


javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,437 @@
...
+            return delegate;
+        }
+
+        C setDelegate(C newDelegate) {
+            if (mutable) {
+                final C previousDelegate = delegate;
+                delegate = newDelegate;
+                return previousDelegate;
+            } else {
+                throw new UnsupportedOperationException();
```
seanf: This incurs a fair bit of unnecessary work building the newDelegate collection in the immutable case.

What about another method `getDelegateForMutation()` which can throw the exception earlier in the immutable case, which will be called by the add/remove/mutator methods instead of `getDelegate()`?

javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,437 @@
...
+
+        @Override
+        public String toString() {
+            return getDelegate().mkString("[", ", ", "]");
+        }
+
+        // -- private helpers
+
+        private boolean setDelegateAndCheckChanged(Seq<T> delegate) {
+            // TODO: or `return previousDelegate != delegate`?
```
seanf: That might be safer, because not every change will change the size, eg `set()`. Especially given the current method name.

javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,437 @@
...
+
+            private int index;
+
+            static <T> ListIterator<T> of(Seq<T> seq, int index) {
+                final IndexedSeq<T> indexedSeq = (seq instanceof IndexedSeq) ? (IndexedSeq<T>) seq : seq.toVector();
+                return new ListIterator<>(indexedSeq, index);
+            }
+
+            ListIterator(IndexedSeq<T> delegate, int index) {
+                super(delegate, true); // TODO: always mutable?
```
seanf: You know what I'm going to say to that question. :-)

Exceptions aren't nice, but silent failures are worse than noisy failures. Pretending an immutable collection is mutable is bound to end in tears. You would be quietly mutating a secret copy of the immutable list. I'm not sure any other code would ever see that mutated copy.

javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,437 @@
...
+        }
+
+        @Override
+        public Object[] toArray() {
+            return getDelegate().toJavaArray();
+        }
+
+        @Override
+        public <T1> T1[] toArray(T1[] array) {
+            return getDelegate().toJavaList().toArray(array);
```
seanf: Now I remember why I objected to `return delegate.toJavaList().toArray(array);` in the other PR - the O(n) call to `toJavaList()`. But I suppose that can always be optimised later with help from the delegate. Maybe a comment to that effect?

javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,437 @@
...
+            } else if (o instanceof java.util.List) {
+                return Collections.areEqual(getDelegate(), (java.util.List<?>) o);
+            } else {
+                return false;
+            }
+        }
+
+        @Override
+        public int hashCode() {
+            return getDelegate().hashCode();
```
seanf: As I said in the other review:

I see that Collections.hash _currently_ implements the algorithm required by `java.util.List.hashCode()`, but I still think it would be safer from a maintainability point of view to use the implementation provided by AbstractList if possible _[to ensure consistency with `equals()`]_. Alternatively, Collections.hash ought to have a comment advising that the implementation must never be changed. But I don't think it would be a good idea to tie **all** the javaslang collections to follow java.util's rules forever for all. Safer just to use AbstractList. _[Assuming that's still possible with the new implementation.]_

* https://docs.oracle.com/javase/8/docs/api/java/util/List.html#hashCode--
* https://docs.oracle.com/javase/8/docs/api/java/util/AbstractList.html#hashCode--

javaslang/src/main/java/javaslang/collection/JavaConverters.java
danieldietrich: You are right, I also thought about it. But when we run into the UnsupportedOperationException case, there is something wrong at the call-site and code has to be fixed. Then it should not matter.

It is nice (from the maintainability side) to have one central spot that contains the relevant code.

javaslang/src/main/java/javaslang/collection/JavaConverters.java
danieldietrich: Yes, we can do better. The next commit will contain it!

javaslang/src/main/java/javaslang/collection/JavaConverters.java
danieldietrich: I took a look at AbstractList and don't see how we can use its implementations of hashCode() and equals(). I don't expect them to change ever because it would have effects on existing code bases.

javaslang/src/main/java/javaslang/collection/JavaConverters.java
danieldietrich: Ok, I will change it. To get sure, we could add unit tests for that.

javaslang/src/main/java/javaslang/collection/JavaConverters.java
danieldietrich: @seanf

> You know what I'm going to say to that question. :-)

Yes :-)

Yep, you are right. In the immutable case we should throw. Here is a test:

```java
import java.util.ArrayList;
import java.util.Arrays;
import java.util.List;
import java.util.ListIterator;

public class Test {

        Test test = new Test();
        test.testMutable();
        test.testImmutable();
    }

    void testMutable() {
        List<Integer> list = new ArrayList<>();
        list.add(1);
        list.add(2);
        list.add(3);
        testListIterator(list);
    }

    void testImmutable() {
        List<Integer> list = Arrays.asList(1, 2, 3);
        testListIterator(list);
    }

    <T> void testListIterator(List<T> list) {
        ListIterator<T> iter = list.listIterator();
        iter.next();
        iter.remove();
        System.out.println(list);
    }
}
```

Output:

```bash
[2, 3]
Exception in thread "main" java.lang.UnsupportedOperationException
	at java.util.AbstractList.remove(AbstractList.java:161)
	at java.util.AbstractList$Itr.remove(AbstractList.java:374)
	at javaslang.Test.testListIterator(Test.java:37)
	at javaslang.Test.testImmutable(Test.java:31)
	at javaslang.Test.main(Test.java:18)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)
```

javaslang/src/main/java/javaslang/collection/JavaConverters.java
seanf: True. But why not let `getDelegateForMutation()` be that one place?

javaslang/src/main/java/javaslang/collection/JavaConverters.java
seanf: I'm pretty sure _the j.u.AbstractList implementation_ won't change from the hashCode implementation which is specified for j.u.List. But I just want to make sure _the implementation used by the javaslang views_ never changes, lest it break that contract. So a comment to that effect in javaslang's hashCode method would be good. A test would be better.

javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,451 @@
...
+                         ? (U[]) new Object[length]
+                         : (U[]) java.lang.reflect.Array.newInstance(newType.getComponentType(), length);
+            } else {
+                target = array;
+            }
+            final Iterator<T> iter = iterator();
+            for (int i = 0; i < length; i++) {
+                target[i] = (U) iter.next();
+            }
+            return target;
```
seanf: Don't forget to set a `null` element if the array is oversized: https://docs.oracle.com/javase/8/docs/api/java/util/List.html#toArray-T:A-

javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,447 @@
...
+                 * call to {@code nextIndex} or {@code previousIndex}.)
+                 */
+                // may throw a ClassCastException accordingly to j.u.ListIterator.add(T)
+                list.add(index, value);
+                dirty = true;
+            }
+
+            private void checkDirty() {
+                if (dirty) {
+                    throw new IllegalStateException();
```
seanf: It's not clear from the JavaDocs - https://docs.oracle.com/javase/8/docs/api/java/util/ListIterator.html#add-E- - but perhaps `IllegalStateException` should be thrown in case of any change, even if the ListIterator wasn't the agent.

javaslang/src/main/java/javaslang/collection/JavaConverters.java
danieldietrich: Yes, I will add a complete unit test suit for j.u.List and j.u.ListIterator!

javaslang/src/main/java/javaslang/collection/JavaConverters.java
danieldietrich: These are already optimizations. Currently we have a minimal impl that does work for all kind of views. When adding Map (and Set) views we will refactor commonalities.

My first goal is to get the (upcoming) unit tests green.

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -567,6 +569,26 @@ private Array(Object[] delegate) {
...
+
+    @Override
+    public java.util.List<T> asJavaMutable() {
+        return JavaConverters.asJava(this, MUTABLE);
+    }
+
+    @Override
+    public Array<T> asJavaMutable(Consumer<? super java.util.List<T>> action) {
+        return Collections.asJava(this, action, MUTABLE);
+    }
```
danieldietrich: implemented for all sequences now

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -567,6 +569,26 @@ private Array(Object[] delegate) {
     }
 
     @Override
+    public java.util.List<T> asJava() {
+        return JavaConverters.asJava(this, IMMUTABLE);
+    }
+
+    @Override
+    public Array<T> asJava(Consumer<? super java.util.List<T>> action) {
```
danieldietrich: this notion has several advantages:

* j.u.List is scoped / does not escape
* fluent API
* we benefit from inference, i.e. not full qualified names required on name clashes between standard Java and Javaslang:

```Java
List.of(1, 2, 3)
    .asJavaMutable(list -> {
        // hah!
        list.add(4);
    })
    .filter(i -> i % 2 == 0); // = List(2, 4)
```

More specifically:

```Java
javaslang.collection.List<Integer> numbers = List.of(1, 2, 3);

// = List(2, 4)
javaslang.collection.List<Integer> filtered = numbers.asJavaMutable(list -> {
    // hah!
    list.add(4);
}).filter(i -> i % 2 == 0);

// numbers is still List(1, 2, 3)!
```

javaslang/src/main/java/javaslang/collection/Collections.java
```diff
@@ -37,6 +37,12 @@ static boolean areEqual(Iterable<?> iterable1, Iterable<?> iterable2) {
         return iter1.hasNext() == iter2.hasNext();
     }
 
+    static <T, C extends Seq<T>> C asJava(C source, Consumer<? super java.util.List<T>> action, ChangePolicy changePolicy) {
+        final ListView<T, C> view = JavaConverters.asJava(source, changePolicy);
+        action.accept(view);
+        return view.getDelegate();
+    }
```
danieldietrich: same for all seqs - set and map versions will follow

javaslang/src/main/java/javaslang/collection/Seq.java
```diff
@@ -139,6 +148,50 @@ default T apply(Integer index) {
...
+     * The difference to conversion methods {@code toJava*()} is that
+     *
+     * <ul>
+     * <li>A view is created in O(1) (constant time) whereas conversion takes O(n) (linear time), with n = collection size.</li>
+     * <li>The operations on a view have the same performance characteristics than the underlying persistent Javaslang collection whereas the performance characteristics of a converted collection are those of the Java standard collections.</li>
+     * </ul>
+     *
+     * @return A new immutable {@link java.util.Collection} view on this {@code Traversable}.
+     */
+    java.util.List<T> asJava();
```
danieldietrich: we can't define Traversable.asJava() that returns j.u.Collection&lt;T> because j.u.Map does not extend Collection.

javaslang/src/test/java/javaslang/collection/JavaConvertersTest.java
```diff
@@ -0,0 +1,178 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class JavaConvertersTest {
+
+    public abstract static class ListViewTest {
```
danieldietrich: The abstract test contains the methods that are common for immutable and mutable versions.
It is parameterized with a List factory, so we can test our test-cases with standard j.u.List impls and Javaslang List views.

It is important to test against j.u. Collections in order to ensure that our tests are correct and our views behave exactly as standard Java (e.g. when concurrent modifications occur, hashCode, equals and toString etc.)

(Now I will add missing tests...)

javaslang/src/test/java/javaslang/collection/JavaConvertersTest.java
```diff
@@ -0,0 +1,178 @@
...
+        public static java.util.Collection<Object[]> data() {
+            // DEV-NOTE: It is sufficient to test
+            // - one standard Java impl (<-- ensures our tests are correct)
+            // - one LinearSeq impl
+            // - one IndexedSeq impl
+            // to ensure the correctness of the immutable ListView impl.
+            return java.util.Arrays.asList(new Object[][] {
+                    { java.util.List.class.getName(), ListFactory.of(java.util.Arrays::asList) },
+                    { List.class.getName(), ListFactory.of(ints -> List.of(ints).asJava()) },
+                    { Vector.class.getName(), ListFactory.of(ints -> Vector.of(ints).asJava()) }
```
danieldietrich: We do not need test all of our collections - we only need to test, if our view impl is correct for all general types that are supported (like IndexedSeq, LinearSeq, Set, SortedSet, ...)

javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,433 @@
...
+ *
+ * @author Daniel Dietrich, Pap Lőrinc, Sean Flanigan
+ * @since 2.1.0
+ */
+class JavaConverters {
+
+    private JavaConverters() {
+    }
+
+    static <T, C extends Seq<T>> ListView<T, C> asJava(C seq, ChangePolicy changePolicy) {
```
danieldietrich: Added exact type parameter C and returning the view type in order to be able to pull out the original delegate later from within a Javaslang collection.

javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,433 @@
...
+            final int length = getDelegate().length();
+            if (array.length < length) {
+                final Class<? extends Object[]> newType = array.getClass();
+                target = (newType == Object[].class)
+                         ? (U[]) new Object[length]
+                         : (U[]) java.lang.reflect.Array.newInstance(newType.getComponentType(), length);
+            } else {
+                if (array.length > length) {
+                    Arrays.fill(array, length, array.length, null);
+                }
```
danieldietrich: filling the 'overlapping' tail now with nulls, as you suggested

javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,433 @@
...
+
+            @Override
+            public void add(T value) {
+                /* TODO:
+                 * The new element is inserted before the implicit
+                 * cursor: a subsequent call to {@code next} would be unaffected, and a
+                 * subsequent call to {@code previous} would return the new element.
+                 * (This call increases by one the value that would be returned by a
+                 * call to {@code nextIndex} or {@code previousIndex}.)
+                 */
```
danieldietrich: we will need good tests for set, add and remove in conjunction with next/prev and hasNext/hasPrev. This is tricky stuff when the ListView is mutable...

javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -104,30 +85,32 @@ boolean isMutable() {
...
             return delegate;
         }
 
-        C setDelegate(C newDelegate) {
+        @SuppressWarnings("unchecked")
+        protected C setDelegate(Traversable<?> newDelegate) {
             if (mutable) {
                 final C previousDelegate = delegate;
-                delegate = newDelegate;
+                delegate = (C) newDelegate;
```
seanf: @danieldietrich Why the signature change? Was the compiler being difficult?

javaslang/src/main/java/javaslang/collection/JavaConverters.java
seanf: The List javadoc only mentioned one null, not filling with nulls. Like a NUL-terminated string in C, I suppose.

danieldietrich: @seanf Thank you Sean, I will fix that! _(Note to myself: do not make assumptions, read the docs)_

javaslang/src/main/java/javaslang/collection/JavaConverters.java
danieldietrich: The types are ok to the point where we call methods on a Javaslang collection of type C:

```java
<T, C extends Seq<T> test(C seq) {
    // we know that the result is of type C but we get a Seq<T>
    seq.map(t -> t);
}
```

I see no other possibility than doing an unsafe cast. My first version had casts all over the View implementation. But we can do that unsafe cast in a central place, which seems to be better.

I agree that this is a fragile contruct. But we deal only with basic j.u.Collection (and j.u. Map) operations like add(), put(), remove(), ... It should be safe.

javaslang/src/main/java/javaslang/collection/Traversable.java
```diff
@@ -228,8 +228,12 @@
...
      */
     default boolean containsAll(Iterable<? extends T> elements) {
-        final HashSet<T> uniqueElements = HashSet.ofAll(elements);
-        return toSet().intersect(uniqueElements).size() == uniqueElements.size();
+        for (T element : elements) {
+            if (!contains(element)) {
+                return false;
+            }
+        }
+        return true;
```
danieldietrich: A mutable j.u.List view ran in a StackOverflowError when a List was added to itself and checked if it containsAll() of its own elements.

This simplification fixes that problem.

javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,432 @@
...
+            return mutable;
+        }
+
+        // DEV-NOTE: used within package to access the (original) persistent collection
+        C getDelegate() {
+            return delegate;
+        }
+
+        @SuppressWarnings("unchecked")
+        protected boolean setDelegateAndCheckChanged(Supplier<? extends Traversable<?>> delegate) {
```
danieldietrich: I feel guilty of writing this dirty-typed method.

javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,432 @@
...
+            final C newDelegate = (C) delegate.get();
+            final boolean changed = newDelegate != previousDelegate;
+            if (changed) {
+                this.delegate = newDelegate;
+            }
+            return changed;
+        }
+
+        @SuppressWarnings("unchecked")
+        protected void setDelegate(Supplier<? extends Traversable<?>> newDelegate) {
```
danieldietrich: dito

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+        final java.util.List<Integer> list = of(1);
+        assertThat(list.add(2)).isTrue();
+        assertThat(list).isEqualTo(javaList(1, 2));
+    }
+
+    @Test
+    public void shouldAddSubtypeToEndOfNonEmptyListView() {
+        abstract class A {
+            @Override
+            public int hashCode() {
```
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+                return 0;
+            }
+        }
+        final class B extends A {
+            @Override
+            public boolean equals(Object o) {
```
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+                return o instanceof B;
+            }
+        }
+        final class C extends A {
+            @Override
+            public boolean equals(Object o) {
```
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+    public void shouldAppendNull() {
+        final java.util.List<Object> list = empty();
+        assertThat(list.add(null)).isTrue();
+        assertThat(list).isEqualTo(javaList((Object) null));
+    }
+
+    @Test
+    public void shouldAddSelf() {
+        final java.util.List<Object> list = empty();
+        assertThat(list.add(list)).isTrue();
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldAddSelf\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+    // -- TODO: addAll(int, Collection)
+
+    // -- TODO: clear()
+
+    // -- contains(Object)
+
+    @Test
+    public void shouldRecognizeListContainsSelf() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldRecognizeListContainsSelf\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+        assertThat(list.contains(list)).isTrue();
+    }
+
+    // -- containsAll(Collection)
+
+    @Test
+    public void shouldRecognizeListContainsAllSelf() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldRecognizeListContainsAllSelf\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+        assertThat(list.containsAll(list)).isTrue();
+    }
+
+    // -- equals(Object)
+
+    @Test
+    public void shouldRecognizeSelfEqualityOfListThatContainsItself() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldRecognizeSelfEqualityOfListThatContainsItself\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+        assertThat(list.equals(list)).isTrue();
+    }
+
+    // -- hashCode()
+
+    @Test
+    public void shouldThrowInsteadOfLoopingInfinitelyWhenComputingHashCodeOfListThatContainsItself() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldThrowInsteadOfLoopingInfinitelyWhenComputingHashCodeOfListThatContainsItself\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+
+    // -- TODO: clear()
+
+    // -- contains(Object)
+
+    @Test
+    public void shouldRecognizeListContainsSelf() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
+        assertThat(list.contains(list)).isTrue();
```
sputnikci: \[FindBugs\] WARNING: DMI: Collections should not contain themselves in call to java\.util\.List\.contains\(Object\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    // -- containsAll(Collection)
+
+    @Test
+    public void shouldRecognizeListContainsAllSelf() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
+        assertThat(list.containsAll(list)).isTrue();
```
sputnikci: \[FindBugs\] INFO: DMI: For any collection c, calling c\.containsAll\(c\) doesn't make sense

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
danieldietrich: Ok: A is abstract, B and C extend A, they inherit hashCode and override equals.

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
danieldietrich: see above

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
danieldietrich: see above

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
danieldietrich: It is intended to test that corner case here!

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
danieldietrich: It is intended to test that corner case here!

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
danieldietrich: It is intended to test that corner case here!

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
danieldietrich: It is intended to test that corner case here!

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
danieldietrich: It is intended to test that corner case here!

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
danieldietrich: It is intended to test that corner case here!

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
danieldietrich: It is intended to test that corner case here!

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldAddSelf\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldRecognizeListContainsSelf\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldRecognizeListContainsAllSelf\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldRecognizeSelfEqualityOfListThatContainsItself\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldThrowInsteadOfLoopingInfinitelyWhenComputingHashCodeOfListThatContainsItself\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
sputnikci: \[FindBugs\] WARNING: DMI: Collections should not contain themselves in call to java\.util\.List\.contains\(Object\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
sputnikci: \[FindBugs\] INFO: DMI: For any collection c, calling c\.containsAll\(c\) doesn't make sense

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+        final java.util.List<Integer> list = of(1);
+        assertThat(list.add(2)).isTrue();
+        assertThat(list).isEqualTo(javaList(1, 2));
+    }
+
+    @Test
+    public void shouldAddSubtypeToEndOfNonEmptyListView() {
+        abstract class A {
+            @Override
+            public final int hashCode() {
```
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+                return 0;
+            }
+        }
+        final class B extends A {
+            @Override
+            public boolean equals(Object o) {
```
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+                return o instanceof B;
+            }
+        }
+        final class C extends A {
+            @Override
+            public boolean equals(Object o) {
```
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+    public void shouldAppendNull() {
+        final java.util.List<Object> list = empty();
+        assertThat(list.add(null)).isTrue();
+        assertThat(list).isEqualTo(javaList((Object) null));
+    }
+
+    @Test
+    public void shouldAddSelf() {
+        final java.util.List<Object> list = empty();
+        assertThat(list.add(list)).isTrue();
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldAddSelf\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+    // -- TODO: addAll(int, Collection)
+
+    // -- TODO: clear()
+
+    // -- contains(Object)
+
+    @Test
+    public void shouldRecognizeListContainsSelf() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldRecognizeListContainsSelf\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+        assertThat(list.contains(list)).isTrue();
+    }
+
+    // -- containsAll(Collection)
+
+    @Test
+    public void shouldRecognizeListContainsAllSelf() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldRecognizeListContainsAllSelf\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+        assertThat(list.containsAll(list)).isTrue();
+    }
+
+    // -- equals(Object)
+
+    @Test
+    public void shouldRecognizeSelfEqualityOfListThatContainsItself() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldRecognizeSelfEqualityOfListThatContainsItself\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+        assertThat(list.equals(list)).isTrue();
+    }
+
+    // -- hashCode()
+
+    @Test
+    public void shouldThrowInsteadOfLoopingInfinitelyWhenComputingHashCodeOfListThatContainsItself() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldThrowInsteadOfLoopingInfinitelyWhenComputingHashCodeOfListThatContainsItself\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+
+    // -- TODO: clear()
+
+    // -- contains(Object)
+
+    @Test
+    public void shouldRecognizeListContainsSelf() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
+        assertThat(list.contains(list)).isTrue();
```
sputnikci: \[FindBugs\] WARNING: DMI: Collections should not contain themselves in call to java\.util\.List\.contains\(Object\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    // -- containsAll(Collection)
+
+    @Test
+    public void shouldRecognizeListContainsAllSelf() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
+        assertThat(list.containsAll(list)).isTrue();
```
sputnikci: \[FindBugs\] INFO: DMI: For any collection c, calling c\.containsAll\(c\) doesn't make sense

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,240 @@
...
+        final java.util.List<Integer> list = of(1);
+        assertThat(list.add(2)).isTrue();
+        assertThat(list).isEqualTo(javaList(1, 2));
+    }
+
+    @Test
+    public void shouldAddSubtypeToEndOfNonEmptyListView() {
+        abstract class A {
+            @Override
+            public final int hashCode() {
```
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,240 @@
...
+                return 0;
+            }
+        }
+        final class B extends A {
+            @Override
+            public boolean equals(Object o) {
```
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,240 @@
...
+                return o instanceof B;
+            }
+        }
+        final class C extends A {
+            @Override
+            public boolean equals(Object o) {
```
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,240 @@
...
+    public void shouldAppendNull() {
+        final java.util.List<Object> list = empty();
+        assertThat(list.add(null)).isTrue();
+        assertThat(list).isEqualTo(javaList((Object) null));
+    }
+
+    @Test
+    public void shouldAddSelf() {
+        final java.util.List<Object> list = empty();
+        assertThat(list.add(list)).isTrue();
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldAddSelf\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,240 @@
...
+    // -- TODO: addAll(int, Collection)
+
+    // -- TODO: clear()
+
+    // -- contains(Object)
+
+    @Test
+    public void shouldRecognizeListContainsSelf() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldRecognizeListContainsSelf\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,240 @@
...
+        assertThat(list.contains(list)).isTrue();
+    }
+
+    // -- containsAll(Collection)
+
+    @Test
+    public void shouldRecognizeListContainsAllSelf() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldRecognizeListContainsAllSelf\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,240 @@
...
+        assertThat(list.containsAll(list)).isTrue();
+    }
+
+    // -- equals(Object)
+
+    @Test
+    public void shouldRecognizeSelfEqualityOfListThatContainsItself() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldRecognizeSelfEqualityOfListThatContainsItself\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,240 @@
...
+        assertThat(list.equals(list)).isTrue();
+    }
+
+    // -- hashCode()
+
+    @Test
+    public void shouldThrowInsteadOfLoopingInfinitelyWhenComputingHashCodeOfListThatContainsItself() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
```
sputnikci: \[FindBugs\] INFO: IL: A collection is added to itself in javaslang\.collection\.JavaConverters\_ListViewMutableTest\.shouldThrowInsteadOfLoopingInfinitelyWhenComputingHashCodeOfListThatContainsItself\(\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,240 @@
...
+
+    // -- TODO: clear()
+
+    // -- contains(Object)
+
+    @Test
+    public void shouldRecognizeListContainsSelf() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
+        assertThat(list.contains(list)).isTrue();
```
sputnikci: \[FindBugs\] WARNING: DMI: Collections should not contain themselves in call to java\.util\.List\.contains\(Object\)

javaslang/src/test/java/javaslang/collection/JavaConverters_ListViewMutableTest.java
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    // -- containsAll(Collection)
+
+    @Test
+    public void shouldRecognizeListContainsAllSelf() {
+        final java.util.List<Object> list = empty();
+        list.add(list);
+        assertThat(list.containsAll(list)).isTrue();
```
sputnikci: \[FindBugs\] INFO: DMI: For any collection c, calling c\.containsAll\(c\) doesn't make sense

javaslang/src/main/java/javaslang/collection/JavaConverters.java
```diff
@@ -0,0 +1,456 @@
...
+        @Override
+        public <U> U[] toArray(U[] array) {
+            Objects.requireNonNull(array, "array is null");
+            final U[] target;
+            final int length = getDelegate().length();
+            if (array.length < length) {
+                final Class<? extends Object[]> newType = array.getClass();
+                target = (newType == Object[].class)
+                         ? (U[]) new Object[length]
+                         : (U[]) java.lang.reflect.Array.newInstance(newType.getComponentType(), length);
```
ruslansennov: @GwtIgnore

danieldietrich: Yes, will do. Have to burn it into my brain windings that relection isn't gwt compatible :)

javaslang/src/test/java/javaslang/collection/JavaConvertersTest.java
```diff
@@ -0,0 +1,1001 @@
...
+            assertThat(list.add(2)).isTrue();
+            assertThat(list).isEqualTo(immutableList(1, 2));
+        });
+    }
+
+    @Test
+    public void shouldAddSubtypeToEndOfNonEmptyListView() {
+        abstract class A {
+            @Override
+            public final int hashCode() {
```
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConvertersTest.java
```diff
@@ -0,0 +1,1001 @@
...
+                return 0;
+            }
+        }
+        final class B extends A {
+            @Override
+            public boolean equals(Object o) {
```
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

javaslang/src/test/java/javaslang/collection/JavaConvertersTest.java
```diff
@@ -0,0 +1,1001 @@
...
+                return o instanceof B;
+            }
+        }
+        final class C extends A {
+            @Override
+            public boolean equals(Object o) {
```
sputnikci: \[PMD\] INFO: Ensure you override both equals\(\) and hashCode\(\)

danieldietrich:
Added this to 2.1.0 again. We are half the way there. It is an important interoperability feature that has to be shipped.

PR #1753, opened 2016-12-11
marks CompletableFuture as gwt incompatible

ruslansennov:
another way is to implement GWT version of `CompletableFuture` but it is very long story


danieldietrich:
Yes, implementing it seems to be overkill. Will merge it - thank you!

I think we still have an open issue triggering the GWT build. I will take a look today!

PR #1756, opened 2016-12-11
Try extended with function recoverOption

lukaszsoszynski:
I extended the Try class with the recoverOption method. The method has two parameters: exception class and function which accepts exception object and returns Option. The new method is very similar to the javaslang.control.Try#recover(java.lang.Class<X>, java.util.function.Function<? super X,? extends T>) but the new function is more convenient for me because it returns Option<T>. 

The new method solves my problems which occurs when I work with Try and Spring RestTemplate. RestTemplate throws HttpClientErrorException with HTTP response status code. If the status code is equal to 401 I am able to recover it and my recovery function can return not empty Option. But if the response code is equal to e.g. 403 the recovery function is not able to do its job and in this case it returns empty Option.

### DIFF
--- a/javaslang/src/main/java/javaslang/control/Try.java
+++ b/javaslang/src/main/java/javaslang/control/Try.java
@@ -5,19 +5,14 @@
  */
 package javaslang.control;
 
+import java.io.Serializable;
+import java.util.*;
+import java.util.function.*;
+
 import javaslang.Value;
 import javaslang.collection.Iterator;
-import javaslang.collection.Seq;
+import javaslang.collection.*;
 import javaslang.collection.Vector;
-
-import java.io.Serializable;
-import java.util.Arrays;
-import java.util.NoSuchElementException;
-import java.util.Objects;
-import java.util.function.Consumer;
-import java.util.function.Function;
-import java.util.function.Predicate;
-import java.util.function.Supplier;
 
 /**
  * An implementation similar to Scala's Try control.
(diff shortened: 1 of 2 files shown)


danieldietrich:
Hi Lukasz,

thank you for, this is a helpful addition!

We need a 'slight' modification. A `recoverWith` method that returns a `Try` instead of an option fits better because the API is 'symmetric' to the existing one.

Here are the existing methods (normal) and the new methods (**bold**):

recover():

* `Try<T> recover(Function<Throwable, T> f)`
* `Try<T> recover(Class<X> exception, T value)`
* `Try<T> recover(Class<X> exception, Function<X, T> f)`

recoverWith():

* `Try<T> recoverWith(Function<Throwable, Try<T>> f)`
* **`Try<T> recoverWith(Class<X> exception, Try<T> value)`**
* **`Try<T> recoverWith(Class<X> exception, Function<X, Try<T>> f)`**

where X extends Throwable (above the generic bounds are left away for better readability).

The implementation would look like this:

```java
    @GwtIncompatible
    default <X extends Throwable> Try<T> recoverWith(Class<X> exception, Try<T> recovered) {
        Objects.requireNonNull(exception, "exception is null");
        Objects.requireNonNull(recovered, "recovered is null");
        return (isFailure() && exception.isAssignableFrom(getCause().getClass()))
               ? recovered
               : this;
    }

    @GwtIncompatible
    default <X extends Throwable> Try<T> recoverWith(Class<X> exception, Function<? super X, ? extends Try<? extends T>> f) {
        Objects.requireNonNull(exception, "exception is null");
        if (isFailure()) {
            final Throwable cause = getCause();
            if (exception.isAssignableFrom(cause.getClass())) {
                try {
                    return (Try<T>) f.apply((X) cause);
                } catch (Throwable t) {
                    return new Failure<>(t);
                }
            }
        }
        return this;
    }
```

Could you add the above please?

---

**Note:** Your use-case (the 403 error) can be recovered the same way. Just use Try.success() and Try.failure() instead of Option.some() and Option.none(). Or Try.of() instead of Option.of().

If you require an Option for some reason then you can simple call

```java
Try.of(...).recoverWith(x, f).toOption();
```

lukaszsoszynski:
Daniel,
I followed your suggestion and I used Try instead of Option in recovery function. I think that this is a very good idea and I am perfectly fine with it. So finally I added two new functions to Try interface:
`
<X> Try<T> recoverWith(Class<X> exception, Function<X, Try<T>> f);
`
`
<X> Try<T> recoverWith(Class<X> exception,  Try<T> recovered)
` 
I am not sure if it is correct but I've assigned merge request back to you.

danieldietrich:
Hi Łukasz,
awesome - it is great to have a more complete Try now!

PR #1758, opened 2016-12-12
trigger ci build script

danieldietrich:


danieldietrich:
Fixes javaslang/javaslang-gwt#1

PR #1759, opened 2016-12-14
Fixes #1709 Gen.frequency uses only generators with positive frequencies

### LINKED ISSUE #1709: Gen.frequency(0, gen) should never return gen
Currently it may occur that a generator is returned that should occur with a frequency of 0.

This is a very theoretical case (because generators with a frequency of zero can be omitted at all) but we have to cover it.

iref:
I added filtering of provided generators based on their frequencies. Resulting generator now uses only generators with positive frequencies.


danieldietrich:
Many thanks, this looks great!

PR #1768, opened 2016-12-22
Unifies toString()

danieldietrich:


danieldietrich:
Note: I marked it as backward-compatible-relevant because the behaviour of Tree.toString changed.

PR #1770, opened 2016-12-23
Added overloaded forEach to Value - accepts ifEmptyAction param.

Opalo:
Hi,

I've provided an overloaded version of `Value`'s `forEach`. It accepts two arguments:
- an action to be run if `Value` is not empty
- another action to be run if `Value` is empty

### DIFF
--- a/javaslang/src/main/java/javaslang/Value.java
+++ b/javaslang/src/main/java/javaslang/Value.java
@@ -148,6 +148,7 @@ import static javaslang.API.*;
  *
  * @param <T> The type of the wrapped value.
  * @author Daniel Dietrich
+ * @author Maciej Opala
  * @since 2.0.0
  */
 public interface Value<T> extends Iterable<T> {
@@ -314,6 +315,26 @@ public interface Value<T> extends Iterable<T> {
         Objects.requireNonNull(action, "action is null");
         for (T t : this) {
             action.accept(t);
+        }
+    }
+
+    /**
+     * Performs an action on each element if this element is defined, otherwise invokes ifEmptyAction
+     *
+     * @param action action {@code Consumer}
+     * @param action ifEmptyAction {@code java.lang.Runnable}
+     * @throws NullPointerException if {@code action} is null
+     * @throws NullPointerException if {@code ifEmptyAction} is null
+     */
+    default void forEach(Consumer<? super T> action, Runnable ifEmptyAction) {
+        Objects.requireNonNull(action, "action is null");
+        Objects.requireNonNull(ifEmptyAction, "ifEmptyAction is null");
+        if (isEmpty()) {
+            ifEmptyAction.run();
+        } else {
+            for (T t : this) {
+                action.accept(t);
+            }
         }
     }
 
--- a/javaslang/src/test/java/javaslang/AbstractValueTest.java
+++ b/javaslang/src/test/java/javaslang/AbstractValueTest.java
@@ -39,6 +39,7 @@ import java.util.function.Supplier;
 import static javaslang.API.*;
 import static javaslang.Serializables.deserialize;
 import static javaslang.Serializables.serialize;
+import static org.junit.Assert.fail;
 
 public abstract class AbstractValueTest {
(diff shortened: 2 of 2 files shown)


danieldietrich:
Hi @Opalo,

thank you for your contribution! I've checked your addition. It is well done but it does not fit into our overall design.

Last month I received a similar PR: #1682 that targeted to add Option.fold(). There I linked an article

> But then I read the following article and I agree: [Option.fold() considered unreadable](http://www.nurkiewicz.com/2014/06/optionfold-considered-unreadable.html)

which cites Martin Odersky:

> "I personally find methods (...) that take two closures as arguments are often overdoing it."

I think this is also the case here. Beside that, a special method for Option does not scale very well. We could place it in a super type, say Value, but the API would overlap with Try.onSuccess(), Try.onFailure() etc.

A mid-term goal is to reduce the Javaslang API. Version 3.0.0 will be a first step to simplify the API. With Project Valhalla we will have primitive generics + void, i.e. `Function<void, void>` will be the new `Runnable` and `Function<T, void>` the new Consumer. Given that we can express side-effects easily with the existing API, e.g. `option.map(consumer).getOrElse(runnable)`.

---

Generally using Values for the control flow of side-effecting programs is discouraged. The only thing I can suggest here is using if-branches:

```java
if (option.isDefined()) {
    option.forEach(t -> {
        // performSideEffects
    });
} else {
   // perform side-effects
}
```

... or of course write your own helper.

Hope you understand that I will close the PR.

\- Daniel

Opalo:
Hi @danieldietrich,

First of all, thanks for having time for looking at my PR and for great OS library 👍  Actually I didn't know about #1682 but my work was also loosely inspired on `Option.fold` from scala. 

In general I do agree with you, however having an overloaded version of `forEach` is something that is returning very often in various discussions with my fellow developers. `forEach` is a very elegant way to finish processing of a particular item (e.g. an `Option` that was mapped multiple times previously). Not having possibility other than using cumbersome `isDefined` and if statement makes the code looks just ugly. Summing up, this is not the way I'd like to go in this particular case.

Nevertheless, thanks once again for you work and I'm looking forward to using JavaSlang v. 3.0. If you need any help, just let me know 😉

danieldietrich:
Hi @Opalo, I understand the need for handling 'emptiness' in a concise way. 

I've thought about adding Value.onEmpty() as counterpart to Value.forEach(). But it complicates things:

* other types like Try (and Future) already have special handlers like onSuccess(), onFailure()
* Scala 2.12 reduces these redundancies by falling back to forEach() again (see [Futures in Scala 2.12](https://github.com/viktorklang/blog/blob/master/Futures-in-Scala-2.12-part-5.md)).

For Either, Future and Try it is already implemented:

* future.failed().forEach()
* try_.failed().forEach()
* either.left().forEach(), either.right.forEach()

It seems that Validation needs a similar way to handle invalid results.

Option is different than the above because None does not contain any value that could be consumed. So the _dual_ forEach() approach does not work here. Therefore I suggest to add a method `onEmpty()` in favour of a forEach() with two arguments:

```java
interface Option<T> {
    default Option<T> onEmpty(Runnable action) {
        if (isEmpty()) {
            action.run();
        }
        return this;
    }
}
```

Example:

```java
option
    .onEmpty(() -> {
        /* perform side-effect */
    })
    .forEach(t -> ...);
```

What do you think?

Opalo:
Well, that looks definitely awesome! I also had thought about returning `this` from `forEach` and providing a kind of `runIfEmpty` method however that was impossible because of `Iterable` contract.

Your idea provides the missing value: handles the _emptiness_ in a concise way. I'm happy to issue an another PR along with tests to provide this _empitness_ handling in `Option` class.

danieldietrich:
@Opalo that would be awesome! Looking forward to your PR

Opalo:
@danieldietrich sorry for the delay, PR will be ready just after New Year 🎉

danieldietrich:
I will re-open this issue and adjust the name...

PR #1771, opened 2016-12-24
Add euler 45 solution

DAllford:


danieldietrich:
Hi Dan,

thank you for your contribution, it looks great! There are some minor issues: we make local variables final where possible and the formatting looks different to our project formatter (see .ide/) - but I will pull it in as-is because it is test code. From time to time we apply our formatter to the whole code-base.

Thx!

PR #1782, opened 2016-12-30
Update HashSet.java

chagmed:
fixed javadoc typo


danieldietrich:
Thanks for the fix :)

Issue #1781, opened 2016-12-30
Add Option.onEmpty()


danieldietrich:
I've thought about adding Value.onEmpty() as counterpart to Value.forEach(). But it complicates things:

* other types like Try (and Future) already have special handlers like onSuccess(), onFailure()
* Scala 2.12 reduces these redundancies by falling back to forEach() again (see [Futures in Scala 2.12](https://github.com/viktorklang/blog/blob/master/Futures-in-Scala-2.12-part-5.md)).

For Either, Future and Try it is already implemented:

* future.failed().forEach()
* try_.failed().forEach()
* either.left().forEach(), either.right.forEach()

It seems that Validation needs a similar way to handle invalid results.

Option is different than the above because None does not contain any value that could be consumed. So the _dual_ forEach() approach does not work here. Therefore I suggest to add a method `onEmpty()` in favour of a forEach() with two arguments:

```java
interface Option<T> {
    default Option<T> onEmpty(Runnable action) {
        if (isEmpty()) {
            action.run();
        }
        return this;
    }
}
```

Example:

```java
option
    .onEmpty(() -> {
        /* perform side-effect */
    })
    .forEach(t -> ...);
```

**See also #1770**

/cc @Opalo

PR #1791, opened 2017-01-03
Added Option.onEmpty() (#1781)

Opalo:
Hi,

As discussed in #1781 I've provided `onEmpty` method to perform side effects for `Option` class.

### DIFF
--- a/javaslang/src/main/java/javaslang/control/Option.java
+++ b/javaslang/src/main/java/javaslang/control/Option.java
@@ -171,6 +171,20 @@ public interface Option<T> extends Value<T> {
     boolean isEmpty();
 
     /**
+     * Runs a Java Runnable passed as parameter if this {@code Option} is empty.
+     *
+     * @param action a given Runnable to be run
+     * @return this {@code Option}
+     */
+    default Option<T> onEmpty(Runnable action) {
+        Objects.requireNonNull(action, "action is null");
+        if (isEmpty()) {
+            action.run();
+        }
+        return this;
+    }
+
+    /**
      * Returns true, if this is {@code Some}, otherwise false, if this is {@code None}.
      * <p>
      * Please note that it is possible to create {@code new Some(null)}, which is defined.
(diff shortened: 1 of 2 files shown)


javaslang/src/main/java/javaslang/control/Option.java
```diff
@@ -32,6 +32,7 @@
  *
  * @param <T> The type of the optional value.
  * @author Daniel Dietrich
+ * @author Maciek Opala
```
ruslansennov: I believe this is too small contribution to add @author annotation

Opalo: @ruslansennov sure! Wasn't sure about contribution rules. Will remove it.

javaslang/src/main/java/javaslang/control/Option.java
```diff
@@ -171,6 +171,19 @@
     boolean isEmpty();
 
     /**
+     * Runs a Java Runnable passed as parameter if this {@code Option} is empty.
+     *
+     * @param action a given Runnable to be run
+     * @return this {@code Option}
+     */
+    default Option<T> onEmpty(Runnable action) {
```
danieldietrich: Please add a null-check, we do it by convention for better readability of stack traces:

```java
Objects.requireNonNull(action, "action is null");
```

Opalo: @danieldietrich you are absolutely right - sorry for not adding it. I've amended the commit with required null-check.

danieldietrich: Great, thank you!

danieldietrich:
Lovely feature, thank you for your patience 🙏

Opalo:
Always happy to help! Thanks for accepting the PR, we really need this feature :)

danieldietrich:
I will ship 2.1.0-beta as soon as possible. However, it will still take some time...

PR #1792, opened 2017-01-04
Fixes deadlock on concurrent class initialization

### LINKED ISSUE #1773: Deadlock on Static Class Initialization of Iterator
javaslang.collection.Iterator has public final static field EMPTY which is initialized by an instance of anonymous class (extends AbstractIterator). If AbsctractorIterator is loaded in another thread, it may lead to deadlock.

See: https://bugs.openjdk.java.net/browse/JDK-8037567
and the snippet for more information.

In the snippet:

- Thread 1: loads Iterator -> AbstractorIterator
- Thread 2: loads A -> AbstractIterator -> Iterator

The snippet almost always hangs.

```
package slang;

import javaslang.collection.AbstractIterator;

/**
 * Created by aeremeev on 25.12.16.
 */
public class DeadlockTest {

        new Thread(Iterator::empty).start();
        new A();
        System.out.println("Success");
    }

    private static class A extends AbstractIterator<Void> {
        @Override
        protected Void getNext() {
            return null;
        }

        @Override
        public boolean hasNext() {
            return false;
        }
    }
}
```

danieldietrich:
Fixes #1773

### DIFF
--- a/javaslang/src/test/java/javaslang/collection/IteratorTest.java
+++ b/javaslang/src/test/java/javaslang/collection/IteratorTest.java
@@ -18,6 +18,8 @@ import org.junit.Test;
 import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.NoSuchElementException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collector;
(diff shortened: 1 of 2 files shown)


danieldietrich:
Hi @eremeev, @igoonich, a final review is greatly appreciated :)

I introduced a new final class EmptyIterator, like you suggested. Also it is derived from Iterator instead of AbstractIterator. I took a look at java.util.Collections.EmptyIterator.

I think the unit test will never fail because we cannot enforce to run the test _before_ the Iterator class was initialized. When running this test in a big test suite (currently &gt; 10000 unit tests) we may need a custom ClassLoader implementation for our test. But I think this is a bit too much and also a bit error prone because we introduce too much additional code for a unit test. In a perfect world a unit test introduces no additional code at all (only calls and assertions).

Thanks in advance,

/- Daniel

javaslang/src/test/java/javaslang/collection/IteratorTest.java
```diff
@@ -515,4 +517,33 @@ public void shouldRecognizeUnequalObjects() {
         // Iterator equality undefined
     }
 
+    // -- class initialization (see #1773)
+
+    @Test(timeout = 5_000)
+    public void shouldNotDeadlockOnConcurrentClassInitialization() throws InterruptedException {
+        final ExecutorService executorService = Executors.newFixedThreadPool(2);
+        executorService.execute(new ClassInitializer(javaslang.collection.Iterator.class.getName()));
```
eremeev: Iterator and AbstractIterator will be loaded in the main thread, thus, no concurrency.
You have to pass string constants here.

danieldietrich: Thank you, you are absolutely right. Will push a fix.

danieldietrich:
@eremeev thanks for the review

hepin1989:
thank you very much for this fix

danieldietrich:
@hepin1989 thx, @eremeev and @igoonin did the real work!

PR #1799, opened 2017-01-05
2016 -> 2017

danieldietrich:
Same procedure as every year...


danieldietrich:
**Note:** Some files changed because the year-update script might also change file endings. Just saying.

PR #1801, opened 2017-01-05
Ensures that Traversable modifications return the same instance if possible

### LINKED ISSUE #1800: Ensure that Traversable modifications return the same instance if possible
# Phase 1

We focus on _empty.op(non-empty)_ and _non-empty.op(empty)_ cases (exception: _unit_).

## Set

* [x] addAll (depends on #1827)
* [x] add(value) if value is already present
* [x] diff / difference
* [x] intersect
* [x] removeAll
* [x] union
* [x] distinct()

## Map

* [x] merge(Map), merge(Map, BiFunction)
* [x] removeAll
* [x] distinct

## Seq

* [x] appendAll
* [x] insertAll
* [x] prependAll
* [x] removeAll

# Phase 2

We focus on _empty.op(f)_ where f in { Function, BiFunction, Predicate, BiPredicate }.

## Traversable

* [x] empty().filter(...)
* [x] empty.distinctBy(Function)
* [x] nonEmpty.drop(0)
* [x] nonEmpty.dropRight(0)
* [x] empty.drop(1)
* [x] empty.dropRight(1)
* [x] empty.take(1)
* [x] empty.takeRight(1)
* [x] empty.takeWhile()
* [x] empty.takeUntil()
* [x] empty.dropWhile()
* [x] empty.dropUntil()

# Phase 3

We focus on _collection.op(f) == collection_, where collection != empty.

## Traversable

* [x] take(size())
* [x] takeRight(size())

## Seq

* [x] subSequence(0), subSequence(0, length)

_Any missing?_

**Note:** We do not over-optimize, e.g. by returning the same instance when updating a collection at a specific index with the same object using update(int, Object).

---

**Note:** Generally we are now able to determine if a Traversable is ordered, i.e. has an underlying Comparator.

In the presence of a Comparator some operations cannot ensure any more that the same instance is returned under some circumstances, e.g. the below sets are inherently different, if the empty set has an ascending order and the non-empty set has a descending order:

```
// = (3, 2, 1)
().union((3, 2, 1))
```

danieldietrich:
Fixes #1800

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
+++ b/javaslang/src/main/java/javaslang/collection/AbstractMultimap.java
@@ -177,19 +177,20 @@ abstract class AbstractMultimap<K, V, M extends Multimap<K, V>> implements Multi
     @Override
     public M distinctBy(Comparator<? super Tuple2<K, V>> comparator) {
         Objects.requireNonNull(comparator, "comparator is null");
-        return (M) createFromEntries(iterator().distinctBy(comparator));
+        return (M) (isEmpty() ? this : createFromEntries(iterator().distinctBy(comparator)));
     }
 
+    @SuppressWarnings("unchecked")
     @Override
-    public <U> Multimap<K, V> distinctBy(Function<? super Tuple2<K, V>, ? extends U> keyExtractor) {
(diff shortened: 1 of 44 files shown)


danieldietrich:
Current status:

screen shot 2017-01-11 at 00 55 34

danieldietrich:
@ruslansennov hah, no merge conflicts :)

Issue #1659, opened 2016-11-04
Add a static transpose to all Traversables


danieldietrich:
Showing it for List here:

```java
class List {

    // transforms the given rows to columns, truncating the columns to the minimum length
    static <T> Array<Seq<T>> transpose(Iterable<? extends Iterable<? extends T>> rows) { ... }

    // transforms the given rows to columns, filling the columns with the defaultValue to the maximum length
    static <T> Array<Seq<T>> transpose(Iterable<? extends Iterable<? extends T>> rows, T defaultValue) { ... }

}
```

Example:

```java
List<List<Integer>> list = List.of(
    List.of(1, 2, 3),
    List.of(4, 5, 6),
    List.of(7, 8, 9));

// = List((1, 4, 7), (2, 5, 8), (3, 6, 9))
List<Seq<Integer>> zipped = List.zip(list);
```

The resulting sequences need to be truncated to the minimum length here.

danieldietrich:
Maybe we should be more specific regarding the component type of the argument in order to make use of the random-access property of IndexedSeq's when transposing.

Also we could make use of `<T>` instead of `<? extends T>` because we only change the structure, not the type, i.e.

```java
    // transforms the given rows to columns, truncating the columns to the minimum length
    static <T> Array<Seq<T>> transpose(Iterable<Array<T>> rows) { ... }

    // transforms the given rows to columns, filling the columns with the defaultValue to the maximum length
    static <T> Array<Seq<T>> transpose(Iterable<Array<T>> rows, T defaultValue) { ... }
```

danieldietrich:
See also [Transposing arbitrary collection-of-collections in Scala](http://stackoverflow.com/questions/10662462/transposing-arbitrary-collection-of-collections-in-scala)

danieldietrich:
We should define static methods in these (and only these) types:

```java
interface Array<T> {
    static <T> Array<Array<T>> transpose(Iterable<? extends Iterable<? extends T>> rows);
    static <T> Array<Array<T>> transpose(Iterable<? extends Iterable<? extends T>> rows, T defaultValue);
}

interface CharSeq<T> {
    static CharSeq<CharSeq> transpose(Iterable<? extends Iterable<Character>> rows);
    static CharSeq<CharSeq> transpose(Iterable<? extends Iterable<Character>> rows, char defaultValue);
}
```

and so on. We do this for all Seq implies, i.e. all types that have indexed elements:

* Array
* CharSeq
* Vector
* List
[...]

PR #1803, opened 2017-01-06
Implemented `transpose` in `Seq`

AlparSzabados:
Fixes: https://github.com/javaslang/javaslang/issues/1659
Note: ~I only implemented it in `Array`, `List` and `Vector`. It made no sense for all Traversables to have it, but if you want, I can implement it in others too. Since I found no common interface for these three, I had to implement it in each one separately.~
I implemented it in `Seq` except `CharSeq` because this would need a different implementation.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Array.java
+++ b/javaslang/src/main/java/javaslang/collection/Array.java
@@ -554,6 +554,21 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
         return Iterator.unfold(seed, f).toArray();
     }
 
+    /**
+     * Transposes the rows and columns of an {@link Array}.
+     *
+     * @param rows to be transposed.
+     * @return a transposed {@link Array}.
+     * <p>
+     * ex: {@code
+     * Array.transpose(Array(Array(1,2,3), Array(4,5,6))) → Array(Array(1,4), Array(2,5), Array(3,6))
+     * Array.transpose(Array(Array(1,2), Array(3))) → Array(Array(1,3), Array(2))
+     * }
+     */
+    static <T> Array<Array<T>> transpose(Array<Array<T>> rows) {
+        return Collections.transpose(rows, Array::ofAll);
+    }
+
     @Override
     public Array<T> append(T element) {
         final Object[] copy = copyOf(delegate, delegate.length + 1);
@@ -1048,11 +1063,15 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
 
     @Override
     public Array<T> reverse() {
-        final Object[] arr = new Object[delegate.length];
-        for (int i = 0; i < delegate.length; i++) {
-            arr[delegate.length - 1 - i] = delegate[i];
+        if (size() <= 1) {
+            return this;
+        } else {
+            final Object[] arr = new Object[delegate.length];
+            for (int i = 0; i < delegate.length; i++) {
+                arr[delegate.length - 1 - i] = delegate[i];
+            }
+            return wrap(arr);
         }
-        return wrap(arr);
     }
 
     @Override
--- a/javaslang/src/main/java/javaslang/collection/Array.java
+++ b/javaslang/src/main/java/javaslang/collection/Array.java
@@ -1066,9 +1066,10 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
         if (size() <= 1) {
             return this;
         } else {
-            final Object[] arr = new Object[delegate.length];
-            for (int i = 0; i < delegate.length; i++) {
-                arr[delegate.length - 1 - i] = delegate[i];
+            final int length = delegate.length;
+            final Object[] arr = new Object[length];
+            for (int i = 0, j = length - 1; i < length; i++, j--) {
+                arr[j] = delegate[i];
             }
             return wrap(arr);
         }
--- a/javaslang/src/test/java/javaslang/collection/AbstractSeqTest.java
+++ b/javaslang/src/test/java/javaslang/collection/AbstractSeqTest.java
@@ -106,6 +106,8 @@ public abstract class AbstractSeqTest extends AbstractTraversableRangeTest {
     @Override
     abstract protected Seq<Long> rangeClosedBy(long from, long toInclusive, long step);
 
+    abstract protected <T> Seq<? extends Seq<T>> transpose(Seq<? extends Seq<T>> rows);
+
     // -- static narrow
 
     @Test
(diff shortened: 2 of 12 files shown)


javaslang/src/test/java/javaslang/collection/ArrayTest.java
```diff
@@ -204,6 +204,132 @@ public void shouldTransform() {
...
+                of(2, 4, 6));
+        assertThat(actual.transpose()).isEqualTo(expected);
+    }
+
+    @Test
+    public void shouldBeEqualIfTransposedTwice() {
+        Array<Array<Integer>> actual = of(
+                of(1, 2, 3),
+                of(4, 5, 6));
+        assertThat(actual.transpose().transpose()).isEqualTo(actual);
```
paplorinc: This property only holds for iterable `T`s, right?

javaslang/src/test/java/javaslang/collection/ListTest.java
```diff
@@ -293,6 +293,131 @@ public void shouldTransform() {
...
+    // -- transpose()
+
+    @Test
+    public void shouldTransposeIfEmpty() {
+        final List<Integer> actual = empty();
+        assertThat(actual.transpose()).isSameAs(actual);
+    }
+
+    @Test
+    public void shouldTransposeIfSingleValued() {
```
paplorinc: should rather be something like:
```java
    @Test
    public void shouldTransposeIfSingleValued() {
        final List<Integer> actual = of(0);
        final List<List<Integer>> expected = of(of(0));
        assertThat(actual.transpose()).isEqualTo(expected);
    }
```

javaslang/src/test/java/javaslang/collection/VectorTest.java
```diff
@@ -318,4 +318,129 @@ public void shouldReturnSelfOnConvertToVector() {
...
+    // -- transpose()
+
+    @Test
+    public void shouldTransposeIfEmpty() {
+        final Vector<Integer> actual = empty();
+        assertThat(actual.transpose()).isSameAs(actual);
+    }
+
+    @Test
+    public void shouldTransposeIfSingleValued() {
```
paplorinc: should rather be something like:
```java
    @Test
    public void shouldTransposeIfSingleValued() {
        final Vector<Integer> actual = of(0);
        final Vector<Vector<Integer>> expected = of(of(0));
        assertThat(actual.transpose()).isEqualTo(expected);
    }
```

javaslang/src/test/java/javaslang/collection/ListTest.java
AlparSzabados: done, and thanks

javaslang/src/test/java/javaslang/collection/VectorTest.java
AlparSzabados: done

danieldietrich:
Hi @AlparSzabados,

your implementation is different than the description of the original issue #1659 in one important point: **transpose() needs to be static**.

The main reason is that Java is missing higher-kinder types / type constructors. transpose() is very much like flatten(), it is defined on nested structures only.

You try to 'heal' that issue by implementing some hidden 'magic', more specifically, the elements are wrapped in an Iterable if they are not of type Seq. **This is not a good idea because it is not type-safe.**

**Example:**

You defined

```java
interface List<T> {
    default <U> List<List<U>> transpose()
}
```

Given that the following will compile:

```java
List<Integer> list = ...;
List<List<String>> transposed = list.transpose();
```

---

I suggest to create static methods as requested in the issue. I think we should discuss the parameter types and return types:

```java
interface List<T> {

    // return type?
    static <T> List<List<T>> transpose(Iterable<? extends Iterable<? extends T>> rows);
    static <T> List<Seq<T>> transpose(Iterable<? extends Iterable<? extends T>> rows);

    // parameter types?
    static <T> List<List<T>> transpose(Iterable<? extends Iterable<? extends T>> rows);
    static <T> List<List<T>> transpose(Iterable<? extends Iterable<T>> rows);
    static <T> List<List<T>> transpose(Iterable<List<? extends T>> rows);
    static <T> List<List<T>> transpose(Iterable<List<T>> rows);

}
```

danieldietrich:
Btw - many thanks for your PR ;-)

AlparSzabados:
@danieldietrich Thank you for your detailed response. I will try to modify it in a few days. Thanks again for the detailed description 👍 😃

danieldietrich:
@AlparSzabados great to hear. I updated the issue: https://github.com/javaslang/javaslang/issues/1659#issuecomment-270940446

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -1048,11 +1048,15 @@ public int length() {
...
-        final Object[] arr = new Object[delegate.length];
-        for (int i = 0; i < delegate.length; i++) {
-            arr[delegate.length - 1 - i] = delegate[i];
+        if (size() <= 1) {
+            return this;
+        } else {
+            final Object[] arr = new Object[delegate.length];
+            for (int i = 0; i < delegate.length; i++) {
+                arr[delegate.length - 1 - i] = delegate[i];
+            }
```
danieldietrich: We could simplify the operations a bit (read delegate length, subtract 1 and subtract i) like this:

```java
final int length = delegate.length;
for (int i = 0, j = length - 1; i < length; i++, j--) {
    arr[j] = delegate[i];
}
```

paplorinc: Indeed, your version makes it obvious that the two iterate in opposite directions, 
We also repeat the `delegate.length - 1`, which is a constant (the JVM might be smart enough to detect it, I don't know).
Your example could be simplified further by inlining `length`: it might also be more optimal for the JVM to know that we go through the whole array (it could unroll the loop better, if it knows the size in advance), and using `i` and `j` in the declaration order:
```java
for (int i = 0, j = delegate.length - 1; i < delegate.length; i++, j--) {
    arr[i] = delegate[j];
}
```

danieldietrich: @paplorinc 

> and using i and j in the declaration order:

highly subjective but I don't mind :) could be also seen as 1st read from source 2nd write to target. It is just subjective

The length in `i < delegate.length` is read from heap. When introducing `final int length = delegate.length`, length is put on the stack. The loop accesses length n times, so it might matter.

paplorinc: If you think it's a bottleneck, we could create a `JMH` for it.
In my experience non-algorithmic complexity cannot be guessed without measurement! The JVM can eliminate simply eliminate it and reorder whatever it wishes, as long as the visible results are the same.

danieldietrich: @paplorinc I not completely agree. The are things defined by the JVM spec that are crystal clear, e.g. heap vs. stack access.  Stack access can't be faster (other than register access), only heap access can be faster. There are two usages of `delegate.length`, which is kind of dedundant. Adding a local length variable also clarifies that the same length is meant.

Regarding the existing issues I would add a priority $numberOfIssues (greater is less important) to adding a benchmark for the use of a local variable.

paplorinc: > The are things defined by the JVM spec that are crystal clear, e.g. heap vs. stack access.

Except that it's not :). `Escape analysis` can cheat and allocate objects on the stack, instead of the heap to unburdon the GC, see: https://docs.oracle.com/javase/8/docs/technotes/guides/vm/performance-enhancements-7.html#escapeAnalysis

danieldietrich: I think this form of optimization makes most sense for code that is not optimal. But it should not prevent us from writing optimal code if it is still concise and maintainable.

It might also reduce the effort for an escape analysis. JIT: "Already on stack? Ok, nothing to do."

paplorinc: I agree, we shouldn't optimize prematurely, for same complexity, start with the version that's more readable :)

danieldietrich: I see, we have different styles of coding and a different approach to conciseness. It's ok.

paplorinc: Not necessarily, I like your version also, I just wanted to point out that it may not be an optimization at all, it may even be slower, so we should either choose one, it because it looks better, or let's measure it and take that into consideration also. :)

danieldietrich: @AlparSzabados please use the following snippet.

```java
final int length = delegate.length;
for (int i = 0, j = length - 1; i < length; i++, j--) {
    arr[j] = delegate[i];
}
```

(Too much discussion for minor things here)

javaslang/src/test/java/javaslang/collection/AbstractSeqTest.java
```diff
@@ -1800,4 +1802,122 @@ public void withDefault() {
...
+    public void shouldTransposeIfEmpty() {
+        final Seq<Seq<Integer>> actual = empty();
+        assertThat(transpose(actual)).isSameAs(actual);
+    }
+
+    @Test
+    public void shouldTransposeIfSingleValued() {
+        final Seq<Seq<Integer>> actual = of(of(0));
+        final Seq<Seq<Integer>> expected = of(of(0));
+        assertThat(transpose(actual)).isEqualTo(expected);
```
paplorinc: https://github.com/javaslang/javaslang/pull/1803/commits/e6fdea3781f76c22eb5ff902e6843c98cc82ebb1#diff-8eddfb25d90b8bf7a56bafbd64d68a74R307

paplorinc: @danieldietrich, would you expect this to be `isSameAs` also?

javaslang/src/test/java/javaslang/collection/AbstractSeqTest.java
```diff
@@ -1800,4 +1802,122 @@ public void withDefault() {
...
+                of(2, 3, 4, 5),
+                of(),
+                of(6, 7, 8));
+        Seq<Seq<Integer>> expected = of(
+                of(0, 2, 6),
+                of(1, 3, 7),
+                of(4, 8),
+                of(5));
+        assertThat(transpose(actual)).isEqualTo(expected);
+    }
```
paplorinc: :+1: awesome suite of tests!

javaslang/src/test/java/javaslang/collection/AbstractSeqTest.java
danieldietrich: @paplorinc We could do that. However, I checked the code. It looks good to me, the original issue should be fixed. I will move the optimization you suggested to #1800.

danieldietrich:
@AlparSzabados Many thanks, well done & awesome tests!

Issue #1811, opened 2017-01-12
Add factory method to Try for Supplier interface


RobWin:
* [x] Try
* [x] Future

---

Hi,

I often use the Supplier interface so that I can decorate a method call and then I want to invoke the Supplier interface. It would be nice if `Try` would have a factory method for the Supplier interface. For example:

```
   Supplier<String> supplier = () -> businessService.doSomething("input");
   Supplier<String> decoratedSupplier = CircuitBreaker.decorateSupplier(circuitBreaker, supplier)
   Try.of(decoratedSupplier).get();
```

Instead of

```
   Try.of(decoratedSupplier::get).get()
```

The same also applies to the `Callable` interface.

What do you think?

danieldietrich:
I'm afraid, this is not possible - at least when using the same method name `of`, because it will raise ambiguities.

From the viewpoint of the compiler

* a lambda `() -> retVal` matches both signatures `Supplier`, `CheckedSupplier` and `Callable`
* a lambda `() -> {}` matches the signatures `Runnable`, `CheckedRunnable`

So only one is possible when having the same method name `of`.

I think it is not worth to introduce additional API, e.g.

* `Try.ofCallable(Callable)`
* `Try.ofSupplier(Supplier)`
* ...

Just calling

* `Try.of(supplier::get)`

should be sufficient.

---

If you agree we should close the ticket.

RobWin:
Damn, that's true. Haven't thought of that, if you use a lamba directly instead of an interface.
You have to decide it. Other libraries like RxJava provide factory methods for this use case, for example:

`Observable. fromCallable()`

danieldietrich:
I thought a moment about removing Try.CheckedSupplier (in 3.0.0) and using Callable instead. But a CheckedSupplier throws Throwable and Callable throws Exception. It does not completely fit our needs.

We use the static methods `fromXxx()` to convert a value _from_ one to another. It is the dual operation to `toXxx()`. Example:

* Validation.fromEither()
* Validation.toEither()

To to disambiguate `of(CheckedSupplier)` and `of(Callable)`, the right name would be `Try.ofCallable(Callable)` (`fromCallable` does not fit because no conversion takes place).

To to disambiguate `of(CheckedRunnable)` and `run(Runnable)`, the right name would be `Try.runRunnable(Runnable)` (which looks a bit weird).

[...]

PR #1814, opened 2017-01-12
Issue #1811: Added factory methods to Try for Supplier, Callable and Runnable

RobWin:
* Added factory methods to Try for Supplier, Callable and Runnable

### DIFF
--- a/javaslang/src/main/java/javaslang/control/Try.java
+++ b/javaslang/src/main/java/javaslang/control/Try.java
@@ -56,6 +56,7 @@ public interface Try<T> extends Value<T> {
      * exception occurs calling {@code supplier.get()}.
      */
     static <T> Try<T> ofSupplier(Supplier<? extends T> supplier) {
+        Objects.requireNonNull(supplier, "supplier is null");
         return of(supplier::get);
     }
 
@@ -68,6 +69,7 @@ public interface Try<T> extends Value<T> {
      * exception occurs calling {@code callable.call()}.
      */
     static <T> Try<T> ofCallable(Callable<? extends T> callable) {
+        Objects.requireNonNull(callable, "callable is null");
         return of(callable::call);
     }
 
@@ -96,6 +98,7 @@ public interface Try<T> extends Value<T> {
      * calling {@code runnable.run()}.
      */
     static Try<Void> runRunnable(Runnable runnable) {
+        Objects.requireNonNull(runnable, "runnable is null");
         return run(runnable::run);
     }
 
--- a/javaslang/src/main/java/javaslang/control/Try.java
+++ b/javaslang/src/main/java/javaslang/control/Try.java
@@ -38,6 +38,7 @@ public interface Try<T> extends Value<T> {
      * exception occurs calling {@code supplier.get()}.
      */
     static <T> Try<T> of(CheckedSupplier<? extends T> supplier) {
+        Objects.requireNonNull(supplier, "supplier is null");
         try {
             return new Success<>(supplier.get());
         } catch (Throwable t) {
@@ -55,11 +56,7 @@ public interface Try<T> extends Value<T> {
      * exception occurs calling {@code supplier.get()}.
      */
     static <T> Try<T> ofSupplier(Supplier<? extends T> supplier) {
-        try {
-            return new Success<>(supplier.get());
-        } catch (Throwable t) {
-            return new Failure<>(t);
-        }
+        return of(supplier::get);
     }
 
     /**
@@ -71,11 +68,7 @@ public interface Try<T> extends Value<T> {
      * exception occurs calling {@code callable.call()}.
      */
     static <T> Try<T> ofCallable(Callable<? extends T> callable) {
-        try {
-            return new Success<>(callable.call());
-        } catch (Throwable t) {
-            return new Failure<>(t);
-        }
+        return of(callable::call);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


javaslang/src/main/java/javaslang/control/Try.java
```diff
@@ -39,6 +45,39 @@
...
+    /**
+     * Creates a Try of a Supplier.
+     *
+     * @param supplier A supplier
+     * @param <T>      Component type
+     * @return {@code Success(supplier.get())} if no exception occurs, otherwise {@code Failure(throwable)} if an
+     * exception occurs calling {@code supplier.get()}.
+     */
+    static <T> Try<T> ofSupplier(Supplier<? extends T> supplier) {
+        try {
```
danieldietrich: Let's return `of(supplier::get)` to reduce duplicate code. The compiler will translate the method reference to a simple method call - there's no additional overhead.

Please do this for all other static factory methods we introduce in this PR, too.

RobWin: Okidoki

danieldietrich:
@RobWin forgot one thing: all methods need to null-check the argument:

```java
Objects.requireNonNull(<param-name>, "<param-name> is null");
```

javaslang/src/main/java/javaslang/control/Try.java
```diff
@@ -38,6 +38,7 @@
      * exception occurs calling {@code supplier.get()}.
      */
     static <T> Try<T> of(CheckedSupplier<? extends T> supplier) {
+        Objects.requireNonNull(supplier, "supplier is null");
```
danieldietrich: Hah, forgot that myself :)

javaslang/src/main/java/javaslang/control/Try.java
```diff
@@ -55,11 +56,7 @@
      * exception occurs calling {@code supplier.get()}.
      */
     static <T> Try<T> ofSupplier(Supplier<? extends T> supplier) {
-        try {
-            return new Success<>(supplier.get());
-        } catch (Throwable t) {
-            return new Failure<>(t);
-        }
+        return of(supplier::get);
```
danieldietrich: We need the

```java
```

everywhere because of the `::get` call. Otherwise it will throw a NPE before we enter `of(CheckedSupplier)`...

javaslang/src/main/java/javaslang/control/Try.java
```diff
@@ -71,11 +68,7 @@
      * exception occurs calling {@code callable.call()}.
      */
     static <T> Try<T> ofCallable(Callable<? extends T> callable) {
-        try {
-            return new Success<>(callable.call());
-        } catch (Throwable t) {
-            return new Failure<>(t);
-        }
+        return of(callable::call);
```
danieldietrich:
```java
Objects.requireNonNull(callable, "callable is null");
```

javaslang/src/main/java/javaslang/control/Try.java
```diff
@@ -102,12 +96,7 @@
      * calling {@code runnable.run()}.
      */
     static Try<Void> runRunnable(Runnable runnable) {
-        try {
-            runnable.run();
-            return new Success<>(null); // null represents the absence of an value, i.e. Void
-        } catch (Throwable t) {
-            return new Failure<>(t);
-        }
+        return run(runnable::run);
```
danieldietrich:
```java
Objects.requireNonNull(runnable, "runnable is null");
```

danieldietrich:
Go, go, go :-))

RobWin:
I thought about it as well, but then we have two null checks? Is that really necessary?

javaslang/src/main/java/javaslang/control/Try.java
RobWin: Sure? I have to test this.

danieldietrich:
It is Ok! I've learned that the JIT compiler is blazing fast. These are the arguments:

* we have developer-friendly exceptions
* requireNonNull calls will be inlined by the JIT compiler => no method calls!
* no increased mem-tradeoff/GC pressure here

javaslang/src/main/java/javaslang/control/Try.java
RobWin: You are right. I didn't expect that.

danieldietrich: Yes, we do it everywhere. Seems like ballast but the developer experience (DX) is better. Like a polished  product.

RobWin:
Done

javaslang/src/test/java/javaslang/control/TryTest.java
```diff
@@ -152,32 +152,100 @@ public void shouldReturnIteratorOfFailure() {
...
+    @Test
+    public void shouldCreateFailureWhenCallingTryRunRunnable() {
+        assertThat(Try.runRunnable(() -> {
+            throw new Error("error");
+        }) instanceof Try.Failure).isTrue();
+    }
+
+    @Test
+    public void shouldThrowNullPointerExceptionWhenCallingTryRunRunnable() {
+        assertThatThrownBy(() -> Try.runRunnable(null)).isInstanceOf(NullPointerException.class).hasMessage("runnable is null");
```
danieldietrich: cool, better than using `@Test(expected = NullPointerException.class)`!

danieldietrich:
Looks great, thank you for the long breath :-)

RobWin:
👍

javaslang/src/test/java/javaslang/control/TryTest.java
RobWin: Yes, Assertj rocks :)

danieldietrich:
Towards #1811 (the Try part)

PR #1816, opened 2017-01-14
HashArrayMappedTrie.equals() fixed #1815

### LINKED ISSUE #1815: HashMap.equals() is broken
```java
HashMap<String, Integer> map = HashMap.<String, Integer> empty().put("Aa", 1).put("BB", 2);
HashMap<String, Integer> map2 = HashMap.<String, Integer> empty().put("BB", 2).put("Aa", 1);
System.out.println(map.equals(map2)); // prints false
```

`map` and `map2` contain the same key-value pairs, but `equals()` returns `false`.

ruslansennov:
Different keys with same hashCodes can come in random order, so we can't use `Collections.areEqual(this, that)` here


danieldietrich:
Nice finding, thanks!!

Issue #1818, opened 2017-01-15
Ensure symmetry property of collections that have no predictable order


danieldietrich:
Especially HashSet, HashMap and HashMultimap need to return a hashcode that is combined of underlying elements in a commutative way (e.g. sum of element hashcodes).

danieldietrich:
see also #1815 (and #1816)

PR #1823, opened 2017-01-16
LeafList.hashCode() fixed

ruslansennov:
@danieldietrich @orionll #1818 
I believe that only `LeafList.hashCode()` should be changed.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/HashArrayMappedTrie.java
+++ b/javaslang/src/main/java/javaslang/collection/HashArrayMappedTrie.java
@@ -572,7 +572,7 @@ interface HashArrayMappedTrieModule {
 
         @Override
         public int hashCode() {
-            return Objects.hash(hash, value, tail);
+            return Objects.hash(hash, value) + tail.hashCode();
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


orionll:
```java
return Objects.hash(hash, value) + tail.hashCode();
```
I think this can fail with StackOverflowError if the map contains large number of keys with the same hash codes.

Try something like this:

```java
class A {
    @Override
    public int hashCode() {
        return 1;
    }

        HashMap<A, Integer> map = HashMap.empty();
        for (int i = 0; i < 10000; i++) {
            map = map.put(new A(), i);
        }
        System.out.println(map.hashCode());
    }
}
```

I suggest rewriting `hashCode()` the same way it is done in `java.util.AbstractMap`. This is simple and safe: http://hg.openjdk.java.net/jdk9/jdk9/jdk/file/37a05a11f281/src/share/classes/java/util/AbstractMap.java#l490

danieldietrich:
I also think adding hashcodes is a viable solution. I created #1818 to ensure all Hash* impls work properly.
The StackOverflow problem is real (using Objects.hash(val) ... + tail.hashCode()). I fixed it recently for List.hashCode...

danieldietrich:
@ruslansennov Looks good to me! Please merge it if no one has any further objections :)

PR #1828, opened 2017-01-18
Transpose result is 'same as' when empty, 1x0 and 1x1

danieldietrich:
Towards #1800

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Array.java
+++ b/javaslang/src/main/java/javaslang/collection/Array.java
@@ -469,6 +469,21 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
     }
 
     /**
+     * Transposes the rows and columns of an {@link Array}.
+     *
+     * @param rows to be transposed.
+     * @return a transposed {@link Array}.
+     * <p>
+     * ex: {@code
+     * Array.transpose(Array(Array(1,2,3), Array(4,5,6))) → Array(Array(1,4), Array(2,5), Array(3,6))
+     * Array.transpose(Array(Array(1,2), Array(3))) → Array(Array(1,3), Array(2))
+     * }
+     */
+    static <T> Array<Array<T>> transpose(Array<Array<T>> rows) {
+        return Collections.transpose(rows, Array::ofAll, Array::ofAll);
+    }
+
+    /**
      * Creates an Array from a seed value and a function.
      * The function takes the seed at first.
      * The function should return {@code None} when it's
@@ -552,21 +567,6 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
      */
     public static <T> Array<T> unfold(T seed, Function<? super T, Option<Tuple2<? extends T, ? extends T>>> f) {
         return Iterator.unfold(seed, f).toArray();
-    }
-
-    /**
-     * Transposes the rows and columns of an {@link Array}.
-     *
-     * @param rows to be transposed.
-     * @return a transposed {@link Array}.
-     * <p>
-     * ex: {@code
-     * Array.transpose(Array(Array(1,2,3), Array(4,5,6))) → Array(Array(1,4), Array(2,5), Array(3,6))
-     * Array.transpose(Array(Array(1,2), Array(3))) → Array(Array(1,3), Array(2))
-     * }
-     */
-    static <T> Array<Array<T>> transpose(Array<Array<T>> rows) {
-        return Collections.transpose(rows, Array::ofAll);
     }
 
     @Override
--- a/javaslang/src/main/java/javaslang/collection/Collections.java
+++ b/javaslang/src/main/java/javaslang/collection/Collections.java
@@ -289,29 +289,32 @@ final class Collections {
         }
     }
 
-    static <T, U extends Iterable<? extends Iterable<T>>> U transpose(U source, Function<Iterable<?>, Iterable<?>> mapper) {
-        if (source.iterator().hasNext()) {
-            return transposeIterables(source, mapper);
-        } else {
+    static <T, U extends Seq<T>, V extends Seq<U>> V transpose(V source, Function<Iterable<U>, V> outerMapper, Function<Iterable<T>, U> innerMapper) {
+        if (source.isEmpty() || (source.length() == 1 && source.head().length() <= 1)) {
             return source;
+        } else {
+            return transposeIterables(source, outerMapper, innerMapper);
         }
     }
 
-    @SuppressWarnings("unchecked")
-    private static <T, U extends Iterable<? extends Iterable<T>>> U transposeIterables(U source, Function<Iterable<?>, Iterable<?>> mapper) {
-        final ArrayList<ArrayList<T>> results = new ArrayList<>();
+    private static <T, U extends Seq<T>, V extends Seq<U>> V transposeIterables(V source, Function<Iterable<U>, V> outerMapper, Function<Iterable<T>, U> innerMapper) {
+        final java.util.List<java.util.List<T>> results = new ArrayList<>();
 
-        for (Iterable<T> row : source) {
-            int c = 0;
+        for (U row : source) {
+            int columnIndex = 0;
             for (T element : row) {
-                if (results.size() == c) {
-                    results.add(new ArrayList<T>());
+                final java.util.List<T> newRow;
+                if (results.size() == columnIndex) {
+                    newRow = new ArrayList<>();
+                    results.add(newRow);
+                } else {
+                    newRow = results.get(columnIndex);
                 }
-                results.get(c).add(element);
-                c++;
+                newRow.add(element);
+                columnIndex++;
             }
         }
 
-        return (U) mapper.apply(Iterator.ofAll(results).map(mapper::apply));
+        return outerMapper.apply(Iterator.ofAll(results).map(innerMapper::apply));
     }
 }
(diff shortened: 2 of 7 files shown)


javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -469,6 +469,21 @@ private Array(Object[] delegate) {
...
+     *
+     * @param rows to be transposed.
+     * @return a transposed {@link Array}.
+     * <p>
+     * ex: {@code
+     * Array.transpose(Array(Array(1,2,3), Array(4,5,6))) → Array(Array(1,4), Array(2,5), Array(3,6))
+     * Array.transpose(Array(Array(1,2), Array(3))) → Array(Array(1,3), Array(2))
+     * }
+     */
+    static <T> Array<Array<T>> transpose(Array<Array<T>> rows) {
```
danieldietrich: re-ordered the transpose methods alphabetically

javaslang/src/main/java/javaslang/collection/Collections.java
```diff
@@ -289,29 +289,32 @@ int size() {
...
+            int columnIndex = 0;
             for (T element : row) {
-                if (results.size() == c) {
-                    results.add(new ArrayList<T>());
+                final java.util.List<T> newRow;
+                if (results.size() == columnIndex) {
+                    newRow = new ArrayList<>();
+                    results.add(newRow);
+                } else {
+                    newRow = results.get(columnIndex);
```
danieldietrich: only call get when necessary

javaslang/src/main/java/javaslang/collection/Collections.java
```diff
@@ -289,29 +289,32 @@ int size() {
...
             return source;
+        } else {
+            return transposeIterables(source, outerMapper, innerMapper);
         }
     }
 
-    @SuppressWarnings("unchecked")
-    private static <T, U extends Iterable<? extends Iterable<T>>> U transposeIterables(U source, Function<Iterable<?>, Iterable<?>> mapper) {
-        final ArrayList<ArrayList<T>> results = new ArrayList<>();
+    private static <T, U extends Seq<T>, V extends Seq<U>> V transposeIterables(V source, Function<Iterable<U>, V> outerMapper, Function<Iterable<T>, U> innerMapper) {
```
danieldietrich: Made it type-safe. may be also helpful if we decide to add `CharSeq.transpose(Seq<CharSeq<T>>)` in future.

javaslang/src/main/java/javaslang/collection/Collections.java
paplorinc: it's a constant time operation, why do we want to avoid calling it? :)
The code got more complicated.

javaslang/src/main/java/javaslang/collection/Collections.java
```diff
@@ -289,29 +289,32 @@ int size() {
...
+        final java.util.List<java.util.List<T>> results = new ArrayList<>();
 
-        for (Iterable<T> row : source) {
-            int c = 0;
+        for (U row : source) {
+            int columnIndex = 0;
```
paplorinc: :+1:

javaslang/src/main/java/javaslang/collection/Collections.java
paplorinc: :+1: for `Seq`, though you requested iterable: https://github.com/javaslang/javaslang/issues/1659

javaslang/src/test/java/javaslang/collection/AbstractSeqTest.java
```diff
@@ -1822,6 +1822,18 @@ public void shouldTransposeIfEmpty() {
     }
 
     @Test
+    public void shouldTransposeIfIs1x0() {
```
paplorinc: do we need the supressions for the other tests, if apparently they're not needed here :)?

javaslang/src/main/java/javaslang/collection/Collections.java
```diff
@@ -289,29 +289,32 @@ int size() {
         }
     }
 
-    static <T, U extends Iterable<? extends Iterable<T>>> U transpose(U source, Function<Iterable<?>, Iterable<?>> mapper) {
-        if (source.iterator().hasNext()) {
-            return transposeIterables(source, mapper);
-        } else {
+    static <T, U extends Seq<T>, V extends Seq<U>> V transpose(V source, Function<Iterable<U>, V> outerMapper, Function<Iterable<T>, U> innerMapper) {
```
paplorinc: How about `rowMapper` / `columnMapper` instead?

javaslang/src/main/java/javaslang/collection/Array.java
AlparSzabados: Thanks for that! My alphabet is rusty...

javaslang/src/main/java/javaslang/collection/Collections.java
danieldietrich: In fact `rowFactory` and `columnFactory` would fit better - a mapper maps the elements, not the wrapper. This one is really minor because it is not public API.

javaslang/src/main/java/javaslang/collection/Collections.java
danieldietrich: Yes. I also requested a default value (which is currently implicitly null).

javaslang/src/main/java/javaslang/collection/Collections.java
paplorinc: I agree that it's minor.
Though `map` functions have a [`mapper` function param also](https://github.com/javaslang/javaslang/blob/master/javaslang/src/main/java/javaslang/collection/List.java#L896), factory would be too general, in my opinion.

javaslang/src/main/java/javaslang/collection/Collections.java
paplorinc: > without having a default value this might be the better choice

In that case I guess this line was misunderstood :)

javaslang/src/main/java/javaslang/collection/Collections.java
danieldietrich: First storing s.th. and then retrieving it again makes no sense to me. It is like sending a letter to my own address. Why should I do that if I hold the letter in my hands?

Having a name 'newRow' clarifies what we are dealing with. In a functional setting you would extract the logic to a new method if it gets too complicated (which is not the case here).

```java
newRow(results, columnIndex).add(element);
```

javaslang/src/main/java/javaslang/collection/Collections.java
AlparSzabados: I made it without a default value (not even implicitly null). There are not values to fill in the missing cells. This makes sense because transpose is supposed to work for matrixes, so if you have missing values, that's not a matrix. In these cases it will still transpose, but the transposition of the transposition will not be the same as the original (a least not in every case).

javaslang/src/test/java/javaslang/collection/AbstractSeqTest.java
danieldietrich: From what I can see the other tests _logically_ serve a different purpose. They test if the contents are the same. The identity test ensure that the references are the same. Technically they overlap but the semantics are different. I would leave them for now.

javaslang/src/main/java/javaslang/collection/Collections.java
paplorinc: > First storing s.th. and then retrieving it again makes no sense to me.

Both variants store and retrieve it via direct access memory (field vs array access, same thing).

danieldietrich: You assume too much on the technical layer. On the semantic layer it makes no sense to send a present mail to myself.

javaslang/src/main/java/javaslang/collection/Collections.java
danieldietrich: I don't understand!? of course map functions have a mapper, the map the elements, not the container.

javaslang/src/main/java/javaslang/collection/Collections.java
danieldietrich: yes, mhh, that might be seen as bug

paplorinc: the bug is in the input:
https://en.wikipedia.org/wiki/Transpose

> In linear algebra, the transpose of a matrix is an operator which flips a matrix over its diagonal

In incomplete matrix is not a matrix, which doesn't have a diagonal.
In my opinion these inputs should simply throw an exception (or return an `Option`)

danieldietrich: Yes, sounds good 😊 
I will push an update that includes

* CharSeq.transpose(Seq&lt;CharSeq>>)
* better names (rowFactory, columnFactory)
* the matrix checks (throwing IllegalArgumentException is ok)

paplorinc: Good news: for regular matrices our debate is not valid anymore: you create the ArrayLists's for the first row (via e.g. a `map` function) and populate the rest via the `get(columnIndex)`.
I will quickly take a look at whether it's feasible :)

danieldietrich: If the size is known I rather have a T[][] resp Object[][] in mind...

paplorinc: Great idea!

PR #1829, opened 2017-01-19
Changed `transpose` to only work for valid matrices

paplorinc:
Continued the work @AlparSzabados and @danieldietrich:
https://github.com/javaslang/javaslang/pull/1828#discussion_r96847626

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Vector.java
+++ b/javaslang/src/main/java/javaslang/collection/Vector.java
@@ -482,18 +482,18 @@ public final class Vector<T> implements Kind1<Vector<?>, T>, IndexedSeq<T>, Seri
     }
 
     /**
-     * Transposes the rows and columns of a {@link Vector}.
+     * Transposes the rows and columns of a {@link Vector} matrix.
      *
      * @param matrix to be transposed.
-     * @return a transposed {@link Vector}.
+     * @return a transposed {@link Vector} matrix.
+     * @throws IllegalArgumentException if the row lengths of {@code matrix} differ.
      * <p>
      * ex: {@code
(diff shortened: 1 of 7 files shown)


danieldietrich:
Well done! ... and fast ;)

javaslang/src/main/java/javaslang/collection/Vector.java
```diff
@@ -482,18 +482,17 @@
     }
 
     /**
-     * Transposes the rows and columns of a {@link Vector}.
+     * Transposes the rows and columns of a {@link Vector} matrix.
      *
      * @param rows to be transposed.
      * @return a transposed {@link Vector}.
```
AlparSzabados: minor:
`* @return a transposed {@link Vector}.` -> `* @return a transposed {@link Vector} matrix.` ?

paplorinc: Seems redundant, @danieldietrich ?

paplorinc:
... that was a long build :)

javaslang/src/main/java/javaslang/collection/Vector.java
danieldietrich: I like to have `matrix` because it tells us something about the domain.

@paplorinc Please add `@throws IllegalArgumentException if rows is not a regular n x m matrix.` (or similar - your call) to the public `transpose` methods.

paplorinc: Done

danieldietrich:
Now this is a clean solution. Thank you for your previous review!

PR #1830, opened 2017-01-20
Added simple-and-inefficient QuickSort example 

paplorinc:
... just to show off

Using the `Advanced Java Folding` plugin and `Fira Code` font with ligatures:
screenshot from 2017-01-20 16-58-00

### DIFF
--- a/javaslang/src/test/java/javaslang/collection/QuickSortTest.java
+++ b/javaslang/src/test/java/javaslang/collection/QuickSortTest.java
@@ -1,0 +1,34 @@
+/*     / \____  _    _  ____   ______  / \ ____  __    _______
+ *    /  /    \/ \  / \/    \ /  /\__\/  //    \/  \  //  /\__\   JΛVΛSLΛNG
+ *  _/  /  /\  \  \/  /  /\  \\__\\  \  //  /\  \ /\\/ \ /__\ \   Copyright 2014-2017 Javaslang, http://javaslang.io
+ * /___/\_/  \_/\____/\_/  \_/\__\/__/\__\_/  \_//  \__/\_____/   Licensed under the Apache License, Version 2.0
+ */
+package javaslang.collection;
+
+import javaslang.Function1;
+import org.junit.Test;
+
+import static java.lang.Integer.signum;
+import static javaslang.API.List;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class QuickSortTest {
+    @Test
+    public void shouldQuickSort() {
+        final List<Integer> values = List(3, 1, 4, 1, 5, 9, 2, 6, 5, 3, 5, 8, 9, 7, 9, 3, 2, 3, 8, 4, 6, 2, 6);
+        assertThat(sort(values.shuffle())).isEqualTo(values.sorted());
+        assertThat(sort2(values.shuffle())).isEqualTo(values.sorted());
+    }
+
+    /** Note: this example is only meant to show off, not to be used in reality: it can have quadratic performance and cause stack overflow */
+    private static Seq<Integer> sort(Seq<Integer> values) {
+        if (values.size() <= 1) return values;
+        return values.tail().partition(v -> v <= values.head())
+                     .apply((less, more) -> sort(less).append(values.head()).appendAll(sort(more)));
+    }
+    private static <T extends Comparable<T>> List<T> sort2(List<T> values) {
+        if (values.size() <= 1) return values;
+        final Function1<Integer, List<T>> parts = values.groupBy(v -> signum(v.compareTo(values.head()))).withDefaultValue(List());
+        return sort2(parts.apply(-1)).appendAll(parts.apply(0)).appendAll(sort2(parts.apply(1)));
+    }
+}


javaslang/src/test/java/javaslang/collection/QuickSort.java
```diff
@@ -0,0 +1,25 @@
...
+    @Test
+    public void shouldQuickSort() {
+        final Vector<Integer> values = Vector(3, 1, 4, 1, 5, 9, 2, 6, 5, 3, 5, 8, 9, 7, 9, 3, 2, 3, 8, 4, 6, 2, 6);
+        assert sort(values).equals(values.sorted());
+    }
+
+    /** Note: this example is only meant to show off, not to be used in reality: it can have quadratic performance and cause stack overflow */
+    private static Seq<Integer> sort(Seq<Integer> values) {
+        if (values.size() <= 1) return values;
+        else return values.tail().partition(v -> v <= values.head())
```
paplorinc: @danieldietrich, could we deconstruct this to
```java
(pivot, rest) -> rest.partition ...
_ -> values
```
somehow :)?

danieldietrich: @paplorinc Yes:

```java
return Match(values).of(
    Case(Patterns.List($(), $()), (x, xs) ->
            xs.partition(v -> v <= x)
              .apply((less, more) -> sort(less).append(values.head()).appendAll(sort(more)))),
    Case($(), () -> values)
);
```

Spoiler alert: We use javaslang.Patterns.List here, which will be obsolete in near future...

paplorinc: Is there something for a general `Seq` instead :)?

danieldietrich: Currently not. I plan to revise the patterns. The `List` Pattern will be deprecated. Instead we will have a `Cons` and a `Nil` pattern, directly encoded in javaslang.API.

You could define your own patterns, directly within Quicksort (I think / depends on the build order and pattern code generator - haven't tested it):

```java
public class Quicksort {

    ... // use Quicksort$Patterns.Cons here <-- not sure about the name of the class, too lazy to look it up

    @Patterns
    static class $ {
        @Unapply
        static <T> Tuple2<T, List<T>> Cons(List.Cons<T> cons) {
            return Tuple.of(cons.head(), cons.tail());
        }

        @Unapply
        static <T> Tuple0 Nil(List.Nil<T> nil) {
            return Tuple.empty();
        }
    }
}
```

For Seq there is a problem. Patterns are generated for ADTs. In particular an ADT is a _sum type_, which means that it is a disjoint union of several (two) implementations.

E.g. the ADT List is the union of List.Cons and List.Nil.

Seq is an interface. There is no unique union of types. But we could create a (head, tail) Pattern 'by hand' for the Traversable type:

```java
public class Quicksort {

    public static <T, _1 extends T, _2 extends Traversable<T>> Pattern2<Traversable<T>, _1, _2> HeadTail(Pattern<_1, ?> p1, Pattern<_2, ?> p2) {
        return new Pattern2<T, T1, T2>() {
                @Override
                public Option<Tuple2<T1, T2>> apply(T obj) {
                    if (obj instanceof Traversable && ((Traversable) obj).isDefined()) {
                        final Traversable<T> traversable = (Traversable<T>) obj;
                        final Tuple2<U1, U2> unapplied = Tuple(traversable.head(), traversable.tail());
                        return unapplied.apply((u1, u2) ->
                                ((Pattern<U1, ?>) p1).apply(u1).flatMap(_1 ->
                                ((Pattern<U2, ?>) p2).apply(u2).map(_2 -> (Tuple2<T1, T2>) unapplied)
                        ));
                    } else {
                        return Option.none();
                    }
                }
            };
    }

}
```

Haven't tried it ... don't know if it compiles.

paplorinc:
screenshot from 2017-01-20 16-58-00

javaslang/src/test/java/javaslang/collection/QuickSort.java
```diff
@@ -0,0 +1,33 @@
...
+
+import static java.lang.Integer.signum;
+import static javaslang.API.List;
+
+public class QuickSort {
+    @Test
+    public void shouldQuickSort() {
+        final List<Integer> values = List(3, 1, 4, 1, 5, 9, 2, 6, 5, 3, 5, 8, 9, 7, 9, 3, 2, 3, 8, 4, 6, 2, 6);
+        assert sort(values).equals(values.sorted());
+        assert sort2(values).equals(values.sorted());
```
danieldietrich: interesting approach to use Java's assert keyword in unit tests - haven't done that, yet :)

javaslang/src/test/java/javaslang/collection/QuickSort.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package javaslang.collection;
+
+import javaslang.Function1;
+import org.junit.Test;
+
+import static java.lang.Integer.signum;
+import static javaslang.API.List;
+
+public class QuickSort {
```
danieldietrich: I think this test will never run. By convention maven-surefire-plugin look for classes that match the following patterns:

* \*\*/Test*.java
* \*\*/*Test.java
* \*\*/*TestCase.java

(source: http://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html)

javaslang/src/test/java/javaslang/collection/QuickSort.java
paplorinc: Done

javaslang/src/test/java/javaslang/collection/QuickSort.java
paplorinc: Thanks, done

danieldietrich:
Thank you!

Issue #1762, opened 2016-12-15
Memoization seems broken on Java 9


paplorinc:
See: https://github.com/javaslang/javaslang/pull/1563#issuecomment-266869091

danieldietrich:
**Note:** Currently it is not clear if it is a bug in Javaslang or JDK9. But it seems to work correct with JDK8. However, Javaslang 2.1.0 needs to work correct with a future version of the JDK. So we need to solve this for the upcoming 2.1.0 release.

paplorinc:
The problematic test case is:
```java
private static final CheckedFunction5<Integer, Integer, Integer, Integer, Integer, Integer> recurrent1 = (i1, i2, i3, i4, i5) -> i1 <= 0 ? i1 : CheckedFunction5Test.recurrent2.apply(i1 - 1, i2, i3, i4, i5) + 1;
private static final CheckedFunction5<Integer, Integer, Integer, Integer, Integer, Integer> recurrent2 = CheckedFunction5Test.recurrent1.memoized();

@Test
public void shouldCalculatedRecursively() throws Throwable {
    assertThat(recurrent1.apply(11, 11, 11, 11, 11)).isEqualTo(11);
    assertThat(recurrent1.apply(22, 22, 22, 22, 22)).isEqualTo(22);
}
```
which does seem broken, i.e. it should modify the `cache`, while it's already modifying it :/.
Do we really need this functionality?

paplorinc:
The tests fail forEuler 67 also:
```java
private final static Function3<Vector<Vector<Integer>>, Integer, Integer, Integer> smart = Function3.of(
        (Vector<Vector<Integer>> tr, Integer row, Integer col) -> {
            int value = tr.get(row).get(col);
            if (row == tr.length() - 1) {
                return tr.get(row).get(col);
            } else {
                return value + Math.max(Euler67Test.smart.apply(tr, row + 1, col), Euler67Test.smart.apply(tr, row + 1, col + 1));
            }
        }
).memoized();
```
since it calls itself also.

---

I think therefore that memoization was working incorrectly before also (i.e. calculating things twice), but in Java 9 they seem to have added a validation check to avoid this.

Will provide a PR to ask for comments.

paplorinc:
@danieldietrich:
It seems to me that we cannot easily support recursive memoization: http://mail.openjdk.java.net/pipermail/core-libs-dev/2015-February/031237.html

danieldietrich:
[...]

PR #1837, opened 2017-01-24
Simplified function memoization

paplorinc:
Addresses https://github.com/javaslang/javaslang/issues/1762

* `null`s are avoided by always using a `Tuple` wrapper
* concurrent modifications are allowed now, work on Java 9 also.

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -1639,36 +1639,16 @@ def generateMainClasses(): Unit = {
                   } else {
                       ${val mappingFunction = (checked, i) match {
                           case (true, 0) => s"() -> $Try.of(this::apply).get()"
-                          case (true, 1) => s"t -> $Try.of(() -> this.apply(t)).get()"
-                          case (true, _) => s"t -> $Try.of(() -> tupled.apply(t)).get()"
+                          case (true, _) => s"t -> $Try.of(() -> apply($params)).get()"
                           case (false, 0) => s"this::apply"
-                          case (false, 1) => s"this"
-                          case (false, _) => s"tupled"
-                        }
-                        val forNull = (checked, i) match {
-                          case (true, 1) => s"$Try.of(() -> apply(null))::get"
-                          case (false, 1) => s"() -> apply(null)"
-                          case _ => null
+                          case (false, _) => s"tupled()"
                         }
                         if (i == 0) xs"""
                           return ($className$fullGenerics & Memoized) Lazy.of($mappingFunction)::get;
-                        """ else if (i == 1) xs"""
-                          final Object lock = new Object();
-                          final ${im.getType("java.util.Map")}<$generics, R> cache = new ${im.getType("java.util.HashMap")}<>();
-                          return ($className$fullGenerics & Memoized) t1 -> {
-                              synchronized (lock) {
-                                  return cache.computeIfAbsent(t1, $mappingFunction);
-                              }
-                          };
                         """ else xs"""
-                          final Object lock = new Object();
                           final ${im.getType("java.util.Map")}<Tuple$i<$generics>, R> cache = new ${im.getType("java.util.HashMap")}<>();
-                          final ${checked.gen("Checked")}Function1<Tuple$i<$generics>, R> tupled = tupled();
-                          return ($className$fullGenerics & Memoized) ($params) -> {
-                              synchronized (lock) {
-                                  return cache.computeIfAbsent(Tuple.of($params), $mappingFunction);
-                              }
-                          };
+                          return ($className$fullGenerics & Memoized) ($params)
+                                  -> Memoized.of(cache, Tuple.of($params), $mappingFunction);
                         """
                       }
                   }
--- a/javaslang/src/main/java/javaslang/λ.java
+++ b/javaslang/src/main/java/javaslang/λ.java
@@ -6,6 +6,7 @@
 package javaslang;
 
 import java.io.Serializable;
+import java.util.Map;
 
 /**
  * This is a general definition of a (checked/unchecked) function of unknown parameters and a return type R.
@@ -74,5 +75,17 @@ public interface λ<R> extends Serializable {
      * Zero Abstract Method (ZAM) interface for marking functions as memoized using intersection types.
      */
     interface Memoized {
+        static <T extends Tuple, R> R of(Map<T, R> cache, T key, Function1<T, R> tupled) {
+            synchronized (cache) {
+                if (cache.containsKey(key)) {
+                    return cache.get(key);
+                } else {
+                    final R value = tupled.apply(key);
+                    cache.put(key, value);
+                    return value;
+                }
+            }
+        }
+
     }
 }
--- a/javaslang/src-gen/main/java/javaslang/CheckedFunction1.java
+++ b/javaslang/src-gen/main/java/javaslang/CheckedFunction1.java
@@ -167,13 +167,9 @@ public interface CheckedFunction1<T1, R> extends λ<R> {
         if (isMemoized()) {
             return this;
         } else {
-            final Object lock = new Object();
-            final Map<T1, R> cache = new HashMap<>();
-            return (CheckedFunction1<T1, R> & Memoized) t1 -> {
-                synchronized (lock) {
-                    return cache.computeIfAbsent(t1, t -> Try.of(() -> this.apply(t)).get());
-                }
-            };
+            final Map<Tuple1<T1>, R> cache = new HashMap<>();
+            return (CheckedFunction1<T1, R> & Memoized) (t1)
+                    -> Memoized.of(cache, Tuple.of(t1), t -> Try.of(() -> apply(t1)).get());
         }
     }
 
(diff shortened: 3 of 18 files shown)


javaslang/src-gen/test/java/javaslang/CheckedFunction2Test.java
```diff
@@ -165,15 +165,6 @@ public void shouldLiftTryPartialFunction() {
         assertThat(unknown.getCause().getMessage()).isEqualToIgnoringCase("Unknown MessageDigest not available");
     }
 
-    private static final CheckedFunction2<Integer, Integer, Integer> recurrent1 = (i1, i2) -> i1 <= 0 ? i1 : CheckedFunction2Test.recurrent2.apply(i1 - 1, i2) + 1;
```
paplorinc: @danieldietrich, @ruslansennov, how can we solve this properly?

paplorinc: http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-

> Some attempted update operations on this map by other threads may be blocked while computation is in progress, so the computation should be short and simple, and must not attempt to update any other mappings of this map.

paplorinc:
Another `0.12%` codecov increase, we're almost at `98%` :)
cc: @AlparSzabados

javaslang/src-gen/test/java/javaslang/CheckedFunction2Test.java
ruslansennov: This PR #513 shows that we should avoid `ConcurrentHashMap`
Also I believe we should keep recursive memoization if it possible

paplorinc: Thank you @ruslansennov, good find indeed.
Unfortunately however, using `HashMap` isn't safe either.
In `Java 9` they added checks for concurrent modifications, i.e.
http://download.java.net/java/jdk9/docs/api/java/util/HashMap.html#computeIfAbsent-K-java.util.function.Function-
> The mapping function should not modify this map during computation.
>
> This method will, on a best-effort basis, throw a ConcurrentModificationException if it is detected that the mapping function modifies this map during computation.

ruslansennov: then we should use non-single operation solution inside sync block (if.. contains.. etc), or I'm wrong?

paplorinc: I thought about that also, but multi-threaded recursion overflows my brain :/
I'm not sure how to do that safely and how to test it properly.
@danieldietrich, do you?

paplorinc: @ruslansennov, @danieldietrich, @zsolt-donca, do you think something like
```java
final Map<Tuple3<T1, T2, T3>, R> cache = new HashMap<>();
return (Function3<T1, T2, T3, R> & Memoized) (t1, t2, t3) -> {
    final Tuple3<T1, T2, T3> key = Tuple.of(t1, t2, t3);
    if (cache.containsKey(key)) {
        return cache.get(key);
    } else {
        synchronized (cache) {
            if (cache.containsKey(key)) {
                return cache.get(key);
            } else {
                final R value = apply(t1, t2, t3);
                cache.put(key, value);
                return value;
            }
        }
    }
};
```
would work?

Seems to work correctly for all (recursive) tests, even on `Java 9`, but I don't think we have a complex enough test suite to be sure.
Also,  it's a pity that we cannot use `ConcurrentHashMap` because of the `non-null` value restriction.

paplorinc: Force pushed a new version, all tests seem to be passing now on `Java 9` also (from Idea only, not yet from Maven).
Thanks for your help @ruslansennov, @zsolt-donca.
@danieldietrich, do you think we should add more tests here?

zsolt-donca: I'm concerned around
```
if (cache.containsKey(key)) {
    return cache.get(key);
} else {
    synchronized (cache) {
        ...
    }
}
```

It seems an awful lot to me like previous discussion on `Lazy`. The key idea: the _then_ branch above doesn't use any volatile variable, neither does it pass through a `synchronized` block, nor does `cache` involve any volatile reads inside as it's a simple `HashMap`. Because no memory barrier is passed, it is unsafe. See [Pitfall: Semi-Synchronized Is Fine](https://shipilev.net/blog/2016/close-encounters-of-jmm-kind/#pitfall-semi-sync).

paplorinc: > doesn't use any volatile variable

No, but it uses an effectively final variable, that's meant to avoid the need for local volatile ... (Shipilev would cry seeing that we still don't understand this).
However, I don't trust `HashMap` in a non synchronized environment either ... maybe I should avoid the double locking and just put everything in a `syncronized` block :/

ruslansennov: > maybe I should avoid the double locking and just put everything in a syncronized block :/

it can be a good first step

paplorinc: Done, thanks.

javaslang/generator/Generator.scala
```diff
@@ -1639,36 +1639,16 @@ def generateMainClasses(): Unit = {
...
-                          final Object lock = new Object();
-                          final ${im.getType("java.util.Map")}<$generics, R> cache = new ${im.getType("java.util.HashMap")}<>();
-                          return ($className$fullGenerics & Memoized) t1 -> {
-                              synchronized (lock) {
-                                  return cache.computeIfAbsent(t1, $mappingFunction);
-                              }
-                          };
                         """ else xs"""
-                          final Object lock = new Object();
                           final ${im.getType("java.util.Map")}<Tuple$i<$generics>, R> cache = new ${im.getType("java.util.HashMap")}<>();
```
paplorinc: `new WeakHashMap<>();` to avoid memory leaks?
@danieldietrich, @ruslansennov, @zsolt-donca?

ruslansennov: > `new WeakHashMap<>();` to avoid memory leaks?

What about this scenario?

```java
Function1<Tuple2<Integer, Integer>, Tuple2<Integer, Integer>> f;
Function1<Tuple2<Integer, Integer>, Tuple2<Integer, Integer>> mem = f.memoized();
Tuple2<Integer, Integer> keyInstance = Tuple.of(1, 2);
Tuple2<Integer, Integer> result = mem.apply(keyInstance);

        //
        //  hard work

keyInstance = null;

        ///  hard work
        ///

Tuple2<Integer, Integer> newInstance = Tuple.of(1, 2);
Tuple2<Integer, Integer> result2 = mem.apply(newInstance); // calculation again (?)
```

zsolt-donca: Yes, that's a concern, as weak references can be eliminated at any time. The problem is even more problematic for `FunctionN` where `N ≥ 2`, as the key (tuple) is built internally, immediately discarding any other reference to it.

The above problem with the GC prematurely dropping the data could be eliminated by using _soft references_; for details, [see this](http://stackoverflow.com/questions/299659/what-is-the-difference-between-a-soft-reference-and-a-weak-reference-in-java):

> SoftReferences aren't required to behave any differently than  WeakReferences, but in practice softly reachable objects are generally retained as long as memory is in plentiful supply. This makes them an excellent foundation for a cache, such as the image cache described above, since you can let the garbage collector worry about both how reachable the objects are (a strongly reachable object will never be removed from the cache) and how badly it needs the memory they are consuming.

javaslang/src/main/java/javaslang/λ.java
```diff
@@ -74,5 +75,17 @@ default boolean isMemoized() {
      * Zero Abstract Method (ZAM) interface for marking functions as memoized using intersection types.
      */
     interface Memoized {
+        static <T extends Tuple, R> R of(Map<T, R> cache, T key, Function1<T, R> tupled) {
```
danieldietrich: I think this is a good solution, we can do that. I currently have not a better solution for the synching problem. It is hard to test.

I'm not that familiar with proper cache invalidation (Weak/Soft-Refs) but we need to keep the tuples in the cache as long as the memoized function is referenced.

danieldietrich:
@paplorinc thank you, this is an important change toward Java 9!

javaslang/generator/Generator.scala
danieldietrich: @paplorinc @zsolt-donca I follow @ruslansennov that the function should not compute a value twice (imagine Math::random). Not sure how we can ensure that a memoized function in GC'ed if not referenced any more but one of its cached values is still in use 🤔

paplorinc: I think memory out of bounds exception is worse, but I'm ok with both :)

danieldietrich: I will merge this version in. Memory leaks may also appear in other cases, e.g. when capturing the outer object in a lambda. Optimizations need to be done carefully. Actually we have a working version, which is great. Thx!

PR #1845, opened 2017-01-26
Avoid copying array twice in appendAll/insertAll

rdegnan:
toArray should check if the argument is an Array, and if so just return the delegate. Otherwise for appendAll/insertAll it ends up allocating an iterator, copying into a new ArrayList then again copying the backing array.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Array.java
+++ b/javaslang/src/main/java/javaslang/collection/Array.java
@@ -736,7 +736,7 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
                     list.add(u);
                 }
             }
-            return wrap(toArray(list));
+            return wrap(list.toArray());
         }
     }
 
@@ -1334,7 +1334,7 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
         if ((index < 0) || (index >= length())) {
             throw new IndexOutOfBoundsException("update(" + index + ")");
         } else {
-            final Object[] arr = toArray(this);
+            final Object[] arr = copyOf(delegate, delegate.length);
             arr[index] = element;
             return wrap(arr);
         }
@@ -1403,9 +1403,12 @@ public final class Array<T> implements Kind1<Array<?>, T>, IndexedSeq<T>, Serial
     }
 
     private static <T> Object[] toArray(Iterable<T> elements) {
-        if (elements instanceof java.util.List) {
-            final java.util.List<T> list = (java.util.List<T>) elements;
-            return list.toArray();
+        if (elements instanceof Array) {
+            final Array<T> array = (Array<T>) elements;
+            return array.delegate;
+        } else if (elements instanceof java.util.Collection) {
+            final java.util.Collection<T> collection = (java.util.Collection<T>) elements;
+            return collection.toArray();
         } else {
             final java.util.Iterator<? extends T> it = elements.iterator();
             final java.util.List<T> list = new java.util.ArrayList<>();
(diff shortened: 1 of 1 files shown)


javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -736,7 +736,7 @@ public T get(int index) {
                     list.add(u);
                 }
             }
-            return wrap(toArray(list));
+            return wrap(list.toArray());
```
paplorinc: You can revert this with the new impl :)

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -1334,7 +1334,7 @@ public int length() {
         if ((index < 0) || (index >= length())) {
             throw new IndexOutOfBoundsException("update(" + index + ")");
         } else {
-            final Object[] arr = toArray(this);
+            final Object[] arr = copyOf(delegate, delegate.length);
```
paplorinc: does `javaslang.collection.ArrayBenchmark#update` show an improvement?

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -1403,9 +1403,12 @@ public String toString() {
     }
 
     private static <T> Object[] toArray(Iterable<T> elements) {
-        if (elements instanceof java.util.List) {
-            final java.util.List<T> list = (java.util.List<T>) elements;
-            return list.toArray();
+        if (elements instanceof Array) {
```
paplorinc: good call, this can save a double copy! :)

javaslang/src/main/java/javaslang/collection/Array.java
paplorinc: You could try `clone` also:
```java
final Object[] arr = delegate.clone();
```

javaslang/src/main/java/javaslang/collection/Array.java
```diff
@@ -1403,9 +1403,12 @@ public String toString() {
...
+            final Array<T> array = (Array<T>) elements;
+            return array.delegate;
+        } else if (elements instanceof java.util.Collection) {
```
paplorinc: There is actually a `withSize` optimization for this, feel free to use it:
```java
private static <T> Object[] toArray(Iterable<T> elements) {
    if (elements instanceof Array) {
        final Array<T> array = (Array<T>) elements;
        return array.delegate;
    } else {
        return withSize(elements).toArray();
    }
}
```

rdegnan: Great! I will use that.

javaslang/src/main/java/javaslang/collection/Array.java
rdegnan: Since we already know this is an arraylist, calling toArray directly avoids creating an IterableWithSize merely to call toArray on it -- probably a micro-optimization, but...

rdegnan:
javaslang.collection.ArrayBenchmark#update shows substantial improvement:

Before:
ArrayBenchmark  Update      slang_persistent          10      5       842,556.65  ±    15.81%   ops/s        0.03×                                              0.09×
ArrayBenchmark  Update      slang_persistent         100      5        10,944.34  ±    10.34%   ops/s        0.00×                                              0.01×
ArrayBenchmark  Update      slang_persistent        1000      5           147.66  ±    11.24%   ops/s        0.00×                                              0.00×

After:
ArrayBenchmark  Update      slang_persistent          10      5     5,389,360.02  ±    72.03%   ops/s        0.18×                                              0.54×
ArrayBenchmark  Update      slang_persistent         100      5       162,789.41  ±     7.09%   ops/s        0.04×                                              0.16×
ArrayBenchmark  Update      slang_persistent        1000      5         1,718.15  ±     5.96%   ops/s        0.00×                                              0.02×

javaslang/src/main/java/javaslang/collection/Array.java
paplorinc: I think it would be more rezistent to change via the general (warmed up) method, but I don't mind leaving it this way either.

danieldietrich:
@rdegnan great spot, thank you!

Issue #1839, opened 2017-01-26
Wish: Greater Arity of functions and tuples


hamnis:
I have written a typesafe Json library, found at https://github.com/hamnis/immutable-json which
heavily uses Tuples and functions for Encoding and Decoding json and types.

Scala has Tuple 22 and Function 22, which is enough for most users, If not you need a HList or something like that ( which is possible to encode in Java, but not really practical )

paplorinc:
Since all these are generated, I pushed a PR containing a partial fix: https://github.com/javaslang/javaslang/pull/1848

paplorinc:
@hamnis, could you please take a look at the PR?

hamnis:
@paplorinc as partial fix, that is fine. I would still really want more =)

paplorinc:
@danieldietrich decided :)

danieldietrich:
@hamnis what is the specific use-case for greater arity in your JSON lib? JSON itself has no notion for tuples or do you encode lists as tuples and fall back to HList at some point?

hamnis:
@danieldietrich encoding and decoding into user defined types.

https://github.com/hamnis/immutable-json/blob/master/codec/src/main/java/net/hamnaberg/json/codec/Codecs.java#L326

https://github.com/hamnis/immutable-json/blob/master/codec/src/main/java/net/hamnaberg/json/extract/Extractors.java#L88

use:
https://github.com/hamnis/immutable-json/blob/master/codec/src/test/java/net/hamnaberg/json/extract/ExtractorsTest.java#L40-L47

danieldietrich:
@hamnis Thank you for the links. I see, Tuples are needed in Codecs (especially in Iso). However, currently the extraction of JSON objects is bound by the max Tuple arity. If there is Tuple22, you can extract only 22 fields. A better scalable solution would be unbounded, i.e. would allow to extract JSON objects of arbitrary size.

I've sketched a HList impl here: https://github.com/javaslang/javaslang/issues/237#issuecomment-171659133

Maybe you could use HList instead of Tuple*?

---

I will close this issue for now. May Javaslang does also need an HList for greater Tuple arity. I would be really interested if you get practical experience with HList!

danieldietrich:
@hamnis Bad news - our javaslang-gwt module does not return from build anymore since we bumped arity from 8 to 10. Maybe the GWT compiler is unable to cope with the generics anymore. We need to revert back to arity 8 until the GWT issue is solved, sorry :-/

[...]

PR #1848, opened 2017-01-28
Bumped Function/Tuple generation up to 10

paplorinc:
The code wasn't tested for values >9, this commit fixes those cases also.
If @danieldietrich thinks this should be bumbed up to `22` or back to `8`, the rest of the fixes still make sense.
Addresses: https://github.com/javaslang/javaslang/issues/1839

### DIFF
--- a/javaslang/generator/Generator.scala
+++ b/javaslang/generator/Generator.scala
@@ -11,7 +11,7 @@ import JavaGenerator._
 import collection.immutable.ListMap
 import scala.language.implicitConversions
 
-val N = 8
+val N = 10
 val VARARGS = 10
 val TARGET_MAIN = "javaslang/src-gen/main/java"
 val TARGET_TEST = "javaslang/src-gen/test/java"
@@ -353,7 +353,7 @@ def generateMainClasses(): Unit = {
               /$javadoc
                * Alias for {@link Function$i#of(Function$i)}
                *
-               ${(0 to i).gen(j => if (j == 0) "* @param <R>             return type" else s"* @param <T$j>            type of the ${j.ordinal} argument")("\n")}
+               ${(0 to i).gen(j => if (j == 0) "* @param <R>             return type" else s"* @param <T$j>${if (j < 10) " " else ""}           type of the ${j.ordinal} argument")("\n")}
                * @param methodReference A method reference
                * @return A {@link Function$i}
                */
@@ -3201,7 +3201,7 @@ def generateTestClasses(): Unit = {
               ${(i > 0).gen(xs"""
                 @$test
                 public void shouldReturnElements() {
-                    final Tuple$i$intGenerics tuple = createIntTuple(${(1 to i).gen(j => s"$j") mkString ", "});
+                    final Tuple$i$intGenerics tuple = createIntTuple(${(1 to i).gen(j => s"$j")(", ")});
                     ${(1 to i).gen(j => s"$assertThat(tuple._$j).isEqualTo($j);\n")}
                 }
               """)}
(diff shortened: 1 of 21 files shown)


javaslang/generator/Generator.scala
```diff
@@ -353,7 +353,7 @@ def generateMainClasses(): Unit = {
               /$javadoc
                * Alias for {@link Function$i#of(Function$i)}
                *
-               ${(0 to i).gen(j => if (j == 0) "* @param <R>             return type" else s"* @param <T$j>            type of the ${j.ordinal} argument")("\n")}
+               ${(0 to i).gen(j => if (j == 0) "* @param <R>             return type" else s"* @param <T$j>${if (j < 10) " " else ""}           type of the ${j.ordinal} argument")("\n")}
```
paplorinc: alignment

javaslang/generator/Generator.scala
```diff
@@ -3201,7 +3201,7 @@ def generateTestClasses(): Unit = {
               ${(i > 0).gen(xs"""
                 @$test
                 public void shouldReturnElements() {
-                    final Tuple$i$intGenerics tuple = createIntTuple(${(1 to i).gen(j => s"$j") mkString ", "});
+                    final Tuple$i$intGenerics tuple = createIntTuple(${(1 to i).gen(j => s"$j")(", ")});
```
paplorinc: `mkString` generated garbage for some reason :/

javaslang/generator/Generator.scala
```diff
@@ -3522,16 +3522,16 @@ object Generator {
 
     // returns i as ordinal, i.e. 1st, 2nd, 3rd, 4th, ...
     def ordinal: String =
-      if (i / 10 == 1) {
-        s"${i}th"
+      s"$i" + (if (i >= 4 && i <= 20) {
+        "th"
       } else {
         i % 10 match {
-          case 1 => "1st"
```
paplorinc: `21 % 10 == 1 -> 1st` was incorrect

javaslang/src/main/java/javaslang/$.java
```diff
@@ -42,6 +42,10 @@
     static <T1, T2, T3, T4, T5, T6, T7> Tuple7<T1, T2, T3, T4, T5, T6, T7> Tuple7(Tuple7<T1, T2, T3, T4, T5, T6, T7> tuple7) { return tuple7; }
     @Unapply
     static <T1, T2, T3, T4, T5, T6, T7, T8> Tuple8<T1, T2, T3, T4, T5, T6, T7, T8> Tuple8(Tuple8<T1, T2, T3, T4, T5, T6, T7, T8> tuple8) { return tuple8; }
+    @Unapply
```
paplorinc: these had to be added manually.
@danieldietrich, shouldn't these be generated instead?

javaslang/generator/Generator.scala
danieldietrich: Right, this looks better. could you please copy this updated code also to javaslang-test/generator/Generator.scala? Currently we have redundant code :-//

I will change the Generator in a (near?) future. Currently the maven build process hinders us from having one general Generator import.

javaslang/src/main/java/javaslang/$.java
danieldietrich: In a future Javaslang the patterns will be part of the API class if possible. Then the methods will be generated.

**Currently I search a way for the following (targeted for 2.1.0):**

$.java will be deprecated (including Patterns.java, which is generated by the annotation processor). The core project 'javaslang' will not depend on javaslang-match any more (currently only compile-time). The patterns will be inlcuded in API, e.g.

* Tuple(1, 2) should create a Tuple
* Tuple($(1), $(2)) should create a Pattern that matches a Tuple

Also some patterns might change. E.g. Cons(head, tail) and Nil() instead of List(head, tail) and List().

javaslang/generator/Generator.scala
paplorinc: Done

danieldietrich:
Awesome, thank you!

PR #1855, opened 2017-02-03
CharSeq.repeat >10× speedup

paplorinc:
original:
```java
Target            Operation   Impl                  Params  Count            Score  ±     Error    Unit  slang_persistent
CharSeqBenchmark  Repeat      slang_persistent          10      5     4,651,788.68  ±    23.81%   ops/s
CharSeqBenchmark  Repeat      slang_persistent         100      5       639,050.67  ±    20.89%   ops/s
CharSeqBenchmark  Repeat      slang_persistent        1000      5       115,828.61  ±     6.77%   ops/s
```
with doubling:
```java
Target            Operation   Impl                  Params  Count            Score  ±     Error    Unit  slang_persistent
CharSeqBenchmark  Repeat      slang_persistent          10      5    10,155,280.50  ±    12.97%   ops/s
CharSeqBenchmark  Repeat      slang_persistent         100      5     6,494,596.25  ±     4.02%   ops/s
CharSeqBenchmark  Repeat      slang_persistent        1000      5     1,204,059.76  ±     6.58%   ops/s
```
with `char[]` doubling:
```java
Target            Operation   Impl                  Params  Count            Score  ±     Error    Unit  slang_persistent
CharSeqBenchmark  Repeat      slang_persistent          10      5    13,406,385.45  ±     4.30%   ops/s                 
CharSeqBenchmark  Repeat      slang_persistent         100      5     9,117,266.10  ±     3.28%   ops/s                 
CharSeqBenchmark  Repeat      slang_persistent        1000      5     1,606,002.67  ±     4.93%   ops/s   
```

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/CharSeq.java
+++ b/javaslang/src/main/java/javaslang/collection/CharSeq.java
@@ -333,11 +333,20 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @return A CharSeq representing {@code this * times}
      */
     public CharSeq repeat(int times) {
-        final StringBuilder builder = new StringBuilder();
-        for (int i = 0; i < times; i++) {
-            builder.append(back);
+        if (times <= 0) return empty();
+        else if (times == 1) return this;
+        else {
+            final int finalLength = length() * times;
+            final StringBuilder result = new StringBuilder(2 * finalLength);
+            result.append(back);
+
+            for (int i = 1; i <= times; i *= 2) {
+                result.append(result);
+            }
+            result.setLength(finalLength);
+
+            return of(result);
         }
-        return of(builder);
     }
 
     //
--- a/javaslang/src/main/java/javaslang/collection/CharSeq.java
+++ b/javaslang/src/main/java/javaslang/collection/CharSeq.java
@@ -333,19 +333,20 @@ public final class CharSeq implements Kind1<CharSeq, Character>, CharSequence, I
      * @return A CharSeq representing {@code this * times}
      */
     public CharSeq repeat(int times) {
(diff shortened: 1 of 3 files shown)


javaslang/src/main/java/javaslang/collection/CharSeq.java
```diff
@@ -333,11 +333,20 @@ public static CharSeq repeat(char character, int times) {
...
+        else if (times == 1) return this;
+        else {
+            final int finalLength = length() * times;
+            final StringBuilder result = new StringBuilder(2 * finalLength);
+            result.append(back);
+
+            for (int i = 1; i <= times; i *= 2) {
+                result.append(result);
+            }
+            result.setLength(finalLength);
```
paplorinc: ... could also be a final substring append, but this is probably faster

danieldietrich: Nice solution! Guava does it directly with arrays instead of using a StringBuilder. Also Guava's size check and the last array copy (to save mem) are nice:

https://github.com/google/guava/blob/c462d69329709f72a17a64cb229d15e76e72199c/guava/src/com/google/common/base/Strings.java#L142-L166

What do you think, give it a try the 'Guava way'?

paplorinc: Good idea indeed, let me see how quick that is :)

paplorinc: Hmm, it seems to me their version is buggy for cases when the repetition count is a power of 2.
Will provide a PR for Guava also (luckly I didn't copy their impl, so I could compare it against mine).

danieldietrich:
Speedup looks \*great\*, thank you!

PR #1857, opened 2017-02-09
Revert generated Tuple ariaty to 8 for GWT compatibility

paplorinc:


danieldietrich:
@paplorinc Thank you! The GWT compiler 'gwt-dev' internally uses Eclipse Java Development Tools (JDT) resp. Eclipse Compiler for Java (ECJ). My experience with these tools is that there were plenty of type inference bugs in the past. The early versions of JDT for Java 8 let Eclipse completely freeze when opening Javaslang's classes in an editor.

There are several issues related to JDT on the GWT issue tracker, maybe this problem is also related to JDT. We should wait for an JDT update and try it again (see https://github.com/gwtproject/gwt/issues/9461#issuecomment-261611288). Maybe we are able to find a minimal example (one file with many generics) that let's the GWT compiler stop working.

PR #1858, opened 2017-02-09
Set.add(existing) does nothing, Map.put(existing, any) overwrites

### LINKED ISSUE #1827: Don't insert a value into a Set, if it is already present
This depends on the underlying Comparator (for sorted sets) or on the underlying equals implementation of elements.

Currently TreeSet.add(value) returns a new instance, even if value is already present (regarding the underlying comparator).

E.g. when having a TreeSet<Integer> with comparator `(i, j) -> i % 2 - j % 2` then 2 and 4 are considered equal. Currently TreeSet(2).add(4) yields TreeSet(4) but it should be TreeSet(2).

Java acts the same way (not inserting values that are already present):

```java
public class Test {

        // = [2]
        test(new java.util.HashSet<>());

        // = [2]
        test(new java.util.TreeSet<>((i, j) -> (i.val % 2) - (j.val % 2)));

    }

    static void test(java.util.Set<Int> set) {
        set.add(new Int(2));
        set.add(new Int(4));
        System.out.println(set);
    }

    static class Int implements Comparable<Int> {

        final int val;

        Int(int val) {
            this.val = val;
        }

        @Override
        public int compareTo(Int that) {
            return this.val - that.val;
        }

        @Override
            return o == this || o instanceof Int && ((Int) o).val % 2 == val % 2;
        }

        @Override
        public int hashCode() {
            return val % 2;
        }

        @Override
        public String toString() {
            return String.valueOf(val);
        }
    }

}
```

danieldietrich:
Ensures proper Set.add and Map.put behaviour. Fixes #1827

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/HashSet.java
+++ b/javaslang/src/main/java/javaslang/collection/HashSet.java
@@ -473,7 +473,7 @@ public final class HashSet<T> implements Kind1<HashSet<?>, T>, Set<T>, Serializa
 
     @Override
     public HashSet<T> add(T element) {
-        return new HashSet<>(tree.put(element, element));
+        return contains(element) ? this : new HashSet<>(tree.put(element, element));
     }
 
     @Override
--- a/javaslang/src/main/java/javaslang/collection/LinkedHashSet.java
+++ b/javaslang/src/main/java/javaslang/collection/LinkedHashSet.java
@@ -477,7 +477,7 @@ public final class LinkedHashSet<T> implements Kind1<LinkedHashSet<?>, T>, Set<T
 
     @Override
     public LinkedHashSet<T> add(T element) {
-        return new LinkedHashSet<>(map.put(element, element));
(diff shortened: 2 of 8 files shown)


javaslang/src/main/java/javaslang/collection/HashSet.java
```diff
@@ -473,7 +473,7 @@ private HashSet(HashArrayMappedTrie<T, T> tree) {
 
     @Override
     public HashSet<T> add(T element) {
-        return new HashSet<>(tree.put(element, element));
+        return contains(element) ? this : new HashSet<>(tree.put(element, element));
```
danieldietrich: If key is present, tree.put(k,v) needs to return a new instance for Maps and the same instance for Sets. Therefore we currently perform an additional 'contains' check for Sets. A future optimization may add an additional flag `replace` to the backing HAMT.put() / RedBlackTree.insert() methods. Sets set it to replace=false, Maps set it to replace=true. But for now this solution is working for us.

Issue #1466, opened 2016-08-01
map() broken for Multimap backed by SortedSet with comparator


docent:
map() throws exception when called on a `Multimap` backed by a `SortedSet` with a custom comparator, e.g.:

``` java
Multimap<String, Thread> m = HashMultimap.withSortedSet(Comparator.comparing(Thread::getName))
                .of("t1", new Thread("t1c"), "t1", new Thread("t1b"));
m.map(t -> t);
```

It complains about V not implementing `Comparable`:

```
java.lang.ClassCastException: java.lang.Thread cannot be cast to java.lang.Comparable
    at javaslang.Tuple2.compareTo(Tuple2.java:81)
    at javaslang.Tuple2.compareTo(Tuple2.java:97)
    at javaslang.Tuple2.compareTo(Tuple2.java:29)
    at javaslang.collection.Comparators.lambda$naturalComparator$7c28a3e$1(Comparators.java:30)
    at javaslang.collection.RedBlackTreeModule$Node.insert(RedBlackTree.java:656)
    at javaslang.collection.RedBlackTree.insert(RedBlackTree.java:97)
    at javaslang.collection.RedBlackTree.ofAll(RedBlackTree.java:84)
    at javaslang.collection.TreeSet.ofAll(TreeSet.java:206)
    at javaslang.collection.TreeSet.map(TreeSet.java:715)
    at javaslang.collection.TreeSet.map(TreeSet.java:720)
    at javaslang.collection.TreeSet.map(TreeSet.java:32)
    at javaslang.collection.AbstractMultimap.lambda$iterator$9(AbstractMultimap.java:303)
    at javaslang.collection.Iterator$35.hasNext(Iterator.java:1487)
    at javaslang.collection.Iterator.map(Iterator.java:1578)
    at javaslang.collection.Multimap.map(Multimap.java:389)
```

This also breaks `toString()` on such multimaps.

valery1707:
I confirm such behavior - trying to fix

valery1707:
Internally TreeSet mapped from `V` into `Tuple<K, V>` and this tuples compared between each other. Tuple trying compare values directly, but values not comparable.

danieldietrich:
@docent thanks for reporting!
@valery1707 thanks for the analysis! Need to check the code - this is a general problem.

I will be on holiday in one day (for 4 weeks...) - I will have my notebook with me ;-)

ruslansennov:
@danieldietrich @valery1707 
[...]

PR #1865, opened 2017-02-14
fixes Multimap.withSortedSet().map() for incomparable values

ruslansennov:
see #1466

### DIFF
--- a/javaslang/src/test/java/javaslang/collection/AbstractMultimapTest.java
+++ b/javaslang/src/test/java/javaslang/collection/AbstractMultimapTest.java
@@ -13,10 +13,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.LinkedList;
-import java.util.NoSuchElementException;
+import java.util.*;
 import java.util.function.BiConsumer;
 import java.util.function.BinaryOperator;
 import java.util.function.Function;
@@ -25,6 +22,7 @@ import java.util.regex.Pattern;
 import java.util.stream.Collector;
 
 import static java.util.Arrays.asList;
+import static javaslang.collection.Comparators.naturalComparator;
 
 @RunWith(Parameterized.class)
 public abstract class AbstractMultimapTest extends AbstractTraversableTest {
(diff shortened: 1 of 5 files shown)


javaslang/src/test/java/javaslang/collection/AbstractMultimapTest.java
```diff
@@ -453,6 +455,57 @@ public void shouldMapNonEmpty() {
...
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+
+            Incomparable that = (Incomparable) o;
+
+            return s != null ? s.equals(that.s) : that.s == null;
```
danieldietrich: can be simplified with

```java
return o == this || (o instanceof Incomparable && Objects.equals(s, ((Incomparable) o).s));
```

javaslang/src/test/java/javaslang/collection/AbstractMultimapTest.java
```diff
@@ -453,6 +455,57 @@ public void shouldMapNonEmpty() {
...
+                .put(1, new Incomparable("1"))
+                .put(1, new Incomparable("2"))
+                .put(2, new Incomparable("3"));
+        assertThat(map.map(v -> v)).isEqualTo(List.of(
+                Tuple.of(1, new Incomparable("1")),
+                Tuple.of(1, new Incomparable("2")),
+                Tuple.of(2, new Incomparable("3"))));
+    }
+
+    private static class Incomparable {
```
danieldietrich: would make it final because of equals() instanceof check

javaslang/src/test/java/javaslang/collection/AbstractMultimapTest.java
```diff
@@ -453,6 +455,57 @@ public void shouldMapNonEmpty() {
...
+            if (o == null || getClass() != o.getClass()) return false;
+
+            Incomparable that = (Incomparable) o;
+
+            return s != null ? s.equals(that.s) : that.s == null;
+        }
+
+        @Override
+        public int hashCode() {
+            return s != null ? s.hashCode() : 0;
```
danieldietrich: same as `Objects.hashCode(s)`;

javaslang/src/test/java/javaslang/collection/HashMultimapTest.java
```diff
@@ -287,7 +287,7 @@ public void shouldCreateSortedMapFrom10Pairs() {
 
     @Test
     public void shouldNarrowMap() {
-        final HashMultimap<Integer, Number> int2doubleMap = this.<Integer, Number> emptyMap().put(1, 1.0d);
+        final HashMultimap<Integer, Number> int2doubleMap = (HashMultimap<Integer, Number>) this.<Integer, Number> emptyMap().put(1, 1.0d);
```
danieldietrich: I would not provide a default implementation for `AbstractMultimapTest.emptyMap()` because now we have to cast every `emptyMap()` result.
Should not be a big problem, we have only three impls in our test classes.

ruslansennov: good point

ruslansennov: No, we have to cast in three places only, but Idea shows 33 usages of `emptyMap()`.
PR updated

danieldietrich:
Thank you, it looks great!

Issue #1873, opened 2017-02-19
Ensure <collection>.ofAll(Iterable) returns the Iterable if it has the correct type


danieldietrich:
Example:

```java
public static <T> PriorityQueue<T> ofAll(Comparator<? super T> comparator, Iterable<? extends T> elements) {
    Objects.requireNonNull(elements, "elements is null");
    if (elements instanceof PriorityQueue) {
        return (PriorityQueue<T>) elements;
    } else {
        ...
    }
}
```

ruslansennov:
I believe this is wrong way (for `PriorityQueue` at least). Someone may want to create instance based on another one with new comparator

danieldietrich:
@ruslansennov absolutely right - I've overseen that.

* SortedSets have a Comparator
* BitSet has an toInt Function
* Multimaps have a value container
* ...

ruslansennov:
We can do it like in `TreeSet`
```java
if (values instanceof TreeSet && ((TreeSet) values).comparator() == comparator) {
    return (TreeSet<T>) values;
} else {
    // ...
}
```

danieldietrich:
@ruslansennov That looks like a good solution! The check isn't costly compared to creating new instances.

pivovarit:
Done some investigating. Summarizing here:

`Array` has it already:
https://github.com/vavr-io/vavr/blob/dcd456536b26df01fcc1930baf7a2a832c808819/src/main/java/io/vavr/collection/Array.java#L147-L149

`HashSet` has it already:
https://github.com/vavr-io/vavr/blob/a53c966a27eb3ddbf88cab3de31801a35d29ea9e/src/main/java/io/vavr/collection/HashSet.java#L157-L159

`CharSeq` has it already:
https://github.com/vavr-io/vavr/blob/b45b755726780534eec8e46ba02b9a80c7b6f3b4/src/main/java/io/vavr/collection/CharSeq.java#L133-L135

`LinkedHashSet` has it already:
https://github.com/vavr-io/vavr/blob/bef625a6b1bce8e5fbfca3a4113ad1fc51fce264/src/main/java/io/vavr/collection/LinkedHashSet.java#L161-L163

`List` has it already:
https://github.com/vavr-io/vavr/blob/dcd456536b26df01fcc1930baf7a2a832c808819/src/main/java/io/vavr/collection/List.java#L237-L239

`PriorityQueue` has it already:
[...]

PR #1876, opened 2017-02-20
orElse javadoc

ruslansennov:
see #1872 #1873 
`PriorityQueue` refactored as @paplorinc suggested

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/PriorityQueue.java
+++ b/javaslang/src/main/java/javaslang/collection/PriorityQueue.java
@@ -191,16 +191,20 @@ public final class PriorityQueue<T> extends AbstractsQueue<T, PriorityQueue<T>> 
         return ofAll(naturalComparator(), elements);
     }
 
+    @SuppressWarnings("unchecked")
     public static <T> PriorityQueue<T> ofAll(Comparator<? super T> comparator, Iterable<? extends T> elements) {
         Objects.requireNonNull(elements, "elements is null");
-
-        int size = 0;
-        Seq<Node<T>> forest = List.empty();
-        for (T value : elements) {
-            forest = insert(comparator, value, forest);
-            size++;
+        if (elements instanceof PriorityQueue && ((PriorityQueue) elements).comparator == comparator) {
+            return (PriorityQueue<T>) elements;
+        } else {
+            int size = 0;
+            Seq<Node<T>> forest = List.empty();
+            for (T value : elements) {
+                forest = insert(comparator, value, forest);
+                size++;
+            }
+            return new PriorityQueue<>(comparator, forest, size);
         }
-        return new PriorityQueue<>(comparator, forest, size);
     }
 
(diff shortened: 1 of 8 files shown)


javaslang/src/test/java/javaslang/collection/AbstractTraversableTest.java
```diff
@@ -1264,9 +1264,15 @@ public void shouldCaclNonemptyOrElseOther() {
     }
 
     @Test
-    public void shouldCaclEmptyOrElseSupplier() {
-        Supplier<Iterable<Integer>> other = () -> of(42);
-        assertThat(empty().orElse(other)).isEqualTo(of(42));
+    public void shouldCaclEmptyOrElseSameSupplier() {
+        Iterable<Integer> other = of(42);
+        Supplier<Iterable<Integer>> supplier = () -> other;
+        assertThat(empty().orElse(supplier)).isSameAs(other);
```
ruslansennov: also adds test to check that `orElse()` result is same as result of evaluating supplier in case of properly iterable type

javaslang/src/main/java/javaslang/collection/PriorityQueue.java
```diff
@@ -419,14 +423,27 @@ public int length() {
...
+     * @param other An alternative {@code Traversable}
+     * @return this {@code PriorityQueue} if it is nonempty,
+     * otherwise {@code PriorityQueue} created from iterable, using existing comparator.
+     */
     @Override
     public PriorityQueue<T> orElse(Iterable<? extends T> other) {
-        return isEmpty()
-                ? other instanceof PriorityQueue ? (PriorityQueue<T>) other : ofAll(comparator, other)
-                : this;
+        return isEmpty() ? ofAll(comparator, other) : this;
```
danieldietrich: Great!

javaslang/src/main/java/javaslang/collection/PriorityQueue.java
```diff
@@ -191,16 +191,20 @@ public boolean isOrdered() {
...
+    @SuppressWarnings("unchecked")
     public static <T> PriorityQueue<T> ofAll(Comparator<? super T> comparator, Iterable<? extends T> elements) {
         Objects.requireNonNull(elements, "elements is null");
-
-        int size = 0;
-        Seq<Node<T>> forest = List.empty();
-        for (T value : elements) {
-            forest = insert(comparator, value, forest);
-            size++;
+        if (elements instanceof PriorityQueue && ((PriorityQueue) elements).comparator == comparator) {
```
paplorinc: :+1: 
This means the exact same instance of the comparator, not the same logic, right?
Seems ok to me :)

ruslansennov: yes, exactly like in `TreeSet`

danieldietrich: @paplorinc "exact same instance" is the best be can do because function equality is [undecidable](http://stackoverflow.com/a/25581222/1110815)

PR #1892, opened 2017-03-05
Serializable

### LINKED ISSUE #1810: Make interfaces Serializable
@ggalmazor noted that we (unintentionally) changed the object serializability. E.g. Option isn't serializable any more (but Some and None are). This hinders us from adding fields of interface types (like Option) to serializable classes.

There are exceptions, stateful types should **not** implement Serializable:

* Future
* Promise
* Iterator

Also API specific types should not implement it:

* For
* Match, Case, Pattern
* ...

danieldietrich:
Fixes #1810

### DIFF
--- a/javaslang/src/test/java/javaslang/AbstractValueTest.java
+++ b/javaslang/src/test/java/javaslang/AbstractValueTest.java
@@ -98,7 +98,7 @@ public abstract class AbstractValueTest {
     @SuppressWarnings("unchecked")
     abstract protected <T> Value<T> of(T... elements);
 
-    // TODO: Eliminate this method. Switching the behavior of unit tests is evil. Tests should not contain additional logic.
+    // TODO: Eliminate this method. Switching the behavior of unit tests is evil. Tests should not contain additional logic. Also it seems currently to be used in different sematic contexts.
     abstract protected boolean useIsEqualToInsteadOfIsSameAs();
 
     // returns the peek result of the specific Traversable implementation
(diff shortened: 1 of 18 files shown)


javaslang/src/test/java/javaslang/AbstractValueTest.java
```diff
@@ -981,14 +981,14 @@ private boolean isSerializable() {
                         instanceOf(Either.LeftProjection.class),
                         instanceOf(Either.RightProjection.class),
                         instanceOf(Future.class),
-                        instanceOf(Iterator.class),
-                        instanceOf(Validation.class)
+                        instanceOf(Iterator.class)
```
danieldietrich: Matches before `instanceOf(Traversable.class)` is checked, so it is ok.

javaslang/src/test/java/javaslang/AbstractValueTest.java
```diff
@@ -981,14 +981,14 @@ private boolean isSerializable() {
...
                 ), false),
                 Case(anyOf(
                         instanceOf(Either.class),
                         instanceOf(Option.class),
                         instanceOf(Try.class),
-                        instanceOf(Traversable.class)
+                        instanceOf(Traversable.class),
```
danieldietrich: In 3.0.0 Iterator will not implement Traversable any more (i.e. all internal `ofAll(iterator)` calls will be substituted with `ofAll(() -> iterator)`). So all Traversables will be Serializable.

PR #1895, opened 2017-03-07
fix gwt

ruslansennov:


danieldietrich:
Thank you, I've overseen that. Maybe we can provide a GWT compatible version, s.th. like

```java
toJavaList().toArray(new Character[size()]);
```

PR #1904, opened 2017-03-09
Based BMT.iterator() on Iterator instead of AbstractIterator

danieldietrich:
@paplorinc could you please review it?

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
+++ b/javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
@@ -6,6 +6,7 @@
 package javaslang.collection;
 
 import java.io.Serializable;
+import java.util.NoSuchElementException;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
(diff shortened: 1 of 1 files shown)


danieldietrich:
@paplorinc I think this is the best we can do for now. Please approve or reject the code review.

In 3.0.0 we will remove AbstractIterator. Iterator will not have a toString() overload any more, which should be ok. However, it is only interesting for the empty case...

javaslang/src/main/java/javaslang/collection/BitMappedTrie.java
```diff
@@ -282,7 +283,11 @@ private Object getLeafGeneral(int index) {
             public boolean hasNext() { return globalIndex < globalLength; }
 
             @Override
-            public T getNext() {
+            public T next() {
```
paplorinc: :+1:, `hasNext` is very simple here, shouldn't really affect the speed :)

danieldietrich:
Thank you, Lorinc :)

PR #1907, opened 2017-03-10
Fixes memory leak of grouped iterator

### LINKED ISSUE #1899: Iterator.grouped leaks memory
Please consider following snippets and try to execute them:

```java

class InfiniteGroupedIterators {

  private double count = 0;

  private class Message {
    private final double count;

    Message(double count) {
      this.count = count;
    }
  }

  InfiniteGroupedIterators() {
    getInfiniteIterator()
        .grouped(100)
        .zipWithIndex()
        .forEach(t -> System.out.println(String.format("Processed %s group", String.valueOf(t._2))));
  }

  private Iterator<Message> getInfiniteIterator() {
    return Iterator.continually(() -> {
      count++;
      return new Message(count);
    });
  }
}
```

vs

```scala
object InfiniteGroupedIterators {

  case class Message(count: Double)

  var index = 0d

  def infiniteIterator = Iterator.continually(() => {
    index = index + 1
    Message(index)
  })

  infiniteIterator
    .grouped(100)
    .zipWithIndex
    .foreach { t => println(s"Processed ${t._2} group") }

}
```

Javaslang's `Iterator.grouped` OOMs, Scala's `Iterator.grouped` doesn't. Used sliding view should be garbage collectable and evaluated elements shouldn't be held indefinitely, which Stream class does [here](https://github.com/javaslang/javaslang/blob/master/javaslang/src/main/java/javaslang/collection/Iterator.java#L1836) I think.

danieldietrich:
Fixes #1899

### DIFF
--- a/javaslang-benchmark/src/test/java/javaslang/collection/IteratorBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/collection/IteratorBenchmark.java
@@ -1,0 +1,75 @@
+/*     / \____  _    _  ____   ______  / \ ____  __    _______
+ *    /  /    \/ \  / \/    \ /  /\__\/  //    \/  \  //  /\__\   JΛVΛSLΛNG
+ *  _/  /  /\  \  \/  /  /\  \\__\\  \  //  /\  \ /\\/ \ /__\ \   Copyright 2014-2017 Javaslang, http://javaslang.io
+ * /___/\_/  \_/\____/\_/  \_/\__\/__/\__\_/  \_//  \__/\_____/   Licensed under the Apache License, Version 2.0
+ */
+package javaslang.collection;
+
+import javaslang.JmhRunner;
+import org.junit.Test;
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.util.Random;
+
+import static java.util.Arrays.asList;
+import static javaslang.JmhRunner.Includes.*;
+import static javaslang.JmhRunner.getRandomValues;
+import static scala.collection.JavaConversions.asScalaBuffer;
+
+@SuppressWarnings({ "ALL", "unchecked", "rawtypes" })
+public class IteratorBenchmark {
+    
+    static final Array<Class<?>> CLASSES = Array.of(
+            Sliding.class
+    );
+
+    @Test
+    public void testAsserts() { JmhRunner.runDebugWithAsserts(CLASSES); }
+
+    public static void main(String... args) {
+        JmhRunner.runDebugWithAsserts(CLASSES);
+        JmhRunner.runNormalNoAsserts(CLASSES, JAVA, SCALA, JAVASLANG);
+    }
+
+    @State(Scope.Benchmark)
+    public static class Base {
+        @Param({ "10", "100", "1000" })
+        public int CONTAINER_SIZE;
+
+        Integer[] ELEMENTS;
+
+        /* Only use these for non-mutating operations */
+        java.util.List<Integer> javaMutable;
+        scala.collection.Iterator<Integer> scalaPersistent;
+        javaslang.collection.Iterator<Integer> slangPersistent;
+
+        @Setup
+        public void setup() {
+            final Random random = new Random(0);
+            ELEMENTS = getRandomValues(CONTAINER_SIZE, false, random);
+            javaMutable = asList(ELEMENTS);
+            scalaPersistent = (scala.collection.Iterator<Integer>) scala.collection.immutable.Seq$.MODULE$.apply(asScalaBuffer(javaMutable)).iterator();
+            slangPersistent = Iterator.of(ELEMENTS);
+        }
+    }
+
+    public static class Sliding extends Base {
+
+        @Benchmark
+        public void scala_persistent(Blackhole bh) {
+            final scala.collection.Iterator.GroupedIterator values = scalaPersistent.sliding(3, 1);
+            while (values.hasNext()) {
+                bh.consume(values.next());
+            }
+        }
+
+        @Benchmark
+        public void slang_persistent(Blackhole bh) {
+            final Iterator<Seq<Integer>> values = slangPersistent.sliding(3);
+            while (values.hasNext()) {
+                bh.consume(values.next());
+            }
+        }
+    }
+}
--- a/javaslang-benchmark/src/test/java/javaslang/collection/IteratorBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/collection/IteratorBenchmark.java
@@ -12,7 +12,6 @@ import org.openjdk.jmh.infra.Blackhole;
 
 import java.util.Random;
 
-import static java.util.Arrays.asList;
 import static javaslang.JmhRunner.Includes.*;
 import static javaslang.JmhRunner.getRandomValues;
 
@@ -38,7 +37,6 @@ public class IteratorBenchmark {
 
         Integer[] ELEMENTS;
 
-        /* Only use these for non-mutating operations */
         scala.collection.Iterator<Integer> scalaIterator;
         javaslang.collection.Iterator<Integer> slangIterator;
 
--- a/javaslang-benchmark/src/test/java/javaslang/collection/IteratorBenchmark.java
+++ b/javaslang-benchmark/src/test/java/javaslang/collection/IteratorBenchmark.java
@@ -15,7 +15,6 @@ import java.util.Random;
 import static java.util.Arrays.asList;
 import static javaslang.JmhRunner.Includes.*;
 import static javaslang.JmhRunner.getRandomValues;
-import static scala.collection.JavaConversions.asScalaBuffer;
 
 @SuppressWarnings({ "ALL", "unchecked", "rawtypes" })
 public class IteratorBenchmark {
@@ -40,17 +39,15 @@ public class IteratorBenchmark {
         Integer[] ELEMENTS;
 
         /* Only use these for non-mutating operations */
-        java.util.List<Integer> javaMutable;
-        scala.collection.Iterator<Integer> scalaPersistent;
-        javaslang.collection.Iterator<Integer> slangPersistent;
+        scala.collection.Iterator<Integer> scalaIterator;
+        javaslang.collection.Iterator<Integer> slangIterator;
 
         @Setup
         public void setup() {
             final Random random = new Random(0);
             ELEMENTS = getRandomValues(CONTAINER_SIZE, false, random);
-            javaMutable = asList(ELEMENTS);
-            scalaPersistent = (scala.collection.Iterator<Integer>) scala.collection.immutable.Seq$.MODULE$.apply(asScalaBuffer(javaMutable)).iterator();
-            slangPersistent = Iterator.of(ELEMENTS);
+            scalaIterator = (scala.collection.Iterator<Integer>) (Object) scala.collection.mutable.WrappedArray$.MODULE$.make(ELEMENTS).iterator();
+            slangIterator = Iterator.of(ELEMENTS);
         }
     }
 
@@ -58,7 +55,7 @@ public class IteratorBenchmark {
 
         @Benchmark
         public void scala_persistent(Blackhole bh) {
-            final scala.collection.Iterator.GroupedIterator values = scalaPersistent.sliding(3, 1);
+            final scala.collection.Iterator.GroupedIterator values = scalaIterator.sliding(3, 1);
             while (values.hasNext()) {
                 bh.consume(values.next());
             }
@@ -66,7 +63,7 @@ public class IteratorBenchmark {
 
         @Benchmark
         public void slang_persistent(Blackhole bh) {
-            final Iterator<Seq<Integer>> values = slangPersistent.sliding(3);
+            final Iterator<Seq<Integer>> values = slangIterator.sliding(3);
             while (values.hasNext()) {
                 bh.consume(values.next());
             }
--- a/javaslang/src/main/java/javaslang/collection/Iterator.java
+++ b/javaslang/src/main/java/javaslang/collection/Iterator.java
@@ -2066,6 +2066,7 @@ interface IteratorModule {
     final class GroupedIterator<T> implements Iterator<Seq<T>> {
 
         private final Iterator<T> that;
+        private final int size;
         private final int step;
         private final int gap;
 
@@ -2076,6 +2077,7 @@ interface IteratorModule {
                 throw new IllegalArgumentException("size (" + size + ") and step (" + step + ") must both be positive");
             }
             this.that = that;
+            this.size = size;
             this.step = step;
             this.gap = Math.max(step - size, 0);
             this.buffer = List.ofAll(that.take(size));
@@ -2091,8 +2093,10 @@ interface IteratorModule {
             if (!hasNext()) {
                 throw new NoSuchElementException();
             }
-            final Seq<T> result = buffer;
-            buffer = that.hasNext() ? that.drop(gap).take(step).toList().prependAll(buffer.drop(step)) : buffer.take(0);
+            Seq<T> result = buffer;
+            buffer = that.hasNext()
+                     ? buffer.drop(step).appendAll(gap > 0 ? that.drop(gap).take(size) : that.take(step))
+                     : buffer.take(0);
             return result;
         }
     }
--- a/javaslang/src/main/java/javaslang/collection/Iterator.java
+++ b/javaslang/src/main/java/javaslang/collection/Iterator.java
@@ -2069,8 +2069,9 @@ interface IteratorModule {
         private final int size;
         private final int step;
         private final int gap;
+        private final int preserve;
 
-        private List<T> buffer;
+        private Object[] buffer;
 
         GroupedIterator(Iterator<T> that, int size, int step) {
             if (size < 1 || step < 1) {
@@ -2080,24 +2081,57 @@ interface IteratorModule {
             this.size = size;
             this.step = step;
             this.gap = Math.max(step - size, 0);
-            this.buffer = List.ofAll(that.take(size));
+            this.preserve = Math.max(size - step, 0);
+            this.buffer = take(that, new Object[size], 0, size);
         }
 
         @Override
         public boolean hasNext() {
-            return !buffer.isEmpty();
+            return buffer.length > 0;
         }
 
         @Override
         public Seq<T> next() {
-            if (!hasNext()) {
+            if (buffer.length == 0) {
                 throw new NoSuchElementException();
             }
-            Seq<T> result = buffer;
-            buffer = that.hasNext()
-                     ? buffer.drop(step).appendAll(gap > 0 ? that.drop(gap).take(size) : that.take(step))
-                     : buffer.take(0);
-            return result;
+            final Object[] result = buffer;
+            if (that.hasNext()) {
+                buffer = new Object[size];
+                if (preserve > 0) {
+                    System.arraycopy(result, step, buffer, 0, preserve);
+                }
+                if (gap > 0) {
+                    drop(that, gap);
+                    buffer = take(that, buffer, preserve, size);
+                } else {
+                    buffer = take(that, buffer, preserve, step);
+                }
+            } else {
+                buffer = new Object[0];
+            }
+            return Array.wrap(result);
+        }
+
+        private static void drop(Iterator<?> source, int count) {
+            for (int i = 0; i < count && source.hasNext(); i++) {
+                source.next();
+            }
+        }
+
+        private static Object[] take(Iterator<?> source, Object[] target, int offset, int count) {
+            int i = offset;
+            while (i < count + offset && source.hasNext()) {
+                target[i] = source.next();
+                i++;
+            }
+            if (i < target.length) {
+                final Object[] result = new Object[i];
+                System.arraycopy(target, 0, result, 0, i);
+                return result;
+            } else {
+                return target;
+            }
         }
     }
 
--- a/javaslang/src/main/java/javaslang/collection/Iterator.java
+++ b/javaslang/src/main/java/javaslang/collection/Iterator.java
@@ -2069,7 +2069,7 @@ interface IteratorModule {
         private final int step;
         private final int gap;
 
-        private Seq<T> buffer;
+        private List<T> buffer;
 
         GroupedIterator(Iterator<T> that, int size, int step) {
             if (size < 1 || step < 1) {
@@ -2078,7 +2078,7 @@ interface IteratorModule {
             this.that = that;
             this.step = step;
             this.gap = Math.max(step - size, 0);
-            this.buffer = Vector.ofAll(that.take(size));
+            this.buffer = List.ofAll(that.take(size));
         }
 
         @Override
@@ -2092,7 +2092,7 @@ interface IteratorModule {
                 throw new NoSuchElementException();
             }
             final Seq<T> result = buffer;
-            buffer = that.hasNext() ? buffer.drop(step).appendAll(that.drop(gap).take(step)) : buffer.take(0);
+            buffer = that.hasNext() ? that.drop(gap).take(step).toList().prependAll(buffer.drop(step)) : buffer.take(0);
             return result;
         }
     }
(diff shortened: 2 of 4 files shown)


javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -1535,9 +1537,9 @@ default T get() {
 
     @Override
     default Iterator<Seq<T>> grouped(int size) {
```
danieldietrich: Scala's grouped() and sliding() functions return a `GroupedIterator<Seq<T>>`. That GroupedIterator has the ability of padding and skipping partial results (both not implemented in our version, yet).

It would be nice if we could provide the same functionality. Returning a `GroupedIterator<Seq<T>>` instead of `Iterator<Seq<T>>` will not be backward compatible. I will target that change for 3.0.0

danieldietrich: **Note:** I internally based the buffer on our persistent Vector, Scala users a mutable ArrayBuffer. Our version is blazing fast (a little slower than the mutable version). Congrats @paplorinc !!!

danieldietrich:
_(hello, travis-ci, why weren't you triggered?)_

danieldietrich:
travis-ci is not triggered: https://twitter.com/traviscistatus/status/840236186497048577

javaslang/src/main/java/javaslang/collection/Iterator.java
paplorinc: nice, could you share some benchmarks :)?

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -2089,6 +2063,72 @@ public String toString() {
...
+            return filled || fill();
+        }
+
+        @Override
+        protected Seq<T> getNext() {
+            filled = false;
+            return buffer;
+        }
+
+        private boolean fill() {
```
paplorinc: this seems very stateful, would it be possible to separate it into smaller, more focused methods?

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -2089,6 +2063,72 @@ public String toString() {
...
+            if (!that.hasNext()) {
+                return false;
+            }
+
+            final int count = buffer.isEmpty() ? size : step;
+            final Seq<T> xs = Vector.ofAll(that.take(count));
+
+            if (xs.isEmpty()) {
```
paplorinc: this is `true` when `count` is `0`, right?

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -2089,6 +2063,72 @@ public String toString() {
...
+                return false;
+            }
+
+            final int prevSize = buffer.size();
+            final boolean isFirst = prevSize == 0;
+            final int len = xs.size();
+            final int howMany = Math.min(len, size);
+
+            if (howMany < 1 || !isFirst && len <= gap) {
```
paplorinc: could we separate the `first` from the rest somehow?
It's difficult for me to understand it if it's intertwined :)

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -2089,6 +2063,72 @@ public String toString() {
...
+            if (isFirst) {
+                available = len;
+            } else {
+                available = Math.min(howMany, len - gap);
+                buffer = buffer.drop(Math.min(step, prevSize));
+            }
+
+            buffer = buffer.appendAll(xs.takeRight(available));
+
+            return filled = true;
```
paplorinc: this will always return true, right?
or did you mean `filled == true`? (in which case it needs some additional tests)

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: Yes, I'm also not satisfied with the fill() method. I will re-write it...

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: Yes, I had `if (count == 0)` but replaced it. But `count == 0` is better.

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: Yes, I separated it first into an if branch but there was duplicate code. Will try to do better.

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: Yes, I fused two statements:

```java
filled = true;
return true;
```

danieldietrich: (Btw - how cool is that: assignment is an expression in Java. Unbelievable!)

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: @paplorinc 

> It's difficult for me to understand it if it's intertwined :)

Look at [this 'beauty'](https://github.com/scala/scala/blob/86889b2707c5ddc466ac198c19f01caad4679de6/src/library/scala/collection/Iterator.scala#L1084-L1213)

javaslang/src/main/java/javaslang/collection/Iterator.java
paplorinc: :)) yeah, I rather wish `if` and `switch` and `for` were expressions...

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -2089,6 +2063,35 @@ public String toString() {
         }
     }
 
+    final class GroupedIterator<T> extends AbstractIterator<Seq<T>> {
```
danieldietrich: @paplorinc I simplified it...

danieldietrich:
I made a Strawman's benchmark (I started an example on the shell and looked on my stopwatch):

## Results first:

```java
Scala 2.12.1:
  Processed 4400000 groups 42 mb
--
Javaslang
  Using AbstractIterator
    // buffer = that.hasNext() ? buffer.appendAll(that.take(step)).drop(step) : buffer.take(0);
    Vector buffer: Processed 3400000 groups 63 mb
    Array  buffer: Processed 3400000 groups 63 mb
    // buffer = that.hasNext() ? buffer.drop(step).appendAll(that.drop(gap).take(step)) : buffer.take(0);
    Vector buffer: Processed 3500000 groups 63 mb
    Array  buffer: Processed 3500000 groups 63 mb
  Using Iterator
    // buffer = that.hasNext() ? buffer.drop(step).appendAll(that.drop(gap).take(step)) : buffer.take(0);
    Vector buffer: Processed 3908630 groups 63 mb
```

## Interpretation

I started `scala` resp. `java` with `-Xmx64m` on the same shell. The example is an infinite Iterator that calls grouped().

The interesting thing is that we face heavy garbage collection (GC), an effect that we can't see during our _real_ benchmarks. After a second the free mem is nearly zero and the GC has to collect garbage, otherwise the Iterator could not process next elements.

We observe that Scala does not go beyond 42 mb, although we gave 64 mb. Seems to be a Scala-internal thing. Java takes 63 mb.

The Scala execution is 1.12 x faster than the Javaslang version. In fact that does not really say something about the CPU performance of the Iterator. Scala uses a mutable ArrayBuffer to store elements, we use an immutable Vector (or Array). Immutable collections might produce more object instances because existing elements cannot be mutated. So the performance-gap between Scala and Javaslang _might_ also have to do with object instantiation and garbage collection.

JMH benchmarks will give us more insight...

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -2089,6 +2063,40 @@ public String toString() {
         }
     }
 
+    final class GroupedIterator<T> implements Iterator<Seq<T>> {
```
danieldietrich: I don't know if we should really switch to Iterator (instead of AbstractIterator) **in version 2.1.0** for slightly better performance...

In 3.0.0 I think we should use the faster Iterator (without a pretty toString()).

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -2089,6 +2063,40 @@ public String toString() {
...
+
+        private final Iterator<T> that;
+        private final int step;
+        private final int gap;
+
+        private List<T> buffer;
```
paplorinc: is `List` faster than `Vector` for this scenario?

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -2089,6 +2063,40 @@ public String toString() {
...
+            return !buffer.isEmpty();
+        }
+
+        @Override
+        public Seq<T> next() {
+            if (!hasNext()) {
+                throw new NoSuchElementException();
+            }
+            final Seq<T> result = buffer;
+            buffer = that.hasNext() ? that.drop(gap).take(step).toList().prependAll(buffer.drop(step)) : buffer.take(0);
```
paplorinc: * `drop(...).take(...)` seems like a `slice`
* you could use `appendAll` also, it's more intuitive to start with the prefix (it's implemented via `prependAll` in `List` anyway)

javaslang/src/main/java/javaslang/collection/List.java
```diff
@@ -1390,12 +1396,12 @@ default String stringPrefix() {
 
     @Override
     default List<T> take(int n) {
-        if (n >= length()) {
-            return this;
-        }
         if (n <= 0) {
             return empty();
         }
+        if (n >= length()) {
```
paplorinc: what's the difference here?

danieldietrich: we do the length() check second now

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: I did appendAll() before with Vector. List.prependAll() performs better in this use-case. See above. I will add a benchmark.

javaslang/src/main/java/javaslang/collection/List.java
paplorinc: I see that, but why is it important? It's a constant

javaslang/src/main/java/javaslang/collection/Iterator.java
paplorinc: `appendAll` uses `prependAll` internally:
```java
@Override
default List<T> appendAll(Iterable<? extends T> elements) {
    return List.<T> ofAll(elements).prependAll(this);
}
```

javaslang/src/main/java/javaslang/collection/List.java
danieldietrich: yes, I forgot, JIT magic

danieldietrich: every runtime env? GWT resp Javascript?

paplorinc: I don't know, but in every case it's a microoptimizatio, I suggest we only change it if it's easier to understand, I really doubt you could measure any speed difference :)

danieldietrich: if I can sleep better we have gained s.th.

paplorinc: This could help you even more: https://rainymood.com

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: I think about adding appendAll() to Traversable and therefor to Iterator. we could then operate entirely on the Iterator and finally call toList(). Maybe it is simpler to read. However, each operation requires a new Iterator instance.

danieldietrich: Mmhh, does not seem to be applicable here...

paplorinc: If you're lucky it will end up on the stack (i.e. escape analysis), you should measure the impact :)
Seems more readable/maintainable though, that's more important for me personally.

javaslang/src/main/java/javaslang/collection/List.java
danieldietrich: 😴

paplorinc: I use it sometimes when the environment is noisy, it's very relaxing :D

danieldietrich: I will try it at $work :)

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: I haven't done such measures yet - no clue which tool to use in order to see how things are organized within the JVM 🙄 do you have a hint?

paplorinc: `JMH` has some nice profiler addons and https://github.com/AdoptOpenJDK/jitwatch is also quite good for pure `Java` (couldn't use it for `Scala` though).
Let me know if you need help.

danieldietrich:
@paplorinc I added a simple sliding benchmark for slang/scala:

```java
Benchmark                                   (CONTAINER_SIZE)   Mode  Cnt         Score         Error  Units
IteratorBenchmark.Sliding.scala_persistent                10  thrpt    7  25055342.232 ± 1204029.823  ops/s
IteratorBenchmark.Sliding.scala_persistent               100  thrpt    7  24922099.560 ± 1634678.049  ops/s
IteratorBenchmark.Sliding.scala_persistent              1000  thrpt    7  23741708.468 ±  463918.240  ops/s
IteratorBenchmark.Sliding.slang_persistent                10  thrpt    7  15487347.449 ±  649896.302  ops/s
IteratorBenchmark.Sliding.slang_persistent               100  thrpt    7  15051798.533 ±  632192.673  ops/s
IteratorBenchmark.Sliding.slang_persistent              1000  thrpt    7  12654430.670 ± 1081082.228  ops/s




Detailed Performance Execution Report
============================================================================================================================================
  (Error: ±99% confidence interval, expressed as % of Score)
  (Outliers removed: 30.00% low end, 5.00% high end)
  (scala_persistent, slang_persistent: read as current row implementation is x times faster than alternative implementation)

Target             Operation   Impl                  Params  Count            Score  ±     Error    Unit  scala_persistent  slang_persistent
IteratorBenchmark  Sliding     scala_persistent          10      5    25,220,372.98  ±     8.49%   ops/s                              1.63×
IteratorBenchmark  Sliding     scala_persistent         100      5    25,036,302.95  ±    13.08%   ops/s                              1.65×
IteratorBenchmark  Sliding     scala_persistent        1000      5    23,748,305.54  ±     3.36%   ops/s                              1.87×
IteratorBenchmark  Sliding     slang_persistent          10      5    15,500,127.68  ±     6.78%   ops/s            0.61×                  
IteratorBenchmark  Sliding     slang_persistent         100      5    15,161,905.19  ±     6.46%   ops/s            0.61×                  
IteratorBenchmark  Sliding     slang_persistent        1000      5    12,718,506.29  ±    16.98%   ops/s            0.54×                  





Performance Ratios
============================================================================================
  (Outliers removed: 30.00% low end, 5.00% high end)

Ratios <alternative_impl> / slang
Target             Operation                              Ratio        10      100     1000 
IteratorBenchmark  Sliding    scala_persistent/slang_persistent     1.63×    1.65×    1.87×


Ratios slang / <alternative_impl>
Target             Operation  Ratio                                    10      100     1000 
IteratorBenchmark  Sliding    slang_persistent/scala_persistent     0.61×    0.61×    0.54×
```

javaslang-benchmark/src/test/java/javaslang/collection/IteratorBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+        scala.collection.Iterator<Integer> scalaPersistent;
+        javaslang.collection.Iterator<Integer> slangPersistent;
+
+        @Setup
+        public void setup() {
+            final Random random = new Random(0);
+            ELEMENTS = getRandomValues(CONTAINER_SIZE, false, random);
+            javaMutable = asList(ELEMENTS);
+            scalaPersistent = (scala.collection.Iterator<Integer>) scala.collection.immutable.Seq$.MODULE$.apply(asScalaBuffer(javaMutable)).iterator();
+            slangPersistent = Iterator.of(ELEMENTS);
```
danieldietrich: @paplorinc
Not sure if it is fair that slang operation on an array while scala operates of a list buffer...
I will force scala's Iterator to operate on an IndexedSeq by using an Array wrapper...

danieldietrich:
After basing the Iterators (slang/scala) on the element array we have the following results. (Note: Scala used a WrappedArray instance for that purpose).

```java
Benchmark                                   (CONTAINER_SIZE)   Mode  Cnt         Score         Error  Units
IteratorBenchmark.Sliding.scala_persistent                10  thrpt    7  21799919.856 ± 1541033.917  ops/s
IteratorBenchmark.Sliding.scala_persistent               100  thrpt    7  24810257.144 ±  876216.887  ops/s
IteratorBenchmark.Sliding.scala_persistent              1000  thrpt    7  23622057.664 ±  950035.353  ops/s
IteratorBenchmark.Sliding.slang_persistent                10  thrpt    7  15451903.983 ± 1101329.863  ops/s
IteratorBenchmark.Sliding.slang_persistent               100  thrpt    7  15032204.539 ±  333253.341  ops/s
IteratorBenchmark.Sliding.slang_persistent              1000  thrpt    7  13460040.034 ±  415575.322  ops/s




Detailed Performance Execution Report
============================================================================================================================================
  (Error: ±99% confidence interval, expressed as % of Score)
  (Outliers removed: 30.00% low end, 5.00% high end)
  (scala_persistent, slang_persistent: read as current row implementation is x times faster than alternative implementation)

Target             Operation   Impl                  Params  Count            Score  ±     Error    Unit  scala_persistent  slang_persistent
IteratorBenchmark  Sliding     scala_persistent          10      5    21,925,497.91  ±    13.54%   ops/s                              1.43×
IteratorBenchmark  Sliding     scala_persistent         100      5    24,958,627.13  ±     5.53%   ops/s                              1.66×
IteratorBenchmark  Sliding     scala_persistent        1000      5    23,770,536.48  ±     4.66%   ops/s                              1.78×
IteratorBenchmark  Sliding     slang_persistent          10      5    15,316,589.82  ±    13.15%   ops/s            0.70×                  
IteratorBenchmark  Sliding     slang_persistent         100      5    14,994,119.35  ±     4.12%   ops/s            0.60×                  
IteratorBenchmark  Sliding     slang_persistent        1000      5    13,373,117.20  ±     2.40%   ops/s            0.56×                  





Performance Ratios
============================================================================================
  (Outliers removed: 30.00% low end, 5.00% high end)

Ratios <alternative_impl> / slang
Target             Operation                              Ratio        10      100     1000 
IteratorBenchmark  Sliding    scala_persistent/slang_persistent     1.43×    1.66×    1.78×


Ratios slang / <alternative_impl>
Target             Operation  Ratio                                    10      100     1000 
IteratorBenchmark  Sliding    slang_persistent/scala_persistent     0.70×    0.60×    0.56×
```

danieldietrich:
@paplorinc Do you think there is something todo or can it be approved/merged?

javaslang-benchmark/src/test/java/javaslang/collection/IteratorBenchmark.java
```diff
@@ -0,0 +1,72 @@
...
+
+        /* Only use these for non-mutating operations */
+        scala.collection.Iterator<Integer> scalaIterator;
+        javaslang.collection.Iterator<Integer> slangIterator;
+
+        @Setup
+        public void setup() {
+            final Random random = new Random(0);
+            ELEMENTS = getRandomValues(CONTAINER_SIZE, false, random);
+            scalaIterator = (scala.collection.Iterator<Integer>) (Object) scala.collection.mutable.WrappedArray$.MODULE$.make(ELEMENTS).iterator();
```
paplorinc: ... yeah ...

javaslang/src/main/java/javaslang/collection/List.java
```diff
@@ -720,6 +720,12 @@
 
     @Override
     default List<T> drop(int n) {
+        if (n <= 0) {
```
paplorinc: :+1:

javaslang-benchmark/src/test/java/javaslang/collection/IteratorBenchmark.java
```diff
@@ -40,33 +39,31 @@ public static void main(String... args) {
         Integer[] ELEMENTS;
 
         /* Only use these for non-mutating operations */
```
paplorinc: isn't this a mutable datastructure?

danieldietrich: oh, forgot to remove the comment

danieldietrich:
@paplorinc thank you, I will further investigate it before merging it.

javaslang-benchmark/src/test/java/javaslang/collection/IteratorBenchmark.java
paplorinc: but if it's mutable, can't we use a simple `ArrayList` in the background, instead of a linked list?

javaslang/src/main/java/javaslang/collection/Iterator.java
```diff
@@ -2080,24 +2081,57 @@ public String toString() {
...
-            if (!hasNext()) {
+            if (buffer.length == 0) {
                 throw new NoSuchElementException();
             }
-            Seq<T> result = buffer;
-            buffer = that.hasNext()
-                     ? buffer.drop(step).appendAll(gap > 0 ? that.drop(gap).take(size) : that.take(step))
-                     : buffer.take(0);
-            return result;
+            final Object[] result = buffer;
```
paplorinc: this seems very low level, is it a lot faster?

danieldietrich:
@paplorinc It was it worth to wait with the merge of this PR. I found one bug (a specific sliding execution that was not covered by tests) and could speed up the GroupedIterator:

```
Target             Operation                              Ratio        10      100     1000 
IteratorBenchmark  Sliding    scala_persistent/slang_persistent     0.10×    0.12×    0.61×
IteratorBenchmark  Sliding    slang_persistent/scala_persistent     9.84×    8.65×    1.65×
```

javaslang/src/main/java/javaslang/collection/Iterator.java
danieldietrich: see below

danieldietrich:
Here is an overview over the other scenarios:

## Object[] buffer

```
Target             Operation                              Ratio        10      100     1000 
IteratorBenchmark  Sliding    scala_persistent/slang_persistent     0.10×    0.12×    0.61×
IteratorBenchmark  Sliding    slang_persistent/scala_persistent     9.84×    8.65×    1.65×
```

## List buffer

```
Target             Operation                              Ratio        10      100     1000 
IteratorBenchmark  Sliding    scala_persistent/slang_persistent     1.43×    1.66×    1.78×
IteratorBenchmark  Sliding    slang_persistent/scala_persistent     0.70×    0.60×    0.56×
```

## Vector buffer

```
Target             Operation                              Ratio        10      100     1000 
IteratorBenchmark  Sliding    scala_persistent/slang_persistent     3.79×    3.70×    4.05×
IteratorBenchmark  Sliding    slang_persistent/scala_persistent     0.26×    0.27×    0.25×
```

## Array buffer

```
Target             Operation                              Ratio        10      100     1000 
IteratorBenchmark  Sliding    scala_persistent/slang_persistent     3.98×    3.53×    4.84×
IteratorBenchmark  Sliding    slang_persistent/scala_persistent     0.25×    0.28×    0.21×
```

paplorinc:
Wow, really nice progress!
For simple iteration I'm not surprised that modifying persistent structures was slow :)
Good job!!! :)

danieldietrich:
Thank you :)

danieldietrich:
I restarted the initial [Strawman's benchmark](https://github.com/javaslang/javaslang/pull/1907#issuecomment-285828972) and got this result (30 seconds run):

```java
// scala
Processed 4400000 groups 42 mb

// javaslang
Processed 5478179 groups 63 mb
```

nfekete:
You guys make a good team 👍 :)

PR #1909, opened 2017-03-10
add Traversable.arrangeBy

emmanueltouzery:
i have a version of this arrangeBy function in my helpers for quite some time, and each time I think to myself that it's probably not generally useful enough to be PR-worthy, but today I used it once too many times.

I find it useful for instance if you have a list of customers and a list of products, and each customer has a productId, and you want to associate the customers with the products...

you could loop over the customers and for each customer loop over the products to find the right one, but that's exponential complexity. Better go through the products once, build a Map<ProductKey,Product> with something like arrangeBy... Then go through the customers and do a fast Map.get over your map.

The function returns Option because you may have several items in the traversable with the same key. If the programmer is sure that's not the case, he can use get().

The implementation is functional-style, I guess it could be optimized... Also, maybe we need to add some wildcard operators like `super` and `extends`... I did not master these yet.

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Traversable.java
+++ b/javaslang/src/main/java/javaslang/collection/Traversable.java
@@ -447,6 +447,21 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
     <C> Map<C, ? extends Traversable<T>> groupBy(Function<? super T, ? extends C> classifier);
 
     /**
+     * Matches each element with a unique key that you extract from it.
+     * If the same key is present twice, the function will return {@code None}.
+     *
+     * @param classifier A function which extracts a key from elements
+     * @param <K>        key class type
+     * @return A Map containing the elements arranged by their keys.
+     * @throws NullPointerException if {@code getKey} is null.
+     */
+    default <K> Option<Map<K, T>> arrangeBy(Function<? super T, K> getKey) {
+        return Option.of(groupBy(getKey).mapValues(Traversable<T>::singleOption))
+            .filter(map -> !map.exists(kv -> kv._2.isEmpty()))
+            .map(map -> map.mapValues(Option::get));
+    }
+
+    /**
      * Groups this {@code Traversable} into fixed size blocks.
      * <p>
      * Let length be the length of this Iterable. Then grouped is defined as follows:
--- a/javaslang/src/main/java/javaslang/collection/Traversable.java
+++ b/javaslang/src/main/java/javaslang/collection/Traversable.java
@@ -443,6 +443,7 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
      * @param <C>        classified class type
      * @return A Map containing the grouped elements
      * @throws NullPointerException if {@code classifier} is null.
+     * @see #arrangeBy(Function)
      */
     <C> Map<C, ? extends Traversable<T>> groupBy(Function<? super T, ? extends C> classifier);
 
@@ -454,11 +455,12 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
      * @param <K>        key class type
      * @return A Map containing the elements arranged by their keys.
      * @throws NullPointerException if {@code getKey} is null.
+     * @see #groupBy(Function)
      */
-    default <K> Option<Map<K, T>> arrangeBy(Function<? super T, K> getKey) {
+    default <K> Option<Map<K, T>> arrangeBy(Function<? super T, ? extends K> getKey) {
         return Option.of(groupBy(getKey).mapValues(Traversable<T>::singleOption))
             .filter(map -> !map.exists(kv -> kv._2.isEmpty()))
-            .map(map -> map.mapValues(Option::get));
+            .map(map -> Map.narrow(map.mapValues(Option::get)));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


javaslang/src/main/java/javaslang/collection/Traversable.java
```diff
@@ -447,6 +447,21 @@ default T single() {
...
+     * @param classifier A function which extracts a key from elements
+     * @param <K>        key class type
+     * @return A Map containing the elements arranged by their keys.
+     * @throws NullPointerException if {@code getKey} is null.
+     */
+    default <K> Option<Map<K, T>> arrangeBy(Function<? super T, K> getKey) {
+        return Option.of(groupBy(getKey).mapValues(Traversable<T>::singleOption))
+            .filter(map -> !map.exists(kv -> kv._2.isEmpty()))
+            .map(map -> map.mapValues(Option::get));
+    }
```
danieldietrich: This is really a lovely method.

## Name

I like the name.

The method is similar to groupBy(). I first thought about naming it groupDistinctBy() but the 'group' part suggests that there may be several elements grouped together. That contradicts the 'distinct' part of the name. 'Arranging' elements is perfect!

## Signature

Because groupBy() and arrangeBy() are so similar I would suggest to 'sync' their signatures, i.e.

```java
<C> Map<C, ? extends Traversable<T>> groupBy(Function<? super T, ? extends C> classifier);

<C> Option<Map<C, T>> arrangeBy(Function<? super T, ? extends C> classifier);
```

In order to make the `? extends C` part work, we need to narrow the Map:

```java
default <C> Option<Map<C, T>> arrangeBy(Function<? super T, ? extends C> classifier) {
    return Option.of(groupBy(classifier).mapValues(Traversable::singleOption))
            .filter(map -> !map.exists(entry -> entry._2.isEmpty()))
            .map(map -> Map.narrow(map.mapValues(Option::get)));
}
```

**Question:** Can something bad happen when treating a `Map<? extends K, V>` as `Map<K, V>`? We know that persistent/immutable objects are allowed to be narrowed (see [this post](http://blog.javaslang.io/why-java-collections-framework-is-a-cats-and-dogs-library)).

But a Map heavily relies on equals/hashCode. If we deal with a type hierarchy, `a.equals(b)` might be true while `b.equals(a)` might be false, for `A a`, `B b`, where `B` extends `A`.

If the classifier type `Function<? super T, ? extends C>` should be `Function<? super T, C>`, we will have to adjust several other methods in Javaslang, starting with `groupBy(Function)`.

## Return type

I thought a minute about returning `Option<? extends Map<C, T>>` because a LinkedHashMap might want to preserve the insertion order of arranged elements. But we can't preserve the sorted property of a TreeMap because the classifier might not be comparable.

I think starting with `Option<Map<C, T>>` is the right way. It behaves like groupBy().

javaslang/src/main/java/javaslang/collection/Traversable.java
```diff
@@ -447,6 +447,21 @@ default T single() {
     <C> Map<C, ? extends Traversable<T>> groupBy(Function<? super T, ? extends C> classifier);
 
     /**
+     * Matches each element with a unique key that you extract from it.
+     * If the same key is present twice, the function will return {@code None}.
+     *
+     * @param classifier A function which extracts a key from elements
+     * @param <K>        key class type
+     * @return A Map containing the elements arranged by their keys.
+     * @throws NullPointerException if {@code getKey} is null.
```
danieldietrich: Could you please add a `@see #groupBy(Function)` to the javadoc of arrangeBy()
and a `@see #arrangeBy(Function)` to the javadoc of groupBy()?

I think it is good to promote the new method because it is not known from other libraries.

javaslang/src/main/java/javaslang/collection/Traversable.java
```diff
@@ -447,6 +447,21 @@ default T single() {
...
+     */
+    default <K> Option<Map<K, T>> arrangeBy(Function<? super T, K> getKey) {
+        return Option.of(groupBy(getKey).mapValues(Traversable<T>::singleOption))
```
danieldietrich: Do you work with Eclipse? IntelliJ says the `<T>` in `Traversable<T>::singleOption` is not needed.

emmanueltouzery:
interesting question on the narrowing and equals/hashcode. I've written a small test program. If the narrowing was a problem, I guess it should output 2/1 (in other words the narrowing made both items in the set become equal). But the output is 2/2 so it seems it's the dynamic type which is taken into account for the dispatch, thankfully:
https://gist.github.com/emmanueltouzery/848b9f8ae163acf87c113f4abdcbd5bf

javaslang/src/main/java/javaslang/collection/Traversable.java
emmanueltouzery: actually I wrote it with the <T> at first, but it doesn't build for me:

    [WARNING] /home/emmanuel/home/javaslang/javaslang/src/main/java/javaslang/collection/Traversable.java:[461,52] found raw type: javaslang.collection.Traversable
  missing type arguments for generic class javaslang.collection.Traversable<T>

and later:

    [ERROR] /home/emmanuel/home/javaslang/javaslang/src/main/java/javaslang/collection/Traversable.java: warnings found and -Werror specified

Maybe they removed the warning in a later version of java, but that sounds unlikely. Using 1.8.0_121.

emmanueltouzery:
ok I think I've applied your suggestions! I've keep "getKey" and "K" as opposed to "classifier" and "C" since I guess we don't "group" in "classes" but "arrange" by "keys".

danieldietrich:
Thank you!

Issue #1853, opened 2017-02-02
HashSet and HashMap have unnecessary Spliterator.ORDERED characteristic


orionll:
Since HashSet/HashMap are unordered collections, the characteristic Spliterator.ORDERED is unnecessary. This can make certain parallel operations on Streams slower. I can post an example if you need.

danieldietrich:
Thank you, you are right. There is already an issue for revisiting the spliterator characteristics: #1635 

We should implement #1635 instead of fixing this issue directly. However, I will keep this issue open as bug.

Issue #1635, opened 2016-10-21
Override Value.splitterator() and Traversable.splitterator()


danieldietrich:
With #1622 we added `Value.toParallelStream()`. To get the most out of it, we need to return a splitterator that reflect the characteristics of the underlying Value.

The are these [Splitterator characteristics](https://docs.oracle.com/javase/8/docs/api/java/util/Spliterator.html#field.summary):
- `CONCURRENT`\- Characteristic value signifying that the element source may be safely concurrently modified (allowing additions, replacements, and/or removals) by multiple threads without external synchronization.
- `DISTINCT` - Characteristic value signifying that, for each pair of encountered elements x, y, !x.equals(y).
- `IMMUTABLE` - Characteristic value signifying that the element source cannot be structurally modified; that is, elements cannot be added, replaced, or removed, so such changes cannot occur during traversal.
- `NONNULL` - Characteristic value signifying that the source guarantees that encountered elements will not be null.
- `ORDERED` - Characteristic value signifying that an encounter order is defined for elements. A Collection has an encounter order if the corresponding Collection.iterator() documents an order.
- `SIZED` - Characteristic value signifying that the value returned from estimateSize() prior to traversal or splitting represents a finite size that, in the absence of structural source modification, represents an exact count of the number of elements that would be encountered by a complete traversal.
- `SORTED` - Characteristic value signifying that encounter order follows a defined sort order.
- `SUBSIZED` - Characteristic value signifying that all Spliterators resulting from trySplit() will be both SIZED and SUBSIZED.

We have two cases:

**1) Value is single-valued (Option, Try, Lazy, Future, ...)**

**2) Value is multi-valued (all Traversables ...)**

---

All types can contain `null` (including Option.Some), so no type has the characteristic:
- `NONNULL`

Because we have persistent types, these characteristics apply to both cases:
- ~~`CONCURRENT`~~ conurrent modifications mean mutable in-place modifications here. Javaslang does not allow this by definition because collections are persistent/immutable.
- `IMMUTABLE`

The single-valued types have these characteristics:
- `DISTINCT`
- `SIZED`
- `SUBSIZED`

The multi-valued types have the following characteristics (in addition to the common as described above):
- Array - `SIZED`, `SUBSIZED`
- BitSet - `DISTINCT`, `ORDERED`, `SIZED`, `SORTED`, `SUBSIZED`
- CharSeq - `ORDERED`, `SIZED`, `SUBSIZED`
- HashMap -
[...]

PR #1911, opened 2017-03-12
issue #1635: spliterator characteristics

v1ctor:
Hi! I've rewritten a spliterator generation for the Value and Traversable classes. We are going to use introspection methods for building. 
#1635 
#1853

### DIFF
--- a/javaslang/src/main/java/javaslang/Value.java
+++ b/javaslang/src/main/java/javaslang/Value.java
@@ -28,6 +28,8 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Spliterator;
+import java.util.Spliterators;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.*;
 import java.util.function.Function;
@@ -1197,6 +1199,12 @@ public interface Value<T> extends Iterable<T> {
      */
     default Vector<T> toVector() {
         return ValueModule.toTraversable(this, Vector.empty(), Vector::of, Vector::ofAll);
+    }
+
+    @Override
+    default Spliterator<T> spliterator() {
+        return Spliterators.spliterator(iterator(), isEmpty() ? 0 : 1,
+                Spliterator.IMMUTABLE | Spliterator.DISTINCT | Spliterator.SIZED | Spliterator.SUBSIZED);
     }
 
     // -- Object
--- a/javaslang/src/main/java/javaslang/Value.java
+++ b/javaslang/src/main/java/javaslang/Value.java
@@ -1204,7 +1204,7 @@ public interface Value<T> extends Iterable<T> {
     @Override
     default Spliterator<T> spliterator() {
         return Spliterators.spliterator(iterator(), isEmpty() ? 0 : 1,
-                Spliterator.IMMUTABLE | Spliterator.DISTINCT | Spliterator.SIZED | Spliterator.SUBSIZED);
+                Spliterator.IMMUTABLE | Spliterator.ORDERED | Spliterator.SIZED | Spliterator.SUBSIZED);
     }
 
     // -- Object
--- a/javaslang/src/main/java/javaslang/collection/Traversable.java
+++ b/javaslang/src/main/java/javaslang/collection/Traversable.java
@@ -1011,9 +1011,9 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
         if (hasDefiniteSize()) {
             characteristics |= (Spliterator.SIZED | Spliterator.SUBSIZED);
             return Spliterators.spliterator(iterator(), length(), characteristics);
+        } else {
+            return Spliterators.spliteratorUnknownSize(iterator(), characteristics);
         }
-        return Spliterators.spliteratorUnknownSize(iterator(), characteristics);
-
     }
 
     /**
--- a/javaslang/src/main/java/javaslang/collection/Traversable.java
+++ b/javaslang/src/main/java/javaslang/collection/Traversable.java
@@ -113,6 +113,7 @@ import java.util.function.Supplier;
  * <li>{@link #hasDefiniteSize()}</li>
  * <li>{@link #isDistinct()}</li>
  * <li>{@link #isOrdered()}</li>
+ * <li>{@link #isSequential()}</li>
  * <li>{@link #isTraversableAgain()}</li>
  * </ul>
  *
@@ -550,6 +551,15 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
      * @return true, if this Traversable is ordered, false otherwise.
      */
     default boolean isOrdered() {
+        return false;
+    }
+
+    /**
+     * Checks if the elements of this Traversable appear in encounter order.
+     *
+     * @return true, if the insertion order of elements is preserved, false otherwise.
+     */
+    default boolean isSequential() {
         return false;
     }
 
@@ -993,7 +1003,10 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
             characteristics |= Spliterator.DISTINCT;
         }
         if (isOrdered()) {
-            characteristics |= (Spliterator.ORDERED | Spliterator.SORTED);
+            characteristics |= (Spliterator.SORTED | Spliterator.ORDERED);
+        }
+        if (isSequential()) {
+            characteristics |= Spliterator.ORDERED;
         }
         if (hasDefiniteSize()) {
             characteristics |= (Spliterator.SIZED | Spliterator.SUBSIZED);
--- a/javaslang/src/main/java/javaslang/collection/Traversable.java
+++ b/javaslang/src/main/java/javaslang/collection/Traversable.java
@@ -10,12 +10,17 @@ import javaslang.Tuple3;
 import javaslang.Value;
 import javaslang.control.Option;
 
-import java.io.Serializable;
 import java.math.BigInteger;
 import java.util.Comparator;
 import java.util.NoSuchElementException;
 import java.util.Objects;
-import java.util.function.*;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
 
 /**
  * An interface for inherently recursive, multi-valued data structures. The order of elements is determined by
@@ -979,6 +984,21 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
     default Option<T> reduceRightOption(BiFunction<? super T, ? super T, ? extends T> op) {
         Objects.requireNonNull(op, "op is null");
         return isEmpty() ? Option.none() : Option.some(reduceRight(op));
+    }
+
+    @Override
+    default Spliterator<T> spliterator() {
+        int characteristics = Spliterator.IMMUTABLE;
+        if (isDistinct()) {
+            characteristics |= Spliterator.DISTINCT;
+        }
+        if (isOrdered()) {
+            characteristics |= (Spliterator.ORDERED | Spliterator.SORTED);
+        }
+        if (hasDefiniteSize()) {
+            characteristics |= (Spliterator.SIZED | Spliterator.SUBSIZED);
+        }
+        return Spliterators.spliterator(iterator(), length(), characteristics);
     }
 
     /**
--- a/javaslang/src/main/java/javaslang/collection/Traversable.java
+++ b/javaslang/src/main/java/javaslang/collection/Traversable.java
@@ -997,8 +997,10 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
         }
         if (hasDefiniteSize()) {
             characteristics |= (Spliterator.SIZED | Spliterator.SUBSIZED);
+            return Spliterators.spliterator(iterator(), length(), characteristics);
         }
-        return Spliterators.spliterator(iterator(), length(), characteristics);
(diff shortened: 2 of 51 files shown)


javaslang/src/main/java/javaslang/Value.java
```diff
@@ -1199,6 +1201,12 @@ default CharSeq toCharSeq() {
         return ValueModule.toTraversable(this, Vector.empty(), Vector::of, Vector::ofAll);
     }
 
+    @Override
+    default Spliterator<T> spliterator() {
+        return Spliterators.spliterator(iterator(), isEmpty() ? 0 : 1,
+                Spliterator.IMMUTABLE | Spliterator.DISTINCT | Spliterator.SIZED | Spliterator.SUBSIZED);
```
danieldietrich: I ask myself if SUBSIZED can be omitted here because there is at most one element in a single-valued type. It cannot be splitted anymore.

But wait... Java 9 added Optional.stream(). I will ask jshell:

```java
jshell> Optional.of(1).stream().spliterator().characteristics() & Spliterator.SUBSIZED
$3 ==> 16384
```

Ok, we need SUBSIZED :)

javaslang/src/main/java/javaslang/collection/Traversable.java
```diff
@@ -981,6 +986,21 @@ default T reduceRight(BiFunction<? super T, ? super T, ? extends T> op) {
...
 
+    @Override
+    default Spliterator<T> spliterator() {
+        int characteristics = Spliterator.IMMUTABLE;
+        if (isDistinct()) {
+            characteristics |= Spliterator.DISTINCT;
+        }
+        if (isOrdered()) {
+            characteristics |= (Spliterator.ORDERED | Spliterator.SORTED);
+        }
```
danieldietrich: I think we need an additional check for sequential collection (= insertion-order):

```java
if (isSequential()) {
    characteristics |= Spliterator.ORDERED;
}
```

---

**Specs:**

```java
interface Traversable<T> {
    /**
     * Checks if the elements of this Traversable appear in encounter order.
     *
     * @return true, if the insertion order of elements is preserved, false otherwise.
     */
    default boolean isSequential() {
        return false;
    }
}
```

Please insert the new method right below Traversable isOrdered().

Please add also a link to this method to the javadoc of the Traversable class:

```java
...
 * Tests:
 *
 * <ul>
 * <li>{@link #existsUnique(Predicate)}</li>
 * <li>{@link #hasDefiniteSize()}</li>
 * <li>{@link #isDistinct()}</li>
 * <li>{@link #isOrdered()}</li>
 * <li>{@link #isSequential()}</li>
 * <li>{@link #isTraversableAgain()}</li>
 * </ul>
...
```

The following collection types need to override `isSequential()` by returning `true`:

* `Seq`
* `LinkedHashSet`
* `LinkedHashMap`
* `LinkedHashMultimap`
* `Tree`
* `Iterator`

javaslang/src/main/java/javaslang/collection/Traversable.java
```diff
@@ -997,8 +997,10 @@ default T reduceRight(BiFunction<? super T, ? super T, ? extends T> op) {
         }
         if (hasDefiniteSize()) {
             characteristics |= (Spliterator.SIZED | Spliterator.SUBSIZED);
+            return Spliterators.spliterator(iterator(), length(), characteristics);
         }
```
danieldietrich: Please add `else` here to make it clear that this `if` should stay the last statement:

```java
if (hasDefiniteSize()) {
    ...
    return ...;
} else {
    return ...;
}
```

javaslang/src/test/java/javaslang/collection/IntMap.java
```diff
@@ -292,6 +292,11 @@ public long estimateSize() {
             public int characteristics() {
                 return spliterator.characteristics();
             }
+
+            @Override
+            public Comparator<? super T> getComparator() {
+                return null;
+            }
```
danieldietrich: The super impl Spliterator.getComparator() throw an IllegalStateException by default.

Is it really necessary to return null? If null is used somewhere it will throw a NPE, which is roughly the same as throwing an IllegalStateException.

I'm just curious - I'm sure there is a reason!

javaslang/src/test/java/javaslang/collection/IntMultimap.java
```diff
@@ -288,6 +288,11 @@ public long estimateSize() {
             public int characteristics() {
                 return spliterator.characteristics();
             }
+
+            @Override
+            public Comparator<? super T> getComparator() {
+                return null;
+            }
```
danieldietrich: dito

javaslang/src/main/java/javaslang/Value.java
v1ctor: That's a really awesome investigation! Thank you :)

javaslang/src/test/java/javaslang/collection/IntMap.java
v1ctor: Yes, there is. In Traversable we are using a spliterator based on a given iterator which is IteratorSpliterator. It has such implementation of ```getComparator()```:
``` 
if (hasCharacteristics(Spliterator.SORTED))
                return null;
            throw new IllegalStateException();
```
For some reason I thought that it is ok to return null. But know I understood that only Linked Maps are sorted that why I have to implement the same logic. And I will :)

danieldietrich: Great, thank you 👍🏼

v1ctor:
Hi @danieldietrich, I've fixed everything you've said :)

javaslang/src/main/java/javaslang/collection/Traversable.java
```diff
@@ -981,6 +996,26 @@ default T reduceRight(BiFunction<? super T, ? super T, ? extends T> op) {
...
+            characteristics |= (Spliterator.SORTED | Spliterator.ORDERED);
+        }
+        if (isSequential()) {
+            characteristics |= Spliterator.ORDERED;
+        }
+        if (hasDefiniteSize()) {
+            characteristics |= (Spliterator.SIZED | Spliterator.SUBSIZED);
+            return Spliterators.spliterator(iterator(), length(), characteristics);
+        }
+        return Spliterators.spliteratorUnknownSize(iterator(), characteristics);
```
danieldietrich: minor: please add the `else` here

javaslang/src/test/java/javaslang/LazyTest.java
```diff
@@ -290,4 +291,26 @@ public void shouldConvertEvaluatedValueToString() {
         lazy.get();
         assertThat(lazy.toString()).isEqualTo("Lazy(1)");
     }
+
+    // -- spliterator
+
+    @Test
+    public void shouldHaveOrderedSpliterator() {
+        assertThat(Lazy.of(() -> 1).spliterator().hasCharacteristics(Spliterator.ORDERED | Spliterator.SORTED));
```
danieldietrich: assertion .isTrue() missing

javaslang/src/test/java/javaslang/LazyTest.java
```diff
@@ -290,4 +291,26 @@ public void shouldConvertEvaluatedValueToString() {
...
+    }
+
+    @Test
+    public void shouldHaveSizedSpliterator() {
+        assertThat(Lazy.of(() -> 1).spliterator().hasCharacteristics(Spliterator.SIZED | Spliterator.SUBSIZED));
```
danieldietrich: dito

javaslang/src/test/java/javaslang/LazyTest.java
```diff
@@ -290,4 +291,26 @@ public void shouldConvertEvaluatedValueToString() {
...
+    }
+
+    @Test
+    public void shouldHaveDistinctSpliterator() {
+        assertThat(Lazy.of(() -> 1).spliterator().hasCharacteristics(Spliterator.DISTINCT));
```
danieldietrich: dito

javaslang/src/test/java/javaslang/LazyTest.java
```diff
@@ -290,4 +291,26 @@ public void shouldConvertEvaluatedValueToString() {
...
+    }
+
+    @Test
+    public void shouldReturnSizeWhenSpliterator() {
+        assertThat(Lazy.of(() -> 1).spliterator().getExactSizeIfKnown()).isEqualTo(1);
```
danieldietrich: dito

javaslang/src/test/java/javaslang/collection/CharSeqTest.java
```diff
@@ -1444,17 +1444,13 @@ public void shouldSplitNonNil() {
 
     @Test
     public void shouldHaveImmutableSpliterator() {
-        assertThat(of('1', '2', '3').spliterator().characteristics() & Spliterator.IMMUTABLE).isNotZero();
-    }
-
-    @Test
-    public void shouldHaveOrderedSpliterator() {
-        assertThat(of('1', '2', '3').spliterator().characteristics() & Spliterator.ORDERED).isNotZero();
+        assertThat(of('1', '2', '3').spliterator().hasCharacteristics(Spliterator.IMMUTABLE));
```
danieldietrich: assertion .isTrue() missing

javaslang/src/test/java/javaslang/collection/CharSeqTest.java
```diff
@@ -1444,17 +1444,13 @@ public void shouldSplitNonNil() {
...
     }
 
     @Test
     public void shouldHaveSizedSpliterator() {
-        assertThat(of('1', '2', '3').spliterator().characteristics() & Spliterator.SIZED).isNotZero();
+        assertThat(of('1', '2', '3').spliterator()
+                .hasCharacteristics(Spliterator.SIZED | Spliterator.SUBSIZED));
```
danieldietrich: dito

danieldietrich:
I've checked it - the build failed because Value creates a Spliterator with characteristics:

```
IMMUTABLE | DISTINCT | SIZED | SUBSIZED
```

And Lazy is tested to have characteristic ORDERED.

I will double-check which characteristics Java 9's Optional has. We can take all of these for our single-valued Value implementations. Please wait a minute...

v1ctor:
@danieldietrich Haven't seen you comment :) 
But if we need to have ORDERED and SIZED characteristics for the Value classes I will add it back.

danieldietrich:
Java 9's Optional has:

```
jshell> Optional.of(1).stream().spliterator().characteristics()
$4 ==> 17488

0100010001010000

16    ORDERED
64    SIZED
1024  IMMUTABLE
16384 SUBSIZED
```

We should do the same and remove DISTINCT from Value. It might be reserved for containers with multiple elements.

danieldietrich:
Mmhh, ORDERED is the same league as DISTINCT - I see no logical reason to have the one but not the other. But we should do it as in Optional.

I will check if the latest Java 9 builds behave the same. Also I will ask one of the JDK committers...

danieldietrich:
I installed the latest JDK 9-ea build 160 and checked it again. Same characteristics. Alright, our Values need to return ORDERED, IMMUTABLE, SIZED, SUBSIZED

danieldietrich:
Thank you! We are lucky that the Lazy test failed :)

v1ctor:
@danieldietrich Yes, we're. But I don't really understand if these characteristics make any difference for the single-valued types? How I understand as we at most only one element, ```trySplit()``` method won't have any effect.

danieldietrich:
Build says failed but I can't see the logs. Maybe a Travis-CI error...

danieldietrich:
@v1ctor 

> But I don't really understand if these characteristics make any difference for the single-valued types?

I also do not understand it. We can merge it as is (can't be that wrong if Java 9 does it the same). But I will try to clarify it by asking someone of the Java committers.

danieldietrich:
Hah, there it is :)

```java
shouldHaveDistinctSpliterator(javaslang.LazyTest)  Time elapsed: 0.047 sec  <<< FAILURE!
org.junit.ComparisonFailure: expected:<[tru]e> but was:<[fals]e>
	at javaslang.LazyTest.shouldHaveDistinctSpliterator(LazyTest.java:304)
```

danieldietrich:
@v1ctor I think that was the final commit. Many thanks for your long breath! I will merge it in after the build finished.

v1ctor:
I also hope that it was the final commit.
Thank you for the support!

P.S. I will take something else ;-)  If you have something in your mind, it would be great!

danieldietrich:
@v1ctor 

> P.S. I will take something else ;-) If you have something in your mind, it would be great!

Great! I think the next logical step for you is to take #1555 (#1635 was a side-story of it).

Please start reading a bit through the log of the issue. There we talk about two methods: `equals` and `eq`. We focus on `equals` and `eq` is out of scope in a first step (we might check later if there is s.th. to do).

We can discuss then how to further process (tomorrow or the next days, no hurry!).

## Summary

This changes the behavior of collection's `equals` method. Traversable should override `equals` and add a javadoc that contains a description for all existing collection types (Seq, Set, Map, Multimap, Tree, ...).

**Collection Equality**

> The collection libraries have a uniform approach to equality and hashing. The idea is, first, to divide collections into sets, maps, and sequences. Collections in different categories are always unequal. For instance, `Set(1, 2, 3)` is unequal to `List(1, 2, 3)` even though they contain the same elements. On the other hand, within the same category, collections are equal if and only if they have the same elements (for sequences: the same elements in the same order). For example, `List(1, 2, 3) == Vector(1, 2, 3)`, and `HashSet(1, 2) == TreeSet(2, 1)`.

_Source: [Scala Documentation](http://docs.scala-lang.org/overviews/collections/equality.html)_

PR #1915, opened 2017-03-14
issue #1555: add equals and hashCode, to Seq, Set, Map, Multimap

v1ctor:
Hi, I'm going to show some intermediate results, maybe you will have some comments and it will help me to tackle problems with tests.

Some explanations:
1. Despite the fact that IntMap and IntMultimap were written to be the replacement of the Map in tests they don't implement ```Traversable<Tuple2<Integer, T>>```. It makes impossible to compare IntMap with real maps. To fix this problem I've overwritten equals and hashCode methods like this: 
```java
@Override
        Object map = o;
        if (o instanceof IntMap) {
            map = ((IntMap) o).original;
        }
        return Collections.equals(this.original, map);
    }

    @Override
    public int hashCode() {
        return Collections.hash(this.original);
    }
```

2. Tests as below are working only because equals and hashCode for IntMap were written using values() and ignoring keys.
```java
@Test
    public void shouldPartitionIntsInOddAndEvenHavingOddAndEvenNumbers() {
        assertThat(of(1, 2, 3, 4).partition(i -> i % 2 != 0)).isEqualTo(Tuple.of(of(1, 3), of(2, 4)));
    }
```
Really we compare here [{0 -> 1, 2 -> 3}, {1 -> 2, 3 -> 4}] and [{0 -> 1, 1 -> 3},{0 -> 2, 1 -> 4}] and it is incorrect.

I'm going to remove such tests from AbstractTraversableTest class and put them in every child class. For AbstractMapTest and AbstractMultimapTest I'm going to rewrite these tests using ```mapOf()``` methods.

3. I decided to write equals method for ```Traversable``` class and using introspection methods ```isSequential``` and ```isDistinct```. In this case, a value of type ```Set<Tuple<K, V>>``` will be equals to a value of type ```Map<K, V>```.

#1555

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Collections.java
+++ b/javaslang/src/main/java/javaslang/collection/Collections.java
@@ -23,11 +23,11 @@ import static javaslang.collection.ArrayType.asArray;
 final class Collections {
 
     @SuppressWarnings("unchecked")
-    static boolean equals(Map<?, ?> source, Object object) {
+    static <K, V> boolean equals(Map<K, V> source, Object object) {
         if (source == object) {
             return true;
         } else if (source != null && object instanceof Map) {
-            Map map = (Map) object;
+            Map<K, V> map = (Map<K, V>) object;
             if (source.size() != map.size()) {
                 return false;
             } else {
@@ -45,11 +45,11 @@ final class Collections {
     }
 
     @SuppressWarnings("unchecked")
-    static boolean equals(Set<?> source, Object object) {
+    static <V> boolean equals(Set<V> source, Object object) {
         if (source == object) {
             return true;
         } else if (source != null && object instanceof Set) {
-            Set set = (Set) object;
+            Set<V> set = (Set<V>) object;
             if (source.size() != set.size()) {
                 return false;
             } else {
@@ -67,11 +67,11 @@ final class Collections {
     }
 
     @SuppressWarnings("unchecked")
-    static boolean equals(Multimap<?, ?> source, Object object) {
+    static <K, V> boolean equals(Multimap<K, V> source, Object object) {
         if (source == object) {
             return true;
         } else if (source != null && object instanceof Multimap) {
-            Multimap multimap = (Multimap) object;
+            Multimap<K, V> multimap = (Multimap<K, V>) object;
             if (source.size() != multimap.size()) {
                 return false;
             } else {
@@ -89,11 +89,11 @@ final class Collections {
     }
 
     @SuppressWarnings("unchecked")
-    static boolean equals(Seq<?> source, Object object) {
+    static <V> boolean equals(Seq<V> source, Object object) {
         if (object == source) {
             return true;
         } else if (source != null && object instanceof Seq) {
-            Seq seq = (Seq) object;
+            Seq<V> seq = (Seq<V>) object;
             return seq.size() == source.size() && areEqual(source, seq);
         }
         return false;
--- a/javaslang/src/main/java/javaslang/collection/Collections.java
+++ b/javaslang/src/main/java/javaslang/collection/Collections.java
@@ -22,7 +22,7 @@ import static javaslang.collection.ArrayType.asArray;
  */
 final class Collections {
 
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({"unchecked", "rawtypes"})
     static boolean equals(Map<?, ?> map1, Object object) {
         if (map1 == null ^ object == null) {
             return false;
@@ -40,7 +40,7 @@ final class Collections {
         return map1.forAll(map::contains);
     }
 
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({"unchecked", "rawtypes"})
     static boolean equals(Set<?> set1, Object object) {
         if (set1 == null ^ object == null) {
             return false;
@@ -58,7 +58,7 @@ final class Collections {
         return set1.forAll(set::contains);
     }
 
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({"unchecked", "rawtypes"})
     static boolean equals(Multimap<?, ?> multimap1, Object object) {
         if (multimap1 == null ^ object == null) {
             return false;
@@ -76,7 +76,7 @@ final class Collections {
         return multimap1.forAll(multimap::contains);
     }
 
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({"unchecked", "rawtypes"})
     static boolean equals(Seq<?> seq1, Object object) {
         if (seq1 == null ^ object == null) {
             return false;
--- a/javaslang/src/main/java/javaslang/collection/Collections.java
+++ b/javaslang/src/main/java/javaslang/collection/Collections.java
@@ -23,28 +23,75 @@ import static javaslang.collection.ArrayType.asArray;
 final class Collections {
 
     @SuppressWarnings("unchecked")
-    static boolean equals(Traversable<?> traversable1, Object object) {
-        if (traversable1 == object) {
+    static boolean equals(Map<?, ?> map1, Object object) {
+        if (map1 == null ^ object == null) {
+            return false;
+        }
+        if (map1 == object) {
             return true;
         }
-        if (traversable1 == null || object == null) {
+        if (!(object instanceof Map)) {
             return false;
         }
-        if (!(object instanceof Traversable)) {
+        Map map = (Map) object;
+        if (map1.size() != map.size()) {
             return false;
         }
-        Traversable traversable = (Traversable) object;
-        if (traversable.length() != traversable1.length()
-                || traversable1.isSequential() != traversable.isSequential()
-                || traversable1.isDistinct() != traversable.isDistinct()) {
+        return map1.forAll(map::contains);
+    }
+
+    @SuppressWarnings("unchecked")
+    static boolean equals(Set<?> set1, Object object) {
+        if (set1 == null ^ object == null) {
             return false;
         }
-        if (traversable.isSequential()) {
-            return areEqual(traversable1, traversable);
-        } else if (traversable.isDistinct()) {
-            return traversable1.forAll(traversable::contains);
+        if (set1 == object) {
+            return true;
         }
-        return false;
+        if (!(object instanceof Set)) {
+            return false;
+        }
+        Set set = (Set) object;
+        if (set1.size() != set.size()) {
+            return false;
+        }
+        return set1.forAll(set::contains);
+    }
+
+    @SuppressWarnings("unchecked")
+    static boolean equals(Multimap<?, ?> multimap1, Object object) {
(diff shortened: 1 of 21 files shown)


javaslang/src/main/java/javaslang/collection/HashMap.java
```diff
@@ -821,20 +821,13 @@ public int size() {
...
-        if (o == this) {
-            return true;
-        } else if (o instanceof HashMap) {
-            final HashMap<?, ?> that = (HashMap<?, ?>) o;
-            return this.trie.equals(that.trie);
-        } else {
-            return false;
-        }
+    public int hashCode() {
+        return Collections.hash(this);
```
ruslansennov: I believe this is wrong way. `HAMT`'s implementations of `equals()` and `hashCode()` are more complex than usual (see `HAMTTest::shouldEqualsIgnoreOrder()` and #1823 #1815). This behavior is not checked in the `HashMap` and this is a bug in the tests

javaslang/src/test/java/javaslang/collection/IntMap.java
```diff
@@ -45,21 +45,17 @@ public boolean isDistinct() {
...
-            final IntMap<?> that = (IntMap<?>) o;
-            return original.equals(that.original) || original.values().equals(that.original.values());
-        } else if (o instanceof Iterable) {
-            final Iterable<?> that = (Iterable<?>) o;
-            return original.values().equals(that);
-        } else {
-            return false;
+            map = ((IntMap) o).original;
         }
+        return Collections.equals(this.original, map);
```
ruslansennov: I'm afraid your solution will be rejected by tests.
Yes, `IntMap` and `IntMultimap` taken from hell. All tests belonging to the `Map` and `Multimap` should be rewritten without using these classes, but this is huge work.

javaslang/src/main/java/javaslang/collection/HashMap.java
v1ctor: Yes, you're right I will fix an implementation of hashCode for the maps.

ruslansennov: Also please add `shouldEqualsIgnoreOrder()` test (which checks `equals` and hashCode` methods) to `AbstractSetTest`

ruslansennov: ... and to `AbstractMapTest`

danieldietrich:
@v1ctor

> To fix this problem I've overwritten equals and hashCode methods like this: ...

Your IntMap.equals()/hashCode() overrides look great!

v1ctor:
@danieldietrich @ruslansennov How I understand ```HashSet.of(1, 2, 3)``` should be equal to ```TreeSet.of(1, 2, 3)```, but what about LinkedHashSet? Do we need to check order or not?
If we check order it means that ```HashSet.of(1, 2, 3)``` won't be equal to ```LinkedHashSet.of(1, 2, 3)``` because first one isn't sequential.

What do you think?

ruslansennov:
Javaslang inspired by Scala...
```scala
var hm = mutable.HashSet(1, 2, 3, 1, 2)
var lhm = mutable.LinkedHashSet(3, 1, 2, 3, 1, 1, 1)
print(hm.equals(lhm))   // true
```

v1ctor:
@danieldietrich @ruslansennov I've found that compareTo for Tuple2 throw npe if right or left is null. For now, I just switched ```shouldZipAllNonEmptyAndNil()``` test for TreeSet, but I believe we need to think how to compare Tuples with nulls correctly.
I've rewritten code, and fixed all the tests, but do not merge it yet, because I would like to write new tests to check different collections equality.

javaslang/src/test/java/javaslang/collection/CollectionsTest.java
```diff
@@ -0,0 +1,68 @@
...
+                TreeSet.ofAll(1, 2, 3)), false);
+    }
+
+    @Test
+    public void shouldNotBeEqualMapAndSet() throws Exception {
+        forAll(List.of(HashSet.of(Tuple.of(1, 2), Tuple.of(2, 3)),
+                TreeMap.of(1, 2, 2, 3)), false);
+    }
+
+    private void forAll(List<Traversable<?>> traversables, boolean value) {
```
ruslansennov: what do you think, does makes sense to create additional private method to check hashCode equality?

javaslang/src/test/java/javaslang/collection/CollectionsTest.java
```diff
@@ -0,0 +1,68 @@
...
+                Vector.ofAll(1, 2, 3),
+                List.ofAll(1, 2, 3),
+                Queue.ofAll(1, 2, 3)), true);
+    }
+
+    @Test
+    public void shouldBeEqualMaps() throws Exception {
+        forAll(List.of(TreeMap.of(1, 2, 2, 3, 3, 4),
+                HashMap.of(1, 2, 2, 3, 3, 4),
+                LinkedHashMap.of(1, 2, 2, 3, 3, 4)), true);
```
ruslansennov: what about shuffling initial entries? :)

javaslang/src/test/java/javaslang/collection/CollectionsTest.java
```diff
@@ -0,0 +1,68 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class CollectionsTest {
+
+    @Test
+    public void shouldBeEqualSets() throws Exception {
+        forAll(List.of(TreeSet.ofAll(1, 2, 3),
+                HashSet.ofAll(1, 2, 3),
+                LinkedHashSet.ofAll(1, 2, 3),
+                BitSet.ofAll(1, 2, 3)), true);
```
ruslansennov: also here: shuffle?

javaslang/src/main/java/javaslang/collection/Collections.java
```diff
@@ -22,6 +22,78 @@
  */
 final class Collections {
 
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    static boolean equals(Map<?, ?> map1, Object object) {
+        if (map1 == null ^ object == null) {
```
v1ctor: As it is static method we could pass for example ```equals(null, TreeSet.of(1, 2, 3))``` that's why we need to check explicitly if none of parameters are null.

danieldietrich: of course - then before size check:

source != null && ... && ...

danieldietrich: no, wait. source is never null by definition. it is `this` / the caller!

v1ctor: Yes, until it is used only in equals, but we shouldn't forget how to use this static method. Maybe annotate source with ```@NotNull```?

danieldietrich: @v1ctor allright, then we will do the check like this:

```java
static <K, V> boolean equals(Map<K, V> source, Object o) {
    if (o == source) {
        return true;
    } else if (source != null && o instanceof Map) {
        final Map<???, ???> that = (Map<???, ???>) o;
        return source.size() == that.size() && ???;
    } else {
        return false;
    }
}
```

I will investigate now how to deal with the sorted collections (as mentioned before).

danieldietrich:
@v1ctor Please take a look at my [proposal](https://github.com/javaslang/javaslang/issues/1555#issuecomment-287611225)

The equals() impl for Map and Multimap should be roughly equivalent to the one for Set.

danieldietrich:
@v1ctor I made more tests. These implementations work fine for me (Seq and Set):

```java
final class Collections {

    static <T> boolean equals(Seq<T> seq1, Object o) {
        if (o == seq1) {
            return true;
        } else if (seq1 != null && o instanceof Seq) {
            final Seq<T> seq2 = (Seq<T>) o;
            return seq1.size() == seq2.size() && areEqual(seq1, seq2);
        } else {
            return false;
        }
    }

    static <T> boolean equals(Set<T> set1, Object o) {
        if (o == set1) {
            return true;
        } else if (set1 != null && o instanceof Set) {
            final Set<T> set2 = (Set<T>) o;
            if (set1.size() != set2.size()) {
                return false;
            } else {
                try {
                    return set1.forAll(set2::contains);
                } catch(ClassCastException x) {
                    return false;
                }
            }
        } else {
            return false;
        }
    }

    ...

    // already existis!
    static boolean areEqual(Iterable<?> iterable1, Iterable<?> iterable2) { ... }
}
```

**Note:** We intentionally reduce the costs by don't putting the size equality check into the try/catch block.

The Map, Multimap equals impls are similar to the one for Set.

---

**Open issue:** We need a new hashing strategy for Set, Map and Multimap. Especially we need to ensure that `equals(a, b) == true => a.hashCode() == b.hashCode()`, e.g. if `a` is a Set/Map and `b` is a SortedSet/LinkedSet/SortedMap/LinkedMap.

Scala uses `scala.util.hashing.MurmurHash3` but I think it is not part of this issue to implement a sophisticated new hashing algorithm. We must just ensure that the equals/hashCode contract is fulfilled.

Currently we do it in HashMap by _summation_ of all elements starting with 0. But I think it is a better idea to start with 1 because

* our empty collections always have a hashCode of 1
* we can distinguish collections that contain empty collections

Please give this one a try for Set, Map, Multimap:

```java
class Collections {
    public int hashUnordered(Iterable<?> iterable) {
        if (iterable == null) {
            return 0;
        } else {
            int hashCode = 1;
            for (Object o : iterable) {
                hashCode += Objects.hashCode(o);
            }
            return hashCode;
        }
    }
}
```

danieldietrich:
@v1ctor thank you for your ongoing effort and your long breath 👍🏼, this is not a trivial issue.

v1ctor:
@danieldietrich You're doing a really great job! Thank you for the help. I am reading through your comments and I will come with new implementation soon. :)

danieldietrich:
@v1ctor take your time!

v1ctor:
Hi @danieldietrich, I've rewritten equals methods and add the check that if collections are equal then they should have the same hashCode. 
There are really a lot of information in this PR, and I feel a little bit lost.

The problem with hashCode that if we will compare SortedSet and non sorted Set if they have the same elements equals will return us ```true```, but hashCode will be different for ordered and unordered Sets. I think that we should leave it like we have right know, because it conforms equals/hashCode contract – hashCode could be the same, but collections could not be equal.

danieldietrich:
Hi @v1ctor 
thank you, you did a great job!
I will take the time this evening to check out you branch and play around with it. On a first sight the changes look good.

danieldietrich:
@v1ctor I already took a look but still need some time (I have too many open tasks).

You will receive a PR for your branch (if that is possible with git!?) the next days. It is necessary because this is a big change and it is easier to get the last few % right.

danieldietrich:
Replaced by a new PR: #1948

Issue #1470, opened 2016-08-02
Extend Predicates with exists/contains method for Collections


lubos-bistak:
# Application Code

```
private Result calculateResult(final Seq<Result> results) {
    return Match(results).of(
      Case(exists(Result::isCritical), CRITICAL),
      Case(exists(Result::isWarning), WARNING),
      Case($(), OK)
    );
}
```
# Javaslang Code

```
public static <T> Predicate<Traversable<T>> exists(final Predicate<T> predicate) {
    Objects.requireNonNull(predicate, "predicate is null");
    return obj -> obj.exists(predicate);
}
```

What do you think?

danieldietrich:
Sounds good! We should do the following:

### Use Value instead of Traversable

```java
<T> Predicate<Value<T>> exists(Predicate<? super T> predicate)
```

### Use `? super T` in the argument

```java
<T> Predicate<Value<T>> exists(Predicate<? super T> predicate)
```

### Also add a `forAll`:

```java
public static <T> Predicate<Value<T>> forAll(Predicate<? super T> predicate) {
    Objects.requireNonNull(predicate, "predicate is null");
    return obj -> obj.forAll(predicate);
}
```

(I've also omitted the `final` keyword for method parameters)

danieldietrich:
Fixed with #1920

PR #1920, opened 2017-03-20
extend predicates with exists forAll (#1470)

mduesterhoeft:


javaslang/src/main/java/javaslang/Predicates.java
```diff
@@ -138,4 +138,28 @@ private Predicates() {
...
+     * A combinator that checks if <strong>all</strong> elements satisfy the {@code predicate}.
+     *
+     * @param predicate The predicate to check
+     * @param <T>        closure over tested object types
+     * @return A new {@code Predicate}
+     */
+    public static <T> Predicate<Value<T>> forAll(Predicate<? super T> predicate) {
+        Objects.requireNonNull(predicate, "predicate is null");
+        return obj -> obj.forAll(predicate);
+    }
```
danieldietrich: I thought it is a good idea to use `Iterable` instead of `Value`. Please change your impl accordingly:

```java
    /**
     * Lifts a given {@code Predicate} to a new {@code Predicate} that tests
     * if an element of an {@code Iterable} satisfies the given {@code Predicate}.
     *
     * @param predicate A {@code Predicate} that tests elements of type {@code T}
     * @param <T> The element type
     * @return a new {@code Predicate} that tests if a given {@code Iterable} contains an element that satisfies the given {@code predicate}
     */
    public static <T> Predicate<Iterable<T>> exists(Predicate<? super T> predicate) {
        return iterable -> Iterator.ofAll(iterable).exists(predicate);
    }

    /**
     * Lifts a given {@code Predicate} to a new {@code Predicate} that tests
     * if all elements of an {@code Iterable} satisfy the given {@code Predicate}.
     *
     * @param predicate A {@code Predicate} that tests elements of type {@code T}
     * @param <T> The element type
     * @return a new {@code Predicate} that tests if a given {@code Iterable} contains an element that satisfies the given {@code predicate}
     */
    public static <T> Predicate<Iterable<T>> forAll(Predicate<? super T> predicate) {
        return iterable -> Iterator.ofAll(iterable).forAll(predicate);
    }
```

I like your introducing javadoc sentence more.

javaslang/src/test/java/javaslang/PredicatesTest.java
```diff
@@ -131,4 +132,27 @@ public void shouldTestIsNotNull_NegativeCase() {
...
+
+    @Test
+    public void shouldTestForAll_PositiveCase() {
+        assertThat(forAll(P1).test(List.of(2, 3))).isTrue();
+    }
+
+    @Test
+    public void shouldTestForAll_NegativeCase() {
+        assertThat(forAll(P1).test(List.of(3, 0))).isFalse();
+    }
```
danieldietrich: Please add the following tests. They ensure the generics are right:

```java
@Test
public void shouldCheckExistsByLiftingPredicateInContravariantPositionToPredicateInCovariantPosition() {
    final List<Integer> list = List(1, 2, 3);
    final Predicate<Number> p = n -> n.intValue() % 2 == 0;
    final boolean actual = Match(list).of(
            Case(exists(p), true),
            Case($(), false)
    );
    assertThat(actual).isTrue();
}

@Test
public void shouldCheckForAllByLiftingPredicateInContravariantPositionToPredicateInCovariantPosition() {
    final List<Integer> list = List(1, 2, 3);
    final Predicate<Number> p = n -> n.intValue() > 0;
    final boolean actual = Match(list).of(
            Case(forAll(p), true),
            Case($(), false)
    );
    assertThat(actual).isTrue();
}
```

mduesterhoeft:
My fault - sorry - I only saw that you assigned the issue after I started. Please feel free to take your solution - I do not want to cause additional work.

danieldietrich:
@mduesterhoeft let's take your solution (+ my review requests). I like to see PRs from other ppl!

danieldietrich:
Many thanks, looks great!

danieldietrich:
Thank you!

PR #1921, opened 2017-03-20
Adds orElseOf methods to Option

happyser:
I've proposing to add new methods. They seems to be useful.
Example to use:

```java
Option<Object> someOption = makeSomeOption();
// than you can stack calls if option is missing

Option<Object> result = someOption.orElseOf(() -> getSomeOtherValue()).orElseOf("can't find a value");

// this code would be equivalent to:
Option<Object> result = someOption;
if (result.isEmpty()) 
{
  result = Option.of(getSomeOtherValue());
}
if (result.isEmpty())
{
  result = Option.of("can't find a value");
}
```

### DIFF
--- a/javaslang/src/main/java/javaslang/control/Option.java
+++ b/javaslang/src/main/java/javaslang/control/Option.java
@@ -255,6 +255,28 @@ public interface Option<T> extends Value<T>, Serializable {
     }
 
     /**
+     * Returns this {@code Option} if it is nonempty, otherwise return option of the value.
+     *
+     * @param value A value
+     * @return this {@code Option} if it is nonempty, otherwise return option of the value.
+     */
+    default Option<T> orElseOf(T value)
+    {
+        return isEmpty() ? Option.of(value) : this;
+    }
+
+    /**
+     * Returns this {@code Option} if it is nonempty, otherwise return option of the result of evaluating supplier.
+     *
+     * @param supplier An alternative supplier
+     * @return this {@code Option} if it is nonempty, otherwise return option of the result of evaluating supplier.
+     */
+    default Option<T> orElseOf(Supplier<? extends T> supplier)
+    {
+        return isEmpty() ? Option.of(supplier.get()) : this;
+    }
+
+    /**
      * Returns the value if this is a {@code Some}, otherwise the {@code other} value is returned,
      * if this is a {@code None}.
      * <p>
--- a/javaslang/src/main/java/javaslang/control/Option.java
+++ b/javaslang/src/main/java/javaslang/control/Option.java
@@ -255,24 +255,22 @@ public interface Option<T> extends Value<T>, Serializable {
     }
 
     /**
-     * Returns this {@code Option} if it is nonempty, otherwise return option of the value.
+     * Returns this {@code Option} if it is nonempty, otherwise return {@code Option} of the value.
      *
      * @param value A value
-     * @return this {@code Option} if it is nonempty, otherwise return option of the value.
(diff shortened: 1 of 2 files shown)


javaslang/src/main/java/javaslang/control/Option.java
```diff
@@ -255,6 +255,28 @@ default T getOrElse(T other) {
     }
 
     /**
+     * Returns this {@code Option} if it is nonempty, otherwise return option of the value.
+     *
+     * @param value A value
+     * @return this {@code Option} if it is nonempty, otherwise return option of the value.
+     */
+    default Option<T> orElseOf(T value)
+    {
```
paplorinc: Hi! :)
Thanks, please format the source code according to our standards.

javaslang/src/main/java/javaslang/control/Option.java
```diff
@@ -255,6 +255,28 @@ default T getOrElse(T other) {
...
+        return isEmpty() ? Option.of(value) : this;
+    }
+
+    /**
+     * Returns this {@code Option} if it is nonempty, otherwise return option of the result of evaluating supplier.
+     *
+     * @param supplier An alternative supplier
+     * @return this {@code Option} if it is nonempty, otherwise return option of the result of evaluating supplier.
```
paplorinc: ... return `Option` of ...

javaslang/src/main/java/javaslang/control/Option.java
happyser: sorry my bad

javaslang/src/test/java/javaslang/control/OptionTest.java
```diff
@@ -190,6 +190,30 @@ public void shouldReturnAlternativeOnOrElseSupplierIfValueIsNotDefined() {
         assertThat(Option.none().orElse(() -> opt)).isSameAs(opt);
     }
 
+    @Test
+    public void shouldReturnSelfOnOrElseOfIfValueIsPresent() {
```
paplorinc: Could you please provide a more realistic use case?
A convincing real-world example, where this new functionality is useful (i.e. no dummy values).
(I don't decide anything, I'm just curious. @danieldietrich is the boss :smiley:)

paplorinc: From what I understood, this is meant to use a provider without an `Option` wrapping?
(using `import static javaslang.API.*;`)
```java
Option(value).orElseOf(0);
```
is the same as
```
Option(value).orElse(Some(0));
```
, right?

@danieldietrich, do you think this simplifies things?

danieldietrich:
@happyser Sorry for answering so late - I needed every minute to prepare the next bugfix release 2.0.6 (which will be released neartime).

I've thought about it and played around with the new syntax.

As @paplorinc mentioned above, this suggestion can be seen as an alternate syntax for the existing `orElse(T)` resp. `orElse(Supplier<T>)` methods.

Long story short: I see no real benefit. The initial example can be expressed like this:

```java
makeSomeOption().orElse(() -> Option(getSomeOtherValue()))
                .orElse(() -> Some("can't find a value"))
```

Compared to the new approach:

```java
makeSomeOption().orElseOf(() -> getSomeOtherValue())
                .orElseOf(() -> "can't find a value")
```

The only difference is, that we currently create the Option instance explicitly. However, both approaches (yours and the existing one) are making use of Suppliers, so instances are only created if necessary.

The new approach broadens the API by adding a different name to existing functionality.

Please understand that I close the PR.

danieldietrich:
Thank you anyway!

PR #1923, opened 2017-03-23
Added Try-with-resources

danieldietrich:
This one adds try-with-resources functionality to our Try.

The following example emerged from [this](http://stackoverflow.com/questions/42979088/how-to-map-a-try-with-resources-exception-type) StackOverflow question.

```java
public class Example {

    static ThreadLocal<Socket> socket = ThreadLocal.withInitial(Socket::new);

    @SuppressWarnings("unchecked")
    public static void main(String[] args) {

        final Try<String> of = Try.withResources(() -> new BufferedReader(new InputStreamReader(socket.get().getInputStream())))
                .of(BufferedReader::readLine)
                .mapFailure(
                        Case(instanceOf(IOException.class), ConnectionError::new)
                );
    }

    static class ConnectionError extends Error {
        ConnectionError(IOException t) {
        }
    }

}
```

**Note:** The mapFailure() function will be added in the next PR.

Issue #1825, opened 2017-01-16
[type hierarchy] Need for semi-sealed interfaces


danieldietrich:
### I'm not sure if this is worth the effort or if it makes completely sense. It is subject to think about before starting to implement it...

I think it might be better to add a clear specification to the API docs than introducing new types, blowing up the hierarchy and so on. Simplicity wins.

With #1801, we should instead update the API for the case the underlying collection is empty. Here for example: `empty.prependAll(list)` will return list if list is instanceof List.

---

Some of our collections are final classes, like Queue:

```java
public final class Queue<T> {
    Queue<T> prependAll(Iterable<T> elements) {
        if (isEmpty() && elements instanceof Queue) {
            return (Queue<T>) elements; // <-- OK
        } else {
            ...
        }
    }
    ...
}
```

But other collections are interfaces, like List:

```java
public interface List<T> {
    List<T> prependAll(Iterable<T> elements) {
        if (isEmpty() && elements instanceof List) {
            return (List<T>) elements; // <-- PROBLEMATIC
        } else {
            ...
        }
    }
    ...


    final class Nil<T> extends List<T> {
        ...
    }

    final class Cons<T> extends List<T> {
        ...
    }
}
```

The PROBLEMATIC line may be a security issue, if someone else implemented List and passed an instance to prependAll().

We are able to ensure that we return only own implementations by introducing a hidden, tagging type:

```java
public interface List<T> {
    List<T> prependAll(Iterable<T> elements) {
        if (isEmpty() && elements instanceof SealedList) {
            return (List<T>) elements; // <-- OK (by convention)
        } else {
            ...
        }
    }
    ...

    final class Nil<T> extends SealedList<T> {
        ...
    }

    final class Cons<T> extends SealedList<T> {
        ...
    }
}

// tagging interface, not visible to the outside
interface SealedList<T> extends List<T> {
}
```

**Note:** We can't prevent users from implementing List. But we can prevent to return a user-implemented type when operating on user implementations. Therefore I call SealedList a 'semi-sealed' interface.

Kreinoee:
Why not make List an abstract class instead of an interface. By giving the abstract List class an private constructor, we can ensure that only inner classes can extend it, and thereby guaranty that all instances of the List class is immutable.

[...]

PR #1924, opened 2017-03-23
Try map failure

danieldietrich:
A follow-up of #1923

### DIFF
--- a/javaslang/src/main/java/javaslang/collection/Traversable.java
+++ b/javaslang/src/main/java/javaslang/collection/Traversable.java
@@ -457,8 +457,8 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
      * Matches each element with a unique key that you extract from it.
      * If the same key is present twice, the function will return {@code None}.
      *
-     * @param classifier A function which extracts a key from elements
-     * @param <K>        key class type
+     * @param getKey A function which extracts a key from elements
+     * @param <K>    key class type
      * @return A Map containing the elements arranged by their keys.
      * @throws NullPointerException if {@code getKey} is null.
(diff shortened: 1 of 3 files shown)


javaslang/src/test/java/javaslang/control/TryTest.java
```diff
@@ -247,6 +253,268 @@ public void shouldThrowNullPointerExceptionWhenCallingTryRunRunnable() {
...
+            this.value = value;
+        }
+
+        @Override
+        public void close() throws Exception {
+            isClosed = true;
+        }
+    }
+
+    private static <T> Resource<T> resource(T value) {
```
paplorinc: It's just my preference: please consider renaming to `Closable` or `TestAutoClosable` or `DummyClosable` etc and provide a static `of` factory, as done in other places.

javaslang/src/main/java/javaslang/collection/Traversable.java
```diff
@@ -457,8 +457,8 @@ default T single() {
      * Matches each element with a unique key that you extract from it.
      * If the same key is present twice, the function will return {@code None}.
      *
-     * @param classifier A function which extracts a key from elements
-     * @param <K>        key class type
+     * @param getKey A function which extracts a key from elements
```
paplorinc: :+1:

danieldietrich:
**Note:** Calls to `mapFailure` will be unchecked in 2.1.0 because

* Try in an interface
* => we can't make `mapFailure()` final
* => we can't mark `mapFailure` as `@SafeVarargs`

Solution: make Try an abstract class (see #1825)

PR #1926, opened 2017-03-25
Restored method alphabetical method order

danieldietrich:


danieldietrich:
(basically nothing to review here - just copy-pasted existing methods in different order)

Issue #1754, opened 2016-12-11
Add more introspection properties


danieldietrich:
The **bold** printed properties are new:

_Value_ has:

* [x] **isAsync() &mdash; is computed asynchronously**
* [x] **isLazy() &mdash; is evaluated lazily (as opposed to eagerly)**
* [x] isSingleValued() &mdash; can contain at most one value

_Traversable_ additionally has:

* [x] hasDefiniteSize() &mdash; has at most a finite number of elements
* [x] isDistinct() &mdash; cannot contain duplicate elements (using equals)
* [x] isOrdered() &mdash; has an underlying Comparator (by implementing Ordered)
* [x] **isSequential() &mdash; elements occur in a predictable, fixed sequence**
* [x] isTraversableOnce() &mdash; can be iterated once (read: is stateful)

---

Note:** It is odd that these properties are observed on an instance of that type, for example:

```java
// = false, but a user may think that is has size 1
Stream(1).hasDefiniteSize();

// = false, but a user may think that the list contains distinct elements
List(1, 2).isDistinct();

// = false, but a user may think that the elements are ordered in the list
List(1, 2).isOrdered();

// = false, but a user may think that there is only a single element
List(1).isSingleValued();
```

We need to keep in mind that these properties are intended to be observed on instances of super types, for example:

```java
if (traversable.hasDefiniteSize()) {
    println("the size is: " + traversable.size());
}

if (traversable.isOrdered()) {
    Comparator<T> comparator = ((Ordered<T>) traversable).comparator();
    // ...
}
```

PR #1952, opened 2017-04-21
Adds java.util.List view to our sequential collections

danieldietrich:
Addresses #1728
Replaces the PR #1752

TODO:

* [x] Separate PRs for Array fixes and isLazy() introspection method (see #1754)
* [x] all tests resp. modifying operations for **Iterator** and **ListIterator** need to check for concurrent modification (List, Iterator, ListIterator)
* [x] re-enable sputnik (see .travis.yml)
* [x] Write AbstractSeqTest tests for `asJava(Consumer<? super java.util.List<T>> action)` and `asJavaImmutable(Consumer<? super java.util.List<T>> action)`
* [x] Currently remove(Object) is not correctly handled for lazy collections (like Stream). It returns true for non-existing elements.

Scala:

```scala
scala> import scala.collection.JavaConverters._
import scala.collection.JavaConverters._

scala> val stream = Stream(1,2,3)
stream: scala.collection.immutable.Stream[Int] = Stream(1, ?)

scala> val list = stream.asJava
list: java.util.List[Int] = [1, 2, 3]

scala> list.remove("A")
res1: Boolean = false
```

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Array.java
+++ b/vavr/src/main/java/io/vavr/collection/Array.java
@@ -15,6 +15,8 @@ import java.util.*;
 import java.util.function.*;
 import java.util.stream.Collector;
 
+import static io.vavr.collection.JavaConverters.ChangePolicy.IMMUTABLE;
+import static io.vavr.collection.JavaConverters.ChangePolicy.MUTABLE;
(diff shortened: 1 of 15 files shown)


danieldietrich:
Ready, steady, ... merge!

Issue #1757, opened 2016-12-12
Fix backward compatibilty issue regarding *Map.of(Object...) in 2.1.0-alpha


danieldietrich:
@talios [mentioned](https://gitter.im/javaslang/javaslang?at=584e0d4fc29531ac5d4f7dba) that the *Map.of(Object...) factory methods disappeared in 2.1.0-alpha (in favor of(KVKVKV)) because the former was unsafe.

For the reason of backward compatibility we need to re-add them, marked as `@Deprecated`.

**Hint:** Wait for #1716, which will generate these automatically throughout all Map types!

ruslansennov:
Daniel, I'm ready to begin, but I'm not sure that we should do this because now we have new groupId, artifactId and artifatcVersion :)

danieldietrich:
@ruslansennov I plan to do many changes in the upcoming release vavr-0.9.0 that affect backward compatibility.

The reason is that because of the new package names the upcoming release is no drop-in replacement anymore.

If someone wants to migrate his/her code to the new version, he needs control over the code. If he/she has control over the code, also the breaking changes can be easily fixed, presupposed a migration path exists.

In fact I see the name change as a lucky coincidence. Finally we are able to fix many issues. The backward compatibility slowed us down immensely.

danieldietrich:
@ruslansennov

> Daniel, I'm ready to begin

Go ahead!

danieldietrich:
Fixed with #1954

PR #1954, opened 2017-04-21
re-add *Map.of(Object...)

ruslansennov:
see #1757

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/HashMap.java
+++ b/vavr/src/main/java/io/vavr/collection/HashMap.java
@@ -139,6 +139,30 @@ public final class HashMap<K, V> implements Map<K, V>, Serializable {
     }
 
     /**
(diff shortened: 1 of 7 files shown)


danieldietrich:
Thank you, it looks good to me!

It may be that I remove all deprecated methods (including these) before the final release of vavr-0.9.0. I will go through the remaining open issues again to see which of them are outdated.

vavr-0.9.0 will not be backward compatible. We will still maintain the javaslang-2.0.x line. But vavr is a new beginning. We will not release a final version 2.1.0. Instead we prepare a 0.9.0 that will be released neartime (within the next two weeks). spring-data does need it before the release on May 25th. I intentionally named the version 0.x because of the time-pressure. We still have the option to release a 1.x afterwards that may contain backward-incompatible changes. But if we are able to make these changes in 0.9 we don't need to do a 1.0. Instead we can target a 0.10. We will see...

PR #1957, opened 2017-04-23
Fix javadoc.

mg6maciej:
I understand that many developers like immutable lists,
but I think this change will be a great improvement
to what is really returned there.


danieldietrich:
Hah, good catch ;)

PR #1960, opened 2017-04-23
Fixed range-check of Seq.subSequence

danieldietrich:
When start &gt; end of a range (start, end) and both start, end are within their bounds, then the result should be an IllegalArgumentException instead of an IndexOutOfBoundsException.

Java's ArrayList.subList behaves the same for example.

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Vector.java
+++ b/vavr/src/main/java/io/vavr/collection/Vector.java
@@ -1078,11 +1078,8 @@ public final class Vector<T> implements IndexedSeq<T>, Serializable {
 
     @Override
     public Vector<T> subSequence(int beginIndex, int endIndex) {
-        if ((beginIndex >= 0) && (beginIndex <= endIndex) && (endIndex <= length())) {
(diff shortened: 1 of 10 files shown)


vavr/src/main/java/io/vavr/collection/Vector.java
```diff
@@ -1078,11 +1078,8 @@ public int lastIndexOf(T element, int end) {
 
     @Override
     public Vector<T> subSequence(int beginIndex, int endIndex) {
-        if ((beginIndex >= 0) && (beginIndex <= endIndex) && (endIndex <= length())) {
-            return slice(beginIndex, endIndex);
-        } else {
-            throw new IndexOutOfBoundsException("subSequence(" + beginIndex + ", " + endIndex + ") on Vector of size " + length());
-        }
+        Collections.subSequenceRangeCheck(beginIndex, endIndex, length());
+        return slice(beginIndex, endIndex);
```
danieldietrich: I know that we do not pull the last bit of performance out of the JVM but it is more robust and removes redundant code.

Issue #1510, opened 2016-08-19
Change Map.traverse(BiFunction) to Map.iterator(BiFunction)


danieldietrich:
Currently we have:

``` java
<U> Seq<U> traverse(BiFunction<K,V,? extends U> mapper)
```

Instead, we use a Iterators for traversal:

``` java
<U> Iterator<U> iterator(BiFunction<K,V,? extends U> mapper)
```

Note: We already have `transform(Function<Tuple2<K, V>, U>)` to transform a map to a specific type.

PR #1966, opened 2017-04-27
Map.traverse -> Map.iterator 

ruslansennov:
see #1510


danieldietrich:
Looks great, thank you!

Issue #1838, opened 2017-01-25
LinkedHashMap.put should replace existing entry and maintain order


thekaratekid05:
I noticed a difference between the java and javaslang 'put' implementations on LinkedHashMap

My assumption is that a 'put' should replace the existing entry and also maintain the original insertion order
```java
@Test
  public void javaPutWithExistingKeyShouldReplaceExistingEntryAndMaintainOrder() {
    java.util.LinkedHashMap<Integer, String> javaMap = new java.util.LinkedHashMap<>();
    javaMap.put(1, "won");
    javaMap.put(2, "two");
    javaMap.put(1, "one");

    assertThat(javaMap).containsExactly(entry(1, "one"), entry(2, "two"));

  }

  @Test
  public void javaslangPutWithExistingKeyShouldReplaceExistingEntryAndMaintainOrder() {
    javaslang.collection.LinkedHashMap<Integer, String> javaslangMap =
          javaslang.collection.LinkedHashMap.<Integer, String>empty()
                .put(1, "won")
                .put(2, "two")
                .put(1, "one");

    // FAILS ASSERTION:
    assertThat(javaslangMap).containsExactly(Tuple.of(1, "one"), Tuple.of(2, "two"));
  }
```

danieldietrich:
I remember that topic popped up a while ago. Before we change it I want to ensure that other immutable implementations act the same. When looking at Java and Scala, there are currently only immutable versions of LinkedHashMap.

thekaratekid05:
alas it seems Scala behaves the same
```scala
scala> import collection.immutable.ListMap
import collection.immutable.ListMap

scala> val map = ListMap(1 -> 1, 2 -> 2, 3 -> 3)
map: scala.collection.immutable.ListMap[Int,Int] = ListMap(1 -> 1, 2 -> 2, 3 -> 3)

scala> val updated = map + (2 -> 20)
updated: scala.collection.immutable.ListMap[Int,Int] = ListMap(1 -> 1, 3 -> 3, 2 -> 20)
```

danieldietrich:
We align to Scala. Changing the behavior is currently out of scope.

yuriykulikov:
Hello everyone.
This is very unfortunate. Java LinkedHashMap retains insertion order in values() iterator.
Is there still any possibility the behavior of Vavr LinkedHashMap will be chagned?

ruslansennov:
@danieldietrich , I agree with @yuriykulikov 
[...]

PR #1983, opened 2017-05-08
LinkedHashMap.put(existing key) should keep order #1838

ruslansennov:
Also equality of `LinkedHashMap` with `LinkedHashMap` fixed (see #1555)

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/LinkedHashMap.java
+++ b/vavr/src/main/java/io/vavr/collection/LinkedHashMap.java
@@ -701,14 +701,14 @@ public final class LinkedHashMap<K, V> implements Map<K, V>, Serializable {
 
     @Override
     public LinkedHashMap<K, V> put(K key, V value) {
-        Queue<Tuple2<K, V>> newList = list;
-        HashMap<K, V> newMap = map;
-        if (containsKey(key)) {
-            newList = newList.filter(t -> !Objects.equals(t._1, key));
-            newMap = newMap.remove(key);
+        final Queue<Tuple2<K, V>> newList;
+        final Option<V> currentEntry = get(key);
+        if (currentEntry.isDefined()) {
+            newList = list.replace(Tuple.of(key, currentEntry.get()), Tuple.of(key, value));
+        } else {
+            newList = list.append(Tuple.of(key, value));
         }
-        newList = newList.append(Tuple.of(key, value));
-        newMap = newMap.put(key, value);
+        final HashMap<K, V> newMap = map.put(key, value);
         return new LinkedHashMap<>(newList, newMap);
     }
 
@@ -901,7 +901,11 @@ public final class LinkedHashMap<K, V> implements Map<K, V>, Serializable {
 
     @Override
     public boolean equals(Object o) {
-        return Collections.equals(this, o);
+        if (o instanceof LinkedHashMap) {
+            return Collections.equals(this.list, ((LinkedHashMap<?, ?>) o).list);
+        } else {
+            return Collections.equals(this, o);
+        }
     }
 
     @Override
--- a/vavr/src/test/java/io/vavr/collection/AbstractMapTest.java
+++ b/vavr/src/test/java/io/vavr/collection/AbstractMapTest.java
@@ -39,11 +39,15 @@ public abstract class AbstractMapTest extends AbstractTraversableTest {
             public IterableAssert<T> isEqualTo(Object obj) {
                 @SuppressWarnings("unchecked")
                 final Iterable<T> expected = (Iterable<T>) obj;
-                final java.util.Map<T, Integer> actualMap = countMap(actual);
-                final java.util.Map<T, Integer> expectedMap = countMap(expected);
-                assertThat(actualMap.size()).isEqualTo(expectedMap.size());
-                actualMap.forEach((k, v) -> assertThat(v).isEqualTo(expectedMap.get(k)));
-                return this;
+                if (actual instanceof LinkedHashMap && obj instanceof LinkedHashMap) {
+                    return super.isEqualTo(expected);
+                } else {
+                    final java.util.Map<T, Integer> actualMap = countMap(actual);
+                    final java.util.Map<T, Integer> expectedMap = countMap(expected);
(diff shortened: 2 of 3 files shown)


vavr/src/main/java/io/vavr/collection/LinkedHashMap.java
```diff
@@ -701,14 +701,14 @@ public boolean isSequential() {
...
     @Override
     public LinkedHashMap<K, V> put(K key, V value) {
-        Queue<Tuple2<K, V>> newList = list;
-        HashMap<K, V> newMap = map;
-        if (containsKey(key)) {
-            newList = newList.filter(t -> !Objects.equals(t._1, key));
-            newMap = newMap.remove(key);
+        final Queue<Tuple2<K, V>> newList;
+        final Option<V> currentEntry = get(key);
+        if (currentEntry.isDefined()) {
```
ruslansennov: maybe we should check that `oldValue.equals(newValue)` is `true` and in this case do nothing (?)

vavr/src/test/java/io/vavr/collection/LinkedHashMapTest.java
```diff
@@ -181,18 +181,23 @@ public void shouldReturnModifiedKeysMapWithNonUniqueMapperAndPredictableOrder()
     // -- put
 
     @Test
-    public void shouldAppendToTheEndWhenPuttingAnExistingKeyAndNonExistingValue() {
+    public void shouldIgnoreOrderOfEntriesWhenComparingForEquality() {
```
danieldietrich: please add an `@Override` here

vavr/src/test/java/io/vavr/collection/AbstractMapTest.java
```diff
@@ -39,11 +39,15 @@
             public IterableAssert<T> isEqualTo(Object obj) {
                 @SuppressWarnings("unchecked")
                 final Iterable<T> expected = (Iterable<T>) obj;
-                final java.util.Map<T, Integer> actualMap = countMap(actual);
-                final java.util.Map<T, Integer> expectedMap = countMap(expected);
-                assertThat(actualMap.size()).isEqualTo(expectedMap.size());
-                actualMap.forEach((k, v) -> assertThat(v).isEqualTo(expectedMap.get(k)));
-                return this;
+                if (actual instanceof LinkedHashMap && obj instanceof LinkedHashMap) {
```
danieldietrich: Please override the `assertThat(Iterable<T> actual)` method in `LinkedHashMapTest` test instead. A `super` call does inherently rely on the type hierarchy, which is not necessary and error prone.

Instead `LinkedHashMapTest` should defined what equality means (even if it looks like duplicate code at a first glance).

```
@Override
protected <T> IterableAssert<T> assertThat(Iterable<T> actual) {
    ...
}

```

vavr/src/main/java/io/vavr/collection/LinkedHashMap.java
danieldietrich: This is not necessary, we will act like `java.util.HashMap`. It will be important when implementing the `MapView` later.

```java
public class Test {

        A x = new A(1, 1);
        A y = new A(1, 2);

        // true
        System.out.println(x.equals(y));

        java.util.Map<Integer, A> map = new java.util.HashMap<>();

        map.put(1, x);
        map.put(1, y);

        // true
        System.out.println(map.get(1) == y);

    }

    static class A {

        final int a;
        final int b;

        A(int a, int b) {
            this.a = a;
            this.b = b;
        }

        @Override
            if (o == this) {
                return true;
            } else if (o instanceof A) {
                final A that = (A) o;
                return this.a == that.a;
            } else {
                return false;
            }
        }

        @Override
        public int hashCode() {
            return a;
        }

        @Override
        public String toString() {
            return "A(a:" + a + ", b:" + b + ")";
        }

    }
}
```

vavr/src/main/java/io/vavr/collection/LinkedHashMap.java
```diff
@@ -901,7 +901,11 @@ public int size() {
 
     @Override
     public boolean equals(Object o) {
-        return Collections.equals(this, o);
+        if (o instanceof LinkedHashMap) {
+            return Collections.equals(this.list, ((LinkedHashMap<?, ?>) o).list);
+        } else {
+            return Collections.equals(this, o);
+        }
```
danieldietrich: We changed the equals behavior.

This is not correct according to the `Traversable.equals()` javadoc. We act like Scala, which is:

```java
scala> import scala.collection.immutable._
import scala.collection.immutable._

scala> val map1 = ListMap((1, 'a'), (2, 'b'))
map1: scala.collection.immutable.ListMap[Int,Char] = ListMap(1 -> a, 2 -> b)

scala> val map2 = ListMap((2, 'b'), (1, 'a'))
map2: scala.collection.immutable.ListMap[Int,Char] = ListMap(2 -> b, 1 -> a)

scala> map1 == map2
res0: Boolean = true
```

_Note: The immutable LinkedHashMap is called ListMap in Scala_

danieldietrich:
Note: We still are able to use `eq()` for structural equality (which takes the order into account)

ruslansennov:
This PR (and issue) is very problemastic. It's OK for me to keep all things as is. I'm offline now, please close both.

danieldietrich:
I will keep it open for a while. Maybe there is some discussion...

> This PR (and issue) is very problemastic.

Does it break s.th. for you?

danieldietrich:
@ruslansennov Just one more word on this: this PR would break the Liskov Substitution Principle.

This should be equal for Maps that contain `(1, "a")`, `(2, "b")` - regardless of the order. Especially for `LinkedHashMap(1, "a", 2, "b")` and `LinkedHashMap(2, "b", 1, "a")`.

```java
final Map<Integer, String> map1 = ...
final Map<Integer, String> map2 = ...
assert map1.equals(map2);
```

danieldietrich:
@ruslansennov oh noes 💩 of course the LinkedHashMap.put(k,v) should preserve existing order (on existing key). I focused on the equals stuff in my review.

Could you please make a PR for the put() fix, only? We will take it into the 0.9.0 release. Nearly all relevant 0.9.0 issues are finished now!

PR #1988, opened 2017-05-10
LinkedHashMap.put(existing key) should keep order #1838

ruslansennov:


yuriykulikov:
You guys are awesome! Thanks!

PR #1992, opened 2017-05-12
fix gwt again

ruslansennov:
See vavr-io/vavr-gwt-example#3

Actually all errors finds each time in the first stage of compilation.
```
[INFO]    Ignored 10 units with compilation errors in first pass.
[INFO] Compile with -strict or with -logLevel set to TRACE or DEBUG to see all errors.
```
Two of these "`Ignored 10 units`" are `Predicates` and `JavaConverters`. But because of very bad tests coverage, at the second stage of compilation GWT includes in dead code all problem methods. And all tests passes.

### DIFF
--- a/vavr/src/main/java/io/vavr/Predicates.java
+++ b/vavr/src/main/java/io/vavr/Predicates.java
@@ -33,6 +33,7 @@ public final class Predicates {
      * @throws NullPointerException if {@code type} is null
      */
     // DEV-NOTE: We need Class<? extends T> instead of Class<T>, see {@link TryTest#shouldRecoverSuccessUsingCase()}
+    @GwtIncompatible
     public static <T> Predicate<T> instanceOf(Class<? extends T> type) {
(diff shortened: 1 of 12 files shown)


danieldietrich:
Thank you!

tbroyer:
Wrt that `-strict` warning and the fact that tests don't catch those errors: the tests in `vavr-gwt` are run with the `org.codehaus.mojo:gwt-maven-plugin`, which defaults to [running tests in devmode](https://gwt-maven-plugin.github.io/gwt-maven-plugin/test-mojo.html#productionMode) (despite GWT having switched to using prod mode by default); and GWT doesn't allow running tests in strict mode (`-strict` or `-failOnError`) anyway, so you need to catch those errors outside of tests (e.g. compiling a sample app, or possibly running the GWT compiler with the `-validateOnly` flag) where `-failOnError` can be used.

ruslansennov:
@tbroyer thank you

PR #1996, opened 2017-05-18
improve Array.permutations() performance #1995

ruslansennov:
@jest I believe that the Narayana algorithm is simple and fast enough

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Array.java
+++ b/vavr/src/main/java/io/vavr/collection/Array.java
@@ -984,19 +984,61 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
 
     @Override
     public Array<Array<T>> permutations() {
-        if (isEmpty()) {
-            return empty();
-        } else if (delegate.length == 1) {
-            return of(this);
+        final Array<T> src = distinct();
+        if (src.isEmpty()) {
+            return Array.empty();
+        } else if (src.size() == 1) {
+            return Array.of(src);
         } else {
-            Array<Array<T>> results = empty();
-            for (T t : distinct()) {
-                for (Array<T> ts : remove(t).permutations()) {
-                    results = results.append(of(t).appendAll(ts));
+            java.util.List<Array<T>> result = new java.util.ArrayList<>();
+            int[] pi = new int[src.size()];
+
+            // see Narayana algo
+            // https://en.wikipedia.org/wiki/Permutation#Generation_in_lexicographic_order
+
+            for (int i = 0; i < pi.length; i++) {
+                pi[i] = i;
+            }
+
+            int i, j;
+            while (true) {
+                Object[] next = new Object[src.size()];
+                for (int k = 0; k < next.length; k++) {
+                    next[k] = src.delegate[pi[k]];
+                }
+                result.add(wrap(next));
+
+                i = pi.length - 2;
+                while (i >= 0 && pi[i] > pi[i + 1]) {
+                    i--;
+                }
+                if (i < 0) {
+                    break;
+                }
+
+                j = pi.length - 1;
+                while (pi[i] > pi[j]) {
+                    j--;
+                }
+                swapIndices(pi, i, j);
+
+                i++;
+                j = pi.length - 1;
+                while (i < j) {
+                    swapIndices(pi, i, j);
+                    i++;
+                    j--;
                 }
             }
-            return results;
+
+            return Array.ofAll(result);
         }
+    }
+
+    private static void swapIndices(int[] arr, int i, int j) {
+        int t = arr[i];
+        arr[i] = arr[j];
+        arr[j] = t;
     }
 
     @Override
--- a/vavr/src/main/java/io/vavr/collection/Array.java
+++ b/vavr/src/main/java/io/vavr/collection/Array.java
@@ -984,61 +984,17 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
 
     @Override
     public Array<Array<T>> permutations() {
-        final Array<T> src = distinct();
-        if (src.isEmpty()) {
-            return Array.empty();
-        } else if (src.size() == 1) {
-            return Array.of(src);
+        if (isEmpty()) {
+            return empty();
+        } else if (delegate.length == 1) {
+            return of(this);
         } else {
-            java.util.List<Array<T>> result = new java.util.ArrayList<>();
-            int[] pi = new int[src.size()];
-
-            // see Narayana algo
-            // https://en.wikipedia.org/wiki/Permutation#Generation_in_lexicographic_order
-
-            for (int i = 0; i < pi.length; i++) {
-                pi[i] = i;
-            }
-
-            int i, j;
-            while (true) {
-                Object[] next = new Object[src.size()];
-                for (int k = 0; k < next.length; k++) {
-                    next[k] = src.delegate[pi[k]];
-                }
-                result.add(wrap(next));
-
-                i = pi.length - 2;
-                while (i >= 0 && pi[i] > pi[i + 1]) {
-                    i--;
-                }
-                if (i < 0) {
-                    break;
-                }
-
-                j = pi.length - 1;
-                while (pi[i] > pi[j]) {
-                    j--;
-                }
-                swapIndices(pi, i, j);
-
-                i++;
-                j = pi.length - 1;
-                while (i < j) {
-                    swapIndices(pi, i, j);
-                    i++;
-                    j--;
-                }
-            }
-
-            return Array.ofAll(result);
+            final Array<Array<T>> zero = empty();
+            return distinct().foldLeft(zero, (xs, x) -> {
+                final Function<Array<T>, Array<T>> prepend = l -> l.prepend(x);
+                return xs.appendAll(remove(x).permutations().map(prepend));
+            });
         }
-    }
-
-    private static void swapIndices(int[] arr, int i, int j) {
-        int t = arr[i];
-        arr[i] = arr[j];
-        arr[j] = t;
     }
 
     @Override


jest:
@ruslansennov Probably it is, although "fast enough" is always usage-dependant.

I thought more general about S-T-J algorithm also in terms of memory reusing (e.g. as in Vector, where swap of two adjacent elements leaves most of trie branches untouched), but after rethinking, nobody will run `permutations()` on sequences long enough for storage sharing to make sense.

I'll check how this PR performs later in the day.

vavr/src/main/java/io/vavr/collection/Array.java
```diff
@@ -984,21 +984,63 @@ public int length() {
 
     @Override
     public Array<Array<T>> permutations() {
-        if (isEmpty()) {
-            return empty();
-        } else if (delegate.length == 1) {
-            return of(this);
+        final Array<T> src = distinct();
```
danieldietrich: I would call distinct() later, in the last if branch

vavr/src/main/java/io/vavr/collection/Array.java
```diff
@@ -984,21 +984,63 @@ public int length() {
...
+                swapIndices(pi, i, j);
+
+                i++;
+                j = pi.length - 1;
+                while (i < j) {
+                    swapIndices(pi, i, j);
+                    i++;
+                    j--;
                 }
             }
```
danieldietrich: **Just a thought, no requirement to change it:** I think the algorithm could be useful for all IndexSeq impls. I think we only need to make the factory methods `wrap` and `ofAll` generic. But I see that the other IndexedSeq impls do not operate on Object[], so maybe it is not possible to make it generic for impls though.

jest:
master ce6f975
```
Benchmark                                       Mode  Cnt    Score    Error  Units
VavrPermutationTest.bench_dummy_no_permutation  avgt   20    2.417 ±  0.037  ms/op
VavrPermutationTest.bench_guava_permutation     avgt   20    2.699 ±  0.022  ms/op
VavrPermutationTest.bench_vavr_permutation      avgt   20  519.320 ± 10.301  ms/op
```

PR 403b7fc
```
Benchmark                                       Mode  Cnt  Score   Error  Units
VavrPermutationTest.bench_dummy_no_permutation  avgt   20  2.454 ± 0.022  ms/op
VavrPermutationTest.bench_guava_permutation     avgt   20  2.517 ± 0.016  ms/op
VavrPermutationTest.bench_vavr_permutation      avgt   20  1.791 ± 0.018  ms/op
```

Hm, this is fast, suspiciously fast...

So I rerun the test with 9 elements in the array:

```
Benchmark                                       Mode  Cnt   Score   Error  Units
VavrPermutationTest.bench_dummy_no_permutation  avgt   20  28.672 ± 0.361  ms/op
VavrPermutationTest.bench_guava_permutation     avgt   20  25.231 ± 0.383  ms/op
VavrPermutationTest.bench_vavr_permutation      avgt   20  17.353 ± 0.164  ms/op
```

And with 10 elements in the array (large variance is probably to GC, not tuned JMH in this aspect in any way):

```
Benchmark                                       Mode  Cnt     Score     Error  Units
VavrPermutationTest.bench_dummy_no_permutation  avgt   20  1444.303 ±  89.729  ms/op
VavrPermutationTest.bench_guava_permutation     avgt   20   848.617 ±  99.800  ms/op
VavrPermutationTest.bench_vavr_permutation      avgt   20   327.846 ± 245.760  ms/op
```

So if the result of permutation is correct, it is indeed much faster in this PR.

danieldietrich:
Wow

jest:
Rerun the last test, within more stabilized environment:

```
VavrPermutationTest.bench_dummy_no_permutation  avgt   20  1519.123 ± 81.149  ms/op
VavrPermutationTest.bench_guava_permutation     avgt   20   776.917 ± 43.621  ms/op
VavrPermutationTest.bench_vavr_permutation      avgt   20   220.733 ±  2.601  ms/op
```

The further, the merrier! :)

vavr/src/main/java/io/vavr/collection/Array.java
ruslansennov: `EMPTY.distinct()` and `SINGLETON.distinct()` operations cost nothing. On other side, used algorithm requires length of source array at least 2, and we should check that result of `distinct()` operation is valid (to cover cases like [1,1,1,1...]). I would leave this as is.

vavr/src/main/java/io/vavr/collection/Array.java
ruslansennov: Yes, I also thought about this. But as you say this is not obvious change, so let's do it (or not) in own PR/issue

ruslansennov:
> So if the result of permutation is correct, it is indeed much faster in this PR.

It is correct, I checked this, trust me :)
Thank you for measurements

danieldietrich:
@ruslansennov It would be good to move the `distinct()` to the 3rd if branch, where it is needed. There are some instances created otherwise, even if the Array is empty:

```java
    @Override
    public Array<T> distinct() {
        return distinctBy(Function.identity());
    }

    @Override
    public <U> Array<T> distinctBy(Function<? super T, ? extends U> keyExtractor) {
        final java.util.Set<U> seen = new java.util.HashSet<>();
        return filter(t -> seen.add(keyExtractor.apply(t)));
    }
```

ruslansennov:
> There are some instances created otherwise, even if the Array is empty

OK, your're right! I'll do it this evening

danieldietrich:
Thx!

ruslansennov:
I added test for permutation of repeated elements and it seems that `lexicographic order` (both: `S-T-J` and `Narayana`) is bad choice in our case. These algorithms do not take into account repeating of elements and produce result different from other Vavr collections (as well as Scala collections) because they operate with indexes but not with elements. For example:
```java
a1 = [1, 1]
Narayana permutations: [[1,1], [1,1]]
Narayana permutations of distinct elements: [[1]]
Other Vavr collections and Scala: [[1,1]]

a1 = [1, 2, 2]
Narayana permutations: [[1, 2, 2], [1, 2, 2], [2, 1, 2], [2, 2, 1], [2, 1, 2], [2, 2, 1]]
Narayana permutations of distinct elements: [[1, 2], [2, 1]]
Other Vavr collections and Scala: [[1, 2, 2], [2, 1, 2], [2, 2, 1]]
```
So sad, but I don't see any solution of this problem and we should revert all changes to very first algorithm (v.2.0.6). It will be not so fast, but right in all cases

ruslansennov:
Or, we can return `Narayana.distinct()` sequence as result of permutations, but this will be very slow for big `N`

danieldietrich:
It does also not help to first calculate the Narayana permutations and then call distinct():

```
[[1, 2, 2], [1, 2, 2], [2, 1, 2], [2, 2, 1], [2, 1, 2], [2, 2, 1]].distinct() == [[1, 2, 2], [2, 1, 2], [2, 2, 1]]
```

For big `N` it will perform not well. The distinct() has to be called first. However, as you said, the result is different.

Let's take the first algorithm for now.

Any other suggestions are welcome.

jest:
Hm, I'm not sure, probably Narayana's can handle the repeated elements (http://www.mafagafogigante.org/lexicographic-permutations/), when the "grater-or-equal" is  used in `while` loops. Of course we are working on indices instead of the elements, but with additional array to `pi` used for comparisons it should work.

E.g. for `[a, b, b, c]` we would have `pi = [0, 1, 2, 3]` used for permutation generation and `ci = [0, 1, 1, 2]` used for comparisons in `while`-s. And swapping of course elements of both `pi` and `ci` simultaneously. I can try it on the weekend.

Anyway, this is IMO a breaking change in the API?

danieldietrich:
@jest 

> Anyway, this is IMO a breaking change in the API?

If the newly added unit test #1998 works, it will be fine.

> I can try it on the weekend.

Any helping hands are welcome :)

jest:
One question: is the order of permutations in result an implementation detail, a "nice-to-have" feature (e.g. lexicographic order) or part of API?

I'm not sure, I'll be easily able to handle repeated elements which are not grouped, like:

```
arr = [b, a, c, a, b]
pi=[0, 1, 2, 3, 4]
ci=[0, 1, 2, 1, 0]
```

Because `ci` is not first in the lexicographical order (of positions, as `pi` is) and I'll never get to permutation `[b, b, a, a, c]` (`ci=[0, 0, 1, 1, 2]`); hence, I can only generate lexicographical order in the sense of `ci` order, by first  grouping repeating elements, like `[b, a, c, a, b] => [b, b, a, a, c]` and generate permutation of that.

danieldietrich:
@jest This reminds me of java.util.Map. A few years ago some users relied on the keySet() element order, even if the API states that it is unpredictable. Then Java changed its internal implementation and applications broke.

The question is, do we expect Seq.permutations() to behave the same along all Seq implementations?
Following the Liskov Substitution Principle I would say yes, unless the Seq.permutations() API states the order is unpredictable.

It is always a good choice to make it clear using types. E.g. returning a `Set<? extends Seq<T>>` of permutations instead of a collection of the same type, e.g. `Array<Array<T>>`.

However, Scala does return an `Iterator`. Because computing permutations is costly (time & mem), this is a good idea. What if we do the same and provide a permutation iterator with a consistent order along all Seq implementations? We could do this change on a 1.0.0 branch and provide the old implementation on master that will get the version 0.9.1.

jest:
Now that I think of it, maybe we complicate too much? Maybe algorithms and data structures should be kept separate when the algorithm is non-trivial and selected implementation depends on the needs of the user (e.g. time vs. memory performance, structure vs. `Iterator` approach, lexicographical order vs. whatever, repeated elements allowed or not)? So, the default implementation should be "fast enough" for casual uses, and additional variants we discussed should be kept in a utility class? And only leave basic, data structure related code inside?

@danieldietrich Do you see a place in Vavr for more complicated algorithms (permutations, combinations), or should it be a separe project?

danieldietrich:
@jest currently we separate (some) algorithms in package-private classes like `Collections`.

Generally spoken, it is a good idea to separate structure and algorithms. However, I would not add another layer of indirection and make algorithms public 'pluggable' (from the user-perspective). Simplicity (~ user experience) is more important.

If another mem/cpu performance is needed, the user should refer to the corresponding collection type. Examples are List vs Vector vs Array for Seq.

PR #1998, opened 2017-05-19
add permutation tests

ruslansennov:


danieldietrich:
Thx! Yes, because of the Liskov Substitution Principle it is important that all Seq's return the same order of elements.

Issue #2007, opened 2017-06-06
Stream.dropUntil can be stuck in an infinite loop


bacey:
Hi!

I think I found a small bug:

```
// THIS PASSES
@Test
public void shouldNotStuckInAnInfiniteLoopWhenUsingDropUntilOnAFiniteStream() {
    final Stream<Integer> naturalNumbers = Stream.iterate(0, i -> i + 1).take(20);
    final Stream<Integer> naturalNumbersBiggerThanTen = naturalNumbers.dropUntil(i -> i > 10);
    final Integer firstNaturalNumberBiggerThanTen = naturalNumbersBiggerThanTen.head();

    assertThat(firstNaturalNumberBiggerThanTen).isEqualTo(11);
}

// THIS DOES NOT PASS
@Test
public void shouldNotStuckInAnInfiniteLoopWhenUsingDropUntilOnAnInfiniteStream() {
    final Stream<Integer> naturalNumbers = Stream.iterate(0, i -> i + 1);
    final Stream<Integer> naturalNumbersBiggerThanTen = naturalNumbers.dropUntil(i -> i > 10);
    final Integer firstNaturalNumberBiggerThanTen = naturalNumbersBiggerThanTen.head();

    assertThat(firstNaturalNumberBiggerThanTen).isEqualTo(11);
}
```

In the second test (`shouldNotStuckInAnInfiniteLoopWhenUsingDropUntilOnAnInfiniteStream`), the `dropUntil` function is stuck in an infinite loop.

Quoting Gregor Trefs @gtrefs:
> The length of an infinite stream is infinite. `dropUntil` relies on `Collections.dropUntil` where `length` is used to iterate over the elements in a `Seq`. The length of a stream is a fold left counting the elements. For an infite stream, this ends in an infinte loop. In Scala, the following works: `Stream.iterate(0)(_ + 1).dropwHile(_ <= 10)` while the counter part in vavr seems not to work `Stream.iterate(0, i -> i + 1).dropWhile(i -> i <= 10);`.

Tested on the v0.9.0 branch.

ruslansennov:
Also we have very bad performance at least for `Stream` and `List` because of `get()` complexity is `O(N)`

#1749 should be reviewed again

gtrefs:
@ruslansennov 
You are right. It looks like unifying the implementation in `Collections` is the culprit.
Before:
```Java
default Stream<T> dropWhile(Predicate<? super T> predicate) {	
[...]

PR #2009, opened 2017-06-07
revert dropUntil*() for Stream and List

ruslansennov:
see #2007

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Collections.java
+++ b/vavr/src/main/java/io/vavr/collection/Collections.java
@@ -51,7 +51,9 @@ final class Collections {
                     .foldLeft(Iterator.of(empty), (product, ignored) -> product.flatMap(el -> seq.map(t -> (S) el.append(t))));
         }
     }
-    
+
+    // DEV-NOTE: Use this method for non-infinite and direct-access collection only
+    // see https://github.com/vavr-io/vavr/issues/2007
     @SuppressWarnings("unchecked")
     static <T, S extends IndexedSeq<T>> S dropRightUntil(S seq, Predicate<? super T> predicate) {
         Objects.requireNonNull(predicate, "predicate is null");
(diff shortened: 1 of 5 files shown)


vavr/src/main/java/io/vavr/collection/Collections.java
```diff
@@ -51,7 +51,9 @@ static boolean areEqual(Iterable<?> iterable1, Iterable<?> iterable2) {
                     .foldLeft(Iterator.of(empty), (product, ignored) -> product.flatMap(el -> seq.map(t -> (S) el.append(t))));
         }
     }
-    
+
+    // DEV-NOTE: Use this method for non-infinite and direct-access collection only
+    // see https://github.com/vavr-io/vavr/issues/2007
```
gtrefs: Could you rather add a little explanation why one should not use this method when using infinite collections? I think this is more convinient than referring to a github issue which might get lost over the time.

vavr/src/main/java/io/vavr/collection/Collections.java
```diff
@@ -63,6 +65,8 @@ static boolean areEqual(Iterable<?> iterable1, Iterable<?> iterable2) {
         return (S) seq.take(0);
     }
 
+    // DEV-NOTE: Use this method for non-infinite and direct-access collection only
+    // see https://github.com/vavr-io/vavr/issues/2007
```
gtrefs: See comment above.

vavr/src/main/java/io/vavr/collection/Collections.java
ruslansennov: knock on wood :smile: 
I added a short description

ruslansennov:
> [Queue|List|Stream].drop[While|Until]
[Queue|List|Stream].dropRightUntil

these tests already exist, see `AbstractTraversableTest`

Issue #2005, opened 2017-06-05
ConcatIterator iteration performance is bad when ConcatIterator is built by nesting other ConcatIterators


nfekete:
ConcatIterator has very bad iteration performance when the concatenated iterators are ConcatIterators themselves to some degree (not necessarily deep).

I've made a JMH test for this use case, and the numbers vs scala's iterators are looking like this:

| Impl               |   Params |          Score  ±  |   Error |   Unit | scala_persistent | vavr_persistent
|--------------------|----------|--------------------|---------|--------|------------------|----------------
| scala_persistent   |       10 |   2,115,225.97  |  ± 14.18% |  ops/s |                  |        188.51×
| scala_persistent   |       20 |   1,180,047.73  |   ± 7.15% |  ops/s |                  |     100537.32×
| vavr_persistent    |       10 |      11,220.68  |  ± 10.62% |  ops/s |           0.01×  |               
| vavr_persistent    |       20 |          11.74  |   ± 2.64% |  ops/s |           0.00×  |               

The test parameter in this case was represents both the nesting level of ConcatIterators *and* the number of values in the final iterator.

This means, that in the last vavr_persistent test with size=20, vavr was able to iterate only ~12 times a second over an iterator of just 20 elements.

There are multiple inefficiencies in the ConcatIterator implementation, like `current.hasNext()` [being called unnecessarily multiple times](https://github.com/vavr-io/vavr/blob/b3a746df440a81d1312f41d7e4431e5e545dd240/vavr/src/main/java/io/vavr/collection/Iterator.java#L2009-L2012), and the value for `hasNext()` could be cached until the next invocation of `next()`, but these account only for a constant factor. The main issue is that ConcatIterator keeps an inefficient structure for the iteration when nesting ConcatIterators.

ruslansennov:
> The main issue is that ConcatIterator keeps an inefficient structure for the iteration when nesting ConcatIterators.

Do you have any thoughts about this?

nfekete:
[...]

PR #2011, opened 2017-06-08
improve ConcatIterator

ruslansennov:
See #2005 
```
Target             Impl                  Params  Count            Score  ±     Error    Unit  scala_persistent  vavr_persistent
IteratorBenchmark  scala_persistent          10      5     2,516,805.84  ±     5.65%   ops/s                             0.89×
IteratorBenchmark  scala_persistent          20      5     1,412,594.83  ±     9.46%   ops/s                             0.77×
IteratorBenchmark  scala_persistent         100      5       286,046.91  ±     1.46%   ops/s                             0.72×
IteratorBenchmark  scala_persistent        1000      5        28,910.28  ±    72.00%   ops/s                             0.73×
IteratorBenchmark  vavr_persistent           10      5     2,820,759.98  ±     4.30%   ops/s            1.12×                 
IteratorBenchmark  vavr_persistent           20      5     1,825,800.14  ±     7.18%   ops/s            1.29×                 
IteratorBenchmark  vavr_persistent          100      5       399,723.11  ±     1.46%   ops/s            1.40×                 
IteratorBenchmark  vavr_persistent         1000      5        39,369.96  ±     3.55%   ops/s            1.36×                 
```

@nfekete please review

### DIFF
--- a/vavr-benchmark/src/test/java/io/vavr/collection/IteratorBenchmark.java
+++ b/vavr-benchmark/src/test/java/io/vavr/collection/IteratorBenchmark.java
@@ -74,7 +74,7 @@ public class IteratorBenchmark {
     @State(Scope.Benchmark)
     public static class Concat {
 
-        @Param({ "10", "20" /*, "100", "1000"*/ })
+        @Param({ "10", "20" , "100", "1000" })
         private int size;
 
         @Benchmark
--- a/vavr/src/main/java/io/vavr/collection/Iterator.java
+++ b/vavr/src/main/java/io/vavr/collection/Iterator.java
@@ -64,7 +64,11 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
         if (iterables.length == 0) {
             return empty();
         } else {
-            return new ConcatIterator<>(Stream.of(iterables).map(Iterator::ofAll).iterator());
+            ConcatIterator<T> res = new ConcatIterator<>();
+            for (Iterable<? extends T> iterable : iterables) {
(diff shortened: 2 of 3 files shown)


ruslansennov:
Need to do more tests...

nfekete:
LGTM. I like it's iterative, besides being faster, stack will not be a limiting factor. 👍

nfekete:
The only drawback I can see is that the `ConcatIterator` became mutable in a new way (was mutable while iterating over it, now it's also mutable while it's being built). This means that the following code is an infinite loop:
```java
        Iterator<Integer> single = of(1);
        Iterator<Integer> concat = single.concat(single);
        concat = concat.concat(concat);
        int count = concat.count(__ -> true);
```
I'm not entirely sure if we should correct this kind of behavior (scala's solution behaves the same). We would need to create a copy of the linked list of `Cell`s.

danieldietrich:
Thank you @ruslansennov and @nfekete.
I will merge it.
The [drawback](https://github.com/vavr-io/vavr/pull/2011#issuecomment-307415120) you mentioned is acceptable because an Iterator is mutable.The javadoc says:

```java
 * <strong>Note:</strong> Iterators encapsulate mutable state.
 * They are not meant to be used concurrently by different threads. Do not reuse Iterators, e.g. after passing to
 * {@linkplain io.vavr.collection.List#ofAll(Iterable)}.
```

Especially the *"Do not reuse Iterators"* part is violated when concatenating an Iterator with itself.

nfekete:
Thank you, guys!

PR #2016, opened 2017-06-20
Add a fold operation to Option

hamnis:
This is useful whenever you need to handle both cases of an option.
This is implemented in terms of map and getOrElse, but most operations
on Option could be implemented in terms of fold. Then the implementation
would of course be different.

### DIFF
--- a/vavr/src/main/java/io/vavr/control/Option.java
+++ b/vavr/src/main/java/io/vavr/control/Option.java
@@ -361,6 +361,18 @@ public interface Option<T> extends Value<T>, Serializable {
     }
 
     /**
+     * Folds either the {@code None} or the {@code Some} side of the Option value.
+     *
+     * @param ifNone  maps the left value if this is a None
+     * @param f maps the value if this is a Some
+     * @param <U>         type of the folded value
+     * @return A value of type U
+     */
+    default <U> U fold(Supplier<? extends U> ifNone, Function<? super T, ? extends U> f) {
+        return this.<U>map(f).getOrElse(ifNone);
+    }
+
+    /**
      * Applies an action to this value, if this option is defined, otherwise does nothing.
      *
      * @param action An action which can be applied to an optional value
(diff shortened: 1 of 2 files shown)


ruslansennov:
Hi @hamnis 
This is obvious method, but could you add tests for him?

hamnis:
@ruslansennov done

hamnis:
Could this please be merged?

danieldietrich:
@hamnis Sorry, I'm really busy at $work. I will take a look this evening!

vavr/src/main/java/io/vavr/control/Option.java
```diff
@@ -361,6 +361,18 @@ default T getOrElse(Supplier<? extends T> supplier) {
...
 
     /**
+     * Folds either the {@code None} or the {@code Some} side of the Option value.
+     *
+     * @param mapperNone  maps the left value if this is a None
+     * @param mapperSome maps the right value if this is a Some
+     * @param <U>         type of the folded value
+     * @return A value of type U
+     */
+    default <U> U fold(Supplier<? extends U> mapperNone, Function<? super T, ? extends U> mapperSome) {
```
danieldietrich: The supplier does not map an element. Could we use the same parameter names like Scala?

E.g. (modulo generics):

```java
fold(Supplier ifEmpty, Function f)
```

or

```java
fold(Supplier ifEmpty, Function mapper)
```

hamnis: Sure. Will fix.

hamnis: I squashed the commits into one, and fixed the parameter names,

danieldietrich:
Thank you!

danieldietrich:
Hi @juanpavergara,
the users want semantic versioning in order to be able to be able to upgrade without regrets (regarding backward compatibility of binary and source).
The 0.9.2 was a patch release. It includes basically bug fixes, no new features.
All new features go to the next major release 1.0.0 because we do not plan a minor release 0.10.0.

juanpavergara:
Oh ok thank you for your reply @danieldietrich :)
I was about to implement it and open a PR but then checked the source and realized @hamnis already did :)

Issue #1945, opened 2017-04-12
Iterator should internally not make use of toStream() anymore


danieldietrich:
Instead we should implement stateful Iterators that achieve the same functionality.

The reason is memory-efficiency.

danieldietrich:
Closing the issue, Iterator now does not contain toStream() anymore. Thx @ruslansennov for fixing it!

danieldietrich:
Will re-open it, also `Stream.ofAll(...)` should not be used anymore by Iterator internally.

PR #2025, opened 2017-06-30
improve Traversable.reduceLeft

ruslansennov:
Towards #1945

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Iterator.java
+++ b/vavr/src/main/java/io/vavr/collection/Iterator.java
@@ -1665,8 +1665,11 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
         if (isEmpty()) {
             throw new NoSuchElementException("reduceLeft on Nil");
         } else {
-            final Stream<T> stream = Stream.ofAll(this);
-            return stream.tail().foldLeft(stream.head(), op);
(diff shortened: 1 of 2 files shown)


vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -1065,11 +1065,7 @@ default Number product() {
     @Override
     default T reduceLeft(BiFunction<? super T, ? super T, ? extends T> op) {
         Objects.requireNonNull(op, "op is null");
-        if (isEmpty()) {
-            throw new NoSuchElementException("reduceLeft on Nil");
-        } else {
-            return tail().foldLeft(head(), op);
```
ruslansennov: `tail()` is an expensive operation for certain collections

vavr/src/main/java/io/vavr/collection/Iterator.java
```diff
@@ -1665,8 +1665,11 @@ default T reduceLeft(BiFunction<? super T, ? super T, ? extends T> op) {
         if (isEmpty()) {
             throw new NoSuchElementException("reduceLeft on Nil");
         } else {
-            final Stream<T> stream = Stream.ofAll(this);
```
ruslansennov: unnecessary stream

PR #2028, opened 2017-07-02
*.fill & *.tabulate refactoring

ruslansennov:
Now all methods uses `Iterator.tabulate` which is moved from `Collections`


ruslansennov:
Sorry. There was no profit in case of `Array.fill` (which I discovered), but we should be more accurate in other cases (like `List` for example)

danieldietrich:
Ok, thank you for all the effort!

PR #2033, opened 2017-07-06
support Either.sequence

emmanueltouzery:
Either.sequence() appears to have been overlooked. This PR adds it.

### DIFF
--- a/vavr/src/main/java/io/vavr/control/Either.java
+++ b/vavr/src/main/java/io/vavr/control/Either.java
@@ -8,6 +8,8 @@ package io.vavr.control;
 
 import io.vavr.Value;
 import io.vavr.collection.Iterator;
+import io.vavr.collection.Seq;
+import io.vavr.collection.Vector;
 
 import java.io.Serializable;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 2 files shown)


vavr/src/main/java/io/vavr/control/Either.java
```diff
@@ -163,6 +165,29 @@
...
+     *
+     * @param values An {@link Iterable} of {@code Either}s
+     * @param <L>    left type of the Eithers
+     * @param <R>    right type of the Eithers
+     * @return A {@code Either} of a {@link Seq} of results
+     * @throws NullPointerException if {@code values} is null
+     */
+    static <L,R> Either<L, Seq<R>> sequence(Iterable<? extends Either<? extends L, ? extends R>> values) {
+        Objects.requireNonNull(values, "values is null");
+        Vector<R> vector = Vector.empty();
```
ruslansennov: Maybe it is better to use other collection here(`java.util.ArrayList`?) for better performance

emmanueltouzery: hmm but in the end we do return the list in case of success. So maybe it's better to create directly the right format rather than paying the conversion cost later? Also note that I took `Try.sequence` as a model, it's done the same there; so if we chnage, we can change both.

ruslansennov: > hmm but in the end we do return the list in case of success. So maybe it's better to create directly the right format rather than paying the conversion cost later? 

But this will be one conversion vs. `N` updates. 

> Also note that I took Try.sequence as a model, it's done the same there; so if we change, we can change both.

I believe that temporary internal mutable collections are the right way for a lot of `vavr` methods. @danieldietrich ?

emmanueltouzery: One thing is for sure, these implementations are totally non-fp like (not using flatmap, filter and so on). So i guess elegancy shouldn't be a concern but only performance. I am not familiar with the system vavr has for benchmarks, hopefully it's not too difficult to measure? The other thing is that clearly memory use will be higher for create+convert compared to only create (spike at ~2x the memory use I guess).

ruslansennov: > hopefully it's not too difficult to measure? 

There is some benchmarks in `vavr-benchmark` module, we can try :) But note that vavr's `Vector` internally consists from `T[]` blocks and every change causes copying this block and change of all blocks tree

danieldietrich:
Hi Emmanuel,

many thanks. Yes, we've overlooked it :)
It would be so nice if we could define such methods in an abstract way in super classes, but that's not possible. Maybe we will find other ways in future to ensure it, like unit tests

```java
given: all public classes
when: extending Value
then: it has a public static ... method named sequence with parameter Iterable...
```

I will pull your version. Both you and Ruslan have a point. Now I have a burning ambition to see the benchmark results :)

Thx!

PR #2038, opened 2017-07-14
Add Seq.takeRightUntil and Seq.takeRightWhile

ummels:
`takeRightUntil` and `takeRightWhile` were conspiciously missing from `Seq`, while `dropRightUntil` and `dropRightWhile` are there.

This pull request adds the missing methods with a standardized implementation for `IndexedSeq`, also fixes/improves some docstrings.

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Array.java
+++ b/vavr/src/main/java/io/vavr/collection/Array.java
@@ -1307,6 +1307,17 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
     }
 
     @Override
+    public Array<T> takeUntil(Predicate<? super T> predicate) {
+        return io.vavr.collection.Collections.takeUntil(this, predicate);
+    }
+
+    @Override
+    public Array<T> takeWhile(Predicate<? super T> predicate) {
+        Objects.requireNonNull(predicate, "predicate is null");
+        return takeUntil(predicate.negate());
+    }
+
+    @Override
     public Array<T> takeRight(int n) {
         if (n >= length()) {
             return this;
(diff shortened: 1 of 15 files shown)


vavr/src/main/java/io/vavr/collection/CharSeq.java
```diff
@@ -1028,23 +1039,14 @@ public CharSeq takeRight(int n) {
     }
 
     @Override
-    public CharSeq takeUntil(Predicate<? super Character> predicate) {
-        Objects.requireNonNull(predicate, "predicate is null");
-        return takeWhile(predicate.negate());
+    public CharSeq takeRightUntil(Predicate<? super Character> predicate) {
+        return io.vavr.collection.Collections.takeRightUntil(this, predicate);
```
ruslansennov: Could you add tests for `CharSeq`? Tests in `AbstractSeqTest` do not cover this collection

vavr/src/main/java/io/vavr/collection/Array.java
```diff
@@ -1320,21 +1331,14 @@ public int length() {
     }
 
     @Override
-    public Array<T> takeUntil(Predicate<? super T> predicate) {
-        Objects.requireNonNull(predicate, "predicate is null");
-        return takeWhile(predicate.negate());
+    public Array<T> takeRightUntil(Predicate<? super T> predicate) {
+        return io.vavr.collection.Collections.takeRightUntil(this, predicate);
```
danieldietrich: minor/just a style-hint (but no need to change it): we generally import Vavr types and fully qualify Java types.

danieldietrich:
Hi Michael,
great addition - looks all good to me!
Many thanks 😊

vavr/src/main/java/io/vavr/collection/Array.java
ummels: Okay. I was wondering because in the existing code I usually saw `io.vavr.collection.Collections` fully qualified.

danieldietrich: Oh :) I think it is not that important...

Issue #2051, opened 2017-08-06
PartialFunction.lift


skestle:
`PartialFunction` should have a lift function that looks something like

```
Function1<T, Option<R>> lift() {
    // return t -> Option.of(t).collect(this); // I don't think reuse makes sense in this case
    return t -> Option.when(isDefinedAt(t), apply(t));
}
```

skestle:
If the second option is chosen, then Option could change `collect` to be `return orElse(partialFunction.lift().apply(get()))`...

Which could be further refactored to 2 methods of `return collect(partialFunction.lift())`
and 
```
Option<R> collect(Function<? super T, Option<? extends R>> function) {
   return orElse(function.apply(get()));
{
```

danieldietrich:
Agreed, that makes sense.

miszasty93:
#2058

PR #2061, opened 2017-08-24
Modified Option.collect to take advantage of PartialFunction.lift() (…

skestle:
…#2051)

@danieldietrich what about this change? Should we include it? (mvn test indicates success to me, but my eclipse (oxygen) is having real trouble sorting out the assertions)

### DIFF
--- a/vavr/src/main/java/io/vavr/control/Option.java
+++ b/vavr/src/main/java/io/vavr/control/Option.java
@@ -177,7 +177,7 @@ public interface Option<T> extends Value<T>, Serializable {
      */
     default <R> Option<R> collect(PartialFunction<? super T, ? extends R> partialFunction) {
         Objects.requireNonNull(partialFunction, "partialFunction is null");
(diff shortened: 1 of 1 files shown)


danieldietrich:
Thank you!

vavr/src/main/java/io/vavr/control/Option.java
```diff
@@ -177,7 +177,7 @@
      */
     default <R> Option<R> collect(PartialFunction<? super T, ? extends R> partialFunction) {
         Objects.requireNonNull(partialFunction, "partialFunction is null");
-        return filter(partialFunction::isDefinedAt).map(partialFunction::apply);
+        return flatMap(partialFunction.lift()::apply);
```
danieldietrich: @skestle Just recognized that the `::apply` is not necessary. Does it compile if you leave it away? Could you please check that, I'm on vacation and have no IDE at hand...
Thx!

PR #2072, opened 2017-09-06
Changed Either.sequence return type

### LINKED ISSUE #2064: Either.sequence should return Either<Seq<L>, Seq<R>>
In the case of Either (as opposed to Try/Option) it makes sense to accumulate also the left values.

We recently added Either.sequence - it was forgotten. Because it is still not released, we are able to change the signature for the upcoming 0.9.1 patch release.

```java
// current signature
static <L,R> Either<L, Seq<R>> sequence(Iterable<? extends Either<? extends L, ? extends R>> values) { ... }

// new signature
static <L,R> Either<Seq<L>, Seq<R>> sequence(Iterable<? extends Either<? extends L, ? extends R>> values) { ... }
```

**Semantics:** `sequence` collects all right values if all given Eithers are Right, else it collects all Left instances (if one or more Eithers are Left).

**Usage example:**

```java
Either<Seq<L>, Seq<R>> result = Either.sequence(either1, either2, ...);

// perform ugly side-effects
result.orElseRun(errors -> ...);
```

danieldietrich:
Fixes #2064

### DIFF
--- a/vavr/src/main/java/io/vavr/control/Either.java
+++ b/vavr/src/main/java/io/vavr/control/Either.java
@@ -9,7 +9,6 @@ package io.vavr.control;
 import io.vavr.Value;
 import io.vavr.collection.Iterator;
 import io.vavr.collection.Seq;
-import io.vavr.collection.Vector;
 
 import java.io.Serializable;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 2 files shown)


vavr/src/main/java/io/vavr/control/Either.java
```diff
@@ -166,25 +165,40 @@
...
+     * Either.sequence(List.of(Either.right(1), Either.right(2)))
      *
-     * @param values An {@link Iterable} of {@code Either}s
-     * @param <L>    left type of the Eithers
-     * @param <R>    right type of the Eithers
-     * @return A {@code Either} of a {@link Seq} of results
-     * @throws NullPointerException if {@code values} is null
+     * // = Left(Seq("x"))
+     * Either.sequence(List.of(Either.right(1), Either.left("x")))
+     * }</pre>
```
danieldietrich: Following the new 'boy scout rule':

> When touching public API, we ensure that the Javadoc contains example code.

vavr/src/main/java/io/vavr/control/Either.java
```diff
@@ -166,25 +165,40 @@
...
-        return Either.right(vector);
+    @SuppressWarnings("unchecked")
+    static <L,R> Either<Seq<L>, Seq<R>> sequence(Iterable<? extends Either<? extends L, ? extends R>> eithers) {
+        Objects.requireNonNull(eithers, "eithers is null");
+        return Iterator.ofAll((Iterable<Either<L, R>>) eithers)
+                .partition(Either::isLeft)
+                .apply((leftPartition, rightPartition) -> leftPartition.hasNext()
+                    ? Either.left(leftPartition.map(Either::getLeft).toVector())
+                    : Either.right(rightPartition.map(Either::get).toVector())
+                );
```
danieldietrich: This simplifies the imperative version that covers the left and the right case, e.g.

```java
Seq rightResult = Seq();
Iterator iter = ...
while (iter.hasNext()) {
    Either e = iter.next();
    if (e.isRight()) {
        /* add e.get() to rightResult */
    } else {
        Seq leftResult = Seq(e);
        while (iter.hasNext()) {
            Either e2 = iter.next();
            if (e2.isLeft()) {
                /* add e2.getLeft() to leftResult */
            }
        }
        return Either.left(leftResult);
    }
}
return Either.right(rightResult);
```

Iterator.partition currently internally uses a Stream to split the original Iterator. This will be improved with #1945 _if possible_.

Issue #1938, opened 2017-04-05
Alternative to Validation.combine()


fagossa:
Hi everyone,

I'm proposing an update to the `Validation` functor.

Let's assume we have this class
```
class Person {
  private String name;
  private String email;
  private int age;
  ...
}
```

As per the official documentation, the correct way to create a validation is as follows:

```
Validation
    .combine(
        validateName(name),
        validateEmail(email),
        validateAge(age))
    .ap(Person::new)
```

`combine` has 8 overrides, supporting as many different validations instances.

However, for classes with many parameters this approach could not be adequate.

What I'm proposing is a workaround to the 8 parameter limitation.

The idea is to create a builder that internally stores a `List` of validations and we are able to trigger the whole thing at will. Like this:

```
Validation<List<String>, Person> validation =
  new Validation.Builder<String, Person>()
        .combine(p -> validateAge(p))
        .combine(p -> validateName(p))
        .combine(p -> validateEmail(p))
        .apply(person);
```

I would be glad to make a PR. But, first I would like to check with you if it is worth it before going any further.

What do you thing?

regards

fagossa:
Perhaps this is somehow related to [Issue 1565](https://github.com/javaslang/javaslang/issues/1565).

danieldietrich:
@fagossa This definitely sounds interesting. What is the return type of the first combine call?

```java
new Validation.Builder<String, Person>().combine(p -> validateAge(p))
```

A List can only hold values of one type, say `T`. How are the List objects correctly typed when calling `.apply(Person::new)`.

(You wrote `.apply(person)` but I think you meant `.apply(Person::new)`)

fagossa:
@danieldietrich 

`combine` is now a fluid `append` for the list containing our different `Validation`'s.

```
public Builder<E, O> combine(Function<O, Validation<List<E>, O>> func) {
    functions.add(func);
    return this;
}
```

I really ment 
```
.apply(person)
```

And this is probably the biggest flaw of this approach: you specify a `T` and the whole validation process is executed on that instance (e.g. an instance of Person).

danieldietrich:
Interesting approach. It is (more or less) equivalent to

```java
static <T, E> List<E> validate(T value, Function<? super T, ? extends Option<? extends E>>... validators) {
    return List.of(validators).flatMap(validator -> validator.apply(value));
}
```

Example:

```java
List<String> errors = validate(person,
        p -> validateAge(p),
        p -> validateName(p),
[...]

Issue #1565, opened 2016-09-13
Add more syntactic sugar to Validation


danieldietrich:
**Update: Follow #2073** and take this comment into account:

> widen Validation's List&lt;E> to List&lt;? extends E> in many cases.

---

Currently we have:

``` java
Validation<E, T1> v1 = ...
Validation<List<E>, T2> v2 = ...
Validation<List<E>, U> result = v1.ap(v2.map(t2 -> t1 -> f(t1, t2)));
```

We additionally need:
- `<T2, U> Validation<List<E>, U> append(Validation<List<E>, T2>, Function<? super T1, ? super T2, ? extends U>)` that _appends_ an error to an error list
- `<T2, U> Validation<List<E>, U> prepend(Validation<List<E>, T2>, Function<? super T1, ? super T2, ? extends U>)` that _prepends_ an error to an error list
- `static <E, T1, T2> Validation<List<E>, Tuple2<T1, T2>> combine(Validation<List<? extends E>, T1>, Validation<List<? extends E>, T2>)`
- ...
- `static <E, T1, T2, ..., T8> Validation<List<E>, Tuple8<T1, T2, ..., T8>> combine(Validation<List<? extends E>, T1>, Validation<List<? extends E>, T2>, ..., Validation<List<? extends E>, T8)`

**Note:** For release 3.0.0 we should widen `List<E>` to `List<? extends E>` in many cases. Also we should consider to change `Validation<List<E>, T>` to `Validation<Seq<E>, T>` and use `Vector` instead of `List`.

io7m:
For the record, I tend to rewrite `flatten :: Validation<List<List<E>>, T> -> Validation<List<E>, T>` every time I use Validation, so perhaps something more generally typed than that would be useful.

danieldietrich:
Java's type system does not allow us such a flatten method to be defined as instance method - it had to be a static method. However, it is uncommon to flatten the 'negative' (error) case. For example `Either.map()` and `Either.flatMap()` operate on the right value.

I think we need to prevent that `Validation<List<E>, Tn>` are combined to a `Validation<List<List<E>>, Tn>`. This could be accomplished by changing `Validation<E, T>` in the way that it implicitly has a list of errors instead of an error. I.e. `Invalid<E, T>.getError()` returns a `List<E>` instead of an `E`. Then we do not need to care about flattening any more at all. It is automatically done on each operation.

[...]

PR #2073, opened 2017-09-07
Replace Validation class with one that allows for nested validation

amygithub:
Replace Validation class with one that allows for nested validation of deep Pojos.  

Using the current Validation class for anything other than flat classes leads to unreadable code. 

For example, I wish to parse and validate a Person object, which contains a name String and an Address object. Trying to combine these validations leads to very unreadable code.

The chief problem is that validation of a single field has a single error, and a combined validation has a sequence of errors. These cannot be combined in a fluent way.  By ensuring that every validation returns a sequence of errors, I can combine any validations I want fluently. This allows for complex validation rules with nested validation.  

I have deprecated the current Validation class and created a new one in another package. As some of the deprecation warnings for the generated code were appearing despite @SuppressWarnings("deprecation") being switched on (a known issue), I have had to disable deprecation warnings. 

The library is still backward compatible, and it's pretty simple to upgrade your deprecated classes and methods with the new ones. 

Any feedback is appreciated.


danieldietrich:
Hi Amy, many thanks for your PR!

You are fast with providing a solution. We already planned to give `Validation` some love (see #1565 and #1938).

We are in the lucky case of working on a major release that will contain changes that affect backward compatibility. We wrote about it [here](http://blog.vavr.io/fast-forward-to-vavr-1-0). It would be better to re-think the existing Validation rather than deprecating it and adding a new one.

Currently I'm preparing the 0.9.1 patch release (targeted for this weekend). After that I will come back to this PR!

Thx,

Daniel

amygithub:
Great to hear, Daniel! I was worried that backward compatibility would be an issue, so good to see you're biting the bullet. I hadn't even realised that there was an open issue about what I changed.

amygithub:
Hi @danieldietrich , have you had a chance to look at this yet?

danieldietrich:
Hi @amygithub, I took a look now. The differences to the 'original' Validation class are hard to spot. As I said, we are able to change the original one. Could you please move your particular changes to control.Validation and remove the control.fluent package? Then it is easier for me to see the difference and to review the PR.
Sorry for being so slow in the past - but now there is room to get this right!
Thx!

amygithub:
Hi @danieldietrich. I've moved the classes now, so should be ready for review.

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -91,24 +91,32 @@
...
+        }
+        return new Invalid<E, T>(errors);
+    }
+
     /**
      * Creates a {@code Validation} of an {@code Either}.
      *
      * @param either An {@code Either}
-     * @param <E>    type of the given {@code error}
+     * @param <E>    type of the given {@code errors}
```
danieldietrich: I think this was a copy-paste error by us. Better would be to name it

```java
* @param <E>    error type
```

or (like below)

```java
* @param <E>    type of errors
```

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -129,13 +137,14 @@
      * or an invalid Validation containing an accumulated List of errors.
      * @throws NullPointerException if values is null
      */
-    static <E, T> Validation<Seq<E>, Seq<T>> sequence(Iterable<? extends Validation<Seq<E>, T>> values) {
+    static <E, T> Validation<E, Seq<T>> sequence(Iterable<? extends Validation<E, T>> values) {
```
danieldietrich: I don't know why we did these generics. It should be

```java
sequence(Iterable<? extends Validation<? extends E, ? extends T>> values)
```

(see Either.sequence for example)

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -129,13 +137,14 @@
...
         Objects.requireNonNull(values, "values is null");
         List<E> errors = List.empty();
         List<T> list = List.empty();
-        for (Validation<Seq<E>, T> value : values) {
+        for (Validation<E, T> value : values) {
             if (value.isInvalid()) {
-                errors = errors.prependAll(value.getError().reverse());
+                Seq<? extends E> otherErrors = value.getErrors().reverse();
```
danieldietrich: (minor: we declare local vars `final` when not assigned again)

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -393,36 +402,36 @@ default boolean isEmpty() {
...
+     * Gets the value if it is a Valid or an value calculated from the errors
      *
-     * @param other a function which converts an error to an alternative value
+     * @param other a function which converts an errors to an alternative value
      * @return the value, if the underlying Validation is a Valid, or else the alternative value
-     * provided by {@code other} by applying the error.
+     * provided by {@code other} by applying the errors.
      */
-    default T getOrElseGet(Function<? super E, ? extends T> other) {
+    default T getOrElseGet(Function<Seq<? extends E>, ? extends T> other) {
```
danieldietrich: The 1st generic type param is in contra-variant (= consumer) position, it should have the following generic type bounds:

```java
default T getOrElseGet(Function<? super Seq<? super E>, ? extends T> other) {
```

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -393,36 +402,36 @@ default boolean isEmpty() {
...
     /**
-     * Gets the error of this Validation if is an Invalid or throws if this is a Valid
+     * Gets the errors of this Validation if is an Invalid or throws if this is a Valid
      *
-     * @return The error of this Invalid
+     * @return The errors of this Invalid
      * @throws RuntimeException if this is a Valid
      */
-    E getError();
+    Seq<? extends E> getErrors();
```
danieldietrich: The lack of declaration-site variance makes type declarations hard in Java. Here, the return type has to be

```java
Seq<E> getErrors();
```

because we do not plan to override the Validation class (and return a subtype of Seq in a method override).

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -393,36 +402,36 @@ default boolean isEmpty() {
...
 
     /**
      * Returns this as {@code Either}.
      *
-     * @return {@code Either.right(get())} if this is valid, otherwise {@code Either.left(getError())}.
+     * @return {@code Either.right(get())} if this is valid, otherwise {@code Either.left(getErrors())}.
      */
-    default Either<E, T> toEither() {
-        return isValid() ? Either.right(get()) : Either.left(getError());
+    default Either<Seq<? extends E>, T> toEither() {
```
danieldietrich: Here, it has to be

```java
default Either<Seq<E>, T> toEither()
```

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -468,11 +477,11 @@ default void forEach(Consumer<? super T> action) {
      * @return an instance of type U
      * @throws NullPointerException if fInvalid or fValid is null
      */
-    default <U> U fold(Function<? super E, ? extends U> fInvalid, Function<? super T, ? extends U> fValid) {
+    default <U> U fold(Function<Seq<? extends E>, ? extends U> fInvalid, Function<? super T, ? extends U> fValid) {
```
danieldietrich: also contra-variant position, i.e. `Function<? super Seq<? super E>, ? extends U>`

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -468,11 +477,11 @@ default void forEach(Consumer<? super T> action) {
...
         Objects.requireNonNull(fInvalid, "fInvalid is null");
         Objects.requireNonNull(fValid, "fValid is null");
         if (isInvalid()) {
-            final E error = this.getError();
+            final Seq<? extends E> error = this.getErrors();
```
danieldietrich: will change to the following after you change the return type of `getErrors()` as stated above:

```java
final Seq<E> error = this.getErrors();
```

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -486,9 +495,9 @@ default void forEach(Consumer<? super T> action) {
      *
      * @return a flipped instance of Validation
      */
-    default Validation<T, E> swap() {
+    default Validation<T, Seq<? extends E>> swap() {
```
danieldietrich: Can be simplified to

```java
default Validation<T, Seq<E>> swap()
```

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -486,9 +495,9 @@ default void forEach(Consumer<? super T> action) {
...
         if (isInvalid()) {
-            final E error = this.getError();
+            final Seq<? extends E> error = this.getErrors();
```
danieldietrich: can be simplified to

```java
final Seq<E> error = this.getErrors();
```

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -521,11 +530,11 @@ default void forEach(Consumer<? super T> action) {
      * @return an instance of Validation&lt;U,R&gt;
      * @throws NullPointerException if invalidMapper or validMapper is null
      */
-    default <E2, T2> Validation<E2, T2> bimap(Function<? super E, ? extends E2> errorMapper, Function<? super T, ? extends T2> valueMapper) {
+    default <E2, T2> Validation<E2, T2> bimap(Function<Seq<? extends E>, ? extends E2> errorMapper, Function<? super T, ? extends T2> valueMapper) {
```
danieldietrich: contra-variant position:

```java
Function<? super Seq<? superE>, ? extends E2>
```

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -521,11 +530,11 @@ default void forEach(Consumer<? super T> action) {
...
         Objects.requireNonNull(errorMapper, "errorMapper is null");
         Objects.requireNonNull(valueMapper, "valueMapper is null");
         if (isInvalid()) {
-            final E error = this.getError();
+            final Seq<? extends E> error = this.getErrors();
```
danieldietrich: can be simplified

```java
final Seq<E> error = this.getErrors();
```

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -534,43 +543,45 @@ default void forEach(Consumer<? super T> action) {
...
      *
-     * @param <U> type of the error resulting from the mapping
-     * @param f   a function that maps the error in this Invalid
+     * @param <U> type of the errors resulting from the mapping
+     * @param f   a function that maps the errors in this Invalid
      * @return an instance of Validation&lt;U,T&gt;
      * @throws NullPointerException if mapping operation f is null
      */
-    default <U> Validation<U, T> mapError(Function<? super E, ? extends U> f) {
+    default <U> Validation<U, T> mapError(Function<Seq<? extends E>, ? extends U> f) {
```
danieldietrich: contra-variant position:

```java
Function<? super Seq<? super E>, ? extends U>
```

**However**, there is one thing I recognized: We now implicitely have errors of type `Seq<U>` when we have a type `Validation<U, T>`. Isn't it natural then to describe how to map _one_ error? I.e.

```java
default <U> Validation<U, T> mapError(Function<? super E, ? extends U> f) {
    if (isInvalid()) {
        return Validation.invalid(f.apply(errors));
    } else {
        return (Validation<U, T>) this;
    }
}
```

I don't know, **please discuss!**

(maybe it would also affect `fold`, `bimap`, ... ?)

Maybe we need also to change the name: `mapErrors` instead of `mapError`.

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -534,43 +543,45 @@ default void forEach(Consumer<? super T> action) {
...
-            return Validation.invalid(f.apply(error));
+            final Seq<? extends E> errors = this.getErrors();
+            return Validation.invalid(f.apply(errors));
         } else {
             return Validation.valid(this.get());
         }
     }
 
-    default <U> Validation<Seq<E>, U> ap(Validation<Seq<E>, ? extends Function<? super T, ? extends U>> validation) {
+    default <U> Validation<E, U> ap(Validation<E, ? extends Function<? super T, ? extends U>> validation) {
```
danieldietrich: we didn't do it before but maybe it needs to be `ap(Validation<? super E, ...> validation)` in the general case?

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -534,43 +543,45 @@ default void forEach(Consumer<? super T> action) {
...
         Objects.requireNonNull(validation, "validation is null");
         if (isValid()) {
             if (validation.isValid()) {
                 final Function<? super T, ? extends U> f = validation.get();
                 final U u = f.apply(this.get());
                 return valid(u);
             } else {
-                final Seq<E> errors = validation.getError();
+                final Seq<? extends E> errors = validation.getErrors();
```
danieldietrich: can be `Seq<E>`

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -534,43 +543,45 @@ default void forEach(Consumer<? super T> action) {
...
                 return invalid(errors);
             }
         } else {
             if (validation.isValid()) {
-                final E error = this.getError();
-                return invalid(List.of(error));
+                final Seq<? extends E> errors = this.getErrors();
```
danieldietrich: can be `Seq<E>`

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -534,43 +543,45 @@ default void forEach(Consumer<? super T> action) {
...
+                return invalid(errors);
             } else {
-                final Seq<E> errors = validation.getError();
-                final E error = this.getError();
-                return invalid(errors.append(error));
+                final Seq<? extends E> errors1 = validation.getErrors();
+                final Seq<? extends E> errors2 = this.getErrors();
+                Seq<E> specificErrors = List.ofAll(errors1);
+                Seq<E> finalErrors = specificErrors.appendAll(errors2);
```
danieldietrich: specificErrors and finalErrors can be declared `final`. (This is some sort of Vavr convention :) )

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -675,8 +686,8 @@ public T get() {
         }
 
         @Override
-        public E getError() throws RuntimeException {
-            throw new NoSuchElementException("error of 'valid' Validation");
+        public Seq<? extends E> getErrors() throws RuntimeException {
```
danieldietrich: can be

```java
public Seq<E> getErrors()
```

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -704,22 +715,26 @@ public String toString() {
...
-     * @param <E> type of the error of this Validation
+     * @param <E> type of the errors of this Validation
      * @param <T> type of the value of this Validation
      */
     final class Invalid<E, T> implements Validation<E, T>, Serializable {
 
         private static final long serialVersionUID = 1L;
 
-        private final E error;
+        private final Seq<? extends E> errors;
```
danieldietrich: I think it can be `Seq<E> errors`. Maybe it needs to be casted when assigned/at creation time in one of our factory methods.

vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -738,18 +753,18 @@ public T get() throws RuntimeException {
         }
 
         @Override
-        public E getError() {
-            return error;
+        public Seq<? extends E> getErrors() {
```
danieldietrich: can be

```java
public Seq<E> getErrors()
```

vavr/src/main/java/io/vavr/control/Validation.java
amygithub: Hi Daniel,
So this was the main decision I made in this change: that all methods that could be used in one fluent block should have the same return type.
The main problem I had when trying to use Vavr validation for nested fields was that sometimes a single error was returned, and sometimes a sequence of errors. This meant that I as a user had to seriously think about how to combine these nested fields: I ended up creating terribly unreadable code, as the business logic (of methods of validation for each field) was inextricably mixed up in low-level logic like fold, bimap, etc.
I created a very simple test, "shouldCreateSuccessNestedCombine()", to illustrate the kind of nested mapping I wanted to simplify. If validation of a set of fields relies on the content of another field, it's very easy to get lost in the code, if you have to map between multiple return 
types. http://www.baeldung.com/vavr mentions this issue.
An example of nested validation is when a customer selects their company type, (which itself is validated), and depending on this type, we either validate their address, or date of birth.
With the change I made, this becomes trivial to do.

danieldietrich: Hi Amy, your main idea of unifying error to be a sequence is great. I will look into it again today and pull it in.
Because of or work on 1.0.0 currently all is 'in the flow'. Minor subsequent changes may follow...

danieldietrich:
@amygithub Thank you for your contribution and sorry that a took a while. I think you brought validation forward and significantly increased its usability! :)

PR #2076, opened 2017-09-10
Prefer Vector over Stream as default Seq

### LINKED ISSUE #2020: Seq returned by Map.values() should know its own size
The implementations of `Map.values()` all return a `Stream`. This makes operations like `map.values().toJavaSet()` unnecessarily expensive on large maps, since (for reasons I admit I'm unclear on) they require the `size()` of the `Stream`, and this needs to be calculated with `foldLeft`.

profiler snapshot

danieldietrich:
Fixes #2020

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Traversable.java
+++ b/vavr/src/main/java/io/vavr/collection/Traversable.java
@@ -212,19 +212,18 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
         if (isEmpty()) {
             return Option.none();
         } else {
-            final Traversable<?> objects = isTraversableAgain() ? this : toStream();
-            final Object o = objects.head();
-            if (o instanceof Number) {
-                final Traversable<Number> numbers = (Traversable<Number>) objects;
+            final Object head = head();
+            if (head instanceof Number) {
+                final java.util.stream.Stream<Number> numbers = (java.util.stream.Stream<Number>) (
+                        isTraversableAgain() ? this : Iterator.of(head).concat(Iterator.ofAll(this))
+                ).toJavaStream();
                 final double d;
-                if (o instanceof Integer || o instanceof Long || o instanceof Byte || o instanceof BigInteger || o instanceof Short) {
-                    d = numbers.toJavaStream()
-                            .mapToLong(Number::longValue)
+                if (head instanceof Integer || head instanceof Long || head instanceof Byte || head instanceof BigInteger || head instanceof Short) {
+                    d = numbers.mapToLong(Number::longValue)
                             .average()
                             .getAsDouble();
                 } else {
-                    d = numbers.toJavaStream()
-                            .mapToDouble(Number::doubleValue)
+                    d = numbers.mapToDouble(Number::doubleValue)
                             .average()
                             .getAsDouble();
                 }
@@ -798,12 +797,7 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
      */
     @SuppressWarnings("unchecked")
     default Option<T> max() {
-        if (isEmpty()) {
-            return Option.none();
-        } else {
-            final Traversable<T> ts = isTraversableAgain() ? this : toStream();
-            return ts.maxBy(Comparators.naturalComparator());
-        }
+        return maxBy(Comparators.naturalComparator());
     }
 
     /**
@@ -860,12 +854,7 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
      */
     @SuppressWarnings("unchecked")
     default Option<T> min() {
-        if (isEmpty()) {
-            return Option.none();
-        } else {
-            final Traversable<T> ts = isTraversableAgain() ? this : toStream();
-            return ts.minBy(Comparators.naturalComparator());
-        }
+        return minBy(Comparators.naturalComparator());
     }
 
     /**
--- a/vavr/src/main/java/io/vavr/collection/TreeMap.java
+++ b/vavr/src/main/java/io/vavr/collection/TreeMap.java
@@ -1349,7 +1349,7 @@ public final class TreeMap<K, V> implements SortedMap<K, V>, Serializable {
 
     @Override
     public Seq<V> values() {
-        return iterator().map(Tuple2::_2).toVector();
+        return map(Tuple2::_2);
     }
 
     // -- Object
--- a/vavr/src/main/java/io/vavr/collection/TreeMap.java
+++ b/vavr/src/main/java/io/vavr/collection/TreeMap.java
@@ -1349,7 +1349,7 @@ public final class TreeMap<K, V> implements SortedMap<K, V>, Serializable {
 
     @Override
     public Seq<V> values() {
-        return iterator().map(Tuple2::_2).toStream();
+        return iterator().map(Tuple2::_2).toVector();
     }
 
     // -- Object
--- a/vavr/src/test/java/io/vavr/collection/AbstractMapTest.java
+++ b/vavr/src/test/java/io/vavr/collection/AbstractMapTest.java
@@ -500,8 +500,8 @@ public abstract class AbstractMapTest extends AbstractTraversableTest {
 
     @Test
     public void shouldBiMapNonEmpty() {
-        final Seq<Tuple2<Integer, String>> expected = Stream.of(Tuple.of(2, "1!"), Tuple.of(3, "2!"));
-        final Seq<Tuple2<Integer, String>> actual = emptyInt().put(1, "1").put(2, "2").bimap(i -> i + 1, s -> s + "!").toStream();
+        final Seq<Tuple2<Integer, String>> expected = List.of(Tuple.of(2, "1!"), Tuple.of(3, "2!"));
+        final Seq<Tuple2<Integer, String>> actual = emptyInt().put(1, "1").put(2, "2").bimap(i -> i + 1, s -> s + "!").toList();
         assertThat(actual).isEqualTo(expected);
     }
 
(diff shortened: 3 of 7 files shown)


vavr/src/test/java/io/vavr/collection/AbstractMapTest.java
```diff
@@ -500,8 +500,8 @@ public void shouldBiMapEmpty() {
 
     @Test
     public void shouldBiMapNonEmpty() {
-        final Seq<Tuple2<Integer, String>> expected = Stream.of(Tuple.of(2, "1!"), Tuple.of(3, "2!"));
-        final Seq<Tuple2<Integer, String>> actual = emptyInt().put(1, "1").put(2, "2").bimap(i -> i + 1, s -> s + "!").toStream();
+        final Seq<Tuple2<Integer, String>> expected = List.of(Tuple.of(2, "1!"), Tuple.of(3, "2!"));
+        final Seq<Tuple2<Integer, String>> actual = emptyInt().put(1, "1").put(2, "2").bimap(i -> i + 1, s -> s + "!").toList();
```
danieldietrich: does not really matter, but List is most efficient for few values

vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -860,12 +854,7 @@ default T last() {
      */
     @SuppressWarnings("unchecked")
     default Option<T> min() {
-        if (isEmpty()) {
-            return Option.none();
-        } else {
-            final Traversable<T> ts = isTraversableAgain() ? this : toStream();
-            return ts.minBy(Comparators.naturalComparator());
-        }
+        return minBy(Comparators.naturalComparator());
```
danieldietrich: * we do not need to wrap a traversable-once into a Stream because no element is touched before reduction
* minBy API say that the empty case is already covered.

vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -798,12 +797,7 @@ default T last() {
      */
     @SuppressWarnings("unchecked")
     default Option<T> max() {
-        if (isEmpty()) {
-            return Option.none();
-        } else {
-            final Traversable<T> ts = isTraversableAgain() ? this : toStream();
-            return ts.maxBy(Comparators.naturalComparator());
-        }
+        return maxBy(Comparators.naturalComparator());
```
danieldietrich: * we do not need to wrap a traversable-once into a Stream because no element is touched before reduction
* maxBy API say that the empty case is already covered.

vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -212,19 +212,18 @@
...
         } else {
-            final Traversable<?> objects = isTraversableAgain() ? this : toStream();
-            final Object o = objects.head();
-            if (o instanceof Number) {
-                final Traversable<Number> numbers = (Traversable<Number>) objects;
+            final Object head = head();
+            if (head instanceof Number) {
+                final java.util.stream.Stream<Number> numbers = (java.util.stream.Stream<Number>) (
+                        isTraversableAgain() ? this : Iterator.of(head).concat(Iterator.ofAll(this))
+                ).toJavaStream();
```
danieldietrich: it is more efficient to use an Iterator than wrapping every element in a Stream

vavr/src/main/java/io/vavr/collection/TreeMap.java
```diff
@@ -1349,7 +1349,7 @@ public int size() {
 
     @Override
     public Seq<V> values() {
-        return iterator().map(Tuple2::_2).toStream();
+        return iterator().map(Tuple2::_2).toVector();
```
danieldietrich: This catches my eye - the only Map impl that overrides values(). Will double-check, if it is really necessary...

vavr/src/main/java/io/vavr/collection/TreeMap.java
```diff
@@ -1349,7 +1349,7 @@ public int size() {
 
     @Override
     public Seq<V> values() {
-        return iterator().map(Tuple2::_2).toStream();
+        return map(Tuple2::_2);
```
danieldietrich: Yep, could be simplified. Now looks like other *Map.values() impls

nfekete: Wouldn't that give up lazyness of the `Seq<V>` returned?

nfekete: Oh, I see it now. Lazyness is causing problems when querying size of the lazy `Seq`, right?

danieldietrich: Inherently: yes - but it was never intended to be lazy. The user should only rely on the properties of the return value: Seq.

However, I've recognized that the Map.values() Javadoc is missing at all. I will add it.

nfekete: So we're trading creation performance and low memory footprint for read performance here. I realize every decision is a tradeoff, but this one isn't cheap at all. Somehow I feel both could be preserved with a lazy `Seq` type that knows it's size.

nfekete: Java spliterators have some means to query the exact number of elements if the spliterator has the [`SIZED`](https://docs.oracle.com/javase/8/docs/api/java/util/Spliterator.html#SIZED) characteristics. [`Spliterator.estimateSize()`](https://docs.oracle.com/javase/8/docs/api/java/util/Spliterator.html#estimateSize--) will report the exact count of elements if the underlying collection is not modified after spliterator creation.

danieldietrich: This would lead to a general purpose Seq that is both lazy and memory efficient. Currently we do not have such a collection (neither Scala or Clojure have for example).

Lazy collections might be infinite, currently they need to evaluate all elements in O(n) in order to compute the size.

If we create a new lazy Seq that is aware of its size (because we know we use it in a 'finite' context), we will have the problem that Seq operations like get(int) will be O(n). This is because a lazy collection needs to evaluate all elements up to the requested one.

Or am I missing something and there could be a way to be lazy and O(1) at the same time (other than wrapping Vector in a lazy, delegating Seq impl)?

nfekete: I'm just thinking out loud here, but it seemed that we're getting rid of the nice lazy and memory efficient `Stream` based `values()` implementation by replacing it with one that duplicates the whole map because our `size()` implementation is expensive. `Seq` (or rather, `Traversable`) doesn't give any guarrantee on the runtime of it's `size()` implementation, and actually it's javadoc states that the number of elements is _computed_. In the same way, there's no guarrantee over the `get(i)` performance.

From #2020 we know that the `size()` implementation inefficiency causes problems when converting to Java Collections.

These are actually just projections of already existing immutable collections. For the map values or keys we could have a projection of the map that is both efficient in determining it's size and iterating over the elements. It might be solved with a special purpose non-public implementation of `Seq`.

I see another way around this, though I think the previous one would be clearer. `Seq` could be augmented like this: if we would know the size of the `Seq` (in case it has a definite size: `isSizeDefinite()`), we could answer `size()` with O(1). Of course, we cannot be sure all `Seq`s will have definite size (i.e. infinite streams or on-the fly generated ones), but for these simple _derived_ `Seq`s we could have both efficient lazy `Seq` and quick `size()`. This second solution could be useful in other cases as well, but it feels to me like an early optimization (root of all evil :)) and complicates the interface.

danieldietrich: @nfekete I like the idea. For example Stream.Cons could have an internal property `int size` that is -1 if it is unknown.

It should be possible to rewrite the Stream factory methods in the way that they gather the size of the input if possible.

However, when we consider the upcoming Java value types, I would love to see value classes.

```java
sealed interface Stream<T> { ... }

// can this be a singleton / e.g. have a private constructor???
final /*value*/ class Empty<T> extends Stream<T> { ... }

final /*value*/ class Cons<T>(T head, Supplier<Stream<T>> tail) extends Stream<T> { ... }
```

(Will data classes be the same? I think there are different things in the Java pipeline.)

I don't know what is required for the upcoming native Java pattern matching. But a size parameter should not part of the public 'pattern' (aka deconstructor).

danieldietrich: It does not seem to be that simple to add a size field to Stream.Cons. For example consider a finite Stream and we call `filter()`. Because it is lazily evaluated, the size information gets lost.

Maybe it is better to suppress the size() call during type conversion.

Therefore I suggest to revert my changes again (`toVector()` vs. `toStream()`) and suppress size calculation for lazy Traversables like Stream when converting them to Java collections.

The change of Value.java:

```diff
     static <T extends java.util.Collection<V>, V> T toJavaCollection(
             Value<V> value, Function<Integer, T> containerSupplier) {
-        final int size = (value instanceof Traversable) && ((Traversable) value).isTraversableAgain()
-                ? ((Traversable<V>) value).size()
-                : 16;
+        final int size;
+        if (value.isSingleValued()) {
+            size = 1;
+        } else if (value instanceof Traversable && ((Traversable) value).isTraversableAgain() && !value.isLazy()) {
+            size = ((Traversable) value).size();
+        } else {
+            size = 16;
+        }
         final T container = containerSupplier.apply(size);
         value.forEach(container::add);
         return container;
    }
```

This will save us one collection traversal. Another one of course is still needed to add all elements to the target collection.

What do you think @nfekete @chronodm @ruslansennov

danieldietrich:
I split this PR into #2078 and #2079 and will close this now.

vavr/src/main/java/io/vavr/collection/TreeMap.java
nfekete: I like this solution, since it won't sacrifice the low cost properties of those derived `Seq`s, while making conversion to java collections efficient.

PR #2077, opened 2017-09-10
Added Either.sequenceRight

### LINKED ISSUE #2074: Add monadic version of Either.sequence
We labeled this issue as 'bug' and not as 'feature' because it is a follow-up of #2064. There we added an _applicative_ `Either<Seq<L>, Seq<R>> sequence(...)`.

But it might come handy to have a _monadic_ **`Either<L, Seq<R>> sequenceM(...)`**, too (see this [issue comment](https://github.com/vavr-io/vavr/issues/2064#issuecomment-327717610)).

danieldietrich:
Fixes #2074

### DIFF
--- a/vavr/src/main/java/io/vavr/control/Either.java
+++ b/vavr/src/main/java/io/vavr/control/Either.java
@@ -22,6 +22,7 @@ package io.vavr.control;
 import io.vavr.Value;
 import io.vavr.collection.Iterator;
 import io.vavr.collection.Seq;
+import io.vavr.collection.Vector;
 
 import java.io.Serializable;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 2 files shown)


vavr/src/main/java/io/vavr/control/Either.java
```diff
@@ -215,6 +216,46 @@
...
+     * @throws NullPointerException if {@code eithers} is null
+     */
+    static <L,R> Either<L, Seq<R>> sequenceRight(Iterable<? extends Either<? extends L, ? extends R>> eithers) {
+        Objects.requireNonNull(eithers, "eithers is null");
+        Vector<R> rightValues = Vector.empty();
+        for (Either<? extends L, ? extends R> either : eithers) {
+            if (either.isRight()) {
+                rightValues = rightValues.append(either.get());
+            } else {
+                return Either.left(either.getLeft());
```
danieldietrich: Internally we prefer an imperative impl over the nice and clean foldLeft because we are able to shortcut the computation as soon as we meet a left value. This saves CPU cycles and mem.

PR #2078, opened 2017-09-10
Simplified numeric Traversable methods.

danieldietrich:
* Fixed: Traversable.min() is NaN if element type is Float/Double and the traversable contains NaN
* min() and max() now are computed according to the natural comparator, especially when the collection is inherently sorted by an underlying comparator
* Implemented Neumaier's modification of the Kahan summation algorithm

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Traversable.java
+++ b/vavr/src/main/java/io/vavr/collection/Traversable.java
@@ -25,17 +25,11 @@ import io.vavr.Tuple3;
 import io.vavr.Value;
 import io.vavr.control.Option;
 
+import java.math.BigDecimal;
 import java.math.BigInteger;
-import java.util.Comparator;
-import java.util.NoSuchElementException;
-import java.util.Objects;
-import java.util.Spliterator;
-import java.util.Spliterators;
-import java.util.function.BiFunction;
-import java.util.function.Consumer;
-import java.util.function.Function;
-import java.util.function.Predicate;
-import java.util.function.Supplier;
+import java.util.*;
+import java.util.function.*;
+import java.util.stream.DoubleStream;
 
 /**
  * An interface for inherently recursive, multi-valued data structures. The order of elements is determined by
@@ -190,51 +184,48 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
     }
 
     /**
-     * Calculates the average of this elements. Returns {@code None} if this is empty, otherwise {@code Some(average)}.
-     * Supported component types are {@code Byte}, {@code Double}, {@code Float}, {@code Integer}, {@code Long},
-     * {@code Short}, {@code BigInteger} and {@code BigDecimal}.
+     * Calculates the average of this elements, assuming that the element type is {@link Number}.
+     *
+     * Since we do not know if the component type {@code T} is of type {@code Number}, the
+     * {@code average()} call might throw at runtime (see examples below).
      * <p>
      * Examples:
-     * <pre>
-     * <code>
-     * List.empty().average()              // = None
-     * List.of(1, 2, 3).average()          // = Some(2.0)
-     * List.of(0.1, 0.2, 0.3).average()    // = Some(0.2)
-     * List.of("apple", "pear").average()  // throws
-     * </code>
-     * </pre>
+     *
+     * <pre>{@code
+     * List.empty().average()                       // = None
+     * List.of(1, 2, 3).average()                   // = Some(2.0)
+     * List.of(1.0, 10e100, 2.0, -10e100).average() // = Some(0.75)
+     * List.of(1.0, Double.NaN).average()           // = NaN
+     * List.of("apple", "pear").average()           // throws
+     * }</pre>
+     *
+     * Please note that Java's {@link DoubleStream#average()} uses the
+     * <a href="https://en.wikipedia.org/wiki/Kahan_summation_algorithm">Kahan summation algorithm</a>
+     * (also known as compensated summation), which has known limitations.
+     * <p>
+     * Vavr uses Neumaier's modification of the Kahan algorithm, which yields better results.
+     *
+     * <pre>{@code
+     * // = OptionalDouble(0.0) (wrong)
+     * j.u.s.DoubleStream.of(1.0, 10e100, 2.0, -10e100).average()
+     *
+     * // = Some(0.75) (correct)
+     * List.of(1.0, 10e100, 2.0, -10e100).average()
+     * }</pre>
      *
      * @return {@code Some(average)} or {@code None}, if there are no elements
      * @throws UnsupportedOperationException if this elements are not numeric
      */
-    @SuppressWarnings({ "unchecked", "OptionalGetWithoutIsPresent" })
     default Option<Double> average() {
-        if (isEmpty()) {
-            return Option.none();
-        } else {
-            final Traversable<?> objects = isTraversableAgain() ? this : toStream();
-            final Object o = objects.head();
-            if (o instanceof Number) {
-                final Traversable<Number> numbers = (Traversable<Number>) objects;
-                final double d;
-                if (o instanceof Integer || o instanceof Long || o instanceof Byte || o instanceof BigInteger || o instanceof Short) {
-                    d = numbers.toJavaStream()
-                            .mapToLong(Number::longValue)
-                            .average()
-                            .getAsDouble();
-                } else {
-                    d = numbers.toJavaStream()
-                            .mapToDouble(Number::doubleValue)
-                            .average()
-                            .getAsDouble();
-                }
-                return Option.some(d);
-            } else {
-                throw new UnsupportedOperationException("not numeric");
-            }
+        try {
+            final double[] sum = TraversableModule.neumaierSum(this, t -> ((Number) t).doubleValue());
+            final double count = sum[1];
+            return (count == 0) ? Option.none() : Option.some(sum[0] / count);
+        } catch(ClassCastException x) {
+            throw new UnsupportedOperationException("not numeric", x);
         }
     }
-
+    
     /**
      * Collects all elements that are in the domain of the given {@code partialFunction} by mapping the elements to type {@code R}.
      * <p>
@@ -790,20 +781,26 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
     <U> Traversable<U> map(Function<? super T, ? extends U> mapper);
 
     /**
-     * Calculates the maximum of this elements according to their natural order.
+     * Calculates the maximum of this elements according to their natural order. Especially the underlying
+     * order of sorted collections is not taken into account.
+     * <p>
+     * Examples:
+     * <pre>
+     * <code>
+     * List.empty().max()             // = None
+     * List.of(1, 2, 3).max()         // = Some(3)
+     * List.of("a", "b", "c").max()   // = Some("c")
+     * List.of(1.0, Double.NaN).max() // = NaN
+     * List.of(1, "a").max()          // throws
+     * </code>
+     * </pre>
      *
      * @return {@code Some(maximum)} of this elements or {@code None} if this is empty
      * @throws NullPointerException if an element is null
      * @throws ClassCastException   if the elements do not have a natural order, i.e. they do not implement Comparable
      */
-    @SuppressWarnings("unchecked")
     default Option<T> max() {
-        if (isEmpty()) {
-            return Option.none();
-        } else {
-            final Traversable<T> ts = isTraversableAgain() ? this : toStream();
-            return ts.maxBy(Comparators.naturalComparator());
-        }
+        return maxBy(Comparators.naturalComparator());
     }
 
     /**
@@ -852,7 +849,23 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
     }
 
     /**
-     * Calculates the minimum of this elements according to their natural order.
+     * Calculates the minimum of this elements according to their natural order in O(n). Especially the underlying
+     * order of sorted collections is not taken into account.
+     * <p>
+     * Examples:
+     * <pre>
+     * <code>
+     * List.empty().min()             // = None
+     * List.of(1, 2, 3).min()         // = Some(1)
+     * List.of("a", "b", "c").min()   // = Some("a")
+     * List.of(1.0, Double.NaN).min() // = NaN
+     * List.of(1, "a").min()          // throws
+     * </code>
+     * </pre>
+     *
+     * There is an exception for {@link Double} and {@link Float}: The minimum is defined to be {@code NaN} if
+     * this contains {@code NaN}. According to the natural order {@code NaN} would be the maximum element
+     * instead.
      *
      * @return {@code Some(minimum)} of this elements or {@code None} if this is empty
      * @throws NullPointerException if an element is null
@@ -860,11 +873,21 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
      */
     @SuppressWarnings("unchecked")
     default Option<T> min() {
+        // DEV-NOTE: minBy(Comparators.naturalComparator()) does not handle (Double/Float) NaN correctly
         if (isEmpty()) {
             return Option.none();
         } else {
-            final Traversable<T> ts = isTraversableAgain() ? this : toStream();
-            return ts.minBy(Comparators.naturalComparator());
+            final T head = head();
+            final T min;
+            if (head instanceof Double) {
+                min = (T) ((Traversable<Double>) this).foldLeft((Double) head, Math::min);
+            } else if (head instanceof Float) {
+                min = (T) ((Traversable<Float>) this).foldLeft((Float) head, Math::min);
+            } else {
+                final Comparator<T> comparator = Comparators.naturalComparator();
+                min = this.foldLeft(head, (t1, t2) -> comparator.compare(t1, t2) <= 0 ? t1 : t2);
+            }
+            return Option.some(min);
         }
     }
 
@@ -1043,6 +1066,8 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
      * </code>
      * </pre>
      *
+     * Please also see {@link #fold(Object, BiFunction)}, a way to do a type-safe multiplication of elements.
+     *
      * @return a {@code Number} representing the sum of this elements
      * @throws UnsupportedOperationException if this elements are not numeric
      */
@@ -1051,18 +1076,20 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
         if (isEmpty()) {
             return 1;
         } else {
-            final Iterator<?> iter = iterator();
-            final Object o = iter.next();
-            if (o instanceof Number) {
-                final Number head = (Number) o;
-                final Iterator<Number> numbers = (Iterator<Number>) iter;
-                if (head instanceof Integer || head instanceof Long || head instanceof Byte || head instanceof BigInteger || head instanceof Short) {
-                    return numbers.toJavaStream().mapToLong(Number::longValue).reduce(head.longValue(), (l1, l2) -> l1 * l2);
+            try {
+                final Iterator<?> iter = iterator();
+                final Object o = iter.next();
+                if (o instanceof Integer || o instanceof Long || o instanceof Byte || o instanceof Short) {
+                    return ((Iterator<Number>) iter).foldLeft(((Number) o).longValue(), (product, number) -> product * number.longValue());
+                } else if (o instanceof BigInteger) {
+                    return ((Iterator<BigInteger>) iter).foldLeft(((BigInteger) o), BigInteger::multiply);
+                } else if (o instanceof BigDecimal) {
+                    return ((Iterator<BigDecimal>) iter).foldLeft(((BigDecimal) o), BigDecimal::multiply);
                 } else {
-                    return numbers.toJavaStream().mapToDouble(Number::doubleValue).reduce(head.doubleValue(), (d1, d2) -> d1 * d2);
+                    return ((Iterator<Number>) iter).toJavaStream().mapToDouble(Number::doubleValue).reduce(((Number) o).doubleValue(), (d1, d2) -> d1 * d2);
                 }
-            } else {
-                throw new UnsupportedOperationException("not numeric");
+            } catch(ClassCastException x) {
+                throw new UnsupportedOperationException("not numeric", x);
             }
         }
     }
@@ -1319,7 +1346,7 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
             return Spliterators.spliteratorUnknownSize(iterator(), characteristics);
         }
     }
-    
+
     /**
      * Calculates the sum of this elements. Supported component types are {@code Byte}, {@code Double}, {@code Float},
      * {@code Integer}, {@code Long}, {@code Short}, {@code BigInteger} and {@code BigDecimal}.
@@ -1334,6 +1361,8 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
      * </code>
      * </pre>
      *
+     * Please also see {@link #fold(Object, BiFunction)}, a way to do a type-safe summation of elements.
+     *
      * @return a {@code Number} representing the sum of this elements
      * @throws UnsupportedOperationException if this elements are not numeric
      */
@@ -1342,22 +1371,24 @@ public interface Traversable<T> extends Foldable<T>, Value<T> {
         if (isEmpty()) {
             return 0;
         } else {
-            final Iterator<?> iter = iterator();
-            final Object o = iter.next();
-            if (o instanceof Number) {
-                final Number head = (Number) o;
-                final Iterator<Number> numbers = (Iterator<Number>) iter;
-                if (head instanceof Integer || head instanceof Long || head instanceof Byte || head instanceof BigInteger || head instanceof Short) {
-                    return numbers.foldLeft(head.longValue(), (n1, n2) -> n1 + n2.longValue());
+            try {
+                final Iterator<?> iter = iterator();
+                final Object o = iter.next();
+                if (o instanceof Integer || o instanceof Long || o instanceof Byte || o instanceof Short) {
+                    return ((Iterator<Number>) iter).foldLeft(((Number) o).longValue(), (sum, number) -> sum + number.longValue());
+                } else if (o instanceof BigInteger) {
+                    return ((Iterator<BigInteger>) iter).foldLeft(((BigInteger) o), BigInteger::add);
+                } else if (o instanceof BigDecimal) {
+                    return ((Iterator<BigDecimal>) iter).foldLeft(((BigDecimal) o), BigDecimal::add);
                 } else {
-                    return numbers.foldLeft(head.doubleValue(), (n1, n2) -> n1 + n2.doubleValue());
+                    return TraversableModule.neumaierSum(Iterator.of(o).concat(iter), t -> ((Number) t).doubleValue())[0];
                 }
-            } else {
-                throw new UnsupportedOperationException("not numeric");
+            } catch(ClassCastException x) {
+                throw new UnsupportedOperationException("not numeric", x);
             }
         }
     }
-
+    
     /**
      * Drops the first element of a non-empty Traversable.
      *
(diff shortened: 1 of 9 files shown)


vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -212,23 +212,25 @@
...
-                    d = numbers.toJavaStream()
-                            .mapToDouble(Number::doubleValue)
-                            .average()
-                            .getAsDouble();
+                    avg = numbers.mapToDouble(Number::doubleValue).average().getAsDouble();
                 }
-                return Option.some(d);
+                return Option.some(avg);
             } else {
                 throw new UnsupportedOperationException("not numeric");
```
chb0github: Better if to check for this condition first (assert precondition) by inverting the predicate `if(!(head instanceof Number)) throw new UnsupportedOperationException)`

Then, you can remove the `else` altogether and reduce complexity and indentation

vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -212,23 +212,25 @@
...
+                head = iterator.next();
+                elements = Iterator.of(head).concat(iterator).toJavaStream();
+            }
+            if (head instanceof Number) {
+                final java.util.stream.Stream<Number> numbers = (java.util.stream.Stream<Number>) elements;
+                final double avg;
+                if (head instanceof Integer || head instanceof Long || head instanceof Byte || head instanceof BigInteger || head instanceof Short) {
+                    avg = numbers.mapToLong(Number::longValue).average().getAsDouble();
                 } else {
-                    d = numbers.toJavaStream()
```
chb0github: An alternative approach would be to define the translation strategy instead of having a long predicate like this. 

Option 1:
```
// top of class
private static Set<Class> longable = Set.of(Integer.class,Long.class,Byte.class,BigInteger.class, Short.class);
if(longable.contains(head))
 avg = numbers.maptToLong(Number::longValue).average().getAsDouble();
```

Option 2: expressly define the function by type:
```
Map<Class<? extends Number>,Function<Stream<Number,? extends Number>> funcs = Map.of(Integer.class,Number::longValue,Byte.class,Number::longValue,Double.class,Number::doubleValue)

Function<Stream<Number>,Stream<Number>> f = funcs.get(head.getClass()).orElseThrow(RuntimeException::new);
number = f.apply(numbers);
```

Option 3 is to recognize the you can handle the simpler case more tersely:
`if(head instanceof Double || head instance of Float)`

vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -212,23 +212,25 @@
         if (isEmpty()) {
             return Option.none();
         } else {
```
chb0github: this `else` is unnecessary. Once you have done `return Option.none();` the only way the rest of the code executes is as the `else`. Remove the else -> remove indent -> reduce complexity

danieldietrich: There was discussion about 'guards' before. We decided to

* Do method parameter checks as guards (, like you suggested)
* Do control flow of the method body the 'functional way' as if the `if` statement where an expression, like in java. Logically the two cases belong to each other. Flattening a method is the imperative way. Some information gets lost. I know that some say the the complexity increases. For me, the readability decreases.

Example of your suggestions below:

screen shot 2017-09-12 at 20 02 48

vavr/src/main/java/io/vavr/collection/Traversable.java
danieldietrich: The if check isn't a problem. Here is a similar Scala code, taken from [NonFatal.scala](https://github.com/scala/scala/blob/v2.12.1/src/library/scala/util/control/NonFatal.scala):

screen shot 2017-09-12 at 20 14 14

Instanceof calls are cheap and take no additional memory like the Map and Function instances you suggested. This is code that is not needed. An if is at the lowest level of the language. I tried to sort the checks in the way that most common primitive types occur first. (However, I have no statistic - it is subjective).

vavr/src/main/java/io/vavr/collection/Traversable.java
chb0github: Imperative isn't a crime OR a dirty word. I noticed there is a cyclomatic complexity analyzer on the build. Reducing that number seems an unalloyed improvement.

vavr/src/main/java/io/vavr/collection/Traversable.java
chb0github: Sure. I am not opposed to the `instanceof` checks (per-se). I am saying 

```
if(head instanceof Double || head instanceof Float) {
    avg = numbers.mapToDouble(Number::doubleValue).average().getAsDouble();
else
   avg = number.mapToLong(Number::longValue).average().getAsDouble();
``
exact same outcome, less code.

Comment (danieldietrich) [reply vavr/src/main/java/io/vavr/collection/Traversable.java]: This looks great, sorry - I didn't recognized it 😅

File: vavr/src/main/java/io/vavr/collection/Traversable.java (hunk 3 above)
Comment (danieldietrich) [reply vavr/src/main/java/io/vavr/collection/Traversable.java]: Thank you for your feedback! Sometimes keeping the order is more important than a number. E.g. imagine we flatten two scoped blocks to one. No-one hinders someone to change the order of the statements later.

Example:

```java
if (condition1) {
    do();
    this();
    and();
    that();
} else {
    throw x;
}
if (condition2) {
    flip();
    flap();
} else {
    throw y;
}
```

The following might be a valid substitute:

```java
if (!condition1) {
    throw x;
}
if (!condition2) {
    throw y;
}
do();
flip();
this();
and();
flap();
that();
```

But some important information got lost. Do you see what I mean?

The cyclomatic complexity analyzer is 'stupid'. It can't look through humans eyes.

vavr/src/main/java/io/vavr/collection/Traversable.java
danieldietrich: @chb0github I thought about the code - I intentionally not checked

```java
if (head instanceof Double || head instanceof Float)
```

because we support all elements that extend Number. For example BigDecimal should be mapped to double instead of long.

However, I'm aware that some types can't be calculated accurately in the double space. I will mention that in the javadoc.

vavr/src/main/java/io/vavr/collection/Traversable.java
danieldietrich: @chb0github you points are valid. I made some tests and looked in the java.util.stream.DoubleStream.average() algorithm. It is accurate, I think we do not need to convert elements to long.

This leads to the following impl:

```java
    default Option<Double> average() {
        if (isEmpty()) {
            return Option.none();
        }
        try {
            final java.util.stream.Stream<Number> numbers = (java.util.stream.Stream<Number>) toJavaStream();
            final double avg = numbers.mapToDouble(Number::doubleValue).average().getAsDouble();
            return Option.some(avg);
        } catch (ClassCastException x) {
            throw new UnsupportedOperationException("not numeric");
        }
    }
```

chb0github:
There are some limitations with this approach that don't seem to be compatible with the vavr _ethos_ or purity. Functional programming is derived from mathematical concepts and ideas. With that in mind, this code has some gotchas

1) It's not type safe and relies on runtime validation.
2) Long ago I had a discussion with core Java people about why `1L` and `1.1f` should not be comparable at the number level. That is "Why is Number not Comparable?" The reason being that there can be loss of precisions and Java has to mechanisms for things like imaginary numbers. I admit, I hadn't thought of it. So, Every current core implementation of Number is Comparable but not number itself. 

What that gets to is that your functions doesn't handle anything you didn't implement now. What about `AtomicLong` ? With the current logic you would apply the `Number::doubleValue` but based on the the fact that Integral types (Long, Int, Byte, Short) you apply `Number::longValue` I assume that you would intend `Number::longValue` to it as well; but that won't happen. 

I admit, your code handles the common types and that's probably fine* - perhaps the real solution would be to have an overloaded function that takes function that does the proper conversion?

```
<N extends Number, S extends Number> public T avg(Function<? extends N,? super S> transformFun); 

Long result = List.of(1,2,3).avg(Number::longValue);
Long result2 = List.of(1,2,3).avg(); // uses logic currently
```

I think, ideally, if these useful methods were to be implemented, then List should get a couple of extra factory methods where in these methods are defined and they are defined properly:

```
IntList ints = List.of(1,2,3);
```
```
private static class IntList extends List {
     public Integer avg() {
        // implement avg specifically or have a 'NumericList' that takes a Function in its constructor
        // That is applied to every value. 
        // Then have this class sub class it and pass the correct function. Averaging a series of numbers is identical functionally
     }
}
```

```
public class List {
    
   public static IntList of(Integer ... nums) {
       
   }
}
```

This approaches allows the compiler to do what it does best and conforms to polymorphic expectations. I like the simplication and I think we can actually get the compiler onboard too

danieldietrich:
@chb0github `IntList` is no option. An upcoming Java version will support primitive generics, e.g. List&lt;int&gt;. Our strategy is therefore to not provide specializations. (Maybe we will also get reified generics, so that we are able to gather the component type at runtime...)

Our average() returns a double. Integrals are converted lossless to double, e.g. 123 -> 123.0. The [Kahan summation](https://en.wikipedia.org/wiki/Kahan_summation_algorithm) algorithm (also used by Java's DoubleStream.average()) is very accurate.

However, you nerd-sniped me. I currently experiment with Neumaier's improvement of Kahan's algorithm. Maybe we can be even more accurate that Java :)

**Example:** `[1.0, +10e100, 1.0, -10e100].average()`

* Kahan (Java) returns 0.0 (false)
* Neumaier returns 1.0 (true)

danieldietrich:
@chb0github 

> There are some limitations with this approach that don't seem to be compatible with the vavr ethos or purity.

That's not correct. Of course we internally are free to choose the best implement possible, especially when 'impure' or 'non-functional'. This does not matter at all.

For example we support GWT. By removing the instanceof checks the average() method can be used now from Javascript. That's an improvement.

Don't be stigmatic about FP. Instead be practicable.

vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -191,47 +188,62 @@
...
-                            .mapToDouble(Number::doubleValue)
-                            .average()
-                            .getAsDouble();
+                    return Option.some(avg);
                 }
-                return Option.some(d);
-            } else {
-                throw new UnsupportedOperationException("not numeric");
             }
+        } catch (ClassCastException x) {
```
chb0github: You should actualy include `x.getMessage()` as this will actually contain a message saying what type it actually had and tried to convert to.

chb0github:
The real crux of the matter is I don't like pushing off to runtime that which can be asserted at compile-time. Hence the specific subtypes for lists. `List.of("Christian","Bongiorno").avg()` fails. And yet, the compiler could have told you this! I don't know when these magical primitive generics will come out (Incase you hadn't been watching, Java 9 is on it's 10th? Delay? Current "we really think it will work" date for java 9 is Sept 22) but whenever it is it's far far into the future. 

I strongly encourage the subtype approach because the compiler catch a typing mistake right away we won't have to hold out until the 7th-of-never for a special java feature.

vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -191,47 +188,62 @@
...
-                            .average()
-                            .getAsDouble();
+        try {
+            @SuppressWarnings("unchecked")
+            final double[] sum = ((Traversable<Number>) this).map(Number::doubleValue)
+                    .foldLeft(new double[4], (acc, d) -> { // acc = [sum, c, count, simpleSum]
+                        final double tmp = acc[0] + d;
+                        acc[1] += (Math.abs(acc[0]) >= Math.abs(d)) ? (acc[0] - tmp) + d : (d - tmp) + acc[0];
+                        acc[0] = tmp;
+                        acc[2]++;
```
chb0github: YIKES!  A side-effect! `acc[2] = acc[2] + 1`

vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -191,47 +188,62 @@
...
+                        acc[3] += d;
+                        return acc;
+                    });
+            final double count = sum[2];
+            if (count == 0) {
+                return Option.none();
+            } else {
+                final double avg = (sum[0] + sum[1]) / count;
```
chb0github: Not that I terribly care, but why declare everything `final`? Are you trying have the scala equivalent of `val` ?

vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -791,19 +803,23 @@ default T last() {
...
      * Calculates the maximum of this elements according to their natural order.
-     *
+     * <p>
+     * Examples:
+     * <pre>
+     * <code>
+     * List.empty().max()            // = None
+     * List.of(1, 2, 3).max()        // = Some(3)
+     * List.of("a", "b", "c").max()  // = Some("c")
+     * List.of(1, "a").max()         // throws
```
chb0github: Technically, this case shouldn't actually compile.

vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -853,19 +869,23 @@ default T last() {
...
     /**
      * Calculates the minimum of this elements according to their natural order.
+     * <p>
+     * Examples:
+     * <pre>
+     * <code>
+     * List.empty().min()            // = None
+     * List.of(1, 2, 3).min()        // = Some(1)
+     * List.of("a", "b", "c").min()  // = Some("a")
+     * List.of(1, "a").min()         // throws
```
chb0github: Shouldn't compile

vavr/src/main/java/io/vavr/collection/Traversable.java
chb0github: Honestly, toss the catch statement. You're trading one exception for another and the `ClassCastException` says everything you need to know

chb0github: Your conditional collapse substitution is not identical to the original sample.

This would be the equivalence:

```
if (condition1) {
    do;
    this;
    and;
    that;
} else {
    throw x;
}
if (condition2) {
    flip;
    flap;
} else {
    throw y;
}
```

```
if (!condition1) {
   throw x;
}
do;
this;
and;
that;

if (!condition2) {
    throw y;
}
flip;
flap;
```

And, actually, I don't see what got lost except extraneous syntax.

vavr/src/main/java/io/vavr/collection/Traversable.java
danieldietrich: Thx for the suggestion, II've also thought about doing it. I agree that it is a good thing to be specific in order to make it easy to find the error reason. I decided not to return the cause, because the cause it is an implementation detail. But you are right, if we have the information then let's do it.

There is still a (low) risk to break applications that use Vavr in the case we decide to change the implementation in future and the exception cause disappears. Users should not rely on the cause. Our API spec tells that the method throws an UnsupportedOperationException.

danieldietrich:
@chb0github regarding your comment, here  a fragment

> A List.of(1L,2L).avg() should produce a Long result. I am not against what your'e doing (infact, I am in favor of simplification) but the typing issues is a very practical concern.

Java's `LongStream` has `OptionalDouble average()`, e.g. the average of long values is a double.

I agree that boxing takes extra time, but that is the way Java works today.

It was a **design decision** to go the unsafe track. We traded safety for simplicity. This is a common way in Java because we can't express it with the type system (except we duplicate the types).

I also like safety, so what led to this decision? 

* Java has `Stream`, `IntStream`, `DoubleStream`, `LongStream`. Me the moment I saw this: 😱
* It is common practice, [e.g. java.util.TreeSet.add(Object)](https://docs.oracle.com/javase/8/docs/api/java/util/TreeSet.html#add-E-) may throw a ClassCastException

vavr/src/main/java/io/vavr/collection/Traversable.java
danieldietrich: 😄 don't expect the internal implementation of functional libraries to be side-effect free.

In fact I advocate that mutable state is perfectly ok. In the ideal case it is in a local scope. _Shared_ mutable state is evil!

For example our [Future](https://github.com/vavr-io/vavr/blob/1e6af93c618198aa722714226a682eca13132d64/vavr/src/main/java/io/vavr/concurrent/FutureImpl.java#L84-L104), [Lazy](https://github.com/vavr-io/vavr/blob/1e6af93c618198aa722714226a682eca13132d64/vavr/src/main/java/io/vavr/Lazy.java#L65-L67) and [Iterator](https://github.com/vavr-io/vavr/blob/1e6af93c618198aa722714226a682eca13132d64/vavr/src/main/java/io/vavr/collection/Iterator.java#L46-L48) implementations would not work at all without internal mutable state!

vavr/src/main/java/io/vavr/collection/Traversable.java
danieldietrich: Yes, this is specific for our internal Vavr implementation (and was part of our contribution guidelines I think but I can't find it anymore).

The only thing that can be remarked about it is that it is verbose. But in my experience programs get more robust.

The argument against it I hear most often is: "But the object of this variable isn't immutable, I can still mutate it." - and I say: "Yes, but the semantics of final is different, the reference to that object is immutable."

Long story short, it leads to more robust programs because it enforces the developer to not use variables if possible.

It does not mean to not use mutable variables at all.

vavr/src/main/java/io/vavr/collection/Traversable.java
danieldietrich: It does...
screen shot 2017-09-14 at 09 10 03

vavr/src/main/java/io/vavr/collection/Traversable.java
danieldietrich: dito

vavr/src/main/java/io/vavr/collection/Traversable.java
danieldietrich: ClassCastEcxeption is an implementation detail. It might change.

danieldietrich: The rule we follow:

1. If we check an argument we do a guard (most often Objects.requireNonNull(), but also if-guards)
2. After checking the arguments, we use if-else

```java
// THE WAY WE DO IT
default Option<T> maxBy(Comparator<? super T> comparator) {
    Objects.requireNonNull(comparator, "comparator is null");
    if (isEmpty()) {
        return Option.none();
    } else {
        final T value = reduce((t1, t2) -> comparator.compare(t1, t2) >= 0 ? t1 : t2);
        return Option.some(value);
    }
}

// NOT ACCORDING TO OUR CODING GUIDELINE
default Option<T> maxBy(Comparator<? super T> comparator) {
    Objects.requireNonNull(comparator, "comparator is null");
    if (isEmpty()) {
        return Option.none();
    }
    final T value = reduce((t1, t2) -> comparator.compare(t1, t2) >= 0 ? t1 : t2);
    return Option.some(value);
}
```

We justify the guideline mentioned above by keeping things together with if-else which logically belog together. I.e. the two-sided conditional is a composition of a fork in our code.

If we have more than one of these logical units (read: if-else) we also benefit from scoping of variables that are needed only within the branches. Otherwise we would expose these variables to contexts which should not know about them.

Summing up, if-else helps us to structure our code and making it more robust.

vavr/src/main/java/io/vavr/collection/Traversable.java
chb0github: I was just being _dogmatic_ for sake of humor

vavr/src/main/java/io/vavr/collection/Traversable.java
chb0github: Ok, that's puzzling then - why choose serializable and comparable? That will take us off topic

chb0github:
re: Java streams: I literally felt the same way and for the same reasons! And, worse, have you looked at the inheritance tree for streams? I am really hoping that Java 9 brings about something close to a total library rewrite!

chb0github:
Well, the discussions seems to be between you and I only - how about this compromise: Overload the `avg` function to take a function that can do the conversion. This allows users to override the default conversion logic should it prove necessary. This keeps the convenience in place for the people who just don't give a damn but allows for full-control to be applied if needed. The implementation is trivial.

danieldietrich:
@chb0github 

> Overload the avg function to take a function that can do the conversion.

I'm not convinced. How should that function look like? Like this (simplified version, without generic bounds):

```java
Option<Double> avg(Function<Traversable<T>, Double> strategy);
```

That version is unnecessary, it is the same as

```java
strategy.apply(traversable);
```

chb0github:
The function would be on `Traversable` as I put previously. 

`default Option<Double> average(Function<T, Double> doubleFun)`

then use it as:

`List.of(1,2,3).average(n -> Math.round(n * 10) / 10.0);`

but keep what you have now with:

`List.of(1,2,3).average();` 

The implementation could be moved to the new method and the current one would just pass `Number::doubleValue`

danieldietrich:
@chb0github I see. You could achieve almost the same by calling

```java
List.of(1,2,3).map(n -> Math.round(n * 10) / 10.0).avg();
```

But most importantly it does not help users from accidentally calling the unsafe avg(). We will keep the API as-is but I understand your remarks.

> I gave you incentive to scratch and itch :)

I like it to scratch and itch :) 😜

danieldietrich:
@chb0github I'm still testing. I managed to create a confusing exception:

```java
// throws UnsupportedOperationException: not numeric
PriorityQueue.of(1.0d, 1.0f).average();
```

The average should be calculated and the result should be 1.0d, but the PriorityQueue uses the natural Comparator by default, which is not able to compare doubles and floats.

As you said, the exception message (without stack trace) does not really help here...

chb0github:
@danieldietrich Your `Traversable` type should not allow for mixed types for precisely reasons like this. It should just take `T` and let the user decide what the highest level supportable type is. 

For this case it's unavoidable and insurmountable. Although much of the Java APIs makes me cringe, the type strictness isn't one of them.

danieldietrich:
@chb0github 

> Your Traversable type should not allow for mixed types

That is not as easy as it sounds. If T (e.g. inferred type of a static factory method call) is Object, all and is possible...

Also we are not able to allow only Comparable types for our sortable collections, because it would break the type hierarchy. They could not extend Set / Map any more.

Unfortunately Java's type system is not expressible enough for our uses-cases.

nfekete:
> `PriorityQueue.of(1.0d, 1.0f).average();`
I can't even make that compile with Eclipse w/ Java8 source level. What does IntelliJ infer there?

danieldietrich:
@nfekete Actually I use `this.<Number> of(1.0, 1.0f).min()` in the unit tests (class: AbstractTraversableTest). That is the reason it works.

```java
// compiles
List.of(1.0d, 1.0f).average()

// compiles
List.of(1.0d, 1.0f).toSortedSet().average();

// does not compile (with IntelliJ IDEA)
PriorityQueue.of(1.0d, 1.0f).average();
```

danieldietrich:
Something like this:

```java
    class RealAwareComparator<T> implements Comparator<T>, Serializable {

        Comparator<T> comparator;

        @Override
        public int compare(T t1, T t2) {
            if (comparator == null) {
                // DEV-NOTE: null will lead to a NPE in each comparator anyway
                final T t = (t1 == null) ? t1 : t2;
                if (t instanceof Double) {
                    comparator = (Comparator<T>) (Comparator<Double> & Serializable) Double::compareTo;
                } else if (t instanceof Float) {
                    comparator = (Comparator<T>) (Comparator<Float> & Serializable) Float::compareTo;
                } else {
                    comparator = Comparators.naturalComparator();
                }
            }
            return comparator.compare(t1, t2);
        }

        @Override
        public boolean equals(Object obj) {
            // DEV-NOTE: it can't be (obj instanceof RealAwareComparator) because this is stateful
            return this == obj;
        }
    }
```

danieldietrich:
Maybe we have to live with that paradoxum;

```java
// = 1.0
TreeSet(1.0, NaN).head()

// = NaN
TreeSet(1.0, NaN).min()
```

🤔

danieldietrich:
Easy to solve - min() and max() work always the same, the do not take an underlying ordering of a sorted collection into account:

```scala
scala> import scala.collection.immutable.TreeSet
import scala.collection.immutable.TreeSet

scala> val set = TreeSet(1.0, Double.NaN)
set: scala.collection.immutable.TreeSet[Double] = TreeSet(1.0, NaN)

scala> set.head
res3: Double = 1.0

scala> set.min
res4: Double = NaN

scala> val set = TreeSet(1.0, 2.0, 3.0)((d1, d2) => if (d1 < d2) 1 else -1)
set: scala.collection.immutable.TreeSet[Double] = TreeSet(3.0, 2.0, 1.0)

scala> set.head
res5: Double = 3.0

scala> set.min
res6: Double = 1.0
```

vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -190,47 +183,101 @@
...
+     * // = OptionalDouble(0.0) (wrong)
+     * j.u.s.DoubleStream.of(1.0, 10e100, 2.0, -10e100).average()
+     *
+     * // = Some(0.75) (correct)
+     * List.of(1.0, 10e100, 2.0, -10e100).average()
+     * }</pre>
+     *
      * @return {@code Some(average)} or {@code None}, if there are no elements
-     * @throws UnsupportedOperationException if this elements are not numeric
+     * @throws ClassCastException if this elements are not numeric
```
danieldietrich: @chb0github I followed you suggestions: throwing ClassCastException and adding a safe `average` variant (see below.

vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -190,47 +183,101 @@
...
+     * // = Some(0.75) (correct)
+     * List.of(1.0, 10e100, 2.0, -10e100).average(Number::doubleValue)
+     * }</pre>
+     *
+     * @param toDouble a function that converts elements to {@code double} values
+     * @return {@code Some(average)} or {@code None}, if there are no elements
+     */
+    default Option<Double> average(ToDoubleFunction<? super T> toDouble) {
+        @SuppressWarnings("unchecked")
+        final double[] sum = foldLeft(new double[4], (acc, element) -> { // acc = [sum, c, count, simpleSum]
```
danieldietrich: removed the `map` call, now we traverse only once instead of two times

vavr/src/test/java/io/vavr/collection/AbstractTraversableTest.java
```diff
@@ -1157,6 +1189,31 @@ public void shouldThrowNPEWhenMaxOfIntAndNull() {
...
+    }
+
+    @Test
+    public void shouldCalculateMaxOfFloatsContainingNaN() {
+        assertThat(of(1.0f, Float.NaN, 2.0f).max().get()).isEqualTo(Float.NaN);
+    }
+
+    @Test
+    public void shouldThrowClassCastExceptionWhenTryingToCalculateMaxOfDoubleAndFloat() {
+        assertThatThrownBy(() -> this.<Number> of(1.0, 1.0f).max()).isInstanceOf(ClassCastException.class);
```
chb0github: In previous tests you use `@Test(expected = ClassCastException.class)` but here you don't.

vavr/src/test/java/io/vavr/collection/PriorityQueueTest.java
```diff
@@ -328,4 +329,14 @@ private void assertMinimumsAreEqual(java.util.PriorityQueue<Integer> oldQueue, P
...
         }
     }
+
+    // -- ignored tests
+
+    @Override
+    @Test
+    @Ignore
+    public void shouldCalculateAverageOfDoubleAndFloat() {
+        // it is not possible to create a PriorityQueue containing unrelated types
```
chb0github: If your priority queue can be constructed with a customer comparator, it can, right?

vavr/src/main/java/io/vavr/collection/Traversable.java
chb0github: I am honored

vavr/src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -190,47 +183,101 @@
...
+     * j.u.s.DoubleStream.of(1.0, 10e100, 2.0, -10e100).average()
+     *
+     * // = Some(0.75) (correct)
+     * List.of(1.0, 10e100, 2.0, -10e100).average(Number::doubleValue)
+     * }</pre>
+     *
+     * @param toDouble a function that converts elements to {@code double} values
+     * @return {@code Some(average)} or {@code None}, if there are no elements
+     */
+    default Option<Double> average(ToDoubleFunction<? super T> toDouble) {
```
chb0github: Gah! The dread `ToDoubleFunction`! I would LOVE to have a compiler switch that turns off primitives

vavr/src/test/java/io/vavr/collection/PriorityQueueTest.java
danieldietrich: That's right, will do that...

vavr/src/test/java/io/vavr/collection/AbstractTraversableTest.java
danieldietrich: Mmhh, interesting - will check that!

vavr/src/main/java/io/vavr/collection/Traversable.java
danieldietrich: I target this change for 1.0.0 (because it affects backward compatibility). The upcoming 0.9.1 (hopefully this weekend) will have the old exception type.

vavr/src/main/java/io/vavr/collection/Traversable.java
danieldietrich: @chb0github The longer I play with sum, product, average, et al. the more I think you are right and we should remove the unsafe ones in 1.0. I like the `average(ToDoubleFunction)`.

`sum()` and `product()` are basically a `fold()`. However, we could provide also `double`-only versions `sum(ToDoubleFunction)` and `product(ToDoubleFunction)`. This way we would be safe.

I would keep `min()` and `max()` which assume that elements are comparable.

vavr/src/test/java/io/vavr/collection/AbstractTraversableTest.java
danieldietrich: Yes, `@Test(expected=...)` is the 'old style' we used. Assertions is now the way to go, we have more control over the check, e.g. instanceOf, hasCause, withMessage etc.

However, I plan to give all tests some love. We will migrate to JUnit 5.0. We will have more internal interfaces/mixins (prefixed with a `_`), like _Foldable, _NumericOps, ... These will have separate unit tests, which are also interfaces (that's possible with JUnit5).

This will us give better structure/control over our tests.

danieldietrich:
@chb0github **Note:** I saved the work of our discussion to a different branch that will go to 1.0. The changes of this PR are going to 0.9.1

PR #2079, opened 2017-09-11
Optimized conversion to Java collection

### LINKED ISSUE #2020: Seq returned by Map.values() should know its own size

danieldietrich:
Fixes #2020 
Replaces PR #2076

### DIFF
--- a/vavr/src/main/java/io/vavr/Value.java
+++ b/vavr/src/main/java/io/vavr/Value.java
@@ -598,7 +598,21 @@ public interface Value<T> extends Iterable<T> {
     }
 
     /**
-     * Converts this to an untyped Java array.
+     * Converts this to a Java array with component type {@code Object}
+     *
+     * <pre>{@code
+     * // = [] of type Object[]
+     * Future.<String> of(() -> { throw new Error(); })
+     *       .toJavaArray()
+     *
+     * // = [ok] of type Object[]
+     * Try.of(() -> "ok")
+     *    .toJavaArray()
+     *
+     * // = [1, 2, 3] of type Object[]
+     * List.of(1, 2, 3)
+     *     .toJavaArray()
+     * }</pre>
      *
      * @return A new Java array.
      */
@@ -617,8 +631,22 @@ public interface Value<T> extends Iterable<T> {
     }
 
     /**
-     * Converts this to a typed Java array.
+     * Converts this to a Java array having an accurate component type.
      *
+     * <pre>{@code
+     * // = [] of type String[]
+     * Future.<String> of(() -> { throw new Error(); })
+     *       .toJavaArray(String.class)
+     *
+     * // = [ok] of type String[]
+     * Try.of(() -> "ok")
+     *    .toJavaArray(String.class)
+     *
+     * // = [1, 2, 3] of type Integer[]
+     * List.of(1, 2, 3)
+     *     .toJavaArray(Integer.class)
+     * }</pre>
+     * 
      * @param componentType Component type of the array
      * @return A new Java array.
      * @throws NullPointerException if componentType is null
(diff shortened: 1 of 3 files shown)


vavr/src/main/java/io/vavr/Value.java
```diff
@@ -1275,12 +1472,20 @@ default CharSeq toCharSeq() {
...
-        final T container = containerSupplier.apply(size);
+    static <T, R extends java.util.Collection<T>> R toJavaCollection(
+            Value<T> value, Function<Integer, R> containerSupplier) {
+        return toJavaCollection(value, containerSupplier, 16);
+    }
+
+    static <T, R extends java.util.Collection<T>> R toJavaCollection(
+            Value<T> value, Function<Integer, R> containerSupplier, int defaultInitialCapacity) {
+        final int size;
+        if (value instanceof Traversable && ((Traversable) value).isTraversableAgain() && !value.isLazy()) {
```
danieldietrich: the additional check `!value.isLazy()` solves #2020

vavr/src/main/java/io/vavr/Value.java
```diff
@@ -657,27 +699,76 @@ default CharSeq toCharSeq() {
...
+     * // = [1, 2, 3]
+     * List.of(1, 2, 3)
+     *     .toJavaList()
+     * }</pre>
      *
      * @return A new {@link java.util.ArrayList}.
      */
     default java.util.List<T> toJavaList() {
-        return ValueModule.toJavaCollection(this, ArrayList::new);
+        return ValueModule.toJavaCollection(this, ArrayList::new, 10);
```
chb0github: Why?

vavr/src/main/java/io/vavr/Value.java
```diff
@@ -735,26 +856,69 @@ default CharSeq toCharSeq() {
...
+     * // = [1, 2, 3]
+     * List.of(1, 2, 3)
+     *     .toJavaSet()
+     * }</pre>
      *
      * @return A new {@link java.util.HashSet}.
      */
     default java.util.Set<T> toJavaSet() {
-        return ValueModule.toJavaCollection(this, java.util.HashSet::new);
+        return ValueModule.toJavaCollection(this, java.util.HashSet::new, 16);
```
chb0github: How did you arrive at this magic number? I am sure it's an improvement.

vavr/src/main/java/io/vavr/Value.java
danieldietrich: Before we used a default initial size of 16. But Java's ArrayList has 10 (hidden in a private field). I think we should use the default as fallback in order to be 'as good as Java'.

screen shot 2017-09-12 at 19 39 30

(_Note: I used IntelliJ IDEA to look into java.util.ArrayList of JDK 1.8.0_144_)

That the fallback value of 10 will only be taken in the following cases:

* We have a single-valued `Value`, like `Option` or `Try`. Please note that we do not use an initial size of 1 for single-valued type. This is because we return a mutable collection that might be modified by the caller (this is absolutely legitimate!). If the ArrayList has to be resized, it would be counter-productive/too costly to have a size of 1. Therefore we take the default size of 10.
* Another case, where 10 is taken as initial capacity, is for collections that are traversable once (like Iterator) or when they are lazy (like Stream). The latter case fixes the original issue #2020.

vavr/src/main/java/io/vavr/Value.java
danieldietrich: Screenshot taken From the `HashSet` source:

screen shot 2017-09-12 at 19 51 35

Screenshot taken From the `HashMap` source:

screen shot 2017-09-12 at 19 52 02

PR #2080, opened 2017-09-13
Added flexibility to build maven modules separately

nbardiuk:
Looks like scala maven plugin has a way to pass project related information to the script. https://davidb.github.io/scala-maven-plugin/example_script.html

When script is executed in scope PLUGIN the plugin injects a variable `project` with type [MavenProject](https://maven.apache.org/ref/3.0.5/apidocs/org/apache/maven/project/MavenProject.html).

PROS:
 * can speedup development by building only module under change
 * can run parallel builds on CI

CONS: 
 * IDE does not recognize `project` variable and marks it as compilation error 
 * cannot run script without maven (limits debugging)


danieldietrich:
That is nice, thank you!

PR #2088, opened 2017-09-15
Removed memoization of sortBy()

### LINKED ISSUE #2039: Seq.sortBy does not work for sequences containing infinite streams
Unfortunately, the following test does not terminate:

```
final Stream<Integer> stream1 = Stream.continually(1);
final Stream<Integer> stream2 = Stream.continually(2);
final Seq<Stream<Integer>> testee = Vector.of(stream2, stream1);
final Seq<Stream<Integer>> actual = testee.sortBy(Stream::head);
final Seq<Stream<Integer>> expected = Vector.of(stream1, stream2);
```

The reason is that `sortBy` memoizes the mapping function, which leads to evaluation of the hash code of the given elements. Indeed, removing the memoization makes the test pass instantly.

The problem is not only with `Vector`, but with all `Seq` implementations I looked at since `sortBy` seems to be implemented in the same way for all these classes.

danieldietrich:
Fixes #2039

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Collections.java
+++ b/vavr/src/main/java/io/vavr/collection/Collections.java
@@ -25,8 +25,10 @@ import io.vavr.control.Option;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.Objects;
 import java.util.function.*;
+import java.util.stream.Collector;
 
 /**
  * Internal class, containing helpers.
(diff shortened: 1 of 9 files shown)


vavr/src/main/java/io/vavr/collection/Collections.java
```diff
@@ -330,6 +333,14 @@ public T next() {
         return scanLeft(reversedElements, zero, (u, t) -> operation.apply(t, u), us -> finisher.apply(reverseIterator(us)));
     }
 
+    static <T, U, R extends Seq<T>> R sortBy(Seq<? extends T> source, Comparator<? super U> comparator, Function<? super T, ? extends U> mapper, Collector<T, ?, R> collector) {
+        Objects.requireNonNull(comparator, "comparator is null");
+        Objects.requireNonNull(mapper, "mapper is null");
+        return source.toJavaStream()
+                .sorted((e1, e2) -> comparator.compare(mapper.apply(e1), mapper.apply(e2)))
+                .collect(collector);
```
danieldietrich: @ummels @ruslansennov it is not possible to keep the memoization. Because our equals() hashCode() methods rely on deep calculations. Even if we fix the sortBy case for `Vector<Stream<T>>`, we can't cover also ``Vector<Option<Stream<T>>>`` and so on. The only solution is to disable memoization.

However, memoization of collections may be not cheap at all, depending on the size of the collection. We do not cache the hashCode, currently it takes O(n) to calculate it - the same for equals().

ummels: Thanks!

Issue #1529, opened 2016-08-24
Add factory method that allows to resolve/reject a Future


danieldietrich:
In ES6 we use Promise like this:

``` javascript
new Promise((resolve, reject) => {
    try {
        resolve(longRunningComputation());
    } catch (err) {
        reject(err);
    }
})
.then(...)
.catch(...);
```

We aligned to Scala when creating Future (for read operations) and Promise (for write operations). But Promise looks too heavy. It is a wrapper around a Future that provides statements to mutate the underlying Future. Promise is not monadic opposed to Future.

When starting async programming with Scala it was first hard for me to get the difference between Future and Promise and when to use which.

I think we could greatly simplify the API by providing ES6 like factory methods and first deprecate, then remove Promise completely (in the next major release).

We write

``` java
return Future.of(resolve -> {
        onComplete(result -> resolve.accept(mapper.apply(result)));
});
```

instead of

``` java
// implementation of Future.mapTry
final Promise<T> promise = Promise.make();
this.onComplete(result -> promise.complete(mapper.apply(result)));
return promise.future();
```

Note: there will be different factory methods with and without reject function, having BiConsumer, two Consumers or one Consumer, with and without ExecutorService. Examples will follow.

danieldietrich:
Let's dig deeper into the topic on how should the new `Future.of((resolve, reject) -> ...)` API make use of ExecutorServices.

### Example 1: Expensive resolve call, e.g. having a long running operation

```java
// starts a new Thread because longRunningCompuration() may not block the current Thread
Future.of(resolve -> resolve.accept(longRunningComputation()));

// uses someExecutor to start a new Thread
Future.of(someExecutor, resolve -> resolve.accept(longRunningComputation()));
```

### Example 2: Cheap resolve call, e.g. for internal Future.map implementation

```java
Future<T> first = Future.of(() -> longRunningComputation());
Function<T, U> mapper = ...;

// creates a new Thread for second Future
Future<U> second = Future.of(resolve ->
        // onComplete reuses Thread of first Future
        first.onComplete(result -> resolve(result.map(mapper))
);
```

The second Future immediately finishes the computation because `first.onComplete()` is cheap.

If we are aware of cheap running operations, we should execute the operation on the current thread:

```java
Future<U> second = Future.of(SAME_THREAD_EXECUTOR_SERVICE, resolve ->
        first.onComplete(result -> resolve(result.map(mapper))
);
```

[...]

PR #2093, opened 2017-09-16
Removed Promise in favor of Future.join(). Made ForkJoinPool.commonPool() default ExecutorService.

### LINKED ISSUE #2085: Future's default executor prevents JVM from terminating
There even is a comment in the Javadocs: "Please note that it may prevent the VM from shutdown." This is a pretty serious shortcoming, because it always happens when doing something asynchronously with the default ExecutorService.

What is there against using an ExecutorService that creates daemon threads? The requirement that there should be one non-daemon thread that does something useful with the async result seems harmless, not a real restriction.

I will try to attach a patch for java.io.concurrent.Future for your consideration.

-- Sebastian

danieldietrich:
Fixes #2085


nfekete:
Does the deadlock only occur on the CI server? Can you reproduce it locally?

danieldietrich:
Only on the CI server, locally all is fine.

I started a new build without junit test forking. It just started: https://travis-ci.org/vavr-io/vavr/builds/276364880

danieldietrich:
It did not work. Now I print the enter/exit information during FutureTest:

```java
    @Rule
    public TestRule watcher = new TestWatcher() {
        @Override
        protected void starting(Description description) {
            System.out.println("Starting test: " + description.getMethodName());
        }
        @Override
        protected void finished(Description description) {
            System.out.println("Finished test: " + description.getMethodName());
        }
    };
```

Let's see, which method causes the deadlock...

danieldietrich:
The blocking guy seems to be `FutureTest.shouldFilterFuture()`. Will digg into it. Maybe we revealed a bug?

nfekete:
I don't know. It would be good to see what are the differences between CI instance and local one. Maybe CI is virtualized so that it only sees 1 CPU, which might cause differences between the default JVM setup of common concurrency primitives? It would be much easier to debug if we could reproduce the issue locally. Can you find out the Travis instance VCPU count?

danieldietrich:
I think there is one CPU with two cores:

screen shot 2017-09-17 at 01 26 04

screen shot 2017-09-17 at 01 24 45

I don't know how to reproduce it. Will for now go step by step (with System.out.println)...

danieldietrich:
After I commented out

```java
    @Test
    public void shouldFilterFuture() {
        final Future<Integer> future = Future.successful(42);
        assertThat(future.filter(i -> i == 42).get()).isEqualTo(42);
        assertThat(future.filter(i -> i == 43).isEmpty()).isTrue();
    }
```

the tests proceeded much further but blocked again at

```java
    @Test
    public void shouldAwaitOnGet() {
        final Future<Integer> future = Future.of(() -> {
            Try.run(() -> Thread.sleep(250L));
            return 1;
        });
        assertThat(future.get()).isEqualTo(1);
    }
```

I think it is the Future.get() method that causes the trouble:

```java
    @Override
    default T get() {
        return await().getValue().get().get();
    }
```

More specifically the await() method:

```java
    /**
     * Blocks the current Thread until this Future completed or returns immediately if this Future is already completed.
     *
     * @return this {@code Future} instance
     */
    Future<T> await();
```

Which is implemented in FutureImpl.java:

```java
    @Override
    public Future<T> await() {
        if (!isCompleted()) {
            final Object monitor = new Object();
            onComplete(ignored -> {
                synchronized (monitor) {
                    monitor.notify();
                }
            });
            synchronized (monitor) {
                if (!isCompleted()) {
                    Try.run(monitor::wait);
                }
            }
        }
        return this;
    }
```

**One possibility is that the `monitor.notify()` is executed before `monitor.wait()`. In that case the Future would never return back...

I will wrap my head around it after getting some coffee :)

danieldietrich:
An execution like the following would explain the failure. But such an execution is not possible because we synchronize wait/notify on `monitor`. But it has to be a similar execution like this...

1) await() is called
2) onComplete(...)
  * this future is not complete, yet
  * therefore the complete-handler is queued
  * onComplete exits
3) synchronized(monitor) is entered and this Future is not completed yet
  * if (!isCompleted()) is entered
  * a context switch happens and completes the Future
  * on that completing thread a sub-thread is spawned that executes the complete handler
  * monitor.notify() is called (context switches still may have taken place)
  * finally the monitor::wait is executed... and hangs for-ever

nfekete:
I pulled your changes from your branch and changed the following:

```patch
Index: pom.xml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- pom.xml	(date 1505604858000)
+++ pom.xml	(revision )
@@ -302,6 +302,7 @@
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>${maven.surefire.version}</version>
                     <configuration>
+                        <argLine>-Djava.util.concurrent.ForkJoinPool.common.parallelism=1</argLine>
                         <parallel>none</parallel>
                         <!--
                         <parallel>all</parallel>
```

With this change I was able to reproduce the Travis deadlock.

danieldietrich:
@nfekete that's great! Maybe it is related to this:

screen shot 2017-09-17 at 02 35 12

(Source: https://stackoverflow.com/a/23480863/1110815)

nfekete:
I think I met a similar problem with RxJava a while ago, that's why I volunteered to help with ForkJoinPool issues. I can't recall precisely what happened there, but there was a similar pattern: tasks in an FJP waiting on other tasks in the same pool.

danieldietrich:
I've now a better understanding of the 'error'. I'm not sure, if we are able to control the fork/join mechanism within one thread pool. There is [ForkJoinPool.ManagedBlocker](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.ManagedBlocker.html) but we can't hard-code it into FutureImpl because we do not know which ExecutorService is used at runtime.

I will sleep over the topic and look into it tomorrow.

danieldietrich:
Maybe I find a way to implement `await()` based on `ExecutorService.invokeAny(...)`...

danieldietrich:
**Status:** I reimplement await() now:

```java
@Override
public Future<T> await() {
    // lock is the guard for all future result related fields (i.e. the completion state)
    synchronized (lock) {
        if (!isCompleted()) {
            try {
                // job is of type j.u.c.Future, get() blocks - and could also receive a timeout parameter
                complete(job.get());
            } catch(Throwable x) {
                complete(Try.failure(x));
            }
        }
    }
    return this;
}
```

However, it is still not working - but should in my opinion :)

danieldietrich:
```bash
[STARTING] 2017-09-17T12:08:10.944 shouldFilterFuture(io.vavr.concurrent.FutureTest)

Future.success(42)
---
[Promise] Thread[main,5,main] tryComplete(Success(42))
[Future] Thread[main,5,main] tryComplete: value = Success(42)
[Future] Thread[main,5,main] tryComplete: isCompleted() = false
[Future] Thread[main,5,main] tryComplete: complete(Success(42))
[Future] Thread[main,5,main] complete: value = Success(42)
[Future] Thread[main,5,main] complete: setting value
[Future] Thread[main,5,main] complete: executing actions
[Future] Thread[main,5,main] tryComplete: completed

asserting...
---
[Future] Thread[main,5,main] filter: io.vavr.concurrent.FutureTest$$Lambda$6/1058025095@27abe2cd
[Future] Thread[main,5,main] filterTry: io.vavr.concurrent.Future$$Lambda$7/1359044626@29444d75
[Future] Thread[main,5,main] filterTry: creating promise
[Future] Thread[main,5,main] filterTry: onComplete(...)
[Future] Thread[main,5,main] perform: spawning new onComplete thread
[Future] Thread[main,5,main] filterTry: return promise.future()
[Future] Thread[main,5,main] await()

get()
---
[Future] Thread[main,5,main] await: synchronized (lock) {
[Future] Thread[main,5,main] await: job = null
[Future] Thread[main,5,main] await: value = None
[Future] Thread[main,5,main] await: !isCompleted()
[Future] Thread[main,5,main] await: job.get()

filtering i == 42
[Promise] Thread[ForkJoinPool.commonPool-worker-1,5,main] tryComplete(Success(42))
[Future] Thread[ForkJoinPool.commonPool-worker-1,5,main] tryComplete: value = Success(42)

#
# Here is the error. The job (j.u.c.Future(() -> Success(42).filter(i -> i == 42))) is finished before the Promise is completed!?
#
[Future] Thread[main,5,main] await: completed

[Future] Thread[ForkJoinPool.commonPool-worker-1,5,main] tryComplete: isCompleted() = false
[Future] Thread[ForkJoinPool.commonPool-worker-1,5,main] tryComplete: complete(Success(42))
[Future] Thread[ForkJoinPool.commonPool-worker-1,5,main] complete: value = Success(42)
[Future] Thread[ForkJoinPool.commonPool-worker-1,5,main] complete: setting value
[Future] Thread[ForkJoinPool.commonPool-worker-1,5,main] complete: executing actions
[Future] Thread[ForkJoinPool.commonPool-worker-1,5,main] tryComplete: completed

[FINISHED] 2017-09-17T12:08:10.988 shouldFilterFuture(io.vavr.concurrent.FutureTest)
```

danieldietrich:
Currently I see the following solution:

1. We switch to ForkJoinPool
2. We deprecate Promise (to be removed in 1.0.0)
3. We enhance Future with a safe/scoped completion feature (see #1529)

```java
static <T> Future<T> of(CheckedConsumer<CheckedConsumer<Try<? extends T>>> complete) {
    return null; // TODO
}

static <T> Future<T> of(CheckedBiConsumer<CheckedConsumer<? extends T>, CheckedConsumer<? extends Throwable>> complete) {
    return null; // TODO
}

static void examples() {

    // calls Future.of(CheckedConsumer<CheckedConsumer<Try<T>>)
    Future.of(complete -> complete.accept(Try.of(() -> null)));

    // calls Future.of(CheckedBiConsumer<CheckedConsumer<T>, CheckedConsumer<Throwable>>)
    Future.of((success, failure) -> success.accept(null));
}


@FunctionalInterface
interface CheckedBiConsumer<T, U> {

    void accept(T t, U u) throws Throwable;

    default CheckedBiConsumer<T, U> andThen(CheckedBiConsumer<? super T, ? super U> after) {
        Objects.requireNonNull(after);
        return (l, r) -> {
            accept(l, r);
            after.accept(l, r);
        };
    }
}
```

Given the [suggested await() / await(timout, timeunit)](https://github.com/vavr-io/vavr/pull/2093#issuecomment-330032875), we will solve here the following issues:

* [ ] (this issue) Future's default executor prevents JVM from terminating
* [ ] [Consider adding javaslang.concurrent.Future#get(timeout)](https://github.com/vavr-io/vavr/issues/1871)
* [ ] [Execute Future operations on the same Thread by default](https://github.com/vavr-io/vavr/issues/1530)
* [ ] [Add factory method that allows to resolve/reject a Future](https://github.com/vavr-io/vavr/issues/1529)

danieldietrich:
This change has a too great impact for a patch release. I will target 1.0 instead of 0.9.1.

danieldietrich:
Pushed a test-version with some /*TODO*/ sections.
Basically we need to add an internal feature that enables us to complete a Future on the same thread with onComplete(). By default onComplete() spawns a new task that executes the follow-up action.
Will do that tomorrow.

danieldietrich:
I've collected some test data and ran the `FutureTest` class for different levels of parallelism: 1, 2, 4, 8

```java
Tests run: 221, Failures: 0, Errors: 0, Skipped: 2, Time elapsed: 13.722 sec - in io.vavr.concurrent.FutureTest
```

I observe that the 'queued submission' of the common ForkJoinPool goes up at some point if the parallelism is low (1, 2). Then the deadlock occurs.

If we increase the parallelism (4, 8), the tests work as expected.

screen shot 2017-09-21 at 22 30 18

screen shot 2017-09-21 at 22 32 05

screen shot 2017-09-21 at 22 32 56

screen shot 2017-09-21 at 22 33 40

chb0github:
@danieldietrich empirical evidence FTW

danieldietrich:
I could imagine that we need to fine-tune the ForkJoinPool behavior and the thread creation. (See also https://github.com/scala/bug/issues/8955).

I could also imagine that the probability of deadlocks increases because we use the blocking `await()` call (also implicitly used in `get()` and `isEmpty()`).

danieldietrich:
We need a different blocking strategy. Currently we have

```java
// FutureImpl
    @Override
    public Future<T> await() throws InterruptedException {
        if (!isCompleted()) {
            final BlockingQueue<Try<T>> queue = new ArrayBlockingQueue<>(1);
            onComplete(queue::add);
            ForkJoinPool.managedBlock(new ForkJoinPool.ManagedBlocker() {
                @Override
                public boolean block() throws InterruptedException {
                    if (value.isEmpty()) {
                        // Possibly blocks until one element is available.
                        // If the current Thread receives the element,
                        // this Future is already completed because
                        // actions are performed after completion.
                        queue.take();
                    }
                    return true;
                }
                @Override
                public boolean isReleasable() {
                    return value.isDefined();
                }
            });
        }
        return this;
    }
```

It may be that the use of LockSupport makes more sense (parking/unparking threads). This would release resources. However, the CompletableFuture implementation is more than ugly. Usages of UNSAFE everywhere.

danieldietrich:
There are some minor changes in here. I will split this PR into smaller ones. However, for now I will go on to do some improvements here (in this PR).

This PR goes into 1.0.0. But I think that I'm able to provide a fix that will go into 0.9.2, namely switching to ForkJoinPool.commonPool() and re-implementing a more effective await() + await(timeout, timeunit).

---

But now to something completely different. Here is an idea on how to optimize the 'blocking' await() method.

In Java there are several [Thread states](https://stackoverflow.com/questions/27406200/visual-vm-thread-states/27406503):

java-thread-states

I want to investigate if it is possible to put an existing (the current) thread in an 'inactive' state in the way that other threads are allowed to run instead (if they otherwise had to wait until more 'slots' are available in the given context).

I see the opportunity to use [LockSupport#park()](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/LockSupport.html#park--) for that purpose. Here is my implementation idea:

* our FutureImpl gets a new field, a queue of threads that wait for this Future
* the FutureImpl complete method will execute `waitingThreads.forEach(Thread::unpark)`
* the FutureImpl.await() method puts `Thread.currentThread()` on the waiting queue and parks it (if the Future isn't completed yet)

Technical details:

* waitingThreads needs to be 'guarded' by the same lock as the other instance variables that are relevant regarding the completion state
* waitingThreads need only to be unparked if their actual state is 'parked' (aka suspended?)
* maybe we need to extra-check interrupted conditions (where threads come into play) and act accordingly
* ...

danieldietrich:
The park/unpark does not seem to help regarding our problem of deadlocks when executing the FutureTest test suite using ForkJoinPool parallelism=1.

I've implement several await() strategies, namely

* using wait/notify using a monitor
* using park/unpark
* using a ManagedBlocker than blocks on a concurrent queue than takes the result

None of them solves the deadlock problem described above.

**Observation:** It is interesting that each test runs fine if started as a single test. Only the whole test suite seams to deadlock, even if we do not allow parallel tests (as surefire configuration).

The next step I take is to use a custom ForkJoinPool that differs in two aspects from ForkJoinPool.commonPool():

* We will set parallelism to default parallelism + 1.
* We create a custom thread factory that gives us more control over the threads that are created. Of course we need daemons in order to not block the system shutdown. Most notably we might need to raise the upper bound of threads that exist. Also we need to investigate if we need the original purpose of the ForkJoinPool to have units of work that are forked, wait for results and then joined. Maybe we can relax the behavior here a bit.

danieldietrich:
Oh oh! 😱

Our unit tests contain Futures that block forever. This might be the reason for deadlocks when having parallelism=1 (or=2):

```java
    @Test
    public void shouldGetValueOfUncompletedFuture() {
        final Future<?> future = Future.of(Concurrent::waitForever);
        assertThat(future.getValue()).isEqualTo(Option.none());
    }
```

PR #2096, opened 2017-09-17
Concurrent recursive function memoization

smillies:
Hi there, I've gone and created a pull request. Changes compared to the original proposal: MemoizedConcurrently has an additional method that accepts a cache implementation, and ConcurrentTrampoliningMemoizer is hidden (package private). 
Regards,
Sebastian


danieldietrich:
Thank you, I will review it tomorrow!

smillies:
I will delete the pull request from GitHub, because it does not compile anymore with the demise of Promise in [#2093](https://github.com/vavr-io/vavr/pull/2093)

danieldietrich:
Mmhh, sorry. I think it is still possible. The new `Future.join` is a functional substitute for the imperative `Promise`. I will take a look!

Sorry for being so slow. There is so much to do - it will take some days until I'm able to check all PRs. Our blog will be deleted on Sep 30th because RedHat stops OpenShift v2. I'm currently creating a new blog on clever-cloud and migrating all the data. In parallel we urgently need a v0.9.2 because some nasty bugs where reported. Also we need to support JDK9. With four kids it is sometimes hard to find enough time beside the daylight job. Sorry for letting you down.

One thing about the PR - I need to check how we could integrate the feature with Vavr. I do not want to create new (public visible) types, our goal is to keep the core lib as thin as possible. We will see...

smillies:
Don't worry, I'm not feeling let down at all. I appreciate all the hard work you're putting into this library. (Four kids - happy you. We were four at home as well.)

danieldietrich:
Thx for the kind words

Issue #2049, opened 2017-08-04
Revert 'sneaky throw' workaround


danieldietrich:
We need to revert #1722. It is a hack. Java is not supposed to treat all exceptions as unchecked (even in the checked case). For example a 3rd party `Proxy` instance that usse Vavr may not work as expected. Also async computations (i.e. ExecutorService) may suffer from unexpected exceptions.

PR #2101, opened 2017-09-19
Adds Option.mapTry and flatMapTry

### LINKED ISSUE #2099: Create Try<Optional> utilities
Given a method `Optional<T> retrieve(Id) throws Exception` and `R process(T) throws Exception`

I'd like to be able to get something close to
```
Try<Option<R>> result = Try.of(() -> Option.ofOptional(retrieve(id)))
    .andThenTry(Processor::process);
```

What I currently have to do is 
```
Try<Option<R> result = Try.of(() -> Option.ofOptional(retrieve(id)))
    .map(retrieved -> retrieved.map(t -> Try.of(() -> Processor.process(t))))
    .flatMap(processed -> Try.of(processed.map(Try::get))); // Try<Option<Try<R>>> -> Try<Option<R>>
```

That flatMap just does my head in (and possibly isn't the best for speed either)

The desired API is possible if we overload `TryOption<T> = Try.of(CheckedOptionFunction<T>)` where `interface CheckedOptionFunction<T> extends CheckedFunction0<Option<T>>`

but even some static utility would be awesome. (I think it could be common enough to be an awesome addition)

danieldietrich:
~~Fixes #2099~~ **Update:** Issue #2099 closed by me

### DIFF
--- a/vavr/src/main/java/io/vavr/control/Option.java
+++ b/vavr/src/main/java/io/vavr/control/Option.java
@@ -19,6 +19,7 @@
  */
 package io.vavr.control;
 
+import io.vavr.CheckedFunction1;
 import io.vavr.PartialFunction;
 import io.vavr.Value;
 import io.vavr.collection.Iterator;
@@ -349,6 +350,15 @@ public interface Option<T> extends Value<T>, Serializable {
 
     /**
      * Maps the value to a new {@code Option} if this is a {@code Some}, otherwise returns {@code None}.
+     * <p>
+     * Examples:
+     * <pre>{@code
+     * // = None
+     * Option.<Integer> none().flatMap(i -> Option.of(i + 1));
+     *
+     * // = Some(2)
+     * Option.some(1).flatMap(i -> Option.of(i + 1));
+     * }</pre>
      *
      * @param mapper A mapper
      * @param <U>    Component type of the resulting Option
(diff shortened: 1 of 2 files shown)


danieldietrich:
@skestle you might review it...

vavr/src/main/java/io/vavr/control/Option.java
```diff
@@ -361,8 +371,51 @@ default T getOrElse(Supplier<? extends T> supplier) {
...
+     */
+    @SuppressWarnings("unchecked")
+    default <U> Option<U> flatMapTry(Function<? super T, ? extends Option<? extends U>> mapper) {
+        Objects.requireNonNull(mapper, "mapper is null");
+        if (isEmpty()) {
+            return none();
+        } else {
+            try {
+                return (Option<U>) mapper.apply(get());
+            } catch(Throwable x) {
```
chb0github: catching `Throwable` is generally bad form as this can be Errors like `OutOfMemoryError` or `ExceptionInInitializerError`. These are considered non-recoverable and it's explicitly encouraged to exit the JVM. How about `catch(Exception e) ` as this covers every conceivable non-error case.

vavr/src/main/java/io/vavr/control/Option.java
```diff
@@ -361,8 +371,51 @@ default T getOrElse(Supplier<? extends T> supplier) {
...
+     * // = None
+     * Option.some(0).flatMapTry(i -> Option.of(1 / i)); // division by zero
+     * }</pre>
+     *
+     * @param mapper A value mapper that may throw
+     * @param <U>    Component type of the resulting Option
+     * @return a new {@code Option}
+     */
+    @SuppressWarnings("unchecked")
+    default <U> Option<U> flatMapTry(Function<? super T, ? extends Option<? extends U>> mapper) {
```
chb0github: This whole function just seems likes short hand for something not to hard to already achieve.

What about changing the signature to `Option.some(0).try(i -> 1/i)`? Seems shorter and more intuitive

vavr/src/main/java/io/vavr/control/Option.java
```diff
@@ -374,6 +427,39 @@ default T getOrElse(Supplier<? extends T> supplier) {
...
+     *
+     * // = None
+     * Option.some(0).mapTry(i -> 1 / i); // division by zero
+     * }</pre>
+     * 
+     * @param mapper A value mapper that may throw an exception
+     * @param <U>    The new value type
+     * @return a new {@code Some} containing the mapped value or {@code None}
+     */
+    default <U> Option<U> mapTry(CheckedFunction1<? super T, ? extends U> mapper) {
```
chb0github: Maybe `Option.tryToMap(...)` ? More natural and fluent.

vavr/src/main/java/io/vavr/control/Option.java
danieldietrich: That is a valid point! We need some kind of re-throw logic in the fatal case that will be re-added when we have solved #2049. See also https://github.com/vavr-io/vavr/blob/master/vavr/src/main/java/io/vavr/control/Try.java#L1522-L1527

vavr/src/main/java/io/vavr/control/Option.java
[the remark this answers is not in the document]
danieldietrich: Sorry, I forgot to drop some lines about the design decision.

Basically we want to do a map in the example (see original issue), but it does not accept a function that throws. Instead of adding new concepts to our 'language', we use the same things we already have in Try and Future: namely mapTry and flatMapTry. That's the most straight-forward way to do it.

vavr/src/main/java/io/vavr/control/Option.java
danieldietrich: (see explanation above)

vavr/src/main/java/io/vavr/control/Option.java
```diff
@@ -361,8 +371,51 @@ default T getOrElse(Supplier<? extends T> supplier) {
...
+     * Examples:
+     * <pre>{@code
+     * // = None
+     * Option.<Integer> none().flatMapTry(i -> 1 / i);
+     *
+     * // = Some(0)
+     * Option.some(2).flatMapTry(i -> Option.of(1 / i));
      *
+     * // = None
+     * Option.some(0).flatMapTry(i -> Option.of(1 / i)); // division by zero
```
skestle: (I'm into  definitions and...) I'm not sure this is a fantastic example (for my use case anyhow) - but quite possibly is for this change...

`i -> Option.of(1/i)` is really a `PartialFunction<Int, Int>`. I'd be writing this as `Option.some(0).flatMap(reciprocalFunction.lift())`. I'm not sure there's a really valid semantic use case for this function - certainly it's a shorter path, but I'm not sure that's always a good thing.

`flatMapTry(function)` is the same as `flatMap(t -> Try.of(() -> function.apply(t)).toOption())` right? Or rather `flatMap(t --> function.liftTry().apply(t).toOption())`, or `map(liftTry(function)).flatMap(Try::toOption)`

Seems like this change is helpful, but I'd want to see the equivalent expansion in the docs.

danieldietrich: Thank you! Yoy are right, it is a partial function.
The discussion in the original issue shows that I've missed the point here. What we really need is a mapOrThrow() that takes a CheckedFunction1 and re-throws any exception.
I will update the PR accordingly.

PR #2104, opened 2017-09-22
Low impact solution for helping migration to Java 9 modules

hamnis:
Module names are suggestions, you decide on which ones are useful.

### DIFF
--- a/vavr/pom.xml
+++ b/vavr/pom.xml
@@ -61,6 +61,13 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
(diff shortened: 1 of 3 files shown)


vavr/pom.xml
```diff
@@ -61,6 +61,13 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>io.vavr.core</Automatic-Module-Name>
```
danieldietrich: Please name it `io.vavr`. I read it is best practice to name the module like the main package of the project.

When we slice Vavr modules for version 1.0.0 we will have a module `io.vavr.core`.

Thanks!

danieldietrich:
(Note: #1994 is still important for 1.0.0 /cc @kittylyst)

danieldietrich:
Awesome, thanks!

PR #2105, opened 2017-09-23
Make `VAVr` Java 8 & 9 compatible

paplorinc:
related to:
* https://github.com/vavr-io/vavr/pull/1563
* https://github.com/vavr-io/vavr/pull/1994
* https://github.com/vavr-io/vavr/pull/2104

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@ We use these goals frequently to keep the dependencies and plugins up-to-date:
         <maven.bundle.version>3.2.0</maven.bundle.version>
         <maven.clean.version>3.0.0</maven.clean.version>
         <maven.install.version>2.5.2</maven.install.version>
-        <maven.compiler.version>3.6.0</maven.compiler.version>
+        <maven.compiler.version>3.7.0</maven.compiler.version>
         <maven.deploy.version>2.8.2</maven.deploy.version>
         <maven.gpg.version>1.6</maven.gpg.version>
         <maven.jacoco.version>0.7.8</maven.jacoco.version>
@@ -82,7 +82,7 @@ We use these goals frequently to keep the dependencies and plugins up-to-date:
         <maven.exec.version>1.5.0</maven.exec.version>
         <maven.gwt.plugin>1.0-rc-6</maven.gwt.plugin>
         <scala.maven.version>3.3.1</scala.maven.version>
-        <scala.version>2.11.7</scala.version>
+        <scala.version>2.12.2</scala.version>
     </properties>
     <dependencyManagement>
         <dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -254,6 +254,8 @@ We use these goals frequently to keep the dependencies and plugins up-to-date:
                             <arg>-Xlint:all</arg>
                             <!-- Workaround for JDK bug https://bugs.openjdk.java.net/browse/JDK-6999068 -->
                             <arg>-Xlint:-processing</arg>
+                            <!-- Enable Java 9 compilation with 1.8 compatibility -->
+                            <arg>-Xlint:-options</arg>
                         </compilerArgs>
                     </configuration>
                 </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -65,18 +65,18 @@ We use these goals frequently to keep the dependencies and plugins up-to-date:
         <junit.version>4.12</junit.version>
         <gwt.version>2.8.0</gwt.version>
         <maven.enforcer.version>1.4.1</maven.enforcer.version>
-        <maven.build-helper.version>1.12</maven.build-helper.version>
-        <maven.bundle.version>3.2.0</maven.bundle.version>
+        <maven.build-helper.version>3.0.0</maven.build-helper.version>
+        <maven.bundle.version>3.3.0</maven.bundle.version>
         <maven.clean.version>3.0.0</maven.clean.version>
         <maven.install.version>2.5.2</maven.install.version>
         <maven.compiler.version>3.7.0</maven.compiler.version>
         <maven.deploy.version>2.8.2</maven.deploy.version>
         <maven.gpg.version>1.6</maven.gpg.version>
-        <maven.jacoco.version>0.7.8</maven.jacoco.version>
+        <maven.jacoco.version>0.7.9</maven.jacoco.version>
         <maven.jar.version>3.0.2</maven.jar.version>
         <maven.javadoc.version>2.10.4</maven.javadoc.version>
         <maven.release.version>2.5.3</maven.release.version>
-        <maven.versions.version>2.3</maven.versions.version>
+        <maven.versions.version>2.4</maven.versions.version>
         <maven.surefire.version>2.19.1</maven.surefire.version>
         <maven.source.version>3.0.1</maven.source.version>
         <maven.exec.version>1.5.0</maven.exec.version>
--- a/pom.xml
+++ b/pom.xml
@@ -32,6 +32,7 @@ We use these goals frequently to keep the dependencies and plugins up-to-date:
     <description>Vavr (formerly called Javaslang) is an object-functional language extension to Java 8+.</description>
     <modules>
         <module>vavr</module>
+        <module>vavr-benchmark</module>
         <module>vavr-match</module>
         <module>vavr-test</module>
     </modules>
@@ -61,7 +62,7 @@ We use these goals frequently to keep the dependencies and plugins up-to-date:
         <assertj.core.version>3.8.0</assertj.core.version>
         <eclipse.lifecycle.mapping.version>1.0.0</eclipse.lifecycle.mapping.version>
         <java.version>1.8</java.version>
-        <jmh.version>1.18</jmh.version>
+        <jmh.version>1.19</jmh.version>
         <junit.version>4.12</junit.version>
         <gwt.version>2.8.0</gwt.version>
         <maven.enforcer.version>1.4.1</maven.enforcer.version>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -32,6 +32,7 @@ We use these goals frequently to keep the dependencies and plugins up-to-date:
     <description>Vavr (formerly called Javaslang) is an object-functional language extension to Java 8+.</description>
     <modules>
         <module>vavr</module>
+        <module>vavr-benchmark</module>
```
danieldietrich: Does a Java 9 build require now this module? Just asking...
I ask myself if now a jar is released for vavr-benchmark, too (i.e. to maven central).
Is the benchmark profile then still needed?

vavr/src/main/java/io/vavr/Lazy.java
```diff
@@ -112,9 +112,10 @@ private Lazy(Supplier<? extends T> supplier) {
      * @return A lazy sequence of values.
      * @throws NullPointerException if values is null
      */
+    @SuppressWarnings("Convert2MethodRef") // TODO should be fixed in JDK 9 and Idea
     public static <T> Lazy<Seq<T>> sequence(Iterable<? extends Lazy<? extends T>> values) {
         Objects.requireNonNull(values, "values is null");
-        return Lazy.of(() -> Vector.ofAll(values).map(Lazy::get));
+        return Lazy.of(() -> Vector.ofAll(values).map(lazy -> lazy.get()));
```
danieldietrich: Does a java 9 build not compile with the method reference?

vavr/src/test/java/io/vavr/AssertionsExtensions.java
```diff
@@ -40,10 +40,10 @@ public static ClassAssert assertThat(Class<?> clazz) {
             this.clazz = clazz;
         }
 
+        @SuppressWarnings("deprecation")
         public void isNotInstantiable() {
-            final Constructor<?> cons;
             try {
-                cons = clazz.getDeclaredConstructor();
+                final Constructor<?> cons = clazz.getDeclaredConstructor();
```
danieldietrich: 👍🏼

vavr/src/test/java/io/vavr/MatchTest.java
```diff
@@ -315,17 +315,12 @@ public void shouldDecomposeListOfTuple3() {
...
-    @SuppressWarnings("UnnecessaryLocalVariable")
     @Test
     public void shouldDecomposeListWithNonEmptyTail() {
         final List<Option<Number>> intOptionList = List.of(Option.some(1), Option.some(2.0));
         final String actual = Match(intOptionList).of(
-                Case($Cons($Some($(1)), $Cons($Some($(2.0)), $())), (x, xs) -> {
-                    // types are inferred correctly!
-                    final Some<Number> head = x;
-                    final List<Option<Number>> tail = xs;
-                    return head + "::" + tail;
```
danieldietrich: I would keep the block. It was intended to make the types explicitly visible (even if vars are not used).
The difference to `(Some<Number> x, List<Option<Number>> xs) -> x + "::" + xs)` is that we have `List<Option<Number>> tail = xs`. I.e. we show which general type is inferred by Match.

pom.xml
paplorinc: It's in a separate commit, because it's obviously not strictly related :)
It should be part of the project so that it is compile-time checked. I will disable the deployment, it's a valid concern.

vavr/src/main/java/io/vavr/Lazy.java
paplorinc: Yeah, it's weird :/
We should probably open a bug for the `JDK` and for `Idea`

vavr/src/test/java/io/vavr/MatchTest.java
paplorinc: Yeah, except that it's not working in `Java 9`, different types are inferred for some reason.
If I specify it explicitly, as done here, it's working.
Not sure how to fix it otherwise.

paplorinc:
The JDK 9 additional build is a very good idea, let me do that quickly :)

pom.xml
paplorinc: Checked it, since it has tests only, there's nothing to deploy.
The `benchmark` profile has a run config also, so it still has its purpose - though we can remove the duplicate `modules` part, thanks for the note! :)

vavr/src/test/java/io/vavr/MatchTest.java
```diff
@@ -315,17 +315,12 @@ public void shouldDecomposeListOfTuple3() {
         assertThat(actual).isEqualTo("(begin, 10, 4.5)::List((middle, 11, 0.0), (end, 12, 1.2))");
     }
 
-    @SuppressWarnings("UnnecessaryLocalVariable")
     @Test
     public void shouldDecomposeListWithNonEmptyTail() {
         final List<Option<Number>> intOptionList = List.of(Option.some(1), Option.some(2.0));
```
paplorinc: The problem is, this isn't an `intOptionList`, it's a `numberOptionList`

vavr/src/test/java/io/vavr/MatchTest.java
paplorinc: Fixed, thanks!

pom.xml
paplorinc: (I had to test this module for Java 9 compatibility also, that's why I modified this part also)

vavr/src/test/java/io/vavr/MatchTest.java
```diff
@@ -318,12 +319,12 @@ public void shouldDecomposeListOfTuple3() {
     @SuppressWarnings("UnnecessaryLocalVariable")
     @Test
     public void shouldDecomposeListWithNonEmptyTail() {
-        final List<Option<Number>> intOptionList = List.of(Option.some(1), Option.some(2.0));
-        final String actual = Match(intOptionList).of(
+        final List<Option<? extends Number>> numberOptionList = List.of(Option.some(1), Option.some(2.0));
```
danieldietrich: I think because of the added `? extends` the build currently breaks. Does it not compile with jdk9 otherwise?
Another fix could be `List.<Option<Number>> of(Option.some(1), Option.some(2.0));`

pom.xml
danieldietrich: The benchmarks should not be part of the main build, even if it is a minimal version tests.
If there are assertions/checks, they should be moved to the vavr core unit tests.
Benchmarks are about measuring. The only valid case to run them as part of the tests would be that we break the build if the performance decreases. But that is not possible.

pom.xml
```diff
@@ -253,6 +254,8 @@ We use these goals frequently to keep the dependencies and plugins up-to-date:
                             <arg>-Xlint:all</arg>
                             <!-- Workaround for JDK bug https://bugs.openjdk.java.net/browse/JDK-6999068 -->
                             <arg>-Xlint:-processing</arg>
+                            <!-- Enable Java 9 compilation with 1.8 compatibility -->
+                            <arg>-Xlint:-options</arg>
```
danieldietrich: Where do we use the options? Is it the implicit `<target>${java.version}</target>`?

paplorinc: this is needed because `java.version` is always `8`, even if the selected JDK is 9. The JDK cannot compile to another bytecode version, giving a warning. This disables the warning, since we're only checking the build, not producing any artifacts.
```
<java.version>1.8</java.version>
```

vavr/src/test/java/io/vavr/MatchTest.java
paplorinc: Dunno, man, this seems to randomly pass or fail on `JDK 9`. There's a wedding I should attend, instead I'm debugging this :p

pom.xml
paplorinc: The benchmarks aren't run during the build. Not even the assertions. just the compilation.
They are just part of the project, to make sure any refactoring is propagated there also.

vavr/src/test/java/io/vavr/MatchTest.java
danieldietrich: Go to the medding, wan - eh I mean wedding, man. We can do this later :)

pom.xml
danieldietrich: I see, that's cool.

vavr/src/test/java/io/vavr/MatchTest.java
paplorinc: darn ... could you take a look, it's hopeless (it just passed locally ...)?

danieldietrich: Yes, will do that later today! Currently I'm in the middle of splitting #2093 in to smaller PRs. After that (and after going to german election)...

vavr/src/test/java/io/vavr/MatchTest.java
```diff
@@ -318,9 +320,9 @@ public void shouldDecomposeListOfTuple3() {
     @SuppressWarnings("UnnecessaryLocalVariable")
     @Test
     public void shouldDecomposeListWithNonEmptyTail() {
-        final List<Option<Number>> intOptionList = List.of(Option.some(1), Option.some(2.0));
-        final String actual = Match(intOptionList).of(
-                Case($Cons($Some($(1)), $Cons($Some($(2.0)), $())), (x, xs) -> {
+        final List<Option<Number>> numberOptionList = List.of(Option.some(1), Option.some(2.0));
+        final String actual = Match(numberOptionList).of(
+                Case($Cons($Some($(1)), $Cons($Some($(2.0)), $())), (Some<Number> x, List<Option<Number>> xs) -> {
```
paplorinc: @danieldietrich, it only compiles on `Java 9` if we explicitly specify the types here. Is that still ok?

danieldietrich:
Just realized that the build is broken for jdk9 (you told me already). It does not matter - I will fix that tomorrow. Thx!

PR #2119, opened 2017-10-04
Fix Future.traverse with ExecutorService argument

ummels:
`Future.traverse` currently ignores the provided `ExecutorService` since it is not passed to `Future.sequence`.


vavr/src/main/java/io/vavr/concurrent/Future.java
```diff
@@ -643,7 +643,7 @@
         Objects.requireNonNull(executorService, "executorService is null");
         Objects.requireNonNull(values, "values is null");
         Objects.requireNonNull(mapper, "mapper is null");
-        return sequence(Iterator.ofAll(values).map(mapper));
+        return sequence(executorService, Iterator.ofAll(values).map(mapper));
```
danieldietrich: Wow, good catch - thank you!

Issue #1640, opened 2016-10-25
Add recursive Tree factory methods


danieldietrich:
I described it in the blog-post [Object-Functional Concurrent Programming with Javaslang](http://blog.javaslang.io/object-functional-concurrent-programming-with-javaslang).

**Signatures:**

``` java
Node<T> of(T seed,  Function<? super T, Iterable<? extends T>> descend);

Future<Node<T>> ofAsync(T seed,  Function<? super T, Future<? extends Iterable<? extends T>>> descend);
```

**Depends on:** #1529 

**Example impl:**

``` java
static <T> Future<Node<T>> ofAsync(T seed,  
        Function<? super T, Future<? extends Iterable<? extends T>>>
        descend) {
    return Future.of(promise -> descend.apply(seed)
        .onComplete(result -> result
            .onSuccess(children -> promise.completeWith(
                Future.sequence(
                    Iterator.ofAll(children).map(child -> ofAsync(child, descend))
                ).map(childNodes -> Tree.of(seed, childNodes))))
            .onFailure(promise::failure)));
}
```

danieldietrich:
Maybe this should be closed. The module vavr-collectionx should not depend on vavr-concurrent.

danieldietrich:
The sync case is fixed with #2121. We will not do the async case.

PR #2121, opened 2017-10-06
Add recursive tree factory method

valery1707:
Add useful factory method described in #1640 

P.S.
My English is still bad and I can make some bad decisions on text in Javadoc - please correct me.

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Tree.java
+++ b/vavr/src/main/java/io/vavr/collection/Tree.java
@@ -205,6 +205,34 @@ public interface Tree<T> extends Traversable<T>, Serializable {
         return io.vavr.collection.Collections.fill(n, s, empty(), Tree::of);
     }
 
+    /**
+     * Recursively builds a non-empty {@code Tree}, starting with the given {@code seed} value and proceeding in depth-first order.
+     * <p>
+     * The children of a node are created by
+     * <ol>
+     * <li>applying the {@code descend} function to the node value</li>
+     * <li>calling this method recursively by using each derived child value as new seed (in iteration order).</li>
+     * </ol>
+     * <p>
+     * Example:
+     * <pre>{@code
+     * // = (1 (2 4 5) 3)
+     * Tree.recurse(1, i ->
+     *   (i == 1) ? List.of(2, 3) :
+     *   (i == 2) ? List.(4, 5) :
+     *   List.empty()
+     * ).toLispString();
+     * }</pre>
+     *
+     * @param seed    The start value for the Tree
+     * @param descend A function to calculate the child values
+     * @param <T>     Value type
+     * @return a new, non-empty {@code Tree} instance
+     */
+    static <T> Node<T> recurse(T seed, Function<? super T, ? extends Iterable<? extends T>> descend) {
+        return Tree.of(seed, Stream.of(seed).flatMap(descend).map(children -> recurse(children, descend)));
+    }
+
     @Override
     default <R> Tree<R> collect(PartialFunction<? super T, ? extends R> partialFunction) {
         return ofAll(iterator().<R> collect(partialFunction));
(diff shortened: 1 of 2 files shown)


valery1707:
Fail in JDK 9 is not my fault

vavr/src/main/java/io/vavr/collection/Tree.java
```diff
@@ -143,6 +143,18 @@
     }
 
     /**
+     * Returns a new Node containing the given value and having the children returned from given function.
+     *
+     * @param seed    The start value for the Tree
+     * @param descend A function to calculate the children values
```
danieldietrich: s/children/child/

vavr/src/main/java/io/vavr/collection/Tree.java
```diff
@@ -143,6 +143,18 @@
...
+     * @param <T>     Value type
+     * @return a {@code Tree} with values build up iteration
```
danieldietrich: s/a {@code Tree} with values build up iteration/a new, non-empy {@code Tree} instance/

vavr/src/main/java/io/vavr/collection/Tree.java
```diff
@@ -143,6 +143,18 @@
     }
 
     /**
+     * Returns a new Node containing the given value and having the children returned from given function.
```
danieldietrich: Please replace the first sentence with the following:

```
Recursively builds a non-empty {@code Tree}, starting with the given {@code seed} value and proceeding in depth-first order.
<p>
The children of a node are created by
<ol>
<li>applying the {@code descend} function to the node value</li>
<li>calling this method recursively by using each derived child value as new seed (in iteration order).</li>
</ol>
```

vavr/src/main/java/io/vavr/collection/Tree.java
```diff
@@ -143,6 +143,18 @@
...
+     *
+     * @param seed    The start value for the Tree
+     * @param descend A function to calculate the children values
+     * @param <T>     Value type
+     * @return a {@code Tree} with values build up iteration
+     */
+    static <T> Node<T> of(T seed, Function<? super T, Iterable<? extends T>> descend) {
```
danieldietrich: Please rename the method to `recurse(T seed, Function<? super T, ? extends Iterable<? extends T>> descend)`.

The main reason for renaming it is that `of` might be ambiguous. For example what happens if you change the value type in the unit test from Integer to String? I'm not sure because the call `Tree.of(seed, Stream.....)` matches two signatures: `Tree.of(T, Iterable)` and `Tree(Integer, Function<Integer, ....>)` in the special case of the unit test. This is because Stream is a Seq which implements PartialFunction&lt;Integer, T>, which implements Function<&lt;Integer, T>.

I think the name `Tree.recurse(...)` fits well because  we have `Stream.iterate(...)` which is the same in the sense that a Stream can be seen as Tree with only one or zero children per node.

We use `? extends Iterable` everywhere. I'm not sure if removing `? extends` would still cover all use-cases. (See also our Traversable.flatMap method for example.)

danieldietrich:
Oh, I forgot - we have a new 'boy scout rule #1': We add usage examples to the javadoc 😅

Please add

```java
<pre>{@code
// = (1 (2 4 5) 3)
Tree.recurse(1, i ->
    (i == 1) ? List.of(2, 3) :
    (i == 2) ? List.(4, 5) : List.empty()
}).toLispString();
}</pre>
```

valery1707:
@danieldietrich, need I reorder position of method `recurse`? As I remember, methods in this project has some rule of sorting

valery1707:
Maybe between `fill` and `collect`?

valery1707:
@danieldietrich I update commit

danieldietrich:
Great thank you!
The order is ok, for now it is sufficient, if the new methods are located in the static method block. We do not have alphabetic order in other classes. Anyway, for Vavr 1.0.0 I will 'clean up' all classes.

PR #2123, opened 2017-10-06
Implements unlift operator for total functions returning an Option result

florian-stefan:
This pull request adds the static method `<T, R> PartialFunction<T, R> unlift(Function1<T, Option<R>> totalFunction)` to the interface `PartialFunction`. This method unlifts a total function returning an Option result into a partial function. The partial function is defined at a value if and only if the given total function doesn't return None at this value.

An example use case of this method is the following:
```
Seq<Option<T>> maybeValues = ...
Seq<T> values = maybeValues.collect(PartialFunction.unlift(Function1.identity()))
```
In Scala, one would use Pattern Matching: `maybeValues.collect { case Some(x) => x }`.

### DIFF
--- a/vavr/src/main/java/io/vavr/PartialFunction.java
+++ b/vavr/src/main/java/io/vavr/PartialFunction.java
@@ -37,6 +37,30 @@ import io.vavr.control.Option;
 public interface PartialFunction<T, R> extends Function1<T, R> {
 
     /**
+     * Unlifts a {@code totalFunction} that returns an {@code Option} result into a partial function.
+     * The total function should be side effect free because it might be invoked twice: when checking if the
+     * unlifted partial function is defined at a value and when applying the partial function to a value.
+     *
+     * @param totalFunction the function returning an {@code Option} result.
+     * @param <T> type of the function input, called <em>domain</em> of the function
+     * @param <R> type of the function output, called <em>codomain</em> of the function
+     * @return a partial function that is not necessarily defined for all input values of type T.
+     */
+    static <T, R> PartialFunction<T, R> unlift(Function1<T, Option<R>> totalFunction) {
+        return new PartialFunction<T, R>() {
+            @Override
+            public R apply(T t) {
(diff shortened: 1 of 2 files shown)


vavr/src/main/java/io/vavr/PartialFunction.java
```diff
@@ -42,6 +42,33 @@
     long serialVersionUID = 1L;
 
     /**
+     * Unlifts a {@code totalFunction} that returns an {@code Option} result into a partial function.
+     * The total function should be side effect free because it might be invoked twice: when checking if the
```
danieldietrich: 👍🏼

vavr/src/main/java/io/vavr/PartialFunction.java
```diff
@@ -42,6 +42,33 @@
...
+     * unlifted partial function is defined at a value and when applying the partial function to a value.
+     *
+     * @param totalFunction the function returning an {@code Option} result.
+     * @param <T> type of the function input, called <em>domain</em> of the function
+     * @param <R> type of the function output, called <em>codomain</em> of the function
+     * @return a partial function that is not necessarily defined for all input values of type T.
+     */
+    static <T, R> PartialFunction<T, R> unlift(Function1<T, Option<R>> totalFunction) {
```
danieldietrich: We should use the more general form

```java
unlift(Function<? super T, ? extends Option<? extends R>> totalFunction)
```

vavr/src/main/java/io/vavr/PartialFunction.java
florian-stefan: Thanks for your review! I will have a look into the issue you mentioned ... buy the way: the Java 9 compilation step within the Travis build seems to be flickering?

danieldietrich: > Java 9 compilation step within the Travis build seems to be flickering?

yes, will take a look...

PR #2128, opened 2017-10-08
add Set.isSubsetOf

emmanueltouzery:


vavr/src/main/java/io/vavr/collection/Set.java
```diff
@@ -150,6 +150,18 @@ default Boolean apply(T element) {
...
     /**
+     * Returns true if this set is a subset of the set you give
+     * (including if both sets are equal)
+     *
+     * @param other a set to compare this set with
+     * @return true if this set is a subset to the parameter, false otherwise.
+     */
+    default boolean isSubsetOf(Set<? extends T> other) {
+        final Set<T> other_ = Set.narrow(other);
+        return forAll(other_::contains);
```
danieldietrich: Narrowing is always ugly. We could use a one-liner

```java
forAll(Set.<T> narrow(other)::contains)
```

but it looks a bit more cryptic.

danieldietrich: I played around with it, e.g.

```java
return Set.<T> narrow(other).containsAll(this);
```

I think both solutions are equivalent regarding speed/mem consumption.

danieldietrich: One more thought: `this.isSubsetOf(that)` is different from `that.containsAll(this)` in the way that `containsAll()` focuses on elements whereas `isSubSetOf()` focuses on sets.

Is it expected behavior that `sortedSet1` is subset of `sortedSet2` even if both have different underlying Comparators? E.g.

```java
Comparator<Integer> c = (i, j) -> 0; // all elements are equal
SortedSet<Integer> set1 = TreeSet.of(1, 2, 3);
SortedSet<Integer> set2 = TreeSet.of(c, 1);

// = true
set1.isSubsetOf(set2);
```

Just a question, it could be clarified in the Javadoc...

emmanueltouzery: yes for the narrow i was afraid of performance, since you say there's no difference then that's the way to go for sure. I'll update the PR tonight if you decide to keep it. Regarding `containsAll`: I missed containsAll completely!! Had I seen it, I would not have submitted this PR. So maybe we should close this PR as it's too similar to an existing feature, your call.

Regarding custom comparators... I didn't consider that. I have no opinion. We could check for TreeSet and return false in case the comparators differ. It's pretty tricky. I have no experience with that use-case, not sure what would be the right answer.

danieldietrich: > It's pretty tricky.

isSubsetOf and containsAll are dual, I would not implement extra-logic regarding the Comparator.

Let's close this PR and stick to the existing containsAll, it is the simplest solution.

Thx for your PR!

PR #2133, opened 2017-10-10
Simplified travis-ci config

danieldietrich:
Fixes a build-problem that currently occurs on the Travis-CI Java9 build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: java
+sudo: false # faster builds
 jdk:
   - oraclejdk8
   - oraclejdk9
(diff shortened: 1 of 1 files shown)


danieldietrich:
@chb0github Ha, yes 😄 We did not automate them, in fact it is currently hard manual work, depending on the release type. Basically, it is described in out [contribution guidelines](https://github.com/vavr-io/vavr/blob/f1d30ac88e4096706911be208d6868417c878fde/CONTRIBUTING.md#releasing).

We currently do not work on multiple branches for different versions. Instead we tag the releases (as usual) and cherry pick commits for the next patch release.

I don't know if it is possible to release to Sonatype in an automated way. We need to GPG-sign the artifacts. However, there are other possibilities than publishing artifacts to Maven Central, e.g. JFrog Bintray. The changelog creation could be automated, but writing it by hand is way better.

chb0github:
I have a bunch of experience in this - enough to say I never got past it!

It's definitely possible to release to sonatype automated - I will create an issue where it can be discussed

PR #2136, opened 2017-10-10
Prototype of graph class design and implementation

chb0github:
This is a prototype of a Graph implementation. Basic goals were/are

1. Immutability
2. Traversal is functional. That is, there isn't a `Graph.dfs()` there is `Graph.traverse(dfs())` to allow for easy replacement of traversal functions
3. Bidirectional relationships are 2 1-way relationships and nothing extra.


vavr/pom.xml
```diff
@@ -20,6 +20,12 @@
     </licenses>
     <dependencies>
         <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
```
nfekete: What is lombok used for?

chb0github: Cuts out all the boilerplay code for `get`, `set` `hash` `equals` etc.

If I have a class like this:

```java
@EqualsAndHash
@Getter
@AllArgsConstructor
@NoArgsConstructor
public class Foo {

    private String name;
    private Integer age;

}
```

Then I get a class generated that has this:

```
Foo f = new Foo("Christian",1234);
String name = f.getName();
Integer age = f.getAge();
Set<Foo> foos = HashSet.of(f);
Foo baz = new Foo("Christian",1234);
System.out.println(foos.contains(baz)); // prints true because hash and equals are created
```

There are a lot of annotations. It really reduces code

ruslansennov: @chb0github we keep Vavr without dependencies on the third-part libraries

nfekete: Lombok would be especially problematic, since it's a _manipulating bytecode_, it's doing totally non-standard things, not just a regular annotation processor, hence it's incompatible with GWT, which Vavr supports.

chb0github:
Hmm,

I have never heard that's it's incompatible with GWT and not sure why it
would be. It generates version agnostic code.

vavr/pom.xml
chb0github: @ruslansennov - it's a `provided` dependency. Which means, it doesn't go with the deployment artifact. Meaning it is functionally no dependency at all.

nfekete: Lombok will hook into the internal compiler code of javac and ECJ and modify their internal in-memory AST representation to convince the compiler to generate bytecode according to the annotations you use. With the way it works, you will get into trouble as soon as you'll try to use the same code in another compilation environment, which requires consistency at the source level, since Lombok generated (byte-)code is invisible to those tools. While that may or may not work with GWT, since GWT also uses ECJ, it will definitely break with any other transpiler or tool that parses java source code. It's a total hack that can break any time also if the javac or ECJ team decides to change some internal implementation details of their compilers. Because of these, for me it's totally out of question to use it in any serious project.

chb0github:
I am sure you didn't mean to suggest that the places I have used it and see
it used were not 'serious' projects.

I actually submitted the PR to get some input on a graph implementation,
not get flamed about lombok. It's tangential to the PR. If you have
something constructive to add, I am all ears.

nfekete:
If I offended you, I am truly sorry, I didn't mean to. 😟

chb0github:
I understand. It's all good

PR #2144, opened 2017-10-22
Adds factory method for creating a partial function that maps a Value type to its underlying value

### LINKED ISSUE #2127: Add PartialFunction factory for Value arguments
We should create a PartialFunction factory that can be used in the presence of Value types, like Option, Try, (right-biased) Either, ...

```java
interface PartialFunction<T, R> extends Function<T, R> {

    static <T, V extends Value<T>> PartialFunction<V, T> getIfDefined() {
        return new PartialFunction<V, T>() {

            @Override
            public T apply(V value) {
                return value.get();
            }

            @Override
            public boolean isDefinedAt(V value) {
                return !value.isEmpty();
            }
        };
    }
}
```

I thought about the name, it could be

* get
* getIfDefined
* getIfNotEmpty
* apply                   &lt;--- I came up first with this - create a PartialFunction that applies a Value

Use case:

```java
Seq<Option<T>> maybeValues = ...
Seq<T> values = maybeValues.collect(PartialFunction::getIfDefined)
```

florian-stefan:
Fixes #2127

### DIFF
--- a/vavr/src/main/java/io/vavr/PartialFunction.java
+++ b/vavr/src/main/java/io/vavr/PartialFunction.java
@@ -65,6 +65,35 @@ public interface PartialFunction<T, R> extends Function1<T, R> {
             public boolean isDefinedAt(T value) {
                 return totalFunction.apply(value).isDefined();
             }
+
+        };
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


florian-stefan:
Close and reopen pull request to restart Java 9 build.

PR #2149, opened 2017-10-24
Added onFailure for specific exception types

### LINKED ISSUE #2142: Unable to catch a specific runtime exception using Vavr  library?
I am trying to catch a specific runtime exception (so not throwable) and just log it (log.error has a void return type). What is the simplest way to do this vavr?

   ```
    try {
        sayHello();
    } catch (MyAppRuntimeException ex) {
        log.error("Error occured"); // log.error returns void not Void so I couldn't find a suitable method in  Vavr library
    }
```

I have tried
```
    Try.run(() -> sayHello())
       .recover(MyAppRuntimeException.class, ex->log.error(Error occured: {}, ex.getMessage()));
```

I get Bad return type void cannot be converted to Void

### PR DESCRIPTION
**Pull request by danieldietrich**
Fixes #2142

* added onFailure(Class, Consumer) as handler for specific failures
* added javadoc examples to onSuccess, onFailure, recover, recoverWIth

### DIFF
--- a/vavr/src/main/java/io/vavr/control/Try.java
+++ b/vavr/src/main/java/io/vavr/control/Try.java
@@ -34,7 +34,6 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 
-import static io.vavr.API.Match;
 import static io.vavr.control.TryModule.isFatal;
 import static io.vavr.control.TryModule.sneakyThrow;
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: vavr/src/main/java/io/vavr/control/Try.java [hunk 1]
```diff
@@ -631,7 +638,45 @@ default boolean isSingleValued() {
...
+     *    .onFailure(Error.class, x -> System.out.println("Error"));
+     * }</pre>
+     * 
+     * @param exceptionType the exception type that is handled
+     * @param action an excpetion consumer
+     * @param <X> the exception type that should be handled
+     * @return this
+     * @throws NullPointerException if {@code exceptionType} or {@code action} is null
+     */
+    @GwtIncompatible
```
danieldietrich: @ruslansennov this time I did not forget it 🎉

ruslansennov: :smile: it doesn't matter, because `vavr-gwt` doesn't working now (caused by changes in `FutureImpl`). Seems we should keep both branches: `master` and, say, `v.0.9.x`. Every commit in `master` which is belongs to minor release should be cherry-picked immediately for obtain actual minor snapshot version.

Issue #1716, opened 2016-11-27
[modules] Seamlessly fuse hand-written and generated code


danieldietrich:
Prerequisites:

* [ ] Find a simple way to run Scala source code from within Java
    * see [generating a class from string and instantiating it in scala 2.10](http://stackoverflow.com/questions/12122939/generating-a-class-from-string-and-instantiating-it-in-scala-2-10/12123609#12123609)
    * see [compile and execute scala code at runtime](http://stackoverflow.com/questions/7477589/compile-and-execute-scala-code-at-runtime)
    * see [eval in scala](http://stackoverflow.com/questions/1183645/eval-in-scala)

Spike:

* [ ] Re-utilize the existing ['block-parser'](https://github.com/danieldietrich/xtext-protected-regions/blob/master/plugins/net.danieldietrich.protectedregions.core/src/main/xtend/net/danieldietrich/protectedregions/parser/Parser.xtend) of [xtext-protexted-regions](https://github.com/danieldietrich/xtext-protected-regions), i.e. re-implement it based on javaslang in a separate project, code-name: 'javaslang-generator'. (A future version of javaslang-generator might be based on javaslang-parser instead of using a proprietary block-parser.)
* [ ] In a first step scan src/main/java/** for *.java files and use the block-parser to parse regions. The result will be [non-generated, region-start, generated, region-end, non-generated, region-start, generated, region-end, non-generated, ...]. The regions need to have internal, intermediate unique ids (e.g. a sequential number).
* [ ] Copy the relevant parts of region-start sections to a Scala file. These regions need to be wrapped in a block containing the unique id mentioned above (because we need to substitute the generated regions of the original file in a later step). Also include the base-generator into that scala file (currently our Generator.scala). The include may be a native Scala include or a copy-paste include, depending on how the generator is run (as Scala script or as compiled Scala file).
* [ ] Run the dynamically built Scala generator (as script or as compiled Scala file)
* [ ] Parse the output of the Scala generator with our block-parser. Write the previously parsed contents [non-generated, region-start, generated, region-end, non-generated, region-start, generated, region-end, non-generated, ...] back to the original file and substitute the generated parts on the way.

---

Next steps:

* [ ] Document the code generator and make it public available by announcement
* [ ] Add support for locales / character encodings
* [ ] Make source folders configurable
* [ ] Create a Maven plugin
* [ ] Create a Gradle plugin

---

[...]

PR #2151, opened 2017-10-25
Annotate generated classes with `@Generated`

### LINKED ISSUE #2081: Annotate generated classes with @Generated
Generated classes like [CheckedFunction0](https://github.com/vavr-io/vavr/blob/ec2d9b4337004792e786b42e0ac3eddc6ddf4bc8/vavr/src-gen/main/java/io/vavr/CheckedFunction0.java#L41) could be annotated with `javax.annotation.Generated` to differentiate them from hand written code.

Furthermore the generator class as well as the actual generator method could be referenced via the comment attribute to ease finding the appropriate generator method.

valery1707:
Fix #2081

### DIFF
--- a/vavr/generator/Generator.scala
+++ b/vavr/generator/Generator.scala
@@ -736,6 +736,7 @@ def generateMainClasses(): Unit = {
               /$javadoc
                * For-comprehension with ${i.numerus("Iterable")}.
                */
+              @javax.annotation.Generated("Generator.scala")
               public static class For$i<$generics> {
 
                   ${(1 to i).gen(j => xs"""private final Iterable<T$j> ts$j;""")("\n")}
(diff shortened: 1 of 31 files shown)


danieldietrich:
Awesome, thank you 😊

talios:
Note that `@Generated` is no longer available in JDK 9 ( its part of the EE jars/modules, which aren't enabled by default, and are deprecated) - but are available via a dependency.

danieldietrich:
@talios thanks for the hint! I will create an issue.

**Note:** Because of #1716 we will have a new code generator. The 'inline' generated code will not be annotated with `@Generated` anymore. See also [this tweet](https://twitter.com/danieldietrich/status/926846450016931840).

talios:
@danieldietrich See https://github.com/immutables/immutables/issues/703#issuecomment-342029844 for a related discussion.  Seems Java 9 added `javax.annotation.processing.Generated` :)

PR #2152, opened 2017-10-25
Remove all Eclipse related comments and workarounds

### LINKED ISSUE #2075: Internal/Library Development: Remove all Eclipse related comments and workarounds
We will support `javac` and `IntelliJ` only in order to keep the code clean.

The reason is that the main contributors work with javac/IntelliJ and we have no automated way to test compatibility with Eclipse / ECJ.

valery1707:
Fix #2075

Successful compiled by maven with JDK8 `1.8.0_45`.

But IDEA `2017.1.5` show errors on this lines
Multimap#flatMap

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,12 +6,6 @@ target
 .gradle
 build
 
-# -- Eclipse
-.classpath
-.project
-.settings
-bin
-
 # -- IntelliJ IDEA
 .idea
 *.iml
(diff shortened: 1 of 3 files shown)


valery1707:
@nfekete but GWT-part of vavr moved into separate project: https://github.com/vavr-io/vavr-gwt
Does the changes in source of `vavr` affect the assembly of `vavr-gwt`?

nfekete:
Yes, all sources need to be compilable by the JDT compiler in order to be usable in GWT. `vavr-gwt` provides only some emulation of some of the concurrency-related classes which are not available in the JavaScript environment as it has no concept of threads or concurrency.

valery1707:
@danieldietrich, if @nfekete is right than please reject PR and maybe close issue #2075 .
@nfekete please pay attention to issue #2075 and make same comments there

danieldietrich:
@valery1707 Yes, @nfekete is right, we have to keep the casts because of GWT 🤔
I will close the issue and reject the PR.
However, many thanks for your work!

PR #2157, opened 2017-11-08
Implement Traversable#reject

### LINKED ISSUE #2156: Add reject(), the inverse of filter
```java
interface Traversable<T> {
    // semantically equivalent to filter(f.negate())
    Traversable<T> reject(Predicate<? super T> predicate);
}
```

valery1707:
Fix #2156.

P.S.
A also replace `Map#removeAll`, `Map#removeKeys` and `Map#removeValues` with `reject*` variant. With deprecation of course.

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/AbstractMultimap.java
+++ b/vavr/src/main/java/io/vavr/collection/AbstractMultimap.java
@@ -254,9 +254,21 @@ abstract class AbstractMultimap<K, V, M extends Multimap<K, V>> implements Multi
     }
 
     @Override
+    public M reject(Predicate<? super Tuple2<K, V>> predicate) {
+        Objects.requireNonNull(predicate, "predicate is null");
+        return filter(predicate.negate());
+    }
+
+    @Override
     public M filter(BiPredicate<? super K, ? super V> predicate) {
         Objects.requireNonNull(predicate, "predicate is null");
         return filter(t -> predicate.test(t._1, t._2));
+    }
+
+    @Override
+    public M reject(BiPredicate<? super K, ? super V> predicate) {
+        Objects.requireNonNull(predicate, "predicate is null");
+        return reject(t -> predicate.test(t._1, t._2));
     }
 
     @Override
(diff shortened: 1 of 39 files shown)


valery1707:
JDK9 build failed with `Missing dependency 'object java.lang.Object in compiler mirror', required by /home/travis/.m2/repository/org/scala-lang/scala-library/2.10.5/scala-library-2.10.5.jar(scala/package.class)`

vavr/src/main/java/io/vavr/collection/AbstractMultimap.java
```diff
@@ -254,39 +254,63 @@ public M filter(Predicate<? super Tuple2<K, V>> predicate) {
...
+    @Override
+    @Deprecated
+    public M removeAll(BiPredicate<? super K, ? super V> predicate) {
+        return reject(predicate);
     }
 
     @Override
+    @Deprecated
     public M removeKeys(Predicate<? super K> predicate) {
-        Objects.requireNonNull(predicate, "predicate is null");
```
ruslansennov: I believe you should save this checking. This will be more concrete if you will throw an exception in the called method but not in its implementation. This also applies to other deprecated methods. There are many methods in Vavr where this rule is not respected, but I believe that it should

valery1707:
@ruslansennov fixed.
Maybe move to milestone [vavr-0.9.2](https://github.com/vavr-io/vavr/milestones/vavr-0.9.2)?

ruslansennov:
> fixed.

thank you!

> Maybe move to milestone vavr-0.9.2?

this PR adds new methods to the `Traversable` interface, so it can't be included to the minor release

vavr/src/main/java/io/vavr/collection/AbstractQueue.java
```diff
@@ -164,9 +164,16 @@ public Q removeAll(Iterable<? extends T> elements) {
         return Collections.removeAll((Q) this, elements);
     }
 
-    @SuppressWarnings("unchecked")
+    @Deprecated
```
danieldietrich: It is good to mark it! Maybe we can remove the method for Vavr 1.0.0 - but for now it can stay as-is. Thx!

vavr/src/main/java/io/vavr/collection/Map.java
```diff
@@ -58,11 +58,11 @@
  * <li>{@link #filter(BiPredicate)}</li>
  * <li>{@link #filterKeys(Predicate)}</li>
  * <li>{@link #filterValues(Predicate)}</li>
+ * <li>{@link #reject(BiPredicate)}</li>
+ * <li>{@link #rejectKeys(Predicate)}</li>
+ * <li>{@link #rejectValues(Predicate)}</li>
```
danieldietrich: great, thx!

PR #2158, opened 2017-11-08
Override some methods in SortedMap for refinement of their result class

### LINKED ISSUE #2155: SortedMap interface does not define 'put' with value merger
The 'SortedMap' interface does not define 'put(K key, U value, BiFunction<? super V, ? super U, ? extends V> merge)'. If one attempts to use it, the 'Map' implementation is used, meaning the return type becomes 'Map'. The 'TreeMap' implementation does define this version of 'Put' with the correct return type.

valery1707:
Fix #2155


ruslansennov:
thank you

valery1707:
@ruslansennov maybe move to milestone [vavr-0.9.2](https://github.com/vavr-io/vavr/milestones/vavr-0.9.2)?

ruslansennov:
> maybe move to milestone vavr-0.9.2

this PR adds new methods to the `SortedMap` interface, so it can't be included to the minor release

Issue #2053, opened 2017-08-14
Include Capsule in our benchmarks


danieldietrich:
https://github.com/usethesource/capsule

CauchyPeano:
What kind of help needed here?

danieldietrich:
We could start by adding a benchmark that compares certain Vavr collection operations with those of the Capsule collections. Especially it will tell us

1. how good our HAMT and BMT perform
2. if we could improve our collections

We already have benchmarks that compare our collections with those of Scala and Clojure. In a first step we could add Capsule to the party.

CauchyPeano:
@danieldietrich what I see that HAMT from capsule is already included into benchmark `io.vavr.collection.MapBenchmark`
Regarding BMT I couldn't find it in capsule library... Do I miss something?

danieldietrich:
Right - there seems nothing to do for us. Capsule seems to implement HAMT only.
Let's close this ticket, okay?

CauchyPeano:
@danieldietrich yeah, totally fine with me :) I was just puzzled if it's me that maybe don't understand something.

danieldietrich:
@CauchyPeano thanks!

PR #2161, opened 2017-11-09
Add Capsule's Set into benchmarks

valery1707:
Ref #2053 

[Capsule](https://github.com/usethesource/capsule), as I see, contains only `Map` and `Set` implementations.
But Vavr doesn't contains benchmarks for `Map` yet.


ruslansennov:
thank you

PR #2164, opened 2017-11-10
Add concurrency tests with JCStress

### LINKED ISSUE #1847: Add `jcstress` concurrency tests
Now that [jcstress](https://wiki.openjdk.java.net/display/CodeTools/jcstress) is finally available on [Maven](https://maven2repo.com/org.openjdk.jcstress/jcstress-core), we should create some stress tests.

Some contender with mutable inner state that should definitely be validated:
* [javaslang.Lazy#computeValue](https://github.com/javaslang/javaslang/blob/master/javaslang/src/main/java/javaslang/Lazy.java#L145)
* [javaslang.λ.Memoized#of](https://github.com/javaslang/javaslang/blob/master/javaslang/src/main/java/javaslang/%CE%BB.java#L78-L78)
* [javaslang.concurrent.Future](https://github.com/javaslang/javaslang/blob/master/javaslang/src/main/java/javaslang/concurrent/Future.java) (`find`, `await`, `cancel`, `complete`, `onComplete`, `isEmpty`, `run`, `tryComplete`)

valery1707:
- [x] Disable `treat warnings as errors` only in stress module
- [x] Run tests as part of the maven build process
- [x] Implement test for `Lazy`
- [x] Implement test for `λ.Memoized#of`
- [ ] Implement test for `Future#find`
- [ ] Implement test for `Future#await`
- [ ] Implement test for `Future#cancel`
- [ ] Implement test for `Future#complete`
- [ ] Implement test for `Future#onComplete`
- [ ] Implement test for `Future#isEmpty`
- [ ] Implement test for `Future#run`
- [ ] Implement test for `Future#tryComplete`

Fixes #1847


vavr-jcstress/src/main/java/io/vavr/LazyConcurrencyTest.java
```diff
@@ -0,0 +1,33 @@
...
+    private Lazy<Long> lazy = Lazy.of(source::incrementAndGet);
+
+    @Actor
+    public void actor1(LongResult1 r) {
+        r.r1 = lazy.get();
+    }
+
+    @Actor
+    public void actor2(LongResult1 r) {
+		r.r1 = lazy.get();
```
PascalSchumacher: formatting error ;) :)

valery1707: @PascalSchumacher, yes. Sorry - will fix now

vavr-jcstress/pom.xml
```diff
@@ -0,0 +1,118 @@
...
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <!--jcstress version to use with this project.-->
+        <jcstress.version>0.1.1</jcstress.version>
+        <!--Name of the test Uber-JAR to generate.-->
+        <uberjar.name>jcstress</uberjar.name>
+    </properties>
+
+    <build>
+        <plugins>
```
ruslansennov: Please skip the `maven-deploy-plugin` for the `vavr-jcstress` module. See [vavr-benchmark](https://github.com/vavr-io/vavr/blob/3365fd2e5d5d4a31d8ed41579f40db37da11fb8b/vavr-benchmark/pom.xml#L100) for example. I believe that replacing `<artifactId>maven-install-plugin</artifactId>` with `<artifactId>maven-deploy-plugin</artifactId>` should work. 

BTW, can the `jcstress` start in a phase different from `install` (I see your hint `mvn clean install -Pjcstress`)? Then we can skip `maven-install-plugin` too.

nfekete: [Maven lifecycle refernce](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference) ➡️ `verify` is just before install

ruslansennov: @nfekete ok, thanks, then we can skip both phases

valery1707: @ruslansennov fixed

valery1707:
Can somebody (@danieldietrich, @ruslansennov, @paplorinc , ...) hint me for the path to check the `Future` methods?
For `Lazy` and `Lambda.Memoized` it was easy: they must call to source method only once within all threads.

But for `Future` methods is not so easy:
* `Future#find` - Take list of Futures and find first completed Future which matches the predicate
* `Future#await` - Blocks the current Thread until this Future completed or returns immediately if this Future is already completed.
* `Future#cancel` - Cancels the Future. A running thread is interrupted.
* `Future#onComplete` - Performs the action once the Future is complete.
* `Future#isEmpty` - Checks, if this future has a value.
* `Future#run` - Starts an asynchronous computation

vavr-jcstress/src/main/java/io/vavr/MemoizedConcurrencyTest.java
```diff
@@ -0,0 +1,415 @@
...
+
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Supplier;
+
+import static org.openjdk.jcstress.annotations.Expect.ACCEPTABLE;
+import static org.openjdk.jcstress.annotations.Expect.FORBIDDEN;
+
+@Outcome(id = "1", expect = ACCEPTABLE, desc = "Source touched only once.")
+@Outcome(id = "2", expect = FORBIDDEN, desc = "Source touched twice.")
+public class MemoizedConcurrencyTest {
```
paplorinc: Really good job @valery1707!
@danieldietrich, should these be added to the code generator?

paplorinc:
@valery1707, if you're stuck, you could create more PRs for the same issue.
Also, check the codebase for synchronizations and concurrent code (including atomic ones) and add those to the suite, too :)
Seem like a good job so far!

vavr-jcstress/src/main/java/io/vavr/MemoizedConcurrencyTest.java
danieldietrich: @paplorinc Good remark! Generally I would say yes but for now we can leave it as is. When we have the new inline code generator we are able to write self-expanding Java files (the generator code will be part of a Java comment). I will then change all Vavr source files accordingly...

danieldietrich:
Hi @valery1707,
I think this issue was closed automatically because I switched branches. (old master is not v0.9.x).
Maybe we need to open a new PR for the v0.9.x branch...

PR #2169, opened 2017-11-15
Implement Value#none the inverse of `forAll`

valery1707:


PascalSchumacher:
Great! Eactly what I requested in https://github.com/vavr-io/vavr/issues/2035

valery1707:
@PascalSchumacher I don't see yours tiket than start to implement this :)
I start thinking about this method after implementing #2156 - sometimes build `negate` version of predicate is too complex than simply call method that will negate my predicate internally.
And currently vavr has methods that check elements that `all` or `some` match with predicate, and I think variant `none match` will cover all needs.

danieldietrich:
We should keep the API as slim as possible. When it comes to convenience methods / syntactic sugar and Scala does not have that API it is a signal for us not to include it.

I prefer to close this PR.

For new features it is always better to open an issue before in order to discuss it. However, I see that one new method is not a big deal but adding here and there one convenience method will blow up the API over time.

Thx!

danieldietrich:
As already stated I close this PR.
@valery1707 keep on the great work, it is always hard to say 'no' - we need to take special care when adding new things.

PR #2172, opened 2017-11-23
Fixes JDK9 build and tests

danieldietrich:
Fixes the build with JDK9 in conjunction with scala-maven-plugin:

screen shot 2017-11-24 at 00 21 17

/cc @szarnekow (works on my machine - hopefully it will work also on your machine)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,8 +83,8 @@ We use these goals frequently to keep the dependencies and plugins up-to-date:
         <maven.exec.version>1.5.0</maven.exec.version>
         <maven.gwt.plugin>1.0-rc-6</maven.gwt.plugin>
         <scala.maven.version>3.3.1</scala.maven.version>
-        <scala.version>2.10.6</scala.version>
-        <scala.compat.version>2.10</scala.compat.version>
+        <scala.version>2.12.3</scala.version>
+        <scala.compat.version>2.12</scala.compat.version>
     </properties>
     <dependencyManagement>
         <dependencies>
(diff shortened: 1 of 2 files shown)


szarnekow:
Hey @danieldietrich - unfortunately it does not work on my box:
```
[ERROR] Failed to execute goal net.alchim31.maven:scala-maven-plugin:3.3.1:script (default) on project vavr: wrap: java.lang.Exception: A NoSuchMethodError exception was thrown: scala.Product.$init$(Lscala/Product;)V -> [Help 1]
```

This is the change that I see in the pom.xml

```
-        <scala.version>2.10.6</scala.version>
-        <scala.compat.version>2.10</scala.compat.version>
+        <scala.version>2.12.3</scala.version>
+        <scala.compat.version>2.12</scala.compat.version>
```

If I revert this, it builds successfully.

danieldietrich:
Hi @szarnekow, interestingly, with JDK9, for me 2.10 does not work but with 2.12 it compiles fine.

I think I will not but put more effort into making Vavr work with the scala maven plugin. Instead I move to Gradle and switch the code generator from Scala to [code.js](https://github.com/vavr-io/code.js), which is already in the make.

Thx for looking into it & have a great weekend!

szarnekow:
Let me double-check this. I guess I ran the build with Java8. You don't use the enforcer plugin to assert the java version?

danieldietrich:
Generally we support both versions, Java 8 and Java 9, therefore we do not enforce the version. The target version is 1.8 (source and binary).

PR #2184, opened 2017-11-26
Build with gradle

Conorrr:
- [x] compile
- [ ] tests (benchmarks not working properly, might move this into the vavr-project)
- [ ] test coverage
- [ ] strip away maven
- [ ] codegen with scala
- [ ] travis
- [ ] benchmarks
- [ ] javadoc
- [ ] sources
- [ ] IntelliJ support
- [ ] scripted-publishing
- [ ] docs - update existing and write new ones
- [ ] clean-up formatting
- [ ] formatting guidelines
- [ ] Look for API-Breakages (like maven-bundle-plugin)
- [ ] Building with Java 9
- [ ] Building on Windows
- [ ] Single version location
- [ ] Check committed generated code is correct

Discussion Points
* Gradle Wrapper - some projects don't like including gradle with the source
* How to handle caching generated code
* Caching dependencies and build cache in travis

Other possible Gradle Features
* Distributed Build Cache
* Parallel Builds
* Build Scans
* Probably lots more


vavr-benchmark/build.gradle
```diff
@@ -0,0 +1,15 @@
...
+  compile project(':vavr')
+    compile group: 'org.scalaz', name: 'scalaz-core_2.12', version:'7.3.0-M8'
+    compile group: 'org.clojure', name: 'clojure', version:'1.9.0-alpha14'
+    compile group: 'org.pcollections', name: 'pcollections', version:'2.1.2'
+    compile group: 'org.functionaljava', name: 'functionaljava-java8', version:'4.6'
+    compile group: 'org.eclipse.collections', name: 'eclipse-collections', version:'8.0.0'
+    compile group: 'io.usethesource', name: 'capsule', version:'0.6.1'
+    compile group: 'org.openjdk.jol', name: 'jol-core', version:'0.7.1'
+    compile group: 'org.jetbrains', name: 'annotations', version:'15.0'
+    testCompile group: 'junit', name: 'junit', version:'4.12'
```
ruslansennov: It may be appropriate to keep all versions in a common `gradle.properties` file. At least for `junit` and `assertj`

settings.gradle
```diff
@@ -0,0 +1,15 @@
...
+include ':vavr-test'
+
+// TODO is this needed?
+project(':vavr').projectDir = "$rootDir/vavr" as File
+project(':vavr-benchmark').projectDir = "$rootDir/vavr-benchmark" as File
+project(':vavr-match').projectDir = "$rootDir/vavr-match" as File
+project(':vavr-test').projectDir = "$rootDir/vavr-test" as File
+
+// Dependency Versions
+ext['assertJ.version'] = '3.8.0'
```
ruslansennov: This is not necessary if `gradle.properties` will be used

settings.gradle
```diff
@@ -0,0 +1,15 @@
+rootProject.name = 'vavr-parent'
+include ':vavr'
+include ':vavr-benchmark'
+include ':vavr-match'
+include ':vavr-test'
+
+// TODO is this needed?
```
ruslansennov: Our modules have the same names with artifacts, so there is no

ruslansennov:
> Gradle Wrapper - some projects don't like including gradle with the source

I believe this is right way. Using wrapper will allow us to have the same (maybe latest) version of the gradle anywhere and not depend on the locally installed version (or even not have it). If we will use a wrapper, please add

```groovy
task wrapper(type: Wrapper) {
    gradleVersion = '4.3.1'
}
```

Conorrr:
I've prototyped running the codegen using gradle in the `vavr-test` sub-project.

I've had to make a couple of changes to the scala that generates the java; primarily I have changed it from a script to a class, I couldn't find a clean way to run the scala script from gradle.  [It wasn't possible](https://stackoverflow.com/a/19941627) using the plugin a couple of years ago and doesn't look like it is at the moment. Let me know if this messes with the general workflow.

I have also changed the generator script to take 2 arguments, the output directories. This may also make your workflow harder, let me know if it does.

Would be great to get some feedback.

vavr-test/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+  testCompile "junit:junit:$junitVersion"
+  testCompile "org.assertj:assertj-core:$assertJVersion"
+
+  codegenCompile "org.scala-lang:scala-library:$scalaVersion"
+  zinc "com.typesafe.zinc:zinc:$zincVersion"
+}
+
+task runScript(type: JavaExec) {
+  inputs.dir file("generator")
+  outputs.dirs file("src/generated"), file("src/generatedTest")
```
ruslansennov: It would be nice to delete the generated directories before each generator startup to be sure that the old generated files will not remain in the repository (test case: we can rename something in generator).

Also, can we keep the generated code in the `src-gen` folder for a while? This will reduce the changes in PR and help the reviewers. Just asking :)

Conorrr:
Should both be fine. `src-gen` is slightly unconventional but there isn't a problem with it.

---- On Thu, 30 Nov 2017 06:03:47 +0000 notifications@github.com wrote ----

@ruslansennov commented on this pull request.

In vavr-test/build.gradle:

> +    }
+  }
+}
+
+dependencies {
+  compile project(':vavr')
+  testCompile "junit:junit:$junitVersion"
+  testCompile "org.assertj:assertj-core:$assertJVersion"
+
+  codegenCompile "org.scala-lang:scala-library:$scalaVersion"
+  zinc "com.typesafe.zinc:zinc:$zincVersion"
+}
+
+task runScript(type: JavaExec) {
+  inputs.dir file("generator")
+  outputs.dirs file("src/generated"), file("src/generatedTest")
It would be nice to delete the generated directories before each generator startup to be sure that the old generated files will not remain in the repository (test case: we can rename something in generator).

Also, can we keep the generated code in the src-gen folder for a while? This will reduce the changes in PR and help the reviewers. Just asking :)

vavr-test/build.gradle
Conorrr: Sure, I'll change both of them. Using `src-gen` is slightly unconventional but there isn't any technical reason it wont work.

build.gradle
```diff
@@ -0,0 +1,32 @@
...
+  
+  repositories {
+    mavenLocal()
+    mavenCentral()
+  }
+
+}
+
+task wrapper(type: Wrapper) {
+    gradleVersion = '4.3.1'
```
szarnekow: Can we use 4.4.x here?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Wed Nov 29 19:15:14 GMT 2017
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.2.1-all.zip
```
szarnekow: Version is different from the one in the build script

.travis/run-tests.sh
```diff
@@ -3,10 +3,10 @@ set -ev
 #mvn test -B
 
 # Break CI build on PRs where committed differ from generated sources
-if [ -n "$(git status -su vavr/src/generated*)" ]; then
+if [ -n "$(git status -su vavr/src-gen)" ]; then
   exit 1
 fi
 
-if [ -n "$(git status -su vavr-test/src/generated*)" ]; then
+if [ -n "$(git status -su vavr-test/src-test)" ]; then
```
szarnekow: that should go to src-gen instead of src-test to reduce the noise

hepin1989:
why not with SBT?

danieldietrich:
@hepin1989 because the Scala code generator will be replaced. We will not use Scala anymore for the build-process.

danieldietrich:
We will stick with Maven and with the Scala code generator.
Moving to another tooling does not provide any value for end-users.
For s committers it is only time-consuming effort and the risk that we will introduce new issues.

Therefore I will close this PR.
Thanks for all the effort!!!

Issue #1807, opened 2017-01-10
Return same instance instead of wrapping Iterator results when possible


danieldietrich:
1. Remove the method `useIsEqualToInsteadOfIsSameAs()` from unit tests and test for `isSameAs()`
2. fix broken unit tests

**Example: take(0)**

```java
// AbstractTraversableTest
@Test
public void shouldTakeNoneOnNil() {
    if (useIsEqualToInsteadOfIsSameAs()) {
        assertThat(empty().take(1)).isEqualTo(empty());
    } else {
        assertThat(empty().take(1)).isSameAs(empty());
    }
}
```

```java
// PriorityQueue (current)
@Override
public PriorityQueue<T> take(int n) {
    return ofAll(comparator, iterator().take(n));
}

// PriorityQueue (time and mem saver)
@Override
public PriorityQueue<T> take(int n) {
    if (n <= 0) {
        return this;
    } else if (n >= size()) {
        return empty(comparator());
    } else {
        return ofAll(comparator, iterator().take(n));
    }
}
```

This was only an example. We have many similar occurrences.

PR #2188, opened 2017-12-08
Eliminated method useIsEqualToInsteadOfIsSameAs() from AbstractTraversableTest class (#1807)

NataliiaPrivezentseva:


danieldietrich:
Hi @NataliiaPrivezentseva, I really appreciate your hard work on this issue. However, I realized that code duplication is no viable alternative for us. It makes the tests hard to maintain over time, the lines of code increased by factor 5.4 for the corresponding tests. Therefore I will close this PR.

There can be only one solution to solve #1807 - all collections need to return the same instance in the corresponding cases. But we need to decide, if this makes really sense in all cases. For example, if the performance (significantly) decreases when performing extra checks, if the same instance should be returned, we will create another instance instead. We will see...

Thanks for contributing and sorry for closing the ticket - this is always a hard decision for me.

NataliiaPrivezentseva:
@danieldietrich, I understand your decision. I don't like code duplication either.

danieldietrich:
Thank you. I think we need to be more specific when creating issues...

Issue #2045, opened 2017-07-26
Add syntactic sugar for Iterator creation


danieldietrich:
It would be great to have an additional factory method

```java
interface Iterator<T> {

    static <T> Iterator<T> iterate(Supplier<Boolean> hasNext, Supplier<? extends T> next) {
        return new AbstractIterator<>() {
            @Override
            public boolean hasNext() {
                return hasNext.get();
            }
            @Override
            public T getNext() {
                return next.get();
            }
        };
    }

}
```

It makes sense to also add this method to `Stream`:

```java
interface Stream<T> {

    static <T> Stream<T> iterate(Supplier<Boolean> hasNext, Supplier<? extends T> next) {
        return Stream.ofAll(Iterator.iterate(hasNext, next));
    }

}
```

**Hint:** The addition might introduce ambiguities regarding existing `iterate()` methods.

nfekete:
Another factory method worth considering could be one accepting a `Supplier<Option<? extends T>>`,
which would repeatedly invoke the supplier while it's a `Some` and end on the first `None`.

chb0github:
I like @nfekete idea because it allows the data and the predicate to be tied together.

chb0github:
what about 
```
interface Iterator<T> {

    static <T> Iterator<T> iterate(Supplier<? extends T> next, Predicate<? extends T> until);
}
```

so you can do `Iterator.iterate(StringUtils::randomString,s -> s.length() > 10)`

chb0github:
what about 
```
interface Iterator<T> {

    static <T> Iterator<T> iterate(Supplier<? extends T> next, Predicate<? extends T> until);
}
```

so you can do `Iterator.iterate(StringUtils::randomString,s -> s.length() > 10)`

valery1707:
@chb0github your `Iterator#iterate(Supplier<T>, Predicate<T>)` can be replaced with exists `Iterator.continually(Supplier<T>).takeWhile(Predicate<T>)`: `Iterator.continually(StringUtils::randomString).takeWhile(s -> s.length() > 10)`

valery1707:
Variant `iterate(Supplier<Option<? extends T>>)` is more suitable as I think.
This situation looks like the https://github.com/vavr-io/vavr/issues/2134#issuecomment-342297157

ruslansennov:
[...]

PR #2193, opened 2017-12-22
new methods: iterate(Supplier<Option>)

ruslansennov:
towards #2045

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Iterator.java
+++ b/vavr/src/main/java/io/vavr/collection/Iterator.java
@@ -979,6 +979,7 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
      *
      * @param supplier A Supplier of iterator values
      * @param <T> value type
+     * @param <A> supplier option type
      * @return A new {@code Iterator}
      * @throws NullPointerException if supplier produces null value
      */
--- a/vavr/src/main/java/io/vavr/collection/Iterator.java
+++ b/vavr/src/main/java/io/vavr/collection/Iterator.java
@@ -974,6 +974,37 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
     }
 
     /**
+     * Creates an iterator that repeatedly invokes the supplier
+     * while it's a {@code Some} and end on the first {@code None}
+     *
+     * @param supplier A Supplier of iterator values
+     * @param <T> value type
+     * @return A new {@code Iterator}
+     * @throws NullPointerException if supplier produces null value
+     */
+    static <T, A extends T> Iterator<T> iterate(Supplier<Option<A>> supplier) {
+        Objects.requireNonNull(supplier, "supplier is null");
+        return new AbstractIterator<T>() {
+            Option<A> nextOption;
+
+            @Override
+            public boolean hasNext() {
+                if (nextOption == null) {
+                    nextOption = supplier.get();
+                }
+                return nextOption.isDefined();
+            }
+
+            @Override
+            public T getNext() {
+                T next =  nextOption.get();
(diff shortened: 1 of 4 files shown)


vavr/src/main/java/io/vavr/collection/Iterator.java
```diff
@@ -973,6 +973,38 @@ public T getNext() {
...
+     * Creates an iterator that repeatedly invokes the supplier
+     * while it's a {@code Some} and end on the first {@code None}
+     *
+     * @param supplier A Supplier of iterator values
+     * @param <T> value type
+     * @param <A> supplier option type
+     * @return A new {@code Iterator}
+     * @throws NullPointerException if supplier produces null value
+     */
+    static <T, A extends T> Iterator<T> iterate(Supplier<Option<A>> supplier) {
```
ruslansennov: @nfekete @danieldietrich please review.
Here we should use `A extends T` to make it more convenient to use

nfekete: Looks good to me, but I admit I don't really know the advantage of this version over `static <T> Iterator<T> iterate(Supplier<Option<? extends T> supplier)`. Would you help me out with this?

vavr/src/main/java/io/vavr/collection/Stream.java
```diff
@@ -234,6 +234,20 @@
...
+     * that provides a {@code Some} for each next value and a {@code None} for the end
+     *
+     * @param supplier A Supplier of iterator values
+     * @param <T> value type
+     * @param <A> supplier option type
+     * @return A new Stream
+     */
+    static <T, A extends T> Stream<T> iterate(Supplier<Option<A>> supplier) {
+        Objects.requireNonNull(supplier, "supplier is null");
+        return Stream.ofAll(Iterator.iterate(supplier));
```
nfekete: Since `Stream` (contrary to `Iterator`) can be iterated multiple times, I think it's worth mentioning in the javadoc that the Stream will constructed in such a way that the `Supplier` will be invoked only that many times until it returns `None`, and repeated iteration over the stream will produce the same values in the same order, without any further invocations to the `Supplier`.

vavr/src/main/java/io/vavr/collection/Iterator.java
ruslansennov: `? extends T` works fine in trivial cases like lambdas, but fails on more complex and frequently occurring cases. For example see `IteratorTest.OptionSupplier`. If it don't implement `Supplier<Option<? extends Integer>>`, a compilation error will be made

vavr/src/main/java/io/vavr/collection/Stream.java
ruslansennov: thanks, done

nfekete:
👍 looks nice 😃

vavr/src/main/java/io/vavr/collection/Iterator.java
danieldietrich: Just stumbled upon this while trying to achieve inbox-zero :) Sorry for the late reply.
We should take care of uniformly using generic declarations across the project.
Maybe the addition of another parameter is a great trick to circumvent the need of type witnesses in the presence of several `? extends T` declarations - I don't know. It looks a bit like Scala's sort of type declaration now (declaration-site variance vs use-site variance).

I will create an issue to investigate it.

danieldietrich: I double-checked it. The correct declaration should look like this:

```java
static <T> Iterator<T> iterate(Supplier<? extends Option<? extends T>> supplier) { ... }
```

Then all compile errors disappear. I will do an updating PR.

Thx!

PR #2197, opened 2017-12-28
Removed dependency to java.util.concurrent.ExecutorService

### LINKED ISSUE #2194: Consider eliminating Future's dependency on ExecutorService
Currently, in order to create an `io.vavr.concurrent.Future` and instance of a `java.util.concurrent.ExecutorService` is required (assuming the `DEFAULT_EXECUTOR_SERVICE` is unacceptable). Since Java's native `java.util.concurrent.CompletableFuture` simply requires a `java.util.concurrent.Executor`, shouldn't it also be sufficient for Vavr's Futures?
The problem can be clearly illustrated when attempting to leverage Vavr's Futures when developing with Akka, since the `akka.dispatch.MessageDispatcher` does *not* implement `java.util.concurrent.ExecutorService` (nor should it).

danieldietrich:
Fixes #2194


vavr/src/main/java/io/vavr/concurrent/FutureImpl.java
```diff
@@ -104,50 +103,67 @@ private FutureImpl(ExecutorService executorService, Option<Try<T>> value, Queue<
...
             }
         }));
     }
 
+    private static Thread createThread(Executor executor, Runnable runnable) throws InterruptedException {
+        final AtomicReference<Thread> thread = new AtomicReference<>(null);
+        executor.execute(() -> {
+            synchronized (thread) {
+                thread.set(Thread.currentThread());
+                thread.notify();
```
danieldietrich: Using monitors in the presence of ForkJoinPool seems to deadlock the JVM (on the CI server). Instead of using a ManagedBlocker I will rewrite this without blocking.

PR #2201, opened 2018-01-01
Removed redundant generic parameter

danieldietrich:
Follow-up of #2193


danieldietrich:
Travis-CI seems to have problems. Local build works find. Will merge the PR now.

Issue #2191, opened 2017-12-15
Quadratic behavior in LinkedHashSet.union() and LinkedHashSet.addAll()


grievejia:
According to the [user guide](http://www.vavr.io/vavr-docs/#_performance_characteristics), `LinkedHashSet.add()` has an effective constant complexity, which implies that LinkedHashSet.addAll() and `LinkedHashSet.union()` should behave no worse than O(N). However, I run into a particular data sequence that would cause the aforementioned methods to have O(N^2) performance in vavr-0.9.2. I am not too familiar with the vavr codebase myself and I don't have too much time to look into this. Maybe it's just a really weird edge case. Still I'd like to share the data anyway if any dev is interested. 

Attached is the source code I used to generate the data sequence and the corresponding JMH codes for performance measurement. The `Test.compute()` method constructs two sequences `s0` and `s1`, which then get transformed into two `LinkedHashSet` and unioned together. Benchmarking the `LinkedHashSet.union()` or `LinkedHashSet.addAll()` method call alone for a couple of `INPUT_SIZE`s would reveal that the running time scales quadratically with the size. On the other hand, the quadratic behavior doesn't exist if I manually write a loop to add all elements from one set to another, or if I switch to `HashSet` instead.

[src.zip](https://github.com/vavr-io/vavr/files/1564031/src.zip)

nfekete:
`LinkedHashSet.add()` and it's backing `LinkedHashMap.put()` _can_ be linear in the worst case, when you're adding the same element over and over again, since, when the element is already contained in the set/map it's doing a replace on the existing element with the new instance in both the set/map and the queue used for keeping insertion order of the elements, and the cost of the latter is linear (see `Queue.replace`).

You can easily avoid this by checking if the set/map already contains the element to be inserted, and skip insertion if you don't actually need to replace the existing instance in the set/map. While this would effectively double the set/map lookup cost part of the insertion cost for new elements, it will be mostly a cache hit for the lookup so the cost would be negligible.

grievejia:
> when the element is already contained in the set/map it's doing a replace on the existing element with the new instance in both the set/map and the queue used for keeping insertion order of the elements

I see. Is this mentioned anywhere in the doc? The replacement behavior is quite surprising to me. Why not just refuse to add the new element if it's already contained?

[...]

PR #2204, opened 2018-01-02
Add some javadoc to clarify the time complexity of certain operations…

grievejia:
… in LinkedHashSet and LinkedHashMap. See issue #2191

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/LinkedHashSet.java
+++ b/vavr/src/main/java/io/vavr/collection/LinkedHashSet.java
@@ -485,11 +485,27 @@ public final class LinkedHashSet<T> implements Set<T>, Serializable {
         return LinkedHashSet.ofAll(Iterator.rangeClosedBy(from, toInclusive, step));
     }
 
+    /**
+     * Add the given element to this set, replacing existing one if it is already contained.
+     * <p>
+     * Note that this method has a worst-case linear complexity.
+     *
+     * @param element The element to be added.
+     * @return A new set containing all elements of this set and also {@code element}.
+     */
     @Override
     public LinkedHashSet<T> add(T element) {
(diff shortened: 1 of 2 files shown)


vavr/src/main/java/io/vavr/collection/LinkedHashSet.java
```diff
@@ -485,11 +485,27 @@ private LinkedHashSet(LinkedHashMap<T, Object> map) {
...
+     */
     @Override
     public LinkedHashSet<T> add(T element) {
         return contains(element) ? this : new LinkedHashSet<>(map.put(element, element));
     }
 
+    /**
+     * Adds all of the given elements to this set, replacing existing one if they are not already contained.
+     * <p>
+     * Note that this method has a worst-case quadratic complexity.
```
ruslansennov: Not applicable to this PR, but it seems that we can do this in linear time. `LinkedHashMap` consists of two objects: `List` and `HashMap`. In fact, we should two things here:
* maybe to remove `n` elements from `HashMap` and add `n` elements to `HashMap`. This can be done in linear time.
* maybe to remove `n` elements from `List` and add `n` elements to `List`. This also can be done in linear time. Something like this: `list.filter(additional::contains).appendAll(additional)`

nfekete: That would _slightly_ change the current behavior, that the elements in the  `List` are kept in the order they were _first inserted_ to the `List`, and thus to the `LinkedHashMap` and `LinkedHashSet` it is backing. Whether that is acceptable or even desirable, that I don't know ⛄️

ruslansennov: Yep, sorry :) Very old story.

ruslansennov:
thanks

Issue #1958, opened 2017-04-23
Make use of Collections.isEmpty(Iterable) where applicable


danieldietrich:
I think there are several methods than can be optimized.

For example CharSeq containts the following method:

```java
@Override
public CharSeq prependAll(Iterable<? extends Character> elements) {
    Objects.requireNonNull(elements, "elements is null");
    final StringBuilder sb = new StringBuilder();
    for (char element : elements) {
        sb.append(element);
    }
    sb.append(back);
    return sb.length() == 0 ? EMPTY : of(sb);
}
```

It can be optimized like this:

```java
@Override
public CharSeq prependAll(Iterable<? extends Character> elements) {
    Objects.requireNonNull(elements, "elements is null");
    if (Collections.isEmpty(elements)) {
        return this; // 1
    } else if (isEmpty()) {
        return CharSeq.ofAll(elements); // 2
    } else {
        final StringBuilder sb = new StringBuilder();
        for (char element : elements) {
            sb.append(element);
        }
        sb.append(back);
        return CharSeq.of(sb);
    }
}
```

Notes:

* 1: We save a StringBuilder instance, and some (empty) operations and checks
* 2: Same benefits as in 1. Also CharSeq.ofAll does contain additional optimizations, like instanceof CharSeq-shortcut etc.

PR #2208, opened 2018-01-31
Improved some methods in CharSeq, Vector, TreeSet and HashSet classes…

NataliiaPrivezentseva:
… using Collections.isEmpty(Iterable) method (#1958)

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/HashSet.java
+++ b/vavr/src/main/java/io/vavr/collection/HashSet.java
@@ -494,10 +494,11 @@ public final class HashSet<T> implements Set<T>, Serializable {
             final HashSet<T> set = (HashSet<T>) elements;
             return set;
         }
-        if (Collections.isEmpty(elements) || this.equals(elements)){
+        final HashArrayMappedTrie<T, T> that = addAll(tree, elements);
+        if (that.size() == tree.size()) {
             return this;
         } else {
-            return new HashSet<>(addAll(tree, elements));
+            return new HashSet<>(that);
         }
     }
 
--- a/vavr/src/main/java/io/vavr/collection/HashSet.java
+++ b/vavr/src/main/java/io/vavr/collection/HashSet.java
@@ -494,11 +494,10 @@ public final class HashSet<T> implements Set<T>, Serializable {
             final HashSet<T> set = (HashSet<T>) elements;
             return set;
         }
-        final HashArrayMappedTrie<T, T> that = addAll(tree, elements);
-        if (that.size() == tree.size()) {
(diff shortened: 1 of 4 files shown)


vavr/src/main/java/io/vavr/collection/HashSet.java
```diff
@@ -494,11 +494,10 @@ private HashSet(HashArrayMappedTrie<T, T> tree) {
             final HashSet<T> set = (HashSet<T>) elements;
             return set;
         }
-        final HashArrayMappedTrie<T, T> that = addAll(tree, elements);
-        if (that.size() == tree.size()) {
+        if (Collections.isEmpty(elements) || this.equals(elements)){
```
danieldietrich: The new way is only better when this HashSet and the elements are equal but it adds an overhead for all other cases. Please restore the previous version.

vavr/src/main/java/io/vavr/collection/TreeSet.java
```diff
@@ -532,17 +532,16 @@
     @Override
     public TreeSet<T> addAll(Iterable<? extends T> elements) {
         Objects.requireNonNull(elements, "elements is null");
+        if (Collections.isEmpty(elements) || this.equals(elements)){
```
danieldietrich: A TreeSet is a SortedSet. We can't check for equality here because insertion depends on the underlying Comparator. Please restore the previous version.

danieldietrich:
🤔 Interestingly the CI tests failed both Java 8 and Java 9 for two Euler tests... Maybe the upcoming changes will heal them again.

NataliiaPrivezentseva:
Thanks for comments! I'll revert those two methods. However it won't heal the problem with test (it is in Vector.ofAll() method). I have an idea, how to fix it.

ruslansennov:
@danieldietrich @NataliiaPrivezentseva it is [compatible](https://github.com/gwtproject/gwt/blob/master/user/super/com/google/gwt/emul/java/util/Collections.java#L331)

danieldietrich:
Thx!

ruslansennov:
Sorry, I posted a non-relevant link :( Here we use `vavr.collection.Collections` class but not implementation of `java.util.Collections`.

But all is OK, this test works fine, I checked it

```java
package io.vavr.collection;

import com.google.gwt.junit.client.GWTTestCase;

public class CollTestGwt extends GWTTestCase {

    public void testCollectionsIsEmpty() {
        List<Integer> list = List.of(42);
        assertFalse(Collections.isEmpty(list));
    }

    @Override
    public String getModuleName() {
        return "TestModule";
    }
}
```

danieldietrich:
Thanks for the changes and sorry for the delay!!

Issue #2171, opened 2017-11-17
For comprehensions on monadic types


imode69:
New to vavr and have recently started looking a for comprehensions to replace `a.flatMap(x => b.map(y => {...})` patterns with `for ( x <- a, y <- b) yield {...}`.

So I have this:

```java
    Option<Integer> a = Option.some(1);
    Option<Integer> b = Option.some(2);

    Option<Integer> s1 = a.flatMap(x -> b.map(y -> x + y));
    Option<Integer> s2 = For(a, b).yield((x, y) -> x + y).toOption();

    assertTrue(s1.eq(s2));
```
the toOption (same for Try) is necessary since For/Yield is defined only for Iterable<T>. Will this change in future versions of vavr?

danieldietrich:
Yes, we will add For methods for our monadic types, see #1629.

PR #2211, opened 2018-02-12
added method to help working with Validation when result is Seq of va…

CauchyPeano:
…lues

I found this method is useful while working with lists of values. 
For example when you need to validate list of values, and return either valid composite, or accumulated errors

### DIFF
--- a/vavr/src/main/java/io/vavr/control/Validation.java
+++ b/vavr/src/main/java/io/vavr/control/Validation.java
@@ -220,6 +220,27 @@ public interface Validation<E, T> extends Value<T>, Serializable {
     }
 
     /**
+     * Appends {@code Validation} instances into a single {@code Validation} by joining two {@code Seq} into
+     * one.
+     *
+     * @param <E>    value type in the case of invalid
+     * @param <T>    value type in the case of valid
+     * @param first first validation to be joined
+     * @param second second validation to be joined
+     * @return Merged validation: if both validations are valid then valid object is returned. If one of them is invalid
+     * then invalid object returned containing an accumulated List of errors.
+     * @throws NullPointerException if values is null
+     */
+    static <E, T> Validation<E, Seq<T>> append(Validation<E, Seq<T>> first, Validation<E, Seq<T>> second) {
+
+        Objects.requireNonNull(first, "first value is null");
+        Objects.requireNonNull(second, "second value is null");
+
+        return first.combine(second)
+                .ap(Seq::appendAll);
+    }
+
+    /**
      * Narrows a widened {@code Validation<? extends E, ? extends T>} to {@code Validation<E, T>}
      * by performing a type-safe cast. This is eligible because immutable/read-only
      * collections are covariant.
--- a/vavr/src/test/java/io/vavr/control/ValidationTest.java
+++ b/vavr/src/test/java/io/vavr/control/ValidationTest.java
@@ -179,6 +179,51 @@ public class ValidationTest extends AbstractValueTest {
         assertThat(actual).isEqualTo(Validation.invalid("error1", "error2", "error3", "error4"));
     }
 
+    // -- Validation.append
+
+    @Test(expected = NullPointerException.class)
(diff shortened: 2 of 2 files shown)


danieldietrich:
Hi Igor, thank you for your PR!

I think the new method does not scale well. It is a very specific use-case to cover `Seq<T>` only. Java users may prefer to use one of `java.util.*` collections instead, an `Iterable` or an array `T[]`.

Additionally, the API is not specific for `Validation` - all Vavr types than represent a Value, like `Option`, `Try`, etc. could have it.

The best way to achieve the result currently is:

```java
validation1.flatMap(seq1 -> validation2.map(seq1::appendAll));
```

Alternatively using Vavr's For-Comprehension (needs #2171 to be implemented):

```java
// currently the result is Iterable<Seq<String>>
// with #2171 the result will be Validation<E, Seq<T>>
For(validation1, validation2).yield(Seq::appendAll);
```

So I think we should close this PR.

CauchyPeano:
Hi @danieldietrich,

Thanks for looking into my PR! I understand your reasoning, but still have a feeling that this direction could be useful. I've actually implemented parser for my project using Validation and find this case of using Seq<> within validation not so rare. 

I've also introduced this not just by my use-case, but looking at this method: 
`static <E, T> Validation<E, Seq<T>> sequence(Iterable<? extends Validation<E, T>> values)`

It introduces into the play validation of sequences, and this concept seems to be reasonable part of validation class. That's why I think this was a logical step to start introducing following method. 
(It was also present in [scalaz](https://github.com/scalaz/scalaz/blob/60719683de098efb7470af19e45b4002614fc2f1/core/src/main/scala/scalaz/Validation.scala#L302))

That's my opinion :)
Best,
Igor

danieldietrich:
Scalaz has a general notion for that method. It isn't defined on a Scala companion object (which is the same as _static_ in Java) but as instance method of the class.

Also Scalaz does not restrict the _append_ method to Seq types but to Semigroups, i.e. all combinable types that obey specific rules.

Scalaz defines it once and it applies to all types. We would need to implement it as static method n-times on each type.

Additionally, it would lead to ambiguities, if we wanted to provide the same functionality for different inputs regarding the generic type argument. More specifically these methods could not exist side-by-side (from the perspective of a call-site):

```java
static <E, T> Validation<E, Seq<T>> append(Validation<E, Seq<T>> first, Validation<E, Seq<T>> second) { ... }

static <E, T> Validation<E, Set<T>> append(Validation<E, Set<T>> first, Validation<E, Set<T>> second) { ... }
```

Validation.sequence is of a different flavor. The return type is baked into the method name. In contrast to the example above different implementations could therefore co-exist. The only drawback is, that we also have to implement it n-times.

```java
static <E, T> Validation<E, Seq<T>> sequence(Iterable<? extends Validation<E, T>> values) { ... }

static <E, T> Validation<E, Set<T>> distinct(Iterable<? extends Validation<E, T>> values) { ... }
```

The static _sequence_ method exists because we do not have any alternatives. For _append_ it is a different case, we have a relatively concise alternative (see my [previous comment](https://github.com/vavr-io/vavr/pull/2211#issuecomment-365192767)).

```java
static <E, T> Validation<E, Seq<T>> sequence(Iterable<? extends Validation<E, T>> values) {
    List<E> errors = List.empty();
    List<T> list = List.empty();
    for (Validation<E, T> value : values) {
        if (value.isInvalid()) {
            errors = errors.prependAll(value.getErrors().reverse());
        } else if (errors.isEmpty()) {
            list = list.prepend(value.get());
        }
    }
    return errors.isEmpty() ? valid(list.reverse()) : invalidAll(errors.reverse());
}
```

---

These static methods generally are a pain in the ass from the perspective of a maintainer. We can't statically check, if all classes implement them (the right way). It is kind of code duplication.

CauchyPeano:
Hi @danieldietrich, 
I agree, here we are bumping into java's limitations. I've also noted that it would make sense to have it not as static method, but it's not feasible
Thanks for explanation!

danieldietrich:
Thanks!

Issue #2183, opened 2017-11-26
Let Future.andThen take a Function instead of a Consumer


danieldietrich:
Replace

```java
Future.andThen(Consumer<? super Try<T>> action)
```

with

```java
andThen(Function<? super Try<T>, ? extends U> action)
```

See [Scala's Future.andThen&#x28;&#x29;](http://www.scala-lang.org/api/current/scala/concurrent/Future.html#andThen[U]&#x28;pf:PartialFunction[scala.util.Try[T],U]&#x29;&#x28;implicitexecutor:scala.concurrent.ExecutionContext&#x29;:scala.concurrent.Future[T]), see also [this deprecation discussion](http://www.scala-archive.org/Future-andThen-should-be-deprecated-td4644489.html)

Issue #1629, opened 2016-10-18
We need more API.For methods


danieldietrich:
As described here: http://blog.vavr.io/redundancy-of-the-java-language/

CauchyPeano:
Hi @danieldietrich , link is broken, where can I find article?

nfekete:
@danieldietrich some of the internal links on the blog are also broken (they're absolute links pointing to javaslang.io instead of relative links). For example:

> In our [previous post](http://blog.javaslang.io/idiomatic-javaslang-the-for-comprehension/) we showed the Grill example...

danieldietrich:
Thank you for the hint! I will fix them!

CauchyPeano:
@danieldietrich following this broken links topic - there are also images links broken:
`These are the results:
![for-benchmark](/content/images/2016/10/for-benchmark-1.png) `

CauchyPeano:
Hi @danieldietrich ,

Can you maybe provide some examples how it should look like?

Do you expect to have `For()` method on every monadic type?

danieldietrich:
Hi @CauchyPeano,

> following this broken links topic - there are also images links broken:

I fixed the image - it wasn't the link, it was a changed behavior of the markdown renderer in the the Ghost blog version.

Before I used this snippet to change the size of an image:

```html
<div style="max-width:406px !important">
![for-benchmark](/content/images/2016/10/for-benchmark-1.png)
</div>
```

Now I have to use this code instead:

```html
<img style="max-width:406px !important" alt="for-benchmark" src="/content/images/2016/10/for-benchmark-1.png">
```

danieldietrich:
@CauchyPeano 

> Can you maybe provide some examples how it should look like?

We need to change the code generator for [vavr/src-gen/main/java/io/vavr/API.java](https://github.com/vavr-io/vavr/blob/f3afd61b84a77df8fa9ff0a87c2cfa10c7d15232/vavr/src-gen/main/java/io/vavr/API.java#L2229-L2732), located in [vavr/generator/Generator.scala](https://github.com/vavr-io/vavr/blob/f3afd61b84a77df8fa9ff0a87c2cfa10c7d15232/vavr/generator/Generator.scala#L682-L779).

Wlog, currently we have:

```java
public static <T1> For1<T1> For(Iterable<T1> ts1) { ... }
```

Basically, we need additional method overloads for all monadic types.

```java
public static <T1> For1<T1> For(Option<T1> monad1) { ... }
```

[...]

PR #2212, opened 2018-02-15
changed signature of Future.andThen (#2183)

CauchyPeano:
After careful reading of related thread discussion and reasoning, I've understood that only this small change is needed to be done.

Please correct me if i'm wrong.

### DIFF
--- a/vavr/src/main/java/io/vavr/concurrent/Future.java
+++ b/vavr/src/main/java/io/vavr/concurrent/Future.java
@@ -651,7 +651,7 @@ public interface Future<T> extends Value<T> {
     /**
      * Support for chaining of callbacks that are guaranteed to be executed in a specific order.
      * <p>
-     * An exception, which occurs when performing the given {@code action}, is not propagated to the outside.
+     * An exception, which occurs when performing the given {@code function}, is not propagated to the outside.
      * In other words, subsequent actions are performed based on the value of the original Future.
      * <p>
      * Example:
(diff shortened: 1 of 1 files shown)


danieldietrich:
Could you please point to the related discussion? I'm not aware of the topic. Thx!

CauchyPeano:
Yep, it's this one: https://github.com/vavr-io/vavr/issues/2183

danieldietrich:
Hi Igor,

I tried to recapitulate what my intention was creating issue #2183.

In Scala, we have different conditions than in Java. Scala does not distinguish between suppliers, consumers, runnables and functions - there are just... functions. This is because there is no distinguished return type 'void'.

This is the reason why Scala's Future.onComplete(), foreach() and andThen() take functions. The returned value is just ignored - it might be a value, or just Unit (see also [the difference between Nothing and Unit in Scala](https://stackoverflow.com/a/13539858/1110815)).

Future's methods onComplete(), forEach() and andThen() are intended to perform side-effects. In Java, the only way to express this is using Consumer as argument. **If we use Function, we effectively have to return something. This is the most important difference to Scala.**

I'm afraid, therefore we have to keep the original forEach(Consumer).

I will close this PR and also my issue. This is not the right way to do it in Java.

Sorry, I really appreciate your effort and your contribution! I hope you understand.

CauchyPeano:
Hi @danieldietrich,

no problem. Reasoning is completely clear for me. Luckily change was not so big :) 

If there any tickets which are bothering you and you need help with, I will be glad to work on them.

Cheers, 
Igor

danieldietrich:
@CauchyPeano thanks!

You could take a look at #1629. I will assist you on any questions!
It involves the code generator.

PR #2214, opened 2018-02-17
Added missing unchecked() methods to functional interfaces

### LINKED ISSUE #2213: vavr 0.9.2: .forEach that takes a CheckedConsumer
Would be convenient if the forEach method was overloaded for a CheckedConsumer or that CheckedConsumer has a default unchecked() method (both would be ideal).

Currently I'm doing

```Java
private static <T> Consumer<T> unchecked(CheckedConsumer<T> f) {
    return arg -> {
        try {
            f.accept(arg);
	} catch (Throwable e) {
            e.printStackTrace();
	}
    };
}

@Override	
public void write(ObjectDataOutput out, Stream<?> stream) throws IOException {
    stream.forEach(unchecked(o -> out.writeObject(o)));		
}
```

danieldietrich:
Fixes #2213 

(also fixed some minor javadoc issues reported by doclint)

### DIFF
--- a/vavr/src/main/java/io/vavr/CheckedConsumer.java
+++ b/vavr/src/main/java/io/vavr/CheckedConsumer.java
@@ -20,6 +20,9 @@
 package io.vavr;
 
 import java.util.Objects;
+import java.util.function.Consumer;
+
+import static io.vavr.CheckedConsumerModule.sneakyThrow;
 
 /**
  * A consumer that may throw, equivalent to {@linkplain java.util.function.Consumer}.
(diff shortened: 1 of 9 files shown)


vavr/src/main/java/io/vavr/CheckedConsumer.java
```diff
@@ -29,6 +32,29 @@
...
+     * }</pre>
+     *
+     * @param methodReference (typically) a method reference, e.g. {@code Type::method}
+     * @param <T> type of values that are accepted by the consumer
+     * @return a new {@code CheckedConsumer}
+     * @see CheckedFunction1#of(CheckedFunction1)
+     */
+    static <T> CheckedConsumer<T> of(CheckedConsumer<T> methodReference) {
+        return methodReference;
+    }
```
danieldietrich: We also need factory methods in order to be able to take advantage of the new unchecked() method.

Issue #2203, opened 2018-01-02
Unit-tests occasionally run forever


danieldietrich:
The latest changes applied to Future #2197 (removed dependency to ExecutorService) seem to have introduced a regression. See https://travis-ci.org/vavr-io/vavr/builds/323967434

It seems that the unit tests sometimes run in a deadlock situation. It is not clear if the tests are the cause or if FutureImpl.java has a bug. This needs to be investigated.

**Note:** In one case the Java 8 build succeeded and the Java 9 build failed with a timeout. In the other case it was the other way round. Therefore it has nothing to do with the JDK version.

danieldietrich:
Our Future implementation will be dropped. There will be no Vavr 1.0 module `vavr-concurrent`.

Reason: Implementing our own concurrency library is too error-prone, concurrency is a hard problem. There are many(!) alternatives, first of all Java's native CompletableFuture

PR #2220, opened 2018-03-13
Fix some typos and formatting

valery1707:


valery1707:
Failed on JDK9 with `FutureTest.shouldNotCrashWhenRecoverFails:852 Condition not met.`
Maybe tests run too slow?

danieldietrich:
Hi @valery1707, I will take a look today. We observed that 'flickering' also sometimes in the JDK8 build, see #2203 but it isn't reproducible on a local machine.

Issue #2215, opened 2018-03-02
Add concat methods to Tuple classes


jlorenzen:
It would be handy if the individual Tuple classes add concat or append methods to them. I've ran into two instances lately where they would have been helpful. The jOOL [Tuple](http://www.jooq.org/products/jOO%CE%BB/javadoc/0.9.5/org/jooq/lambda/tuple/Tuple2.html) classes have them. I would use them but the jOOQ Tuple classes don't have the `apple` method which I've come to love in the vavr Tuple classes. And I'm trying to standardize on using just vavr.

nfekete:
@jlorenzen do you mean _widening_ and _narrowing_ `Tuple`s from the right?

jlorenzen:
@nfekete Yeah I think so. Basically the following:
```java
final Tuple2<String, String> user = Tuple.of("John", "Doe");
final Tuple3<String, String, String> userWithEmail = user.concat("john.doe@example.com");
```

jlorenzen:
@nfekete @danieldietrich I would be interested in making these changes if you guys would welcome them and offer any suggestions?

ruslansennov:
@jlorenzen The `append` and `prepend` methods were added and removed twice. It was a difficult decision, see the [last discussion](https://github.com/vavr-io/vavr/pull/1559#discussion_r80514299).

jlorenzen:
@ruslansennov That last discussion didn't really go deep into why it was removed. There is hint of people not liking that `Tuple8` (the last numbered Tuple) had to throw an exception since there isn't a `Tuple9` class.
Looking at the last jOOL Tuple class, Tuple16, they just don't add a `concat` method. That would seem like a reasonable compromise.
Here is the reason I'd like to see the vavr Tuple classes have at least a `concat` or `append` method. Lately I've been using the `Validation` class to perform validation on HTTP request parameters. I then convert those to a Tuple to be used in the next step. So something like the following:
```java
private SignInResult signIn(final String email, final String password) {
    validate(email, password)
        .flatMap(emailAndPassword -> emailAndPassword.apply(signInService::signIn));
}

private Either<Throwable, Tuple2<Email, Password> validate(final String email, final String password) {
    // create validation objects for each param
    return Validation.combine(emailValidation, passwordValidation)
        .ap(Tuple::of)
        .toEither();
}
```

[...]

PR #2221, opened 2018-03-14
Add concat methods to Tuple classes

jlorenzen:
Updates the `Generator.scala` script to auto-generate the `concat` methods for the various Tuple classes for #2215.

This is the first stab at it. It's not perfect but I wanted to get some early feedback before I went further. Might be easier if you just checkout the branch and run the scrip to see the output. Below is a quick summary of what is done though.

What's done:
* `Tuple0` has a `concat(T1 t1)` method
* `Tuple1-Tuple7` have the following methods:
  * `concat(TN tn)`
  * multiple `concat(TupleN tuple)` methods

What's left:
* Need to finish adding the various `concat(TupleN tuple)` methods to `Tuple0`.
* Need javadocs on each method
* Tests

### DIFF
--- a/vavr/generator/Generator.scala
+++ b/vavr/generator/Generator.scala
@@ -2013,14 +2013,12 @@ def generateMainClasses(): Unit = {
 
             ${(i == 0).gen(xs"""
               public <T1> Tuple1<T1> concat(T1 t1) {
-                  Objects.requireNonNull(t1, "t1 is null");
                   return ${im.getType("io.vavr.Tuple")}.of(t1);
               }
             """)}
 
             ${(i > 0 && i < N).gen(xs"""
               public <T${i+1}> Tuple${i+1}<${(1 to i+1).gen(j => s"T$j")(", ")}> concat(T${i+1} t${i+1}) {
-                  Objects.requireNonNull(t${i+1}, "t${i+1} is null");
                   return ${im.getType("io.vavr.Tuple")}.of(${(1 to i).gen(k => s"_$k")(", ")}, t${i+1});
               }
             """)}
@@ -2031,7 +2029,6 @@ def generateMainClasses(): Unit = {
                * j=$j
                */
               public <${(i+1 to i+j).gen(k => s"T$k")(", ")}> Tuple${i+j}<${(1 to i+j).gen(k => s"T$k")(", ")}> concat(Tuple$j<${(i+1 to i+j).gen(k => s"T$k")(", ")}> tuple) {
-                  Objects.requireNonNull(tuple, "tuple is null");
                   return ${im.getType("io.vavr.Tuple")}.of(${(1 to i).gen(k => s"_$k")(", ")}, ${(1 to j).gen(k => s"tuple._$k")(", ")});
               }
             """)("\n\n")}
--- a/vavr/generator/Generator.scala
+++ b/vavr/generator/Generator.scala
@@ -3421,20 +3421,11 @@ def generateTestClasses(): Unit = {
                   assertThat(actual).isEqualTo(Tuple0.instance());
               }
 
-              ${(i == 0).gen(xs"""
+              ${(i < N).gen(xs"""
                 @$test
                 public void shouldAppendValue() {
-                    final Tuple1<Integer> actual = Tuple0.instance().append(1);
-                    final Tuple1<Integer> expected = Tuple.of(1);
-                    assertThat(actual).isEqualTo(expected);
-                }
-              """)}
-
-              ${(i > 0 && i < N).gen(xs"""
-                @$test
-                public void shouldAppendValue() {
-                    final Tuple${i+1}${if (i == 0) "" else s"<${(1 to i+1).gen(j => s"Integer")(", ")}>"} actual = Tuple.of(${(1 to i).gen(j => xs"$j")(", ")}).append(${i+1});
-                    final Tuple${i+1}${if (i == 0) "" else s"<${(1 to i+1).gen(j => s"Integer")(", ")}>"} expected = Tuple.of(${(1 to i+1).gen(j => xs"$j")(", ")});
+                    final Tuple${i+1}<${(1 to i+1).gen(j => s"Integer")(", ")}> actual = ${ if (i == 0) "Tuple0.instance()" else s"Tuple.of(${(1 to i).gen(j => xs"$j")(", ")})"}.append(${i+1});
+                    final Tuple${i+1}<${(1 to i+1).gen(j => s"Integer")(", ")}> expected = Tuple.of(${(1 to i+1).gen(j => xs"$j")(", ")});
                     assertThat(actual).isEqualTo(expected);
                 }
               """)}
--- a/vavr/generator/Generator.scala
+++ b/vavr/generator/Generator.scala
@@ -2010,26 +2010,20 @@ def generateMainClasses(): Unit = {
                 """}
             }
 
-            ${(i == 0).gen(xs"""
-              public <T1> Tuple1<T1> concat(T1 t1) {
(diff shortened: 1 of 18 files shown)


valery1707:
@jlorenzen you need add generated files into commit too

jlorenzen:
@valery1707 Ah ok. Sorry. I didn't realize those were checked in. I'll do that now.

jlorenzen:
Pushed up the generated files now

vavr/src-gen/main/java/io/vavr/Tuple0.java
```diff
@@ -95,6 +95,11 @@ public int compareTo(Tuple0 that) {
         return List.empty();
     }
 
+    public <T1> Tuple1<T1> concat(T1 t1) {
+        Objects.requireNonNull(t1, "t1 is null");
```
valery1707: Do we need this check here?

vavr/src-gen/main/java/io/vavr/Tuple1.java
```diff
@@ -142,6 +142,74 @@ public T1 _1() {
         return List.of(_1);
     }
 
+    public <T2> Tuple2<T1, T2> concat(T2 t2) {
+        Objects.requireNonNull(t2, "t2 is null");
```
valery1707: Do we need this check here?

vavr/src-gen/main/java/io/vavr/Tuple0.java
jlorenzen: Yeah I guess you are right. I'll remove those.

vavr/src-gen/main/java/io/vavr/Tuple1.java
```diff
@@ -143,7 +143,6 @@ public T1 _1() {
     }
 
     public <T2> Tuple2<T1, T2> concat(T2 t2) {
-        Objects.requireNonNull(t2, "t2 is null");
         return Tuple.of(_1, t2);
```
valery1707: We don't need to check simple values for `null` - this is user side.

vavr/src-gen/main/java/io/vavr/Tuple1.java
```diff
@@ -152,7 +151,6 @@ public T1 _1() {
      * j=1
      */
     public <T2> Tuple2<T1, T2> concat(Tuple1<T2> tuple) {
-        Objects.requireNonNull(tuple, "tuple is null");
         return Tuple.of(_1, tuple._1);
```
valery1707: But we must check `tuple` for `null` because we will go deeper with them

jlorenzen: Oh you're right. Good call. I'll put just those back.

vavr/src-gen/main/java/io/vavr/Tuple1.java
```diff
@@ -142,6 +141,73 @@ public T1 _1() {
         return List.of(_1);
     }
 
+    public <T2> Tuple2<T1, T2> concat(T2 t2) {
```
ruslansennov: It can be ambiguous with `concat(Tuple1<T2> tuple)` (in case if `T2` not shown intentionally). What do you think about rename this method to `add(T2)` and keep others as `concat(Tuple*<***>)`?

danieldietrich: Yes, we had problems in the past with overloads that have a generic value parameter. However, my tests do not lead to ambiguities. Unless we find a counter-example, I would keep same names.

```java
class Tuple2<T1, T2> {

    // ...

    public final <T3> Tuple3<T1, T2, T3> concat(T3 value) {
        return new Tuple3<>(_1, _2, value);
    }

    public final <T3> Tuple3<T1, T2, T3> concat(Tuple1<T3> tuple) {
        return new Tuple3<>(_1, _2, tuple._1);
    }

}

// both tests compile fine
final Tuple3<String, Boolean, Integer> testee1 = Tuple.of("a", true).concat(1);
final Tuple3<String, Boolean, Integer> testee2 = Tuple.of("a", true).concat(Tuple.of(1));
```

ruslansennov:
```java
// fail
Tuple3<String, Boolean, Integer> testee3 =
        Tuple.of("a", true).concat(Tuple.of(1)).map3(third -> third._1);
// compiles fine
Tuple3<String, Boolean, Integer> testee4 = 
        Tuple.of("a", true).<Tuple1<Integer>>concat(Tuple.of(1)).map3(third -> third._1);
```

jlorenzen:
Any other feedback? I'll probably work on this some more tonight.

vavr/src-gen/main/java/io/vavr/Tuple1.java
danieldietrich: Yes, Ruslan is right. In the case of values T1, ... Tn we should choose another name. I would prefer **`append`** over `add` because our sequential types also have `append` (in contrast to Set which has `add`).

danieldietrich:
All the other things conceptually look great!

* [x] Tuple0 has concat and append methods
* [x] Tuple1..Tuple8 do not concat Tuple0
* [x] we omit impossible methods instead at throwing at runtime
* [x] we use `concat` for Tuple args and `append` for value args

jlorenzen:
@danieldietrich @ruslansennov @valery1707 Ok I've made all the recommended changes. It's ready for a final review.

vavr/generator/Generator.scala
```diff
@@ -2011,6 +2010,32 @@ def generateMainClasses(): Unit = {
                 """}
             }
 
+            ${(i < N).gen(xs"""
+              /$javadoc
+               * Append a value to this tuple.
+               *
+               * @param t${i+1} the value to append
+               * @return a new Tuple with the value appended
```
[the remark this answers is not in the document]
jlorenzen: @valery1707 @danieldietrich I understand if this is a convention vavr has taken for their javadocs, but this seems odd and confusing to me. It makes it appear that the method takes two parameters when there really is only one parameter. Where does this convention come from? When I look at the javadocs for `java.util.List.add(E e)`, I don't see an additional javadoc line for the `E` type.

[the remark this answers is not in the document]
nfekete: Because that `<E>` type is documented at the class level since it's the generic type argument for the class itself, the `java.util.List.add(E e)` doesn't introduce a new type argument in the equation. See [`Tuple2.map1`](https://static.javadoc.io/io.vavr/vavr/0.9.2/io/vavr/Tuple2.html#map1-java.util.function.Function-) for an example on what is needed to be documented here.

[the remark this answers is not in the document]
valery1707: @jlorenzen JDK itself use JavaDoc for generic attibutes.
You can find this at `java.util.stream.Stream#map`: [JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#map-java.util.function.Function-) / [Source](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/stream/Stream.java#l178)

jlorenzen:
@danieldietrich Ok I updated the script to set the javadocs for the type parameters. I think it's ready for a final review and merge.
However, it failed in CI but the failure it unrelated to my changes. The test that failed was: `FutureTest.shouldNotCrashWhenRecoverFails:852 Condition not met.`. I'm guessing this might just be an anomaly. I would kick off the job again but I'm not sure how to do that.

danieldietrich:
Awesome, thank you! 👍🏼

_(the build is known to flicker - it is a CI problem with our concurrent tests, it isn't currently prio 1 for me to fix that)_

PR #2223, opened 2018-03-15
added type witness to make make GWT compiler happy

nfekete:


valery1707:
Sorry, this my fall :(
May be add some comment for this line?

nfekete:
It's ok, no worries. 
I don't believe a comment on this would have a good signal to noise ratio. Devs could just assume type witnesses are intentional and not remove them.

PR #2224, opened 2018-03-15
Fix missing generated test sources

nfekete:
this fixes the missing source folder for annotation processing generated test classes folder in Eclipse


danieldietrich:
Great, thank you Nándor!

PR #2229, opened 2018-03-19
Refactored Stream.flatMap internal iterator to a static class

audunhalland:
`Stream.flatMap` uses an inner class for iteration, with the effect of the result stream holding an unnecessary indirect reference to the head of the source stream, resulting in a "temporary" memory leak.

In my project which uses vavr I wrote a unit test to verify that memory is in fact cleaned up correctly during iteration of lazy streams, but it failed because of this bug.

I did a quick search for unit tests in vavr related to memory usage/leaks, but found nothing of relevance. So no test has been written for this change; it does not alter the behaviour of any _functions_ as such.

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Stream.java
+++ b/vavr/src/main/java/io/vavr/collection/Stream.java
@@ -1049,25 +1049,7 @@ public interface Stream<T> extends LinearSeq<T> {
     @Override
     default <U> Stream<U> flatMap(Function<? super T, ? extends Iterable<? extends U>> mapper) {
         Objects.requireNonNull(mapper, "mapper is null");
-        return isEmpty() ? Empty.instance() : Stream.ofAll(new Iterator<U>() {
-
-            final Iterator<? extends T> inputs = Stream.this.iterator();
-            java.util.Iterator<? extends U> current = java.util.Collections.emptyIterator();
-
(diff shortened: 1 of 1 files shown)


danieldietrich:
This is great, many thanks!

I was aware of the fact that (method local, anonymous) inner classes bind the implicit this (within a non-static context) and thought we only use inner classes in static methods. I will seize the opportunity to double-check the whole code-base.

Thanks again! 👍🏼

audunhalland:
Glad to be of help 👍

danieldietrich:
**Update:** There are 6 occurrences of new Iterator()... left in the codebase. All of them look ok to me (they produce no memory leaks). Especially Stream.cycle, which needs to hold a pointer to _this_.

screen shot 2018-03-19 at 22 54 40

audunhalland:
The others looked ok, yes.

Especially for `Stream`, I think it's important to not have unneeded references, because I figure it is commonly used to process large amounts of data, like in my case. When I think about it, the leak observed here would lead to the entire source stream being unable to become garbage collected during an ongoing consumption, because of the memoized elements, and the first Stream node indirectly referencing all items after it.

danieldietrich:
I will prepare a bugfix release 0.9.3 the next days.

PR #2241, opened 2018-04-09
Add peekInvalid on Validation

### LINKED ISSUE #2237: Validation peekInvalid
Hi guys,

Is there any reason why `Validation` doesn't have a `peekInvalid` method? Just like `Either#peekLeft` or `Try#onFailure`.

This is typically useful to log when a validation fails.

Example: 
```java
    default Validation<E, T> peekInvalid(Consumer<? super E> action) {
        if (isInvalid()) {
            action.accept(getError());
        }
        return this;
    }
```

Sir4ur0n:
Closes #2237

### DIFF
--- a/vavr/src/main/java/io/vavr/control/Validation.java
+++ b/vavr/src/main/java/io/vavr/control/Validation.java
@@ -740,6 +740,19 @@ public interface Validation<E, T> extends Value<T>, Serializable {
     }
 
     /**
+     * Consumes the errors if this is an Invalid.
+     *
+     * @param action The action that will be performed on the errors.
+     * @return this instance
+     */
+    default Validation<E, T> peekInvalid(Consumer<? super Seq<E>> action) {
+        if (isInvalid()) {
+            action.accept(getErrors());
+        }
+        return this;
+    }
+
+    /**
      * A {@code Validation}'s value is computed synchronously.
      *
      * @return false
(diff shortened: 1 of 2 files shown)


vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -739,6 +739,19 @@ default void forEach(Consumer<? super T> action) {
...
+     *
+     * @param action The action that will be performed on the errors.
+     * @return this instance
+     */
+    default Validation<E, T> peekInvalid(Consumer<? super Seq<E>> action) {
+        if (isInvalid()) {
+            action.accept(getErrors());
+        }
+        return this;
+    }
```
danieldietrich: I tested these:

```java
    default void test() {
        Validation<String, String> v = Validation.valid("A");

        Consumer<Seq<String>> action1 = System.out::println;
        v.peekInvalid(action1); // OK

        Consumer<Seq<CharSequence>> action2 = System.out::println;
        v.peekInvalid(action2); // TYPE ERROR

        Consumer<Iterable<String>> action3 = System.out::println;
        v.peekInvalid(action3); // OK

        Consumer<Iterable<CharSequence>> action4 = System.out::println;
        v.peekInvalid(action4); // TYPE ERROR
    }
```

However, I think your type signature is correct! Using `Consumer<? super Seq<? super E>> action` does not work.

PR #2242, opened 2018-04-26
Added merge operation on Either

jarekratajski:
I copied approach from Scalaz \/ disjunction, where I can end computations in a nice way - mapping both Left and Right side of Either to the same type. Thus eventually assigning it to a result (http response or String in a single step. 

At  the moment almost same can be achieved by `getOrElseGet` but if both sides are already mapped to the same type (like String. HttpEntity) it  is unnecessary complex call.

see merge at https://static.javadoc.io/org.scalaz/scalaz_2.12/7.3.0-M22/scalaz/$bslash$div.html

### DIFF
--- a/vavr/src/main/java/io/vavr/control/Either.java
+++ b/vavr/src/main/java/io/vavr/control/Either.java
@@ -110,12 +110,8 @@ public interface Either<L, R> extends Value<R>, Serializable {
      * @param <T> commonly assignable type
      * @return left or right - whichever is resolved
      */
-    static <T> T  merge(Either<T,T> either) {
-        if (either.isLeft()) {
-            return either.getLeft();
-        } else {
-            return either.get();
-        }
+    static <T> T  merge(Either<? extends T,T> either) {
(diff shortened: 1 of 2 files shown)


danieldietrich:
Hi Jarek, thank you for your PR!

A while ago we decided against a static `flatten` function in favor of the workaround `flatMap(t -> t)`.

Main reasons:

* we want to keep the amount of static methods small. Most of them exist only because of missing features within Java‘s type system
* we want to keep the API as thin as possible. We decide on a per case basis if ti is worth to add syntactic sugar
* ...

The characteristics of the `merge` function are very similar to those of `flatten`, so I will close this PR.

I hope you understand!

jarekratajski:
Thank you for the answer. Understood.  Not a problem at all. 
getOrElseGet works for me.
and indeed such specific "sugar" operations can be added per project to not pollute vavr api,
(in that case maybe I will use kotlin extension functions)

danieldietrich:
@jarekratajski thank you!

PR #2256, opened 2018-06-14
add Either.filterEither

emmanueltouzery:
so the Either.filter that vavr currently offers doesn't seem super useful to me (it returns an Option<Either>).
I much prefer Try.filter and the various Try.filterTry variants. This is bringing that to Either. We could have two overloads, with one accepting a simple value and the other one a callback, for now I did only the callback one as I'm not a fan of overloading, but we can add that one too. We could also name the method Either.filter directly.

Unfortunately we can't get a perfect symmetry between Try and Either, since Either needs that second parameter, and since we already have that different Either.filter.


enolive:
I'd suggest filterOrElse as the name of this new functionality to be conform with the implementation in Scala and other languages that were inspired by it. See also 
https://www.scala-lang.org/api/current/scala/util/Either.html#filterOrElse[A1%3E:A](p:B=%3EBoolean,zero:=%3EA1):scala.util.Either[A1,B]

emmanueltouzery:
no issue in renaming to `filterOrElse` if that's what vavr authors prefer.

vavr/src/main/java/io/vavr/control/Either.java
```diff
@@ -414,6 +414,44 @@ default void orElseRun(Consumer<? super L> action) {
...
+     * Filters this right-biased {@code Either} by testing a predicate.
+     * If the {@code Either} is a Right and the predicate doesn't match, the
+     * {@code Either} will be turned into a {@code Left} with contents computed by applying
+     * the filterVal function to the {@code Either} value.
+     * <p>
+     *
+     * <pre>{@code
+     * import static io.vavr.API.*;
+     *
+     * class Example {
```
danieldietrich: I would either remove the class declaration (and keep the example/body) or use a static initializer because the current code is not valid Java.

vavr/src/main/java/io/vavr/control/Either.java
```diff
@@ -414,6 +414,44 @@ default void orElseRun(Consumer<? super L> action) {
...
+     *
+     * }
+     * }</pre>
+     *
+     * @param predicate A predicate
+     * @return an {@code Either} instance
+     * @throws NullPointerException if {@code predicate} is null
+     */
+    default Either<L,R> filterEither(Predicate<? super R> predicate,
+                                     Function<? super R, ? extends L> filterVal) {
```
danieldietrich: Let's rename `filterEither` to `filterOrElse` in order to align a bit more to Scala (see [Scala API](https://www.scala-lang.org/api/2.12.5/scala/util/Either.html#filterOrElse[A1>:A](p:B=>Boolean,zero:=>A1):scala.util.Either[A1,B])).

Maybe we should rename `filterVal` also to `zero`, it puts more emphasis that it is used in the empty case.

Scala does have a lazy zero provider `zero: => L`. But I think your choice is better in terms of flexibility. One does not have to use the right value in order to derive a left value. In other cases we use additional overloads (_Supplier&lt;L> zero_ and _L zero_), but I'm not a big fan of adding such a verbosity to the API.

vavr/src/main/java/io/vavr/control/Either.java
```diff
@@ -414,6 +414,44 @@ default void orElseRun(Consumer<? super L> action) {
...
+        Objects.requireNonNull(predicate, "predicate is null");
+        Objects.requireNonNull(filterVal, "filterVal is null");
+        if (isLeft()) {
+            return this;
+        } else if (!predicate.test(get())) {
+            return Either.left(filterVal.apply(get()));
+        } else {
+            return this;
+        }
```
danieldietrich: This could be simplified to

```java
if (isLeft() || predicate.test(get())) {
    return this;
} else {
    return Either.left(filterVal.apply(get()));
}
```

or even

```java
return isLeft() || predicate.test(get()) ? this : left(zero.apply(get()));
```

emmanueltouzery:
Thanks for the feedback! I'm currently in holidays, i can implement the changes from august 4. I'll get to it as soon as i can.

danieldietrich:
Many thanks! I'm in holidays, too :) Have fun!

emmanueltouzery:
ok i have applied the requested changes. this `class Example` in the javadoc was in several spots in Either.java so I made a separate commit to remove them, but you can squash both commits if you prefer. Let me know you wish for further changes :-)

emmanueltouzery:
good point, fixed that bit of the apidoc by updating the second commit!

agrison:
I have the exact same problem, I can see that this has been implemented even in 0.9.x but I can't find it in the official 0.9.2 release which appears to have not been updated since Nov 2017.

Is there any release planned for the 0.9.x branch which includes these improvements ? I'm stuck with Java 8 and can't move to Java 11 at work for the moment and I'd really need this feature to remove some ugly code :smiley:

danieldietrich:
Hi @agrison,
yes, I understand the need of releasing some of the new features.
I think it makes sense to release a 0.9.3 or 0.10.0 before 1.0.0 happens (which will still take some time).
Btw - 1.0.0 will be Java 8 compatible :) JDK 11 makes no sense for us, most users are still on JDK 8.

PR #2257, opened 2018-06-14
add either, option, try, validation traverse. improve the signature of validation.sequence

emmanueltouzery:
vavr has Future.traverse and Future.sequence, but for option, try, validation and either it has only sequence.

This PR adds traverse for all of these. Traverse is a pretty standard FP function (arguably even more standard than sequence).

The PR also slightly improves the Validation.sequence signature, regarding variance of the parameters (in a separate commit).

### DIFF
--- a/vavr/src/main/java/io/vavr/control/Validation.java
+++ b/vavr/src/main/java/io/vavr/control/Validation.java
@@ -220,6 +220,24 @@ public interface Validation<E, T> extends Value<T>, Serializable {
     }
 
     /**
+     * Maps the values of an iterable to a sequence of mapped values into a single {@code Validation} by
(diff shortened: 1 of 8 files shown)


vavr/src/main/java/io/vavr/control/Validation.java
```diff
@@ -205,11 +205,11 @@
      * or an invalid Validation containing an accumulated List of errors.
      * @throws NullPointerException if values is null
      */
-    static <E, T> Validation<E, Seq<T>> sequence(Iterable<? extends Validation<E, T>> values) {
+    static <E, T> Validation<E, Seq<T>> sequence(Iterable<? extends Validation<? extends E, ? extends T>> values) {
```
danieldietrich: Great, we missed that :) Thx!

PR #2275, opened 2018-07-30
Modularization: moving hash calculation to tuples, removing dead code

danieldietrich:


danieldietrich:
(sporadic jdk 9 build error...)

PR #2276, opened 2018-07-31
Modularization: made PartialFunction independent of Option and Value.

danieldietrich:
Applied the boy-scout rule (a good deed every day) and aligned PartialFunction to Scala 2.12.

_(Note: Scala's companion objects generally might contain additional functionality that we did not consider, yet. Not only for PartialFunction.)_

### DIFF
--- a/vavr/src/main/java/io/vavr/PartialFunction.java
+++ b/vavr/src/main/java/io/vavr/PartialFunction.java
@@ -19,7 +19,8 @@
  */
 package io.vavr;
 
-import io.vavr.control.Option;
+import java.util.function.Consumer;
+import java.util.function.Function;
 
 /**
  * Represents a partial function T -&gt; R that is not necessarily defined for all input values of type T.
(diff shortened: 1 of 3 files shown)


vavr/src/main/java/io/vavr/PartialFunction.java
```diff
@@ -42,86 +43,116 @@
...
-     * @return a partial function that is not necessarily defined for all input values of type T.
      */
-    static <T, R> PartialFunction<T, R> unlift(Function1<? super T, ? extends Option<? extends R>> totalFunction) {
-        return new PartialFunction<T, R>() {
-
-            private static final long serialVersionUID = 1L;
-
-            @Override
-            public R apply(T t) {
-                return totalFunction.apply(t).get();
```
danieldietrich: This was really eval. The main purpose of a PartialFunction is to provide _fast_ tests before applying the function.

danieldietrich: However, we don't really lose s.th. here. If we decided that we really need this functionality, we could add static methods to Option, e.g.

```java
final class Option<T> {
    static <T, R> PartialFunction<T, R> unlift(Function1<? super T, ? extends Option<? extends R>> totalFunction) {
        ...
    }
}
```

But for now I don't see the necessity.

PR #2281, opened 2018-08-03
Initial implementation of new For() constructs (#1629)

CauchyPeano:
I've decided for beginning do it straightforward:
There are (Optional, Try, Future, List) x (1..8) =  32 classes. Maybe it's possible to simplify this amount a bit but I haven't found this yet.

### DIFF
--- a/vavr-benchmark/src/test/java/io/vavr/idiom/ForBenchmark.java
+++ b/vavr-benchmark/src/test/java/io/vavr/idiom/ForBenchmark.java
@@ -59,13 +59,13 @@ public class ForBenchmark {
     public static class Base {
         int CONTAINER_SIZE = 1000;
         int AGGREGATE;
-        List<Integer> ELEMENTS;
+        io.vavr.collection.List<Integer> ELEMENTS;
 
         final BiFunction<Integer, Integer, Integer> AGGREGATOR = (i, j) -> i ^ j;
 
         @Setup
         public void setup() {
-            ELEMENTS = Arrays.asList(getRandomValues(CONTAINER_SIZE, 0, true));
+            ELEMENTS = io.vavr.collection.List.ofAll(Arrays.stream(getRandomValues(CONTAINER_SIZE, 0, true)));
 
             AGGREGATE = 0;
             for (Integer i : ELEMENTS) {
--- a/vavr/generator/Generator.scala
+++ b/vavr/generator/Generator.scala
@@ -100,6 +100,8 @@ def generateMainClasses(): Unit = {
       val PredicateType = im.getType("java.util.function.Predicate")
       val SupplierType = im.getType("java.util.function.Supplier")
 
+      val monadicTypesFor = List(OptionType, FutureType, TryType, ListType)
+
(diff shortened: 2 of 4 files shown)


danieldietrich:
Thanks! This looks great for now.

I will increase the pace a bit and merge it. Then I'm able to go on with my Vavr 1.0 tasks.
We will add more/change the For comprehensions later, if necessary.
For now I also have no better solution to abstract over the overloaded methods in order to reduce the method amount.

We would need a common Monadic interface that is implemented by our 'sealed' classes (like Option, Try, List, ...). This type has to be a recursive self-type, like:

```java
interface Monadic<M extends Monadic<M, ?>, T> {

    <U> Monadic<M, U> flatMap(Function<? super T, ? extends Monadic<? extends M, ? extends U>> mapper);

    <U> Monadic<M, U> map(Function<? super T, ? extends U> mapper);
}
```

I will remove the common `Value` interface, because it does not work well in conjunction with checked functions. E.g. Option will have `flatMap(Function)`, Try will have `flatMap(CheckedFunction)`.

Given that, we would also need a CheckedMonadic interface:

```java
interface Monadic<M extends Monadic<M, ?>, T> {

    <U> Monadic<M, U> flatMap(CheckedFunction<? super T, ? extends Monadic<? extends M, ? extends U>> mapper);

    <U> Monadic<M, U> map(CheckedFunction<? super T, ? extends U> mapper);
}
```

Using the above interfaces, we could reduce the overall amount of For methods to (Monadic, CheckedMonadic) x (1..8) = 16 classes. This amount will stay constants when supporting more types, like Either, Future, Stream, ...

However, the flatMap API will be uglified regarding the sketched solution because the _mapper_ requires a recursive self-type (e.g. `CheckedMonadic<Try<?>, U>`) instead of a concrete type `Try<U>`. In turn, the return types of _map_ and _flatMap_ could be overloaded with concrete types.

I will defer the decision a bit if we use recursive self-types for map and flatMap (but I think, it is the right solution and it will open a whole new space for creating new APIs/better functionality).

CauchyPeano:
Hi @danieldietrich,

Cool, I'm looking forward to continue helping here.

danieldietrich:
Thx! I will take a look this evening where to go next! 😊

danieldietrich:
Hi @CauchyPeano,

I've spend some time investing the 'catamorphism' space a bit and added a Monadic interface as described above. I don't want to push Vavr in that direction. It would add an API that opens Vavr in a way that allows us to write algebraic APIs. For example we could expand the algebraic surfance by adding Monoids, Functors, etc.

It is not a goal of Vavr to let user know about the existence of such abstractions. Not only that the API gets uglified (by requiring instances of Monadic&lt;M, T> instead of M&lt;T>), but also because we currently need it only for our For-comprehensions. Those are part of a separate module that gets obsolete if Java language architects decide to add a similar feature. Additionally we have the fallback solution of adding manifold overloads.

Java is not the right language for purely functional programming.

---

I think for now, the support for Try, Option, List and Future is sufficient. I will close the original issue.

PR #2282, opened 2018-08-04
Modularization: decoupled Function* from Option, Try.

danieldietrich:
Removed Lambda and sneakyThrow / unchecked functionality.


danieldietrich:
@CauchyPeano I will wait with pulling this, until we merged #2281 (in order to prevent merge conflicts regarding your branch).

danieldietrich:
I think the function memoization re-implementation saved us 1-2 minutes build time. That's round about 5-10% build time.

PR #2294, opened 2018-09-05
Split vavr-match into two modules: annotation and processor (#2146)

### LINKED ISSUE #2146: [modules] The actual annotations and annotation processors should be in separate jar files
Then we can add `vavr-match.jar` (which includes only the annotations) to the classpath and the `vavr-match-processor` to the processor-classpath.

See also https://github.com/vavr-io/vavr/issues/985#issuecomment-335446401

CauchyPeano:
I guess I haven't missed here anything. 
Cheers!

Fixes #2146


CauchyPeano:
Hm, somehow now build for java8 is only failing. Can we restart this build?

ruslansennov:
> Can we restart this build?

done

CauchyPeano:
Hi @danieldietrich ,

No need to excuse, I really appreciate your time. I've fixed that name. 
Btw how this relates to migration to gradle? I've realized that recently...

danieldietrich:
Hi @CauchyPeano, thanks for applying the changes!
We will move to Gradle in baby steps.
First the control package, second the collection package, then API + annotations + processor. So there is no hurry.
The project is too complex in order to do it in a ‚big bang‘. Complexity can be handled best by slicing things into smaller, simple parts. Namely: modules.

danieldietrich:
Many thanks!

PR #2312, opened 2018-10-09
Deprecate Value.toJavaArray(Class<T>), add toJavaArray(IntFunction<T[]>)

### LINKED ISSUE #2311: Deprecate Value.toJavaArray(Class<T>) in favor of toJavaArray(IntFunction<T[]>)
Besides `Value.toJavaArray(Class<T>)` using reflection to create the array with the correct component type, it's very clumsy when you need to create arrays of generic type. For example, given:
```java
List<Class<? extends Number>> list = List.of(1, 2f).map(Number::getClass);
```
one cannot specify the component type to satisfy the compiler type checker. So the following statement will result in a syntax error:
```java
Class<? extends Number>[] classes = list.toJavaArray(Class<? extends Number>.class);
```
And direct casting `Class.class` to the needed component type (`Class<Class<? extends Number>>` in this case) isn't possible either, so one needs to do a double cast like this:
```java
Class<? extends Number>[] classes = list.toJavaArray(
                                  (Class<Class<? extends Number>>) (Object) Class.class);
```
The proposed signature would allow for
```
Class<? extends Number>[] classes = list.toJavaArray(Class[]::new);
```
which is much more elegant and that's what the JDK Stream API is also using.

nfekete:
Fixes #2311

### DIFF
--- a/vavr-benchmark/src/test/java/io/vavr/collection/MapBenchmark.java
+++ b/vavr-benchmark/src/test/java/io/vavr/collection/MapBenchmark.java
@@ -78,7 +78,7 @@ public class MapBenchmark {
         public void setup() {
             ELEMENTS = getRandomValues(CONTAINER_SIZE, 0);
             sampleTreeMap = index(ELEMENTS);
-            KEYS = predicableShuffle(sampleTreeMap.keySet().toJavaArray(Integer.class));
+            KEYS = predicableShuffle(sampleTreeMap.keySet().toJavaArray(Integer[]::new));
             REMOVAL = predicableShuffle(KEYS.clone());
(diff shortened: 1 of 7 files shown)


danieldietrich:
@nfekete for me it is ok to ignore deprecation warnings on the class level, but only in src/test (in order to fix the build). We should not switch off this compiler warning in the pom.

nfekete:
Yes, I can mark individual tests to suppress deprecation warnings as well. And I could migrate most of the tests to the new method too, so that the new one is also hit by the test coverage, maybe leave one test on the deprecated one so that it stays covered by tests. What do you think?

danieldietrich:
That sounds good!

danieldietrich:
Thank you, Nándor!

nfekete:
Thank you too!

PR #2322, opened 2018-11-03
Describe possible build problems and how to overcome them

jbrains:
Addresses issue https://github.com/vavr-io/vavr/issues/2321, but doesn't "fix" it, because it doesn't change the version of the Scala plugin for Maven. It's not clear to me how many other people have encountered/will encounter this issue.


danieldietrich:
Great!
Many thanks for your effort and the detailed description & scripts!

PR #2323, opened 2018-11-03
sum() for more "interesting" value types

jbrains:
~~**This pull request isn't really ready yet**~~ (see below), but I'm offering it looking for feedback before trying to go any further. I've built everything alongside Vavr without actually moving it into Vavr yet--the production code belongs in `io.vavr.collection` and some of it would move onto `Traversable` (or even `Foldable`). If you folks don't want it at all, then I won't spend the energy trying to integrate it.

I try to show Vavr to every Java programmer I meet in my work as a trainer and coach.

I was working on one of my TDD teaching examples and wanted to be able to invoke `sum()` on a collection of value objects that acted like numbers: `MonetaryAmount`. I noticed that `Traversable.sum()` only operates on `Number` and primitive numeric types. I thought it might be nice to try to contribute a simple implementation to Vavr for other people who want this behavior, so that people don't have to roll their own every time they encounter this situation. Since Vavr encourages a functional style and Whole Value Types are typical of the functional style, this seemed like a nice fit.

Of course, `sum()` needs a description of how to sum things, which is a `Monoid`. I modeled this as an interface that knows the identity element and "add" function for the value object type. The tests include two examples: one numeric and one non-numeric (based on `String`). If you folks accept this idea, then I could write a short article on how to write `Monoid`s for various typical value object types. It's not hard and the way I've done it in `IntegerValue` and `StringValue` shows a simple convention: a value type class, its monoid class, and a function `monoid()` on the value type class that returns a singleton instance of the monoid.

I chose the word "monoid" because it is accurate, but I know that not every programmer is familiar with the formal definition of a monoid, even if they broadly understand that `sum = reduce identity add` and that `add` doesn't really have to be "add", but any associative binary operation. I played around with other, more-intuitive, less-formal-sounding names and nothing really sounded better than "monoid", so I went with "monoid".

I have also provided a kind of contract test template for checking `sum()` (see `SumArbitraryValueObjectsTest`). It could pull double duty as a contract test for people to write their own monoid implementations as well as to use a few sample Value Object types and their monoids to test `sum()` itself.

If you folks don't want this at all, then I probably won't bother finishing it. If you'd like me to hammer it into shape, then I'm happy to accept your demands and hear your suggestions. I probably didn't follow your coding conventions very well, but I tried to do reasonable things.

Thanks.


jbrains:
The first commit involves bumping the version of the Scala plugin for Maven. I'd obviously clean that up in the final pull request.

danieldietrich:
Thank you! I will wait for your signal when the PR is final/ready.

I don't know were the 'generic' reduction methods like `sum` will go. I promised to make Vavr 1.0 safer in that regard...

jbrains:
I'll hammer it into shape, then, and put it where it seems to make sense to go. Thanks.

jbrains:
I now submit this pull request "for real". I accept your feedback. Thank you.

jbrains:
I now submit this pull request "for real". I accept your feedback. Thank you.

danieldietrich:
Hi J.B.!

it seems I totally missed the purpose of your PR. When you added the PR, I only looked at the code changes and it seemed to me that you wanted to add only some changes to the src/test/java part.

I'm afraid, we will not add algebraic abstractions like Monoid (or Functor, Monad, Applicative, Kleisli, ...) to Vavr. In fact I did it myself in the early days of Vavr (formerly Javaslang). Where to start, where to stop?

I soon recognized, that Java is not the right language for such algebraic extensions. In order to define Monad.flatMap, higher kinded types (HKTs) are helpful. However, HKT's can't be expressed with Java's type system.

To that time the overall strategy for Vavr emerged - aligning tightly to Scala. Scala itself does not contain algebraic extensions. For example, a concrete class like String or List should not extend the Monoid interface because the Monoid property is not unique. There are many possibilities to define the 'combine' operation. It could be (but not necessarily have the be) concatenation for example.

Having algebraic abstractions means, to express programs on that level. The code will operate (read: be lifted) on Monoids, Functors, Monads etc. Then, it does not matter if we operate on Lists or Strings or whatever. It can be easily formally proved that the code will be correct, if the corresponding types obey the algebraic abstractions mentioned above.

I do not want to push Vavr in that direction. It is too abstract for (literally spoken) 99% of the Java users. The remaining 1% will most probably **not** chose Java, because it is too restrictive/not able to express algebraic thoughts in an appropriate way.

Given that, the introduction of Monoid et al look more like a playground or proof of concept instead of covering real-world use cases. I fully understand your interest in that topic. I also played around with it. But it is far from being a practical approach in Java.

I'm not the only one who thinks that way. Most original committers of "Functional Java" turned away from their Java repository. Many of them also have roots in the Scalaz project, which is exactly that: algebraic structures for Scala. But as I said, in Java it does not make much sense. Soon we are confronted with shortcomings of Java.

I hope you understand that I have to close the PR.

I want to apologize because I did not recognize the direction of this PR (and that there does not exist a readme, yet, that reflects our overall strategy).

\- Daniel

jbrains:
I understand. I didn't intend Monoid to go in that direction, but rather to provide a convenience for expressing `sum()` as `fold()` that "better" (purely my opinion) expresses intent. When I looked for a name for the thing that combines an identity element and an "add" function, I realized that it should be "Monoid", even though I really wanted to avoid the arguably misleading formalism. Your reaction confirms my suspicion.

I see it the other way: a method like `sum(Monoid)` might help would-be functional programmers in Java learn about fold/reduce without being forced to use arcane-sounding words to describe a familiar concept. Perhaps it works better for `sum()` to be a literal alias for `foldLeft()`.

On the one hand, I would have preferred to know this before I invested energy into doing it. That is why I asked for feedback before going too far. On the other hand, should I have the urge to try to contribute to this code base in the future, I now know how to do it.

danieldietrich:
> Perhaps it works better for sum() to be a literal alias for foldLeft().

I think that users should learn about fold / reduce. These are the most important methods of our collections. Nearly all other operations can be expressed based on them.

The main problem is that sum() (without arguments) is unsafe. It will disappear in Vavr 1.0.

```java
// safe
List(1, 2, 3).fold(0, (a, b) -> a + b);

// unsafe
List("a", "b", "c").sum();
```

If Java would support tuples as first class citizens it would be straight forward:

```java
// = Tuple2<int, BiOperator<int>>
var IntSum = (0, (a, b) -> a + b);

// fold takes 2 arguments, which is effectively a Tuple2
var sum = List(1, 2, 3).fold(IntSum);
```

> That is why I asked for feedback before going too far.

Yes, that's my bad. I owe you a ☕️ (or two)

danieldietrich:
One more thought: We avoid introducing aliases in favor of keeping the API surface area small. It is a strategic decision because of simplicity (user perspective) and maintainability (maintainer perspective).

jbrains:
Thanks for explaining. I think I understand. No worries.

Issue #2137, opened 2017-10-11
Why are Checked* classes declared to throw Throwable


mrpotes:
All `Throwable` instances must either be an `Exception` or an `Error`. The javadoc for `Error` states that:

> An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch.

However, because Vavr lambda classes generally declare to throw `Throwable`, developers end up writing this sort of thing:

```java
public void doSomething(CheckedRunnable thing) throws MyException {
  try {
    thing.run();
  } catch (SomeExpectedException e) {
     // recover/log/rethrow/whatever
  } catch (Throwable e) { // have to catch because CheckedRunnable declares to throw
    throw new IllegalStateException("I only expected SomeExpectedException", e);
  }
}
```

But we've now ended up with any `Error` being caught and handled by the application. Obviously this can be fixed by doing something like:

```java
  } catch (Exception e) {
    throw new IllegalStateException("I only expected SomeExpectedException", e);
  } catch (Throwable t) {
    throw (Error) t; // we know this must now be an Error
  }
```

But as `Error` doesn't have to (or indeed, based on the javadoc, shouldn't) be caught, we really shouldn't have to be doing this.

This would be avoided if the `Checked*` signatures instead were:

```java
@FunctionalInterface
public interface CheckedRunnable {
    void run() throws Exception;
}
```

And the resulting calling code would be safer.

Aside: I'd actually prefer that the signatures were something like:

```java
@FunctionalInterface
public interface CheckedRunnable<E extends Exception> {
    void run() throws E;
}
```

or perhaps, to remind users that a lambda may throw a `RuntimeException`:

```java
@FunctionalInterface
public interface CheckedRunnable<E extends Exception> {
    void run() throws E, RuntimeException;
}
```

chb0github:
I don't like how `Throwable` is caught either. It's can hide real significant problems - problems like what you point out. using `Try.of(() -> connectToDb())` makes sense to catch errors and try again or us a circuit breaker. But an `ExceptionInInitializerError` means that a static block in a class threw an error and the class failed to load. Based experience it's nearly impossible to determine why (the underlying problem usually vanishes).

Errors are _bad_ and are by designed meant to not be caught: 
1. `OutOfMemoryError` - You can't increase JVM Heap space from inside the JVM, so now what?
2. `UnsatifiedLinkError` - An attempt at a native library load failed. Was the library missing? Is it binary incompatible (which, based on experience reports as not found). 
[...]

PR #2352, opened 2019-01-19
Reverted CheckedFunction*.apply throws Exception -> throws Throwable

danieldietrich:


nfekete:
@danieldietrich #2137 doesn't apply anymore?

danieldietrich:
@nfekete it still applies but the v0.9.3 unit tests fail if I run them with the v0.10.0 version. I would break applications.

In v0.10.0 CheckedFunctions will behave like in v0.9.3. We will ship that feature/change in v1.0.0.

nfekete:
Oh, I see, my mistake. I thought this was already in the 0.9.x line. 🤦‍♂️

PR #2360, opened 2019-01-25
Let Try's functional interfaces handle Throwable instead of Exception

danieldietrich:
Reported: https://twitter.com/odrotbohm/status/1088725153423478784
Related to #2137

### DIFF
--- a/src/main/java/io/vavr/control/CheckedConsumer.java
+++ b/src/main/java/io/vavr/control/CheckedConsumer.java
@@ -21,7 +21,8 @@ package io.vavr.control;
 import java.util.Objects;
 
 /**
- * A function that consumes an argument, returns nothing and may throw an {@link Exception}.
+ * A function that consumes an argument, returns nothing and may throw a {@link Throwable}.
+ * <p>
  * Such a function is expected to perform side-effects, i.e. altering the outer world.
  *
  * @param <T> the type of the consumed argument
@@ -35,9 +36,9 @@ public interface CheckedConsumer<T> {
      * Consumes the given argument {@code t} by performing an action.
      *
      * @param t the input argument
-     * @throws Exception if an error occurs when performing the action
+     * @throws Throwable if an error occurs when performing the action
      */
-    void accept(T t) throws Exception;
+    void accept(T t) throws Throwable;
 
     // -- .andThen(CheckedConsumer)
 
--- a/src/main/java/io/vavr/control/CheckedFunction.java
+++ b/src/main/java/io/vavr/control/CheckedFunction.java
@@ -44,9 +44,9 @@ public interface CheckedFunction<T, R> {
      *
      * @param t argument of type {@code T}
      * @return the result of the function application
-     * @throws Exception if something goes wrong applying this function to the given argument
+     * @throws Throwable if something goes wrong applying this function to the given argument
      */
-    R apply(T t) throws Exception;
+    R apply(T t) throws Throwable;
 
     /**
      * Returns a composed function that first applies this to the given argument and then applies
--- a/src/main/java/io/vavr/control/CheckedPredicate.java
+++ b/src/main/java/io/vavr/control/CheckedPredicate.java
@@ -47,9 +47,9 @@ public interface CheckedPredicate<T> {
      *
      * @param t the input argument
      * @return {@code true} if the input argument matches the predicate, otherwise {@code false}
-     * @throws Exception if an error occurs
+     * @throws Throwable if an error occurs
      */
-    boolean test(T t) throws Exception;
+    boolean test(T t) throws Throwable;
 
     /**
      * Combines this predicate with {@code that} predicate using logical and (&amp;&amp;).
--- a/src/main/java/io/vavr/control/CheckedSupplier.java
+++ b/src/main/java/io/vavr/control/CheckedSupplier.java
@@ -19,16 +19,19 @@
 package io.vavr.control;
 
 /**
- * A {@linkplain Runnable} which may throw.
+ * A function that takes no input, returns a value and may throw a {@link Throwable}.
+ * <p>
+ * The function is not required to return the same result each time it is called..
+ *
+ * @param <T> the type of supplied results
  */
 @FunctionalInterface
-public interface CheckedRunnable {
+public interface CheckedSupplier<T> {
 
     /**
-     * Performs side-effects.
+     * Gets a result.
      *
-     * @throws Throwable if an error occurs
+     * @return a result
      */
-    void run() throws Throwable;
-    
+    T get() throws Throwable;
 }
--- a/src/main/java/io/vavr/control/Try.java
+++ b/src/main/java/io/vavr/control/Try.java
@@ -24,7 +24,6 @@ import java.io.Serializable;
 import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.concurrent.Callable;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
@@ -51,8 +50,8 @@ import java.util.stream.Stream;
  *
  * <h2>Creation</h2>
  *
- * Try is intended to be used as value which contains the result of a computation. For that purpose, {@link #of(Callable)}
- * is called. See also {@link #success(Object)} and {@link #failure(Throwable)}.
+ * Try is intended to be used as value which contains the result of a computation. For that purpose,
+ * {@link #of(CheckedSupplier)} is called. See also {@link #success(Object)} and {@link #failure(Throwable)}.
  * <p>
  * However, some use {@code Try} as syntactic sugar for try-catch blocks that only perform side-effects. For that purpose,
  * {@link #run(CheckedRunnable)} is called. This variant does not contain a value but is still able to observe, handle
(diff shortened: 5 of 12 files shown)


nfekete:
@danieldietrich the conclusion of #2137 no longer applies? I believe declaring `throws Exception` is better than `throws Throwable`. Could you explain the rationale behind this change?

tlinkowski:
Sorry to butt in, but here's what SO says about it: [Is “throws Throwable” good practice](https://stackoverflow.com/questions/12564428/is-throws-throwable-good-practice) :wink:

danieldietrich:
I think here is a bit of confusion.

* I think you talk about best practices regarding the _producer_ site. "One should not throw Throwable", which is perfectly ok.
* I talk about the _consumer_ site. I need to handle Throwable because it is possible.

`Checked{Predicate,Supplier,Runnable,...}` are part of the API definition of `Try`. It is the _consumer_ site. It needs to handle Throwable, otherwise we would have a hole / a partially defined API. The library _needs_ to support what is possible in Java, otherwise it would be incomplete and therefore a bad experience. We would leave the user in a stuck situation, he would need to implement workarounds, like wrapping exceptions. The only purpose of Try is to solve that wrapping problem.

screenshot 2019-01-26 at 00 32 45

I think you wish programs (by convention) that are located in the upper right corner. But without wrapping because everyone only throws Exception or Error (which behaves like a runtime exception). But that is Utopia.

The Try API needs functional interfaces that handle the Throwable case.

---

Wlog, io.vavr.(Checked)Function* is used out of the context of Try. Here I can understand that it is better to throw Exception instead of Throwable.

```java
void f(CheckedFunction1<T,  R> in) {
    try {
        in.apply(...);
    } catch(Exception x) {
        handle(x);
    } catch(Error x) {
        throw x;
    } catch(Throwable x) {
        // don't know what to do here because it should never happen
    }
}
```

tlinkowski:
Thanks for such a detailed explanation, Daniel!

If I understand it right, the only reason you need `throws Throwable` is because your users have some code that declares `throws Throwable`.

AFAIK, they don't really need to declare `throws Throwable`, because by declaring `throws Exception` they could still throw exactly the same exceptions. I actually wonder why `throws Throwable` is legal in Java (as opposed to `throws Exception, Error`).

But as soon as these users do declare `throws Throwable`, you indeed need `throws Throwable` too in order not to make their life harder.

If they didn't declare `throws Throwable`, you could have only `throws Exception` in your interfaces, but still catch **all** `Throwables`:
```java
Try<R> wrap(CheckedFunction0<R> in) {
  try {
    return Try.success(in.apply());
  } catch (Throwable t) {
    return Try.failure(t);
  }
}

interface CheckedFunction0<R> {
  R apply() throws Exception;
}
```

PS. Beside the point, but I admit I still have trouble accepting that consumers *should* handle all `Throwable`s. E.g. wrapping `OutOfMemoryError` in a `Try` seems like not such a good idea to me.

----

Just one related funny example:
```java
interface Base {
  int get() throws Throwable;
}
interface Super extends Base {
  @Override
  int get() throws Exception;
}

  Base b = null;
  try {
    b.get(); // Unhandled exception: Throwable
  } catch (Exception e) {
    e.printStackTrace();
  }
  
  Super s = null;
  try {
    s.get(); // OK
  } catch (Exception e) {
    e.printStackTrace();
  }
}
```

So javac:
- lets you override a method from `throws Throwable` to `throws Exception`
- doesn't let you catch `Exception` only when there's `throws Throwable`

danieldietrich:
@tlinkowski I had a use-case in mind that your examples do not cover.

```java
class TertiumDatur extends Throwable {{
    assert !((Throwable) this instanceof Error);
    assert !((Throwable) this instanceof Exception);
}}

interface CheckedFunction0<R> {
    R apply() throws Exception;
}

class Try<T> {
    static <T> Try<T> of(CheckedFunction0<? extends T> supplier) {
        return null;
    }
}

class TryTest {
    void shouldCreateTry() {
        // ❌ COMPILER ERROR: Unhandled exception: TertiumDatur
        Try.of(() -> { throw new TertiumDatur(); });
    }
}
```

I provide a library for Java. So I have to cover all use-cases.

---

Your main argument is that there exist only Exception and Error as subclasses of Throwable. I provided the counter example.

What you really mean is, that it is good practice to only use one of Exception and Error. But the reality looks different.

---

Again, I think about library design (_consumer_ in the image above). You think about application code (_producer_) in the image above. You write _idiomatic code_ that does a common task in the common way for Java. I design a library that is able to handle all kind of code.

Maybe it is easier to accept for you, if you see (wlog) the CheckedSupplier interface only as _consumer_ part of the Try API. You do not need to use CheckedSupplier in your code. (Maybe I can clarify this in the Javadoc).

Instead you are still able to write code that declares only `throws Exception`:

```java
// lambda example
CheckedFunction0<T> f1 = () -> t;
Try<T> try1 = Try.of(f1::apply); // compiles fine

// method example
T f2() throws Exception {
    throw new Error();
}
Try<T> try2 = Try.of(this::f2); // compiles fine
```

---

I don't want to blame someone here, ...but... the relatively popular [Calypso Trading System](https://www2.calypso.com) for example extends Throwable with their own type of exceptions.

There is real-world Java software out there that does stupid things.

Why?

Because Java is designed that way and gives the users the ability to do so.

That's bad. In a (more) perfect world, Java would have **type aliases** and **union types** (or unchecked exceptions):

```java
type Throwable = Exception | Error;
```

¯\_(ツ)_/¯

danieldietrich:
From the library design perspective it is questionable to let CheckedFunctionN throw Exception instead of Throwable.

When Java will get primitive generics, I want to migrate the following:

```java
io.vavr.control.CheckedFunction<T, R> -> io.vavr.CheckedFunction1<T, R>
io.vavr.control.CheckedSupplier<R> -> io.vavr.CheckedFunction1<void, R>
io.vavr.control.CheckedConsumer<T> -> io.vavr.CheckedFunction1<T, void>
io.vavr.control.CheckedPredicate<T> -> io.vavr.CheckedFunction1<T, boolean>
io.vavr.control.CheckedRunnable -> io.vavr.CheckedFunction1<void, void>

// io.vavr.CheckedFunction0<R> will not be needed anymore
```

This will only work, if io.vavr.CheckedFunctionN throw Throwable.

nfekete:
After thinking again about it I tend to agree with this decision. Thanks for elaborating, @danieldietrich.

danieldietrich:
@nfekete thank you. We still change io.vavr.CheckedFunction0..8 the way that they throw Exception. D’accord?

nfekete:
Wait, I might be missing something here. So there is going to be `CheckedFunction0-8` in `io.vavr` that declare to `throw Exception` and `Checked(Consumer|Function|Predicate|Runnable|Supplier)` in `io.vavr.control` declare to `throw Throwable`?

tlinkowski:
@danieldietrich, thank you again for your detailed explanation!

First of all, I apologize for my too quick (and actually stupid) example:
1. It had wrong names (`Super` should be `Sub`).
2. It claimed to prove inconsistence in javac, but it didn't, because the overridden `int get() throws Exception` simply **narrowed down** the thrown exception range from `Throwable` to `Exception` (covariance).
3. As you pointed out, I should've actually considered *consuming* a throwing lambda instead of handling a thrown exception directly.

Secondly, you're absolutely right: I haven't considered a custom subclass of `Throwable`. And it wasn't even about it probably being a wrong practice - I actually missed that fact that you **can** do that! 😛

In light of this, I'm definitely **convinced** that vavr's `Try` **should** handle such a case. So thank you again for explaining this! 👍 

BTW. Of course, you're right that for an "average" user of `Try` (one throwing `Exception`s only), it makes no difference if `CheckedConsumer` declares to throw `Exception` or `Throwable`. For them, the functionality of `Try` remains the same.

----

Finally, my only concern remains with catching `Error`s. I'd hate to waste any more of your time, but perhaps you could just give me a hint of why it's better to catch them in a `Try` than propagate them?

I mean: isn't it too risky, in general? E.g. catching [`ExceptionInInitializerError`](https://docs.oracle.com/javase/8/docs/api/java/lang/ExceptionInInitializerError.html) will lead to [`NoClassDefFoundError`](https://docs.oracle.com/javase/8/docs/api/java/lang/NoClassDefFoundError.html) when trying to access the same class again, and the latter may be confusing for the users.

In other words, I wonder why it's better to implement [`Try.of`](https://github.com/vavr-io/vavr/blob/99e52c67cdbd36df5813a0389e74153cc3274048/vavr/src/main/java/io/vavr/control/Try.java#L72) as:
```java
try {
  return new Success<>(supplier.apply());
} catch (Throwable t) {
  return new Failure<>(t);
}
```
instead of:
```java
try {
  return new Success<>(supplier.apply());
} catch (Error e) {
  throw e;
} catch (Throwable t) {
  return new Failure<>(t);
}
```

The same applies to [`Try.run`](https://github.com/vavr-io/vavr/blob/99e52c67cdbd36df5813a0389e74153cc3274048/vavr/src/main/java/io/vavr/control/Try.java#L115).

tlinkowski:
@nfekete Thank you, that explains it!

I admit I didn't think to check the constructor of `Failure`, because I didn't expect any logic there.

Perhaps there's a very good reason to have it that way (e.g. making *sure* this constraint is always enforced), but let me just mention an alternative design (just in case you haven't considered it):

```java
public static <T> Try<T> of(CheckedSupplier<? extends T> supplier) {
  // ...
  try {
    return success(supplier.get());
  } catch (Throwable t) {
    return failure(t);
  }
}

public static <T> Try<T> failure(Throwable exception) {
  if (cause instanceof LinkageError || cause instanceof ThreadDeath || cause instanceof VirtualMachineError) { 
     throw (Error) cause; 
  } 
  return new Failure<>(exception); // only call to the constructor & no check inside
}
```

nfekete:
@tlinkowski The check being in the constructor will never allow `Failure` to be constructed with those error types, and perhaps that was the original intent.

Issue #2364, opened 2019-01-26
vavr-match should only be a compile time dependency of vavr


danieldietrich:
The vavr-match dependency 'leaked' into the library with the PR #2294.
We need to revert that.

![screenshot 2019-01-26 at 17 04 13](https://user-images.githubusercontent.com/743833/51789735-84489c80-218c-11e9-8d88-fe2b552eace9.png)

This should not be a backward compatibility problem.
Migration to the fix: add vavr-match as compile time dependency to the actual application build.

CauchyPeano:
Oops, is it done? or I will fix it...

danieldietrich:
Hi @CauchyPeano, I think it was my fault. Yes, the `compile`scope is default in Maven. I've confused it with Gradle's `compileOnly` scope. See [my comment](
https://github.com/vavr-io/vavr/issues/2339#issuecomment-457859613) on #2339.

What I really need is that the compile time annotations (included in vavr-match) are only loaded when compiling the vavr library. The exist no other use-cases than that.

What's the right way for a _library_ to express this in Maven?

* [ ] optional
* [ ] provided

Both seem to have similar semantics. However, some tools like the assembly plugin or the shade plugin do not seem to honor the scope correctly ... I don't know.

Maybe it isn't that important. Vavr 1.0 will not use the annotations internally anymore.

**We should just leave Vavr 0.10.0 as-is in order to stay backward compatible.**

Therefore I will close this issue. Sorry for the confusion!!

PR #2365, opened 2019-01-28
Making vavr-match optional dependency for vavr (#2364)

CauchyPeano:
Please see my comments in #2364


danieldietrich:
Thank you, that was a fast 'fix' :)
As I stated in #2364, it is better to stay backward compatible.
Sorry for all the confusion 🙈

danieldietrich:
Thx!

PR #2374, opened 2019-02-09
Moved Iterable.exists/.forAll to Traversable

danieldietrich:
This is an important design decision.

Iterable is at the top of Vavr's type hierarchy.

iterable

Unlike Value before, Iterable's scope is restricted to provide a rich io.vavr.collection.**Iterator**.
Additionally we have a generic conversion method **.to** (experimental).

I move .exists and .forAll to Traversable because they suggest to operate on multiple values and in order to keep Iterable small and clean. **An** Iterable is not necessarily a collection.

Other Iterable types are Option, Try and Either. There we use **fold** to achieve the same as **exists** and **forAll**. This way we keep the API surface area small and remove redundancies.

```java
// Predicate<T> exists = ...
option.fold(() -> false, exists::test)

// Predicate<T> forAll = ...
option.fold(() -> true, forAll::test)
```

### DIFF
--- a/src/main/java/io/vavr/Iterable.java
+++ b/src/main/java/io/vavr/Iterable.java
@@ -22,7 +22,6 @@ import io.vavr.collection.Iterator;
 
 import java.util.Objects;
 import java.util.function.Function;
-import java.util.function.Predicate;
 
 /**
  * Extension of the well-known Java {@link java.lang.Iterable} in the sense that a rich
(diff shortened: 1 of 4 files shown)


danieldietrich:
Note: coverage temporarily decreased because I created JUnit 5 test interfaces that will used as mixins for 'real' test classes

PR #2379, opened 2019-03-19
Avoid swallowing exceptions in `Try`.

Abnaxos:
This is just a proposal.

The current implementation swallows exceptions in two places:

- `sequence()` / `traverse()`: only the first failure is taken into account, more failures are silently ignored
- `null` checks

This pull request fixes this. It adds a method `failWith()` and uses it in `sequence()` to collect all failures (the first encountered failure is the main failure, further failures will be added as suppressed exceptions).

It also introduces a new `requireNonNull()` method in `TryModule` that throws the NullPointerException just like `Objects.requireNonNull()`, but in case of a failure, it adds the cause of the failure as suppressed to the NPE.

**Issues with this proposal:**

1. *(Major)* It modifies the exception (`addSuppressed()`) → that's a side-effect and it's not thread safe due to the implementation in Java. However, silently swallowing exceptions is a complete no-go, that's a killer – I'd rather delete these methods. Another possible solution would be to introduce some `MultipleFailuresException`. This, however, would make proper exception handling more difficult because it introduces a new exception that wouldn't be thrown otherwise.

2. It's disputable whether `failWith()` should be exposed as public API.

### DIFF
--- a/vavr/src/main/java/io/vavr/control/Try.java
+++ b/vavr/src/main/java/io/vavr/control/Try.java
@@ -146,14 +146,15 @@ public interface Try<T> extends Value<T>, Serializable {
      */
     static <T> Try<Seq<T>> sequence(Iterable<? extends Try<? extends T>> values) {
         Objects.requireNonNull(values, "values is null");
-        Vector<T> vector = Vector.empty();
+        Try<Vector<T>> vector = success(Vector.empty());
         for (Try<? extends T> value : values) {
             if (value.isFailure()) {
-                return Try.failure(value.getCause());
+                vector = vector.failWith(value.getCause());
+            } else {
+                vector = vector.map(v -> v.append(value.get()));
             }
-            vector = vector.append(value.get());
         }
-        return Try.success(vector);
+        return narrow(vector);
     }
 
     /**
@@ -219,7 +220,7 @@ public interface Try<T> extends Value<T>, Serializable {
      * @throws NullPointerException if {@code consumer} is null
      */
     default Try<T> andThen(Consumer<? super T> consumer) {
-        Objects.requireNonNull(consumer, "consumer is null");
+        TryModule.requireNonNull(this, consumer, "consumer is null");
         return andThenTry(consumer::accept);
     }
 
@@ -242,7 +243,7 @@ public interface Try<T> extends Value<T>, Serializable {
      * @throws NullPointerException if {@code consumer} is null
      */
     default Try<T> andThenTry(CheckedConsumer<? super T> consumer) {
-        Objects.requireNonNull(consumer, "consumer is null");
+        TryModule.requireNonNull(this, consumer, "consumer is null");
         if (isFailure()) {
             return this;
         } else {
@@ -264,7 +265,7 @@ public interface Try<T> extends Value<T>, Serializable {
      * @throws NullPointerException if {@code runnable} is null
      */
     default Try<T> andThen(Runnable runnable) {
-        Objects.requireNonNull(runnable, "runnable is null");
+        TryModule.requireNonNull(this, runnable, "runnable is null");
         return andThenTry(runnable::run);
     }
 
@@ -301,7 +302,7 @@ public interface Try<T> extends Value<T>, Serializable {
      * @throws NullPointerException if {@code runnable} is null
      */
     default Try<T> andThenTry(CheckedRunnable runnable) {
-        Objects.requireNonNull(runnable, "runnable is null");
+        TryModule.requireNonNull(this, runnable, "runnable is null");
         if (isFailure()) {
             return this;
         } else {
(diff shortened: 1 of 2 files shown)


danieldietrich:
Hi & thank you for your proposal.

What you do here is to work around the _monadic_ nature of Try. Especially the shortcut circuit (reacting on the first failure) is the right semantics of _sequence()_.

What you really want is a form of error handling that accumulates errors. It could be something like


### Syntax

```
static <T> Either<Seq<Throwable>, Seq<T>> accumulate(Try<T>... values);

static <T> Either<Seq<Throwable>, Seq<T>> accumulate(Iterable<? extends Try<? extends T>> values);
```

(Modulo naming, i.e. it might change. We also need to look at the Scala world. Someone most probably already has thought about it!)

### Semantics

* In the error case all throwables are accumulated an Either `Left<Seq<Throwable>>` is returned
* Otherwise all successful values are accumulated an Either `Right<Seq<T>>` is returned

Abnaxos:
That looks like a very reasonable approach which leaves it open to the user how to proceed with multiple exceptions (handling each of them separately, reducing them to one via `addSuppressed()`, whatever). I can also implement this one, I'm fine with that.

Also, I think a clear warning that `sequence()` / `traverse()` may discard exceptions silently with a pointer to the new `accumulate()` variants in the Javadoc would be appropriate.

Abnaxos:
You cannot accumulate several Throwables using the cause:

1. In any scenario involving multiple `Try`s, the exceptions are probably not causally related, as they probably originate from multiple not directly related operations that have failed, possibly for completely different reasons. Claiming exception 1 was the cause for exception 2 would be untrue.
2. There may already be a cause (probably, actually), in which case `initCause()` will throw.

But yes, maybe it's better to have people do it manually:

```java
v.filter(Try::isFailure).map(Try::getCause).reduceLeftOption((l, r) -> {
  l.addSuppressed(r);
  return l;
}).forEach(e -> {
  // handle e
});
```

(Or whatever fits the situation)

danieldietrich:
I see. However, my intuition tells me that it is a bit odd that dependent operations (in the sense that one failure may be the cause of another) are separated in two Try instances that are sequenced. Instead I would assume that Try calls are (deeply) cascaded instead using flatMap or map (as you showed above).

Could you give me a real-world use-case where two _separate_ Try/Failure instances relate on each other by _cause_?

Abnaxos:
A more or less real-world example, `cat` in Java:

```java
public static void main(String[] args) throws Exception {
  Try<String> concatenated = Try.sequence(Vector.of(args)
      .map(Paths::get)
      .map(p -> Try.of(() -> Files.readString(p))))
      .map(s -> s.reduceLeft((l, r) -> l + r));
}
```

This will concatenate the contents of all the files given, the result is a `Try<String>`. If several files couldn't be found, the `Try` will currently contain only the *first* `FileNotFoundException`, the others are discarded. You should *never* discard exceptions.

Now I think, we have to align our understanding of suppressed exceptions. Javadoc of `addSuppressed()`:

> Note that when one exception causes [link to `initCause()`] another exception, the first exception is usually caught and then the second exception is thrown in response. In other words, there is a causal connection between the two exceptions. In contrast, there are situations where two independent exceptions can be thrown in sibling code blocks, in particular in the try block of a try-with-resources statement and the compiler-generated finally block which closes the resource. In these situations, only one of the thrown exceptions can be propagated. In the try-with-resources statement, when there are two such exceptions, the exception originating from the try block is propagated and the exception from the finally block is added to the list of exceptions suppressed by the exception from the try block. As an exception unwinds the stack, it can accumulate multiple suppressed exceptions.

So, in contrast to `initCause`, there is not necessarily a causal relationship between suppressed exceptions, as in the cat example. I might be wrong, but this is my understanding of the above paragraph.

Because I can only throw one exception, there are two things I can do if more than one file couldn't be found:

1. (before Java7) throw some `MultipleErrorsException`
2. (since Java7) throw one exception and record the other exceptions as suppressed

My patch implements the latter behaviour. It's disputable whether it does it in the right way, but IMHO, there should be an easy way to achieve this. It throws the first `FileNotFoundException` and amends: "BTW, these other files also could't be found" by attaching them as a suppressed exceptions, because there can be only one thrown exception.

The other aspect of the patch (`requireNonNull`) does the same thing, but with different weighting. It considers the illegal use of the API as more important, so it throws the `NullPointerException`. But if the `Try` is a failure, there is is exception, that was thrown before, which cannot be thrown anymore, because there was a more important error during error handling → it's attached to the NPE as a suppressed exception, so it doesn't get lost.

Note that in Scala 2.12 this (relatively) new mechanism of suppressing exceptions is completely ignored. I'm doubt we can just have a look at how Scala does it. Scala also doesn't do any try-with-resources (yet).

danieldietrich:
@Abnaxos thanks you! I created an issue for further discussion: #2380
I will come back to that issue later today.

PR #2384, opened 2019-03-21
Possibility to rerun failed execution.

slawekjaranowski:
We can in simple way to run again failed execution it is useful especially when we call remote service, eg by rest client.

We can write:

    Try.of(() -> restClient.call()).retry(5, 5, TimeUnit.SECONDS)

Which means: try execute 5 times and wait 5 second between each call.


danieldietrich:
Thank you for engaging in VAVR! Please create an issue in order to discuss changes.

* we tightly align to Scala's Try
* retry logic is a separate concern, I would not bake it into Try.
* there are several lightweight libraries that already solve the problem

I will close this PR

Issue #2381, opened 2019-03-20
Add println(Object...) to io.vavr.API


danieldietrich:
VAVR already has println(Object):

```java
import static io.vavr.API.*;
...
// prints test
println("test")
```

**Goal:** Add add println(Object...)

```java
import static io.vavr.API.*;
...
// prints this and that
println("this", "and", "that")
```

Please note, that

* the objects are stringified
* the objects are separated by space ' '

_The attentive reader may have noticed that there is some resemblance to JavaScript's console.log ;)_

thadumi:
Into the note section, you said that the objects are stringified. Are we assuming that the method toString is defined or should we implement a stringify function which works similarly to the JS one?
I already saw that the actual version of println just calls the System.out.println method and therefore is using the toString solution.

danieldietrich:
Yes, we use toString() - or more specifically: String.valueOf(), which handles the null case.

A solution could look like this:

```java
// io.vavr.collection.Iterator
Iterator.of(xs).map(String::valueOf).mkString(" ")
```

agebhar1:
This issue might be closed due to #2386.

danieldietrich:
true, thank you!

PR #2386, opened 2019-03-21
[Feature] [#2381] Add println(Object...) to io.vavr.API

thadumi:
Implemented the request of the issue #2381

### DIFF
--- a/vavr/generator/Generator.scala
+++ b/vavr/generator/Generator.scala
@@ -1356,7 +1356,7 @@ def generateMainClasses(): Unit = {
            * @param objs The list of <code>Object</code> to be printed
            */
           public static void println(Object ...objs) {
-              println($ListType.of(objs).map(Object::toString).collect(java.util.stream.Collectors.joining(" ")));
+              println($ListType.of(objs).map(String::valueOf).collect(java.util.stream.Collectors.joining(" ")));
           }
         """
       }
--- a/vavr/generator/Generator.scala
+++ b/vavr/generator/Generator.scala
@@ -99,6 +99,7 @@ def generateMainClasses(): Unit = {
       val BiFunctionType = im.getType("java.util.function.BiFunction")
       val PredicateType = im.getType("java.util.function.Predicate")
       val SupplierType = im.getType("java.util.function.Supplier")
+      // val CollectorsType = im.getType("java.util.stream.Collectors")
 
       val monadicTypesFor = List("Iterable", OptionType, FutureType, TryType, ListType)
 
@@ -1346,6 +1347,16 @@ def generateMainClasses(): Unit = {
            */
           public static void println() {
               System.out.println();
+          }
+
+          /**
+           * Using the {@code System.out.println()} will print all the objects passed as argument
+           * separating them by a space character (' ').
+           *
+           * @param objs The list of <code>Object</code> to be printed
+           */
+          public static void println(Object ...objs) {
+              println($ListType.of(objs).map(Object::toString).collect(java.util.stream.Collectors.joining(" ")));
           }
         """
       }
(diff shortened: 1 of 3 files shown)


vavr/generator/Generator.scala
```diff
@@ -2849,6 +2860,11 @@ def generateTestClasses(): Unit = {
           public void shouldCallprintln() {
               assertThat($captureStdOut(()->println())).isEqualTo("\\n");
           }
+
+          @$test
+           public void shouldCallprintlnWithArguments() {
+              assertThat($captureStdOut(()->println("this", "and", "that"))).isEqualTo("this and that\\n");
+           }
         """
```
danieldietrich: we use spaces here for better readability

```diff
- ()->println
+ () -> println
```

vavr/generator/Generator.scala
```diff
@@ -99,6 +99,7 @@ def generateMainClasses(): Unit = {
       val BiFunctionType = im.getType("java.util.function.BiFunction")
       val PredicateType = im.getType("java.util.function.Predicate")
       val SupplierType = im.getType("java.util.function.Supplier")
+      // val CollectorsType = im.getType("java.util.stream.Collectors")
```
danieldietrich: clean code policy: no commented out code

vavr/generator/Generator.scala
```diff
@@ -1347,6 +1348,16 @@ def generateMainClasses(): Unit = {
...
           }
+
+          /**
+           * Using the {@code System.out.println()} will print all the objects passed as argument
+           * separating them by a space character (' ').
+           *
+           * @param objs The list of <code>Object</code> to be printed
+           */
+          public static void println(Object ...objs) {
+              println($ListType.of(objs).map(String::valueOf).collect(java.util.stream.Collectors.joining(" ")));
```
danieldietrich: The implementation is semantically correct but not optimal regarding the performance characteristics.

* List.of() creates a VAVR list in O(n)
* map() takes also O(n) because it creates a completely new list. The previous list elements are garbage collected.
* collect() creates a lazy Java Stream. That is a little bit late here. It would be better to call `mkString(" ")` instead

Please do it the 'lazy' way by using an Iterator. It is created in O(1) time and memory. VAVR's implementations creates a lightweight wrapper around the given array argument. The terminal operation iterates the original array.

```java
println($IteratorType.of(objs).map(String::valueOf).mkString(" "))
```

(Maybe you need the IteratorType variable in scope, if it does not exist already.)

vavr/generator/Generator.scala
```diff
@@ -1347,6 +1348,16 @@ def generateMainClasses(): Unit = {
           public static void println() {
               System.out.println();
           }
+
+          /**
+           * Using the {@code System.out.println()} will print all the objects passed as argument
+           * separating them by a space character (' ').
+           *
+           * @param objs The list of <code>Object</code> to be printed
```
danieldietrich: I would write

```
@param objs The objects to be printed
```

because it is no list. It could be 'array' or 'sequence' - but I think just 'objects' is enough

vavr/generator/Generator.scala
```diff
@@ -1347,6 +1348,16 @@ def generateMainClasses(): Unit = {
           public static void println() {
               System.out.println();
           }
+
+          /**
+           * Using the {@code System.out.println()} will print all the objects passed as argument
+           * separating them by a space character (' ').
```
danieldietrich: I would change it this way:

```
Prints the given objects as space ' ' separated string using {@code System.out.println()}.
```

vavr/generator/Generator.scala
thadumi: Sorry, you are right. I wasn't aware of the mkString method. I'm going to fix it right now.

PR #2400, opened 2019-03-29
Improve performance of last() call on TreeMap

craffit:


danieldietrich:
Awesome, thank you - I think we missed that one :)

PR #2404, opened 2019-03-31
Adding OpenJDK12 to Travis CI build matrix

### LINKED ISSUE #2406: Different generics behavior when using JDK12
The following does not compile with JDK12 (it worked with JDK8 - JDK11):

```java
@Test
public void shouldComputeUpperBoundOfReturnValue() {
    final Number num = Match(3).of(
            Case($(is(1)), 1),
            Case($(is(2)), 2.0),
            Case($(), i -> new BigDecimal("" + i))
    );
    assertThat(num).isEqualTo(new BigDecimal("3"));
}
```

error:

```bash
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:testCompile (default-testCompile) on project vavr: Compilation failure
[ERROR] /Users/daniel/git/vavr-io/vavr-0.x/vavr/src/test/java/io/vavr/MatchTest.java:[191,39] incompatible types: cannot infer type-variable(s) T,R,T
[ERROR]     (argument mismatch; bad return type in lambda expression
[ERROR]       java.math.BigDecimal cannot be converted to java.lang.Number&java.lang.Comparable<? extends java.lang.Number&java.lang.Comparable<?>&java.lang.constant.Constable&java.lang.constant.ConstantDesc>&java.lang.constant.Constable&java.lang.constant.ConstantDesc)
[ERROR] 
```

---

The following example works with JDK12:

file: **Test.java**

```java
interface Test {
    static void main(String... args) {
        Number num = switch (3) {
            case 1 -> 1;
            case 2 -> 2.0;
            default -> new java.math.BigDecimal("3");
        };
        System.out.println("Number: " + num);
    }
}
```

compile & run it:

```bash
javac --enable-preview --release 12 Test.java
java --enable-preview Test
```

output:

```bash
Number: 3
```

danieldietrich:
Fixes #2406

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ matrix:
   - jdk: oraclejdk9
   - jdk: openjdk10
   - jdk: openjdk11
+  - jdk: openjdk12
   - jdk: openjdk-ea
   allow_failures:
   - jdk: openjdk-ea
--- a/vavr/src/test/java/io/vavr/MatchTest.java
+++ b/vavr/src/test/java/io/vavr/MatchTest.java
@@ -188,7 +188,7 @@ public class MatchTest {
 
     @Test
     public void shouldComputeUpperBoundOfReturnValue() {
-        final Number num = Match(3).of(
+        final Number num = Match(3).<Number> of(
                 Case($(is(1)), 1),
                 Case($(is(2)), 2.0),
                 Case($(), i -> new BigDecimal("" + i))


danieldietrich:
This PR revealed a JDK 12 bug: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8221742

danieldietrich:
The JDK 12 build currently fails because `java.lang.constant.Constable` was introduced to  the type hierarchy. We need to fix our tests accordingly.

Screenshot 2019-07-01 at 14 21 09

PR #2426, opened 2019-07-04
Fixes DistictIterator to not eat null values

### LINKED ISSUE #2425: DistinctIterator eats null values
I ran across a problem with the .distinct()-implementation on Iterator. In addition to removing duplicate elements, it also removes all null values from the input. Other distinct()-implementations (e.g. on List) do not show this behaviour.

Code:
`System.out.println(io.vavr.collection.Iterator.of((String)null).toList().distinct())`
`System.out.println(io.vavr.collection.Iterator.of((String)null).distinct().toList())`
Result:
`List(null)`
`List()`

The problem appears to be located in the implementation of DistinctIterator.hasNext() (see https://github.com/vavr-io/vavr/blob/master/vavr/src/main/java/io/vavr/collection/Iterator.java#L2146). The loop uses `next == null` to indicate that a new distinct element has not yet been found, while it may also occur when the input is actually a null element.

danieldietrich:
Fixes #2425

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Iterator.java
+++ b/vavr/src/main/java/io/vavr/collection/Iterator.java
@@ -2164,8 +2164,10 @@ interface IteratorModule {
 
         @Override
         public T getNext() {
+            final T result = next;
             nextDefined = false;
-            return next;
+            next = null;
+            return result;
         }
     }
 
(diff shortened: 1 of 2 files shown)


vavr/src/main/java/io/vavr/collection/Iterator.java
```diff
@@ -2144,22 +2145,27 @@ public T getNext() {
...
+            return false;
         }
 
         @Override
         public T getNext() {
-            final T result = next;
-            next = null;
-            return result;
+            nextDefined = false;
+            return next;
```
danieldietrich: We still hold the reference on the last consumed `next` value.
Corner case: a reference to this iterator remains. Then the last iterator value that was consumed can't be garbage collected.

I think I will set `next = null` in another commit to be clean here.

PR #2454, opened 2019-07-21
Improve shuffle test

mincong-h:
This PR aims to improve the shuffle test, as described in issue https://github.com/vavr-io/vavr/issues/2453. Compared to existing test case, it has some improvements on the following fields:

1. Ensure the size of the shuffled collection is the same as the original collection
2. Ensure the elements stayed the same
3. Ensure the shuffled collection does not equal to the original collection. If this is the case, repeat up to 10 times (test will fail after that).

Actually, about the 1st point, it's not exactly true. The size has been asserted in another test just above, called `shouldShuffleHaveSameLengthDeterministic()`. Not sure it should be deleted or not.


mincong-h:
> I thought about what you did say - and about the random seed. Actually, the latter is already present! It is the argument that is passed to the Random constructor.

I didn't know that. Just learnt something new. Thank you!

Issue #2458, opened 2019-07-25
Add simple examples to Javadoc


danieldietrich:
This is an ongoing issue. I would love to see contributions in form of several small PRs that successively add examples to our Javadoc.

The Javadoc should have the following general structure:

```diff
/**
 * <text text text>
 *
+* <pre>{@code
+* <example(s)>
+* }</pre>
 *
 * <@annotations>
 */
```

Example:

```java
/**
    * Consumes the cause if this is a {@link Try.Failure} and the cause is instance of the given {@code exceptionType}.
    *
    * <pre>{@code
    * // (does not print anything)
    * Try.success(1).onFailure(Error.class, System.out::println);
    *
    * // prints "java.lang.Error"
    * Try.failure(new Error()).onFailure(Error.class, System.out::println);
    *
    * // (does not print anything)
    * Try.failure(new Exception()).onFailure(Error.class, System.out::println);
    * }</pre>
    *
    * @param <X>           Exception type
    * @param exceptionType The specific exception type that should be handled
    * @param action        An exception consumer
    * @return this
    * @throws NullPointerException if one of {@code exceptionType} or {@code action} is null
    */
public <X extends Throwable> Try<T> onFailure(Class<X> exceptionType, Consumer<? super Throwable> action) {
    Objects.requireNonNull(exceptionType, "exceptionType is null");
    Objects.requireNonNull(action, "action is null");
    if (isFailure() && exceptionType.isInstance(getCause())) {
        action.accept(getCause());
    }
    return this;
}
```

Please note that `src-gen/**/*.java` must not be edited directly. You can either skip these classes or edit `generator/Generator.sc` instead. Run `./gradlew generateSources` to re-generate the code, if you do so.

Please run `./gradlew check && ./gradlew javadoc` before submitting your PR.

FuncGuy:
i'm ready to take this issue but i'm confused from where to start?

danieldietrich:
Great! It is meant to be on a fine-grained basis. Just pick methods of your choice. I expect several persons will participate over time!

FuncGuy:
thanks for adding me to the contributor list.

FuncGuy:
I'm gonna pick up next please suggest.

pepijno:
Hi, I would like to contribute. Can I add examples for `Option` or is someone already working on this?

thiyagu-7:
[...]

PR #2465, opened 2019-07-27
Examples for Either.bimap, Either.fold and Either.traverse (#2458)

zbyszkop:
Continuing with PRs for Either - code examples in javadoc for bimap and fold

### DIFF
--- a/src/main/java/io/vavr/control/Either.java
+++ b/src/main/java/io/vavr/control/Either.java
@@ -192,9 +192,9 @@ public abstract class Either<L, R> implements io.vavr.Iterable<R>, io.vavr.Value
 
     /**
      * Maps either the left or the right side of this disjunction.
-     * <pre>{@code
      *
-     * Either<?, AtomicInteger> success = Either.right(new AtomicInteger(42));
+     * <pre>{@code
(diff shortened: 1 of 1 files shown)


src/main/java/io/vavr/control/Either.java
```diff
@@ -193,6 +193,18 @@ private Either() {
     /**
      * Maps either the left or the right side of this disjunction.
      *
+     * <pre>{@code
+     *Either<?, AtomicInteger> success = Either.right(new AtomicInteger(42));
```
danieldietrich: one space missing at the beginning

src/main/java/io/vavr/control/Either.java
```diff
@@ -193,6 +193,18 @@ private Either() {
...
+     *
+     * //prints "Right(42)"
```
danieldietrich: Please add one space after `//` as we do it across the whole code base for reasons of consistency and readability.

zbyszkop: Fixed, also in previous change.

src/main/java/io/vavr/control/Either.java
zbyszkop: Fixed.

PR #2471, opened 2019-07-30
Adds missing transform methods.

danieldietrich:


goerge:
@danieldietrich Any reason why `Seq` didn't get a `transform` method? Or did I miss anything?

danieldietrich:
In Java, the type system does not allow to declare a generic API for the `transform()` method _across a type hierarchy_. I had the same problem with other methods, like `flatMap()` and therefore had to make design choices.

Screenshot 2019-07-31 at 12 24 08

Regarding flatMap, it is practicable to use `Iterable`as target type...

```java
interface List<T> extends LinearSeq<T> {
    <U> List<U> flatMap(Function<? super T, ? extends Iterable<? extends U>> mapper);
}
```

...instead of the concrete collection type...

```java
interface List<T> extends LinearSeq<T> {
    <U> List<U> flatMap(Function<? super T, ? extends List<? extends U>> mapper);
}
```

The latter would leave to the same problems stated above. The former, our design decision, comes at the cost of violating some monad laws for collections. However, in practice it does not matter. Btw, Scala did the same.

Back to topic, we can't make such a design decision for `transform()´ because the argument _has to be_ of the correct type. Otherwise the semantics of the method would not make sense anymore.

I hope you understand my decision.

Thx for your question.

PR #2473, opened 2019-07-31
Feature/leanient validation javadoc amends

cesarcastillob:


danieldietrich:
Sorry, but this is not how open source works. This isn't your personal playground here.

First, an issue is created and the maintainers of this library will decide whether a new feature is needed or not. Then, _maybe_, eventually a PR will be created. This _might_ be someone from the outside or, more often, an internal committer.

cesarcastillob:
Sorry for the inconvenient, my intention was not to create this PR, it was incomplete at that moment and happened accidentally, that is the reason I closed almost immediately. And, thanks for the guidelines about the process to follow.

danieldietrich:
Hi @cesarcastillob, thanks for reaching out to me!

I think I should also clarify it in the contribution docs.

Ideally, there would be only one Validation type instead of N over time.
It would be nice to first to  what is missing in the existing Validation. That is what I call the „requirement space“. If and how it is solved is part of the „solution space“.

PR #2474, opened 2019-07-31
Lenient Validation

fabianboeller:
We introduced a LenientValidation class that helps to validate a data structure that is only partially invalid.
While a Validation can either be a Valid or an Invalid, an instance of LenientValidation can contain errors and a value at the same time. This is especially helpful in situations where you want to validate a sequence of values. While the validation of a single value might fail, it can still be reasonable to continue with all the values of the sequence that were successfully validated.

```
LenientValidation<String,Integer> valid1 = LenientValidation.valid(2);
LenientValidation<String,Integer> valid2 = LenientValidation.valid(3);
LenientValidation<String,Integer> valid3 = LenientValidation.invalid(List.of("error"));

LenientValidation<String,Seq<Integer>> result = 
  LenientValidation.sequence(List.of(valid1, valid2, valid3));
```

The resulting LenientValidation here is a value `[2, 3]` together with the occurred error `["error"]`.
With using the Validation class, you would only be able to get the error `["error"]`.

In our use case, this more lenient behaviour helps us a lot since we can not rely on good data quality but want to process as much as possible.

We put this class into the existing Vavr Value hierarchy and added much of the functionality you have in the Validation class as well. We did not put it directly as a subclass under Validation to not relax the existing constraints of having either a valid or invalid instance of the Validation class.


danieldietrich:
I really appreciate your work and your interest in VAVR.

However, I do not plan to add more validation related types into the core library. In fact I already stated in existing issues that I thought about deprecating the existing validation types.

Adding new functionality does require much more than only "added much of the functionality you have in the Xyz class". I take care about interoperability, obeying monad and functor laws, etc.

Our main design rule is to align tightly to Scala. Extending the VAVR core library takes place in separate repositories. These are generally maintained by those who want to extend VAVR. Don't underestimate the effort! You need to take care of compatibility to each new release.

I can't accept PRs, sorry. Strategic decisions are made internally.

PR #2484, opened 2019-08-02
Fix integer overflow in range

mincong-h:
This is a draft pull request for handling https://github.com/vavr-io/vavr/issues/2263 Integer overflow problem in `Stream.rangeClosedBy`. Note that, this draft PR does not cover all acceptance criteria from the issue. It's mainly for discussion purpose. It means:

- It only contains changes for integer, need to extend to long, char, big integer and other types.
- It doesn't address overflow problem in `Integer.signum(from - toInclusive)`.
- It doesn't check / fix other similar APIs where addition / subtraction is performed.

In term of implementation, this PR uses the same overflow detection mechanism as method [`java.lang.Math#addExact(int, int)`](https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#addExact-int-int-), available since Java 8:

```java
int r = x + y;
// HD 2-12 Overflow iff both arguments have the opposite sign of the result
if (((x ^ r) & (y ^ r)) < 0) {
    // overflow
} else {
    ...
}
```

Compared to Scala, which does convert the int values to long and performs some checks (https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/immutable/Range.scala#L84-L87), this approach avoid changing type, thus the solution can be applied to all other similar numeric types.

### DIFF
--- a/src/main/java/io/vavr/collection/Iterator.java
+++ b/src/main/java/io/vavr/collection/Iterator.java
@@ -597,9 +597,9 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
             throw new IllegalArgumentException("step cannot be 0");
         }
         if (step > 0) {
-            return new IncrementalIntRangeIterator(from, toExclusive, step, false);
+            return new RangeIntForwardIterator(from, toExclusive, step, false);
         } else {
-            return new DecrementalIntRangeIterator(from, toExclusive, step, false);
+            return new RangeIntBackwardIterator(from, toExclusive, step, false);
         }
     }
 
@@ -644,8 +644,14 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
      * @throws IllegalArgumentException if {@code step} is zero
      */
     static Iterator<Long> rangeBy(long from, long toExclusive, long step) {
-        final long toInclusive = toExclusive - (step > 0 ? 1 : -1);
-        return rangeClosedBy(from, toInclusive, step);
+        if (step == 0) {
+            throw new IllegalArgumentException("step cannot be 0");
+        }
+        if (step > 0) {
+            return new RangeLongForwardIterator(from, toExclusive, step, false);
+        } else {
+            return new RangeLongBackwardIterator(from, toExclusive, step, false);
+        }
     }
 
     /**
@@ -745,9 +751,9 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
             throw new IllegalArgumentException("step cannot be 0");
         }
         if (step > 0) {
-            return new IncrementalIntRangeIterator(from, toInclusive, step, true);
+            return new RangeIntForwardIterator(from, toInclusive, step, true);
         } else {
-            return new DecrementalIntRangeIterator(from, toInclusive, step, true);
+            return new RangeIntBackwardIterator(from, toInclusive, step, true);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/vavr/collection/Iterator.java
```diff
@@ -742,39 +742,48 @@ public BigDecimal next() {
...
             if (step > 0) {
                 return new Iterator<Integer>() {
-                    int i = from - step;
+                    int next = from;
+                    boolean overflow = false;
 
                     @Override
                     public boolean hasNext() {
-                        return i <= end;
+                        return !overflow && next <= toInclusive;
```
danieldietrich: thumbs up for your solution 👍🏼

src/main/java/io/vavr/collection/Iterator.java
```diff
@@ -742,39 +742,48 @@ public BigDecimal next() {
...
                     public Integer next() {
                         if (!hasNext()) {
                             throw new NoSuchElementException();
                         }
-                        return i += step;
+                        int curr = next;
+                        int r = curr + step;
+                        overflow = ((curr ^ r) & (step ^ r)) < 0;
+                        next = r;
+                        return curr;
```
danieldietrich: Well done, Mincong, your solutions (not only this one) are well thought through!

I would love if I may invite you to join the group of official VAVR committers! It is meant as kind of achievement :)

mincong-h: Thank you @danieldietrich for your invitation. It's really a pleasure to hear that! I would love to contribute more in the future :)

mincong-h:
Since Daniel approved my strategy, I will continue my fix on the same direction and submit a complete PR in the future. Closing this draft now.

danieldietrich:
Thx! We could also reuse this PR, rebase the git branch (by squashing commits) and change the draft status...

mincong-h:
I feel more comfortable to push (forced) often in my own branch and submit a fix once it's ready, but I can also try your approach the next time :)

danieldietrich:
It is okay, you can do it the way you feel most comfortable!

src/main/java/io/vavr/collection/Iterator.java
```diff
@@ -742,39 +742,48 @@ public BigDecimal next() {
         } else if (Integer.signum(step) == Integer.signum(from - toInclusive)) {
```
danieldietrich: @mincong-h this will still fail for `from = Integer.MAX_VALUE`, `toInclusive = -5` and `step = -1`.

danieldietrich: It would be better to completely remove the `if (Integer.signum(step) == Integer.signum(from - toInclusive))` check.

Instead we could perform these checks in the `step > 0` and `step < 0` if branches:

```java
if (step == 0) {
} else if (from == toInclusive) {
    return of(from);
} if (step > 0) {
    if (from > toInclusive) {
        return empty();
    } else {
        return new Iterator<Integer>() { ... };
    }
} else {
    if (from < toInclusive) {
        return empty();
    } else {
        return new Iterator<Integer>() { ... };
    }
}    
```

mincong-h: Yes, thanks for your hint. I'm working on a fix now. Will add test(s) for this as well.

mincong-h:
The new implementation fixes the integer overflow problem by avoiding operations (addition, subtraction) whenever possible. Instead, it uses boolean variables and arithmetic operators. Boolean variables: variable `inclusive` to determine whether the range is inclusive, variable `overflow` to determine if there are more "next" value(s). Also, use arithmetic operators (>, ≥, <, ≤) to handle different cases.

danieldietrich:
Hi @mincong-h, thank you, that was fast!

`<short-excursion>`
This topic really interested me. Scala has a dedicated [Range]() type. I plan to introduce it in VAVR 2.0.0.

A `Range` is `Iterable`. That way we are able to remove all(!) static `range`, `rangeBy`, `rangeClosed` and `rangeClosedBy` (multiplied with { char, double, long, int }, multiplied with all collection types).

Because it is `Iterable`, we just call `List.ofAll(range)` for example.

Please have a look, how I solved the inclusive/exclusive topic. I've implemented one abstract Iterator that contains redundant code.
`</short-excursion>`

I will do a full code review of your changes tomorrow. Thx!

PR #2490, opened 2019-08-03
Added Range type

danieldietrich:


src/main/java/io/vavr/collection/Range.java
```diff
@@ -0,0 +1,114 @@
...
+public interface Range<T> extends io.vavr.Iterable<T> {
+
+    static Range<Integer> inclusive(int from, int toInclusive) {
+        return Range.inclusiveBy(from, toInclusive, from <= toInclusive ? 1 : -1);
+    }
+
+    static Range<Integer> inclusiveBy(int from, int toInclusive, int step) {
+        if (step == 0) {
+            throw new IllegalArgumentException("step cannot be 0");
+        } else if (from == toInclusive) {
```
mincong-h: Do you consider removing some `else`?

```java
if (step == 0) {
}
if (from == toInclusive) {
    return () -> Iterator.of(from);
}
if (step > 0) {
    ...
} else {
    ...
}

File: src/main/java/io/vavr/collection/Range.java [hunk 2]
```diff
@@ -0,0 +1,114 @@
...
+
+    static Range<Integer> exclusive(int from, int toExclusive) {
+        return Range.exclusiveBy(from, toExclusive, from <= toExclusive ? 1 : -1);
+    }
+
+    static Range<Integer> exclusiveBy(int from, int toExclusive, int step) {
+        var signum = Integer.signum(step);
+        var toInclusive = toExclusive - signum;
+        if (Integer.signum(toInclusive) != Integer.signum(toExclusive)) {
+            // because of abs(signum) <= abs(step) and overflow detection, toExclusive will not be included
```
mincong-h: Do we have all the existing tests running? I think I understand your approach, but it's very complex... Personally, I prefer not rely on `inclusiveBy`...

src/main/java/io/vavr/collection/Range.java
```diff
@@ -0,0 +1,114 @@
...
+            return Range.inclusiveBy(from, toExclusive, step);
+        } else {
+            return Range.inclusiveBy(from, toInclusive, step);
+        }
+    }
+
+}
+
+abstract class AbstractRangeIterator implements Iterator<Integer> {
```
mincong-h: This is elegant, combined with the abstract `hasNext()` :+1:

src/main/java/io/vavr/collection/Range.java
```diff
@@ -0,0 +1,114 @@
...
+                    public boolean hasNext() {
+                        return !overflow && next >= toInclusive;
+                    }
+                };
+            }
+        }
+    }
+
+    static Range<Integer> exclusive(int from, int toExclusive) {
+        return Range.exclusiveBy(from, toExclusive, from <= toExclusive ? 1 : -1);
```
mincong-h: Do you think about using another implementation for step=1? The overflow detection can be avoided, which can improve performance, as I believe most people are using range without giving step (step=1). JDK does it without overflow detection with step=1, see `java.util.stream.Streams.RangeIntSpliterator`.

ullenboom:
I really like a `Range` type. Some wishes and ideas:

a. Write Javadoc.

b. Store `from`, `to` and `step` into final variables. This has two advantages:

1. a proper `toString()` method can return all values and current state (useful for debugging)
2. introduce a message and pass into the constructor of `throw new NoSuchElementException(msg)`; it is best practise to give as much context as possible.

c. Something controversial:

1. The default notation for ranges in Java is: `start` inclusive and `end` exclusive. Because of that, a method can be simplifies to `Range.of(...)`.

2. A `contains(...)` method would be convenient if its intended to evolve the data type. Other methods would be convenient too, how about https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-int-progression/index.html?

danieldietrich:
@ullenboom Thanks for your suggestions. The PR is marked as Draft / WIP.

> Write Javadoc.

This is part of our definition of done. You can expect something, otherwise the PR will not be pulled :)

> Store from, to and step into final variables.

If you mean making Range a class instead of an interface: yes, I plan to do that. Probably I will align to Scala. See https://www.scala-lang.org/api/current/scala/collection/immutable/Range.html

Screenshot 2019-12-12 at 13 52 23

> The default notation for ranges in Java is (...)

Yes, in Scala an exclusive end is also the default. There the syntax is `1 until 10`, which is syntactic sugar for `Range(1, 10)`, which is syntactic sugar for `Range.apply(1, 10)`. Indeed, we could use `Range.of` for that purpose in Vavr.

The inclusive end is denoted `Range.inclusive(1, 10)`, which is the same as `1 to 10` there.

(Additionally, there are method overloads for the step parameters).

> A contains(...) method

pivovarit:
I assume you won't be continuing this :) closing for now

Issue #2497, opened 2019-08-25
Add more tests for boundary conditions in Array.java


mincong-h:
When changing conditional boundary for some cases, all existing tests still passed. For example, changing relation operator from `>=` to `>` in class `io.vavr.collection.Array`, the mutation survived:

```diff
@@ -711,7 +711,7 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
     @SuppressWarnings("unchecked")
     @Override
     public T get(int index) {
-        if (index < 0 || index >= length()) {
+        if (index < 0 || index > length()) {
             throw new IndexOutOfBoundsException("get(" + index + ")");
         }
         return (T) delegate[index];
```

This is discovered using pitest (http://pitest.org). Actually, it happens because Java throws the same `java.lang.IndexOutOfBoundsException` when accessing array using out-of-bound index. In other words, the exception can be thrown by the `delegate` array.

I would like to add some tests to ensure this mutation is killed.

mincong-h:
Done, see 7395524be6a01327b7b8a81f6a00e69b3665e999

PR #2498, opened 2019-08-25
Add test for Array#get()

mincong-h:
Add test for `io.vavr.collection.Array#get(int)` as described in #2497. The two screenshots below show the pitest result before and after the commit changes.

Before:

Screenshot 2019-08-24 at 20 55 50

After:

Screenshot 2019-08-25 at 14 06 25


danieldietrich:
Thank you!

PR #2507, opened 2019-10-03
Add Seq.getOption (fixes #2505)

### LINKED ISSUE #2505: Add getOption() to Seq
`Seq` has `get(int index)`, however, this is a partial function which throws an exception if the `index` is out of bonds.

As far as I can tell, there is no total function that allows you to get a specific element out of a sequence, returning an optional result. This is a method that is widespread in other contexts, for example, multiple Haskell packages have a function `(!?)` which does exactly that.

Given other method pairs like `init()` and `initOption()`, `getOption(int index)` seems like a natural inclusion to the `Seq` interface, which I found myself missing when using vavr `Array`s.

spencerwi:


danieldietrich:
I close #2505, for details see [this comment](https://github.com/vavr-io/vavr/issues/2505#issuecomment-557848136)

danieldietrich:
Btw - thanks @spencerwi for the PR, I really appreciate it! If an issue isn't labeled, yet, it is not approved.

PR #2508, opened 2019-10-06
Adjust Javadoc of LinkedHashSet::add.

korotkevics:
Hello, I'd like to commit a minor adjustment.

I find this particular Javadoc misleading since it would not replace but ignore an element being added if it is already contained.

```java
    @Override
    public LinkedHashSet<T> add(T element) {
        return contains(element) ? this : new LinkedHashSet<>(map.put(element, element));
    }
```


danieldietrich:
Great, thank you!

PR #2511, opened 2019-10-11
removed duplication in method hash of interface Tuple

googl3r:


danieldietrich:
We looked into the original requirement and prepared additional changes. I will close this PR (because it is stale) in favor of #2522.

Nevertheless, thx for your engagement!

PR #2519, opened 2019-11-29
Avoid unnecessary copy in 'ofAll()' methods

mincong-h:
Fix https://github.com/vavr-io/vavr/issues/2516


mincong-h:
Thank you @danieldietrich !

tv-pt:
Hi, 

would it be possible to port that fix back to 0.10.x branch, too? Stable 1.0.0 is not in sight and it wouldn't be backward (binary) compatible with 0.10.x anyway. That would be a important performance improvement for us. 

Thank You.

danieldietrich:
@tv-pt we could include it in the next patch release 0.10.3

PR #2521, opened 2019-12-05
[draft] Removed Comparable from the list of interfaces implemented by Tuple

### LINKED ISSUE #2446: Tuple should not implement Comparable
In #605, we made Tuple* comparable in order to be used as element type in conjunction with SortedSet and SortedMap.

However, it might lead to a class cast exception at runtime, if the tuple components aren't comparable.

Screenshot 2019-12-07 at 17 34 08

Because we strive for more safety, we will remove the Comparable interface from the Tuple types.

pivovarit:
Closes #2446


pivovarit:
It's a draft. Clicked the wrong button.

pivovarit:
To be replaced with a proper `draft`

danieldietrich:
Ok, thanks! This is an interesting topic, I hope I find some time tomorrow to think a bit about it...

pivovarit:
That's definitely way harder than I thought :) the lack of `Comparable` bubbles up pretty aggressively. I will come back with a proposal (will play around with deriving internal comparators, mostly something along the lines:

```
Tuple2.comparator(comparator(), Comparators.naturalComparator())
```

danieldietrich:
Yes, we had a reason for introducing it. But maybe there is a better/safer solution possible...

danieldietrich:
I will assign you to the issue in order to mark it as 'occupied'. It helps me organizing the remaining issues for 1.0.0. No pressure, even if you decide not to do it..

PR #2522, opened 2019-12-05
Simplify internal implementation of Tuple#hash static methods

pivovarit:
Since `TupleN` hash tests use `Objects#hash()` to validate calculated hashes, why not just reuse it in the internal implementation? 

```assertThat(t.hashCode()).isEqualTo(Objects.hash(t._1, t._2, t._3, t._4, t._5, t._6, t._7, t._8));```

I assume that we need these helper methods only to provide an abstraction layer in case someone wants to swap the hashing strategy.

### DIFF
--- a/generator/Generator.sc
+++ b/generator/Generator.sc
@@ -2255,7 +2255,10 @@ def generateMainClasses(): Unit = {
            * Return the order-dependent hash of the ${i.numerus("given value")}.
            ${(0 to i).gen(j => if (j == 0) "*" else s"* @param o$j the ${j.ordinal} value to hash")("\n")}
            * @return the same result as {@link $Objects#${if (i == 1) "hashCode(Object)" else "hash(Object...)"}}
+           *
+           * @deprecated use {@link Objects#hash} instead
            */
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by pivovarit** [comment]
I was working on #2446 and fixed https://github.com/vavr-io/vavr/pull/2511 meanwhile

**Comment by pivovarit** [comment]
Done, keep in mind that `Tuple1` is a unique scenario - it used `Objects.hashCode` instead of `Objects.hash` logic

File: generator/Generator.sc [hunk 1]
```diff
@@ -2195,7 +2195,7 @@ def generateMainClasses(): Unit = {
 
             @Override
             public int hashCode() {
-                return ${if (i == 0) "1" else s"""Tuple.hash(${(1 to i).gen(j => s"_$j")(", ")})"""};
+                return ${if (i == 0) "1" else if (i == 1) "Objects.hashCode(_1)" else s"""Objects.hash(${(1 to i).gen(j => s"_$j")(", ")})"""};
```
danieldietrich: 👍🏼

Issue #2190, opened 2017-12-13
Unify bi-operators throughout the library and add missing methods


jlorenzen:
Be nice if vavr's `Either` class had a `bipeek` method like AOL's cyclops [Either](https://static.javadoc.io/com.aol.cyclops/cyclops-reactor/9.0.0-MI7/cyclops/control/lazy/Either.html#bipeek-java.util.function.Consumer-java.util.function.Consumer-) class. This would replace having to chain a `peek` and `peekLeft`.

nfekete:
If this gets added, you might consider overloading `peek` instead of introducing `bipeek`. I don't see a reason to name it differently.

danieldietrich:
Currently we have to disjoint union types: Either and Validation.

* both types are biased, Either operates on Right, Validation operates on Valid
* both currently have **map**, **mapLeft** and **bimap**
* other **bixxx** methods are not present
* the non-biased / alternate methods are named differently, e.g. Validation.getErrors() and Either.getLeft(). This makes it currently impossible to have a unified interface for both

We should unify the names and have a hidden common interface, maybe similar to Value?

For the Map collection we use similar methods. E.g. Map.bimap that takes a Key and a Value mapper.

I like the idea to provide similar-named overloads throughout the library, like @nfekete suggested.

danieldietrich:
On help let's discuss it first.

thadumi:
Should we introduce a BiTransformable<T1,T2>  (or Transformable2<T1,T2>) with the following methods?
```java
peek(Consumer<? super T1> c1, Consumer<? super T2> c2)
map(Function<? super T1,? extends X> mapper1, Function<? super T2,? extends Y> mapper2)
```
The name Trasformable2 would be more aligned with the names of the other Vavr's interfaces, as Function1, Function2,...

Should this interface be implemented also by Try? 

> We should unify the names and have a hidden common interface, maybe similar to Value?
I think that we could introduce a common interface, not just for Either and Validation, but also for Try and Option. This interface could be a Value2<L, R>. Or what about ForkValue<L, R> which represents the fact that a class is projected in two difference subclasses?

danieldietrich:
For now I would not add additional interfaces. If a 3rd party app needs a common interface, it could define a functional interface on its own and just use method references to the Vavr methods.

I would just start adding overloads `peek` and `map` for the types you mentioned: Try, Option, Either and Validation.

PR #2532, opened 2019-12-11
Unify bi-operators throughout the library #2190 

thadumi:
Added missing methods to Try, Option, Either, and Validation according to issue #2190

### DIFF
--- a/src/main/java/io/vavr/control/Either.java
+++ b/src/main/java/io/vavr/control/Either.java
@@ -796,14 +796,18 @@ public abstract class Either<L, R> implements Iterable<R>, io.vavr.Value<R>, Ser
     @Override
     public final Either<L, R> peek(Consumer<? super R> action) {
         Objects.requireNonNull(action, "action is null");
-
-        return peek(l -> {}, action);
+        if (isRight()) {
+            action.accept(get());
+        }
+        return this;
     }
 
     public final Either<L, R> peekLeft(Consumer<? super L> action) {
         Objects.requireNonNull(action, "action is null");
-
-        return peek(action, r -> {});
+        if (isLeft()) {
+            action.accept(getLeft());
+        }
+        return this;
     }
 
     /**
--- a/src/main/java/io/vavr/control/Either.java
+++ b/src/main/java/io/vavr/control/Either.java
@@ -573,6 +573,31 @@ public abstract class Either<L, R> implements Iterable<R>, io.vavr.Value<R>, Ser
     }
 
     /**
+     * Maps either the left, or the right side of this disjunction.
+     *
+     * <pre>{@code
+     * Either<?, AtomicInteger> success = Either.right(new AtomicInteger(42));
+     *
+     * // prints "Right(42)"
+     * System.out.println(success.map(Function1.identity(), AtomicInteger::get));
+     *
+     * Either<Exception, ?> failure = Either.left(new Exception("error"));
+     *
+     * // prints "Left(error)"
+     * System.out.println(failure.map(Exception::getMessage, Function1.identity()));
+     * }</pre>
+     *
+     * @param leftMapper  maps the left value if this is a Left
+     * @param rightMapper maps the right value if this is a Right
+     * @param <X>         The new left type of the resulting Either
+     * @param <Y>         The new right type of the resulting Either
+     * @return A new Either instance
+     */
+    public final <X, Y> Either<X, Y> map(Function<? super L, ? extends X> leftMapper, Function<? super R, ? extends Y> rightMapper) {
+        return bimap(leftMapper, rightMapper);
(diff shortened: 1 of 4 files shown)


src/main/java/io/vavr/control/Try.java
```diff
@@ -694,6 +694,26 @@ public final boolean isSingleValued() {
...
+     * @param mapSuccess A mapping function taking the current value
+     * @param <U> The new component type
+     * @return
+     */
+    public final <U> Try<U> map(Function<? super Throwable, ? extends U> mapFailure, Function<? super T, ? extends U> mapSuccess) {
+        Objects.requireNonNull(mapFailure, "mapFailure is null");
+        Objects.requireNonNull(mapSuccess, "mapSuccess is null");
+
+        if (isFailure()) {
+            return Try.of(() -> mapFailure.apply(getCause()));
```
thadumi: Map with both parameters for a Try could have two behaviours. In the first one the mapFailure parameter acts as a recover function, while in the second one mapFailure acts as a mapping from Throwable to ? extends Throwable. 

In the second case, the behaviour will be the same as calling `toEither().map(mapFailure, mapSuccess)`, and, at the same time, we can ensure that the left contains only a Throwable value. Instead in the first scenario calling `toEither().map(mapFailure, mapSuccess)` feels like having an Either<Recover_Type, Success_Type>.

What do you think about it?

danieldietrich: Hi Theodor, that‘s a good question. I‘m currently sitting in a meetup and will take a look later. But I see parallels to the existing fold() and transform() methods. Maybe the map() methods are kind of redundant here...

danieldietrich: @thadumi I've double-checked the 2.0.0 branch, where I already hardened the upcoming API.

Several methods will change, especially [`transform`](https://github.com/vavr-io/vavr/blob/v2.0.0/src/main/java/io/vavr/control/Try.java#L701-L711):

```java
// Vavr 1.0
<U> U transform(Function<Try<T>, U> f)

// Vavr 2.0
<U> Try<U> transform(CheckedFunction<Throwable, Try<U>> ifFailure, CheckedFunction<T, Try<U>> ifSuccess)
```

`fold` is already there and will stay that way:

```java
<U> U fold(Function<Throwable, U> ifFailure, Function<T, U> ifSuccess)
```

By using `fold` and `transform` we are already able to map the underlying values:

```java
transform(err => Failure(new Error("")), x => Success("whatever"))
```

When Java Records & Switch Expressions & Pattern Matching & Object Deconstruction will be available, a convenient way in Java would be:

```java
var mappedTry = switch(myTry) {
    case Success(val) -> mapSuccess.apply(val)
    case Failure(err) -> mapFailure.apply(err)
}
```

I know that these solutions are a bit more verbose than the map overload we thought about. But it is important to me that we keep the API surface area small. There are already several options to achieve a similar result.

So let's keep the new `peek` and abandon the `map` overloads. Okay?

thadumi: Yes, I see your point. 
I just removed them.

src/main/java/io/vavr/control/Either.java
```diff
@@ -795,6 +816,10 @@ private LeftProjection(Either<L, R> either) {
             return either.<L2, R2> bimap(leftMapper, rightMapper).left();
         }
 
+        public <L2, R2> LeftProjection<L2, R2> map(Function<? super L, ? extends L2> leftMapper, Function<? super R, ? extends R2> rightMapper) {
+            return bimap(leftMapper, rightMapper);
+        }
```
danieldietrich: Please remove this map method

src/main/java/io/vavr/control/Either.java
```diff
@@ -1050,6 +1075,10 @@ private RightProjection(Either<L, R> either) {
             return either.<L2, R2> bimap(leftMapper, rightMapper).right();
         }
 
+        public <L2, R2> RightProjection<L2, R2> map(Function<? super L, ? extends L2> leftMapper, Function<? super R, ? extends R2> rightMapper) {
+            return bimap(leftMapper, rightMapper);
+        }
```
danieldietrich: Please remove this map method

src/main/java/io/vavr/control/Validation.java
```diff
@@ -777,8 +777,30 @@ public final void forEach(Consumer<? super T> action) {
...
+        } else {
+            validAction.accept(get());
+        }
+
+        return this;
+    }
+
     @Override
     public final Validation<E, T> peek(Consumer<? super T> action) {
+        Objects.requireNonNull(action, "action is null");
```
danieldietrich: Great!

PR #2539, opened 2019-12-15
Return source BitSet if creating new BitSet based on it via BitSet#ofAll(Iterable)

pivovarit:
https://github.com/vavr-io/vavr/issues/1873


pivovarit:
Nope, that doesn't make sense because that would sneakily propagate `toInt` function of the original `BitSet`

danieldietrich:
Yep, that's right. We can't sneaky propagate injected behavior. (We did take care of that for sorted collections in general...)

PR #2548, opened 2020-01-03
Add reverse() in Set, And Implement it in TreeSet and LinkedHashSet

ankushchaudhary:
**What does this PR do?**
* It adds `reverse()` in `Set`, And Implement it in `LinkedHashSet` and `TreeSet`.

**What issue does this fix or reference to?**
* It reference to [Issue-2514](https://github.com/vavr-io/vavr/issues/2514)

### DIFF
--- a/src/main/java/io/vavr/collection/Set.java
+++ b/src/main/java/io/vavr/collection/Set.java
@@ -164,6 +164,15 @@ public interface Set<T> extends Traversable<T>, Function1<T, Boolean>, Serializa
     Set<T> removeAll(Iterable<? extends T> elements);
 
     /**
+     * Reverse the order of all given elements in this set, if present.
+     *
+     * @return A new set consisting of the elements in reverse order of this set
+     */
+    default Set<T> reverse() {
+        throw new UnsupportedOperationException("Operation not supported");
+    }
+
+    /**
      * Converts this Vavr {@code Set} to a {@code java.util.Set} while preserving characteristics
      * like insertion order ({@code LinkedHashSet}) and sort order ({@code SortedSet}).
      *
--- a/src/main/java/io/vavr/collection/LinkedHashSet.java
+++ b/src/main/java/io/vavr/collection/LinkedHashSet.java
@@ -22,10 +22,7 @@ import io.vavr.*;
 import io.vavr.control.Option;
 
 import java.io.*;
-import java.util.ArrayList;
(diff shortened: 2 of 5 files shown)


src/main/java/io/vavr/collection/Set.java
```diff
@@ -163,6 +163,15 @@ default Boolean apply(T element) {
...
     Set<T> removeAll(Iterable<? extends T> elements);
 
+    /**
+     * Reverse the order of all given elements in this set, if present.
+     *
+     * @return A new set consisting of the elements in reverse order of this set
+     */
+    default Set<T> reverse() {
+        throw new UnsupportedOperationException("Operation not supported");
+    }
```
danieldietrich: This makes are life harder than it needs to be. A Set is unordered by definition. Please remove this method from Set and add it to SortedSet, but with no default implementation. We need to duplicate it then to LinkedHashSet but that's ok, it is the best we can do.

Could you please add it also to SortedMap and LinkedHashMap? We are then able to close the original issue.

ankushchaudhary: > This makes are life harder than it needs to be. A Set is unordered by definition. Please remove this method from Set and add it to SortedSet, but with no default implementation. We need to duplicate it then to LinkedHashSet but that's ok, it is the best we can do.
> 

Thank you for pointing out my mistake. Will make the suggested changes. But I have one concern regarding`SortedSet`. It is also implemented by `BitSet`. How are we going to handle it. Should I throw `UnsupportedOperartionException` in `BitSet#reverse()` or you have any other suggestion?

danieldietrich:
I studied the BitSet - I think it isn't possible because `BitSet<T>` allows to operate on generic types `T` in the presence of mappers `fromInt: Integer -> T` and `toInt: T -> Integer`.

When reversing a BitSet, the mappers need also to be changed, which is not possible in an automatic way.

My stomach says that there are too many caveats with this whole issue of adding a `reverse()` method. It does not fit to the type hierarchy, as we already saw, and we have problems to implement it for specific types.

Please let us leave this functionality away, I have no good feeling about it and it will mean maintenance hell because we need to copy paste it to LinkedHashMap and LinkedHashSet.

I hope you understand and I really appreciate your effort. At least we learned more about it. The issue history will help us answering future questions.

Many thanks!

\- Daniel

PR #2562, opened 2020-03-02
support collection creation from java enumerables

pertl:
It would be great to have support for creating vavr collections from instances of java.util.Enumerable<T>

### DIFF
--- a/src/main/java/io/vavr/collection/Array.java
+++ b/src/main/java/io/vavr/collection/Array.java
@@ -154,6 +154,23 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
         return wrap(toArray(elements));
     }
 
+	/**
+	 * Creates a Array that contains the elements of the given {@link Enumeration}.
+	 *
+	 * @param enumeration A {@link Enumeration}
+	 * @param <T>        Component type of the Enumeration.
+	 * @return A Array containing the given elements in the same order.
+	 */
+		public static <T> Array<T> ofAll(Enumeration<? extends T> enumeration) {
+			Objects.requireNonNull(enumeration, "enumeration is null");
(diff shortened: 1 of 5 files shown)


danieldietrich:
Hi & thank you for your PR.
Creating an issue would have been the better choice in the case of an API addition.
I think creating a collection based on an enumeration isn't a common use-case.
My current efforts go in direction of having `ofAll(Iterable)` as only API in the ofAll family.
Therefore I will close this ticket - but I also want to give you an alternative at hand: please take a look at Guava's `forEnumeration(Enumeration enumeration)` (maybe in conjunction with the method reference ::iterator). It will save us a lot of additional factory methods!
Thanks
\- Daniel

pertl:
Hi Daniel,

I was thinking of java API calls like

https://docs.oracle.com/javase/7/docs/api/java/net/NetworkInterface.html#getInetAddresses()

which unfortunately return an enumerable.

But you are right - wrapping it inside a utility method to convert the enumerable to an iterable is straightforward and keeps vavr slim.

Thanks for giving us vavr - I enjoy it daily and it makes living with java in companies that did not change to kotlin or scala bearable again :-)

danieldietrich:
@pertl Thank you for your kind words, Peter! 🙏

carnott-snap:
> [P]lease take a look at Guava's `forEnumeration(Enumeration enumeration)` (maybe in conjunction with the method reference ::iterator). It will save us a lot of additional factory methods!

The Servlet API is similarly dated, with `HttpServletRequest` returning a number of `Enumeration`s, and I am curious on what your suggestion entails.
- Can you explain in more depth?
- Is this a new central `forEnumeration` method?
- Should we leverage Guava?
- Something else.

PR #2563, opened 2020-03-02
space detection looks wrong

ritschwumm:
just '\n' like i added is probably not enough in the presence of a CRLF, but as it is this would not do as said in the comment:
for example if part were `foo\ttest test` then space would now be `" "` instead of `"\t"`.

### DIFF
--- a/generator/Generator.sc
+++ b/generator/Generator.sc
@@ -4101,7 +4101,7 @@ object Generator {
         (part, arg) <- sc.parts zip args.map(s => if (s == null) "" else s.toString)
       } yield {
         // get the leading space of last line of current part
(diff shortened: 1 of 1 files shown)


danieldietrich:
Mmhh, you might be right. I'm currently trying to figure out why the CI fails, the log says 'successfully finished'...

The generated code looks correct, so we need a (counter-)example, which fails with the current implementation.

ritschwumm:
i used something like this
```
		code"""
			static const bool v46[] = {
				${ "true\n,true\,true" }
			}
		"""
```
and was rendered like this:
```
static const bool v46[] = {
	true,

PR #2564, opened 2020-03-02
remove bad blank

ritschwumm:
at least if it is there, the code does not do as advertised above in the comment


danieldietrich:
Mmhh, this build also fails, I need to figure out why, first...

ritschwumm:
if my fix is correct, then this would express what's meant in an even cleaner way:
```"""([ \t]*\r?\n){2,}"""```

Issue #2559, opened 2020-01-31
partition method seems to not work correctly, comparing to scala


kefasb:
Hello, I spotted a weird behaviour of partition method.
I include code examples from Java and Scala which I used to make the comparison.

Java
```
public class PartitionVavr {
    private static final Map<String, Eat> fruitsBeingEaten = new HashMap<>();

    public static void main(String[] args) {

        final Set<String> fruitsToEat = HashSet.of("apple", "banana");

        final var partition = fruitsToEat.partition(PartitionVavr::biteAndCheck);

        System.out.println(partition);
    }

    private static boolean biteAndCheck(String name) {
        final var eat = fruitsBeingEaten.getOrDefault(name, Eat.prepare(name)).bite();
        fruitsBeingEaten.put(name, eat);
        return eat.isEaten();
    }

    private static class Eat {
        final int bitesLimit = 2;
        final int bites;
        final String name;

        public static Eat prepare(String name) {
            return new Eat(0, name);
        }

        private Eat(int bites, String name) {
            this.bites = bites;
            this.name = name;
        }

        public Eat bite() {
            return new Eat(bites + 1, name);
        }

        public boolean isEaten() {
            return bites >= bitesLimit;
        }
    }
}
```

Scala
```
object PartitionScala extends App {
  private val fruitsBeingEaten: mutable.Map[String, Eat] = mutable.Map()

  private val fruitsToEat = Set("apple", "banana")

  private val partition = fruitsToEat.partition(name => biteAndCheck(name))

  println(partition)

  private def biteAndCheck(name: String): Boolean = {
    val eat = fruitsBeingEaten.getOrElse(name, Eat.prepare(name)).bite
    fruitsBeingEaten += (name -> eat)
    eat.isEaten
  }

  private object Eat {
    def prepare(name: String) = new Eat(0, name)
  }

  private class Eat private(val bites: Int, val name: String) {
    final private val bitesLimit = 2

    def bite = new Eat(bites + 1, name)

    def isEaten: Boolean = bites >= bitesLimit
  }
}
```

The Java output is:
`(HashSet(), HashSet())`

The Scala output is:
`(Set(),Set(apple, banana))`

I think the difference is because vavr traverses the collection and checks the predicate twice while scala does it once.

Is this a bug or expected behaviour?

kefasb:
I used vavr 0.10.2

danieldietrich:
You are right! We need to change that.

[...]

PR #2577, opened 2020-05-05
Iterate once to create two iterators in partition

mincong-h:
Fix https://github.com/vavr-io/vavr/issues/2559

### DIFF
--- a/src/main/java/io/vavr/collection/Iterator.java
+++ b/src/main/java/io/vavr/collection/Iterator.java
@@ -23,6 +23,7 @@ import io.vavr.control.Option;
 
 import java.math.BigDecimal;
 import java.util.*;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.*;
 
 import static io.vavr.collection.BigDecimalHelper.areEqual;
@@ -1956,6 +1957,74 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
             final Stream<T> that = Stream.ofAll(this);
             return Tuple.of(that.iterator().takeWhile(predicate), that.iterator().dropWhile(predicate));
         }
+    }
+
+    /**
+     * Creates two new iterators that both iterates over the same elements as
+     * this iterator and in the same order. The duplicate iterators are
+     * considered equal if they are positioned at the same element.
+     * <p>
+     * Given that most methods on iterators will make the original iterator
+     * unfit for further use, this methods provides a reliable way of calling
+     * multiple such methods on an iterator.
+     *
+     * @return a pair of iterators
+     */
+    default Tuple2<Iterator<T>, Iterator<T>> duplicate() {
+        final java.util.Queue<T> gap = new java.util.LinkedList<>();
+        final AtomicReference<Iterator<T>> ahead = new AtomicReference<>();
+        class Partner implements Iterator<T> {
+
+            @Override
+            public synchronized boolean hasNext() {
+                return (this != ahead.get() && !gap.isEmpty()) || Iterator.this.hasNext();
+            }
+
+            @Override
+            public synchronized T next() {
+                if (gap.isEmpty()) {
+                    ahead.set(this);
+                }
+                if (this == ahead.get()) {
+                    final T element = Iterator.this.next();
+                    gap.add(element);
+                    return element;
+                } else {
+                    return gap.poll();
+                }
+            }
+
+            private boolean hasSame(java.util.Queue<T> queue) {
+                return gap == queue;
+            }
+
+            @Override
+            public boolean equals(Object obj) {
+                if (obj == this) {
+                    return true;
+                }
+                if (obj == null) {
+                    return false;
+                }
+                if (obj.getClass().isAssignableFrom(Partner.class)) {
+                    @SuppressWarnings("unchecked")
+                    final Partner that = (Partner) obj;
+                    /*
+                     * Two iterators are equal if and only if they are partners
+                     * (use the same queue for iteration) and they iterated
+                     * over the same number of elements, i.e. no gaps.
+                     */
+                    return that.hasSame(gap) && gap.isEmpty();
+                }
+                return super.equals(obj);
+            }
+
+            @Override
+            public int hashCode() {
+                return gap.hashCode();
+            }
+        }
+        return Tuple.of(new Partner(), new Partner());
     }
 
     @Override
--- a/src/main/java/io/vavr/collection/Iterator.java
+++ b/src/main/java/io/vavr/collection/Iterator.java
@@ -2217,36 +2217,6 @@ interface IteratorModule {
                     return gap.poll();
                 }
             }
-
-            private boolean hasSame(java.util.Queue<T> queue) {
-                return gap == queue;
-            }
-
-            @Override
-            public boolean equals(Object obj) {
-                if (obj == this) {
-                    return true;
-                }
-                if (obj == null) {
-                    return false;
-                }
-                if (obj.getClass().isAssignableFrom(Partner.class)) {
-                    @SuppressWarnings("unchecked")
-                    final Partner that = (Partner) obj;
-                    /*
-                     * Two iterators are equal if and only if they are partners
-                     * (use the same queue for iteration) and they iterated
-                     * over the same number of elements, i.e. no gaps.
-                     */
-                    return that.hasSame(gap) && gap.isEmpty();
-                }
-                return super.equals(obj);
-            }
-
-            @Override
-            public int hashCode() {
-                return gap.hashCode();
-            }
         }
         return Tuple.of(new Partner(), new Partner());
     }
--- a/src/main/java/io/vavr/collection/Iterator.java
+++ b/src/main/java/io/vavr/collection/Iterator.java
@@ -1723,7 +1723,7 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
         if (!hasNext()) {
             return Tuple.of(empty(), empty());
         } else {
-            final Tuple2<Iterator<T>, Iterator<T>> dup = duplicate();
+            final Tuple2<Iterator<T>, Iterator<T>> dup = IteratorModule.duplicate(this);
             return Tuple.of(dup._1.filter(predicate), dup._2.filterNot(predicate));
         }
     }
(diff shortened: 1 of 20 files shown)


mincong-h:
CI: https://travis-ci.org/github/vavr-io/vavr/builds/683313113

src/main/java/io/vavr/collection/Iterator.java
```diff
@@ -1722,10 +1722,16 @@ public U next() {
...
         } else {
-            final Stream<T> that = Stream.ofAll(this);
-            final Iterator<T> first = that.iterator().filter(predicate);
-            final Iterator<T> second = that.iterator().filter(predicate.negate());
-            return Tuple.of(first, second);
+            Stream<T> first = Stream.empty();
+            Stream<T> second = Stream.empty();
+            for (T t : this) {
+                if (predicate.test(t)) {
+                    first = first.append(t);
```
ruslansennov, line 1729: As far as I remember, Stream.append is an O(n) operation, so the whole op will be O(n^2). 
I believe you should use two java's ArrayList objects to partition.

mincong-h: Thanks for your pointer @ruslansennov . This should be fixed now.

src/main/java/io/vavr/collection/Iterator.java
```diff
@@ -1722,10 +1722,16 @@ public U next() {
         if (!hasNext()) {
             return Tuple.of(empty(), empty());
         } else {
-            final Stream<T> that = Stream.ofAll(this);
-            final Iterator<T> first = that.iterator().filter(predicate);
-            final Iterator<T> second = that.iterator().filter(predicate.negate());
-            return Tuple.of(first, second);
+            final java.util.List<T> first = new ArrayList<>();
```
kefasb, line 1725: I do not know if it is an issue here, but...
This solution is filtering eagerly and previous solution was filtering lazily.

If we would like to keep the same behaviour there is an option of creating a memoized predicate and then use it during filtering, e.g.
```java
final Predicate<? super T> memoizedPredicate = Function1.of(predicate::test).memoized()::apply;
final Iterator<T> first = that.iterator().filter(memoizedPredicate);
final Iterator<T> second = that.iterator().filter(memoizedPredicate.negate());
```

kefasb: I found that my solution with memoized predicate will not work in case when a mutable object mutates to the value that corresponds to another object to be filtered. Then another object is not evaluated but taken from cache with wrong result for it.
Example test:
```java
final HashSet<MutableInteger> integers = HashSet.of(MutableInteger.of(1), MutableInteger.of(2));
final Tuple2<HashSet<MutableInteger>, HashSet<MutableInteger>> partition = integers.partition(mutableInteger -> mutableInteger.incrementAndCheckGreaterThan(2));
assertThat(partition._1).isEqualTo(HashSet.of(MutableInteger.of(3)));
assertThat(partition._2).isEqualTo(HashSet.of(MutableInteger.of(2)));
```
Where 
```java
    private static class MutableInteger {
        private int integer;
...
        boolean incrementAndCheckGreaterThan(int i) {
            return integer++ > i;
        }
...
    }
```
```
In this case the result of evaluation `MutableInteger.of(1).incrementAndCheckGreaterThan(2)` is put into cache with entry `MutableInteger.of(2), false`.
Then when the next object from set `MutableInteger.of(2)` is filtered and it is found in the cache and evaluated to `false`. While normally it would evaluate to `MutableInteger.of(3) -> true.


But the question stays whether we want to keep the partitioning lazy or not.

File: src/main/java/io/vavr/collection/Iterator.java [hunk 3]
```diff
@@ -1722,10 +1722,16 @@ public U next() {
         if (!hasNext()) {
             return Tuple.of(empty(), empty());
         } else {
-            final Stream<T> that = Stream.ofAll(this);
```
kefasb, line 1725: There is also `partition` method implemented by `Stream` which works wrongly either.

I do not know another places. Maybe @danieldietrich could point them out.

mincong-h: I was not aware of the problem of `Stream#partition`. Thanks for pointing it out! :bow: I checked all data-structures in VAVR, apart from Stream, all other implementations work correctly. About the Stream itself, I'm afraid that I don't know how to fix it and I'm not sure if it's really a bug. In Scala 2.13.2, the Stream implementation uses the predicates twice ([link](https://github.com/scala/scala/blob/v2.13.2/src/library/scala/collection/immutable/Stream.scala#L151)) as below:

```scala
override def partition(p: A => Boolean): (Stream[A], Stream[A]) = (filter(p(_)), filterNot(p(_)))
```

If I modify your example in issue #2559 by changing Set into Stream, you will see result in Scala as follow:

```scala
  "Stream" should "partition correctly" in {
    val fruitsToEat = Stream("apple", "banana")
    val partition = fruitsToEat.partition(name => biteAndCheck(name))

    partition._1 shouldEqual Stream()
    partition._2 shouldEqual Stream()  // not Stream("apple", "banana")

    fruitsBeingEaten.get("apple").get.name shouldEqual "apple"
    fruitsBeingEaten.get("apple").get.bites shouldEqual 2  // not 1
    fruitsBeingEaten.get("banana").get.name shouldEqual "banana"
    fruitsBeingEaten.get("banana").get.bites shouldEqual 2  // not 1
  }
```

So I would say the VAVR is aligned with Scala on Stream's behaviors. Also, my current approach (ArrayList) does not fit the Stream requirement, because Stream is lazy sequence of elements which may be infinitely long. So I don't know how to use only one predicate to achieve that... So my suggestion is let's keep stream as it is and avoid modifying it.

src/main/java/io/vavr/collection/Iterator.java
mincong-h: This is a bit too technical for me... Maybe @danieldietrich can have an answer about it.

src/main/java/io/vavr/collection/Iterator.java
ruslansennov: > Stream is lazy sequence of elements which may be infinitely long

Oh sorry guys I completely forgot about it :(

mincong-h:
Thank you Daniel for your comments (both here and in the issue). I will take a look this weekend.

danieldietrich:
@mincong-h thank you! please find my answers to your 3 concerns below.

> the correctness in concurrent situation

An Iterator is stateful. It does not offer thread-safety on its own. The user has two options in a concurrent setting:

* explicitly synchronize iterator access from the outside
* alternatively use a persistent collection instead

Please remove the occurrences of synchronized. The `AtomicReference` also isn't needed.

> whether or not the duplicate method should be part of the public API

For now, I would hide it in the IteratorModule you mentioned. We offer a static method:

```
public Iterator<T> {
    // ...
}

interface IteratorModule {
  static <T> Tuple2<Iterator<T>, Iterator<T>> duplicate(Iterator<T> iterator) {
    // ...
  }
}
```

> whether you see improvement about testing

For Iterators, `equals` and `hashCode` aren't overwritten, we use Object equality because an Iterator may be infinite and there is no way to compare the elements.

Therefore, please remove `hasSame`, `equals` and `hashCode` from `Partner`.

The unit tests cannot use the `isEqualTo` method on Iterators. We need to first consume the elements, e.g. by storing them in a `List` or by joining them in a `String`. These then can be compared. Maybe the cheapest solution is to compare joined strings.

It is great to see that we will get the `duplicate` method :)

src/main/java/io/vavr/collection/Iterator.java
```diff
@@ -1952,6 +1951,74 @@ public boolean hasNext() {
...
+     * @return a pair of iterators
+     */
+    default Tuple2<Iterator<T>, Iterator<T>> duplicate() {
+        final java.util.Queue<T> gap = new java.util.LinkedList<>();
+        final AtomicReference<Iterator<T>> ahead = new AtomicReference<>();
+        class Partner implements Iterator<T> {
+
+            @Override
+            public synchronized boolean hasNext() {
+                return (this != ahead.get() && !gap.isEmpty()) || Iterator.this.hasNext();
```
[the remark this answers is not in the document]
danieldietrich: > But we need to use the "parent" iterator

The static duplicate method receives the parent Iterator as parameter:

IteratorModule.duplicate(Iterator iter)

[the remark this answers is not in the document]
mincong-h: Ah yes 😂 👍

mincong-h:
@danieldietrich , thanks for your suggestion and comments, really appreciated! I addressed most of the comments, here are something left.

> Please remove the occurrences of synchronized. The `AtomicReference` also isn't needed.

I cannot remove `AtomicReference`, because I need a mutable object holder for holding the reference. It's not for being thread safe here. Assigning the reference to `Iterator<T> ahead` is not an option, because it fails the compilation with warning: _"Variable 'ahead' is accessed from within innerclass, needs to be final or effectively final."_

> The unit tests cannot use the `isEqualTo` method on Iterators. We need to first consume the elements, e.g. by storing them in a `List` or by joining them in a `String`. These then can be compared. Maybe the cheapest solution is to compare joined strings.

Vavr has it own implementation about `assertThat(Iterator)` in `IteratorTest` as `IterableAssert`, which transforms the actual iterator and expected iterator into two lists and then compare them. So I think it's fine to use `isEqualsTo`. But just in case of something I didn't think about, I use what you suggested, i.e. joining them into a `String` and compare.

> It is great to see that we will get the `duplicate` method :)

+1 :)

mincong-h:
Hi Daniel. Thanks for your review, I learnt a lot from them. I'm looking forwards to contributing more in the future 😀

danieldietrich:
Hi Mincong, I‘m also looking forward to your future contributions, it is fun to work with you!

Currently I have to finish some heavy duty tasks, such as restructuring the core Gradle project a bit and pulling some Git projects back into the core multi-module project. We need a clean version baseline in order to prevent versioning conflicts.

Additionally, I will change Try, Option and Either accordingly to the current 2.0.0 branch (which will be deleted afterwards, it is just an arbitrary name).

Finally, after some polishing, we will be able to release Vavr 1.0. After that, I will focus on a rework of the web page and the documentation. I expect also that there will be many questions and migration issues.

Additionally, I expect several new feature requests. Here come new PRs into play but we will see.

Thanks!

mincong-h:
The future looks exciting and your plan makes sens. Thanks for doing this Daniel! Do not hesitate to ping me if you need any help. I will try to grab some issues whenever I can as well.

Best,
Mincong

PR #2588, opened 2020-05-16
No more failures in the build

mincong-h:
Thanks to the following PRs, the build are no longer failing on JDK 14 and JDK EA:

- https://github.com/vavr-io/vavr/pull/2578 Upgrade to Gradle 6.3
- https://github.com/vavr-io/vavr/pull/2581 Resolving 'yield may become a restricted identifier in a future releases' warning
- https://github.com/vavr-io/vavr/pull/2585 Javadoc error fix 'heading used out of sequence' (JDK 13.0.2)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,9 +19,6 @@ matrix:
     env: DEPLOY=true
   - jdk: openjdk14
   - jdk: openjdk-ea
-  allow_failures:
-  - jdk: openjdk14
-  - jdk: openjdk-ea
 
 before_cache:
 - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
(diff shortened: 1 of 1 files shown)


sleepytomcat:
Great news, glad to be part of it!

.travis.yml
```diff
@@ -19,9 +19,6 @@ matrix:
     env: DEPLOY=true
   - jdk: openjdk14
   - jdk: openjdk-ea
-  allow_failures:
-  - jdk: openjdk14
-  - jdk: openjdk-ea
```
danieldietrich, line 24: We still allow failures for openjdk-ea, which is a floating release in the sense that it is always the actual JDK development and not tied to a specific version.

mincong-h: You are right Daniel. I hesitated when I deleted this line. I thought it would be easier to detect failure because when the EA version is changed and introduce new warnings or errors, the build status will be :x: directly. But your argument makes more sense. I added back the lines.

danieldietrich: I see you point! Thank you, I will merge it

PR #2594, opened 2020-05-16
fixes #2584 -- Add Seq.getOption(int) and Seq.getOrElse(int, T)

### LINKED ISSUE #2584: Add Seq.getOption(int) and Seq.getOrElse(int, T)
Hi, would a PR adding Traversable.getOption be accepted? It would return `Option<T>` when getting a certain index instead of `T`.

 In fact it could be argued that this should be the default and [from what I can tell](https://stackoverflow.com/questions/4981689/get-item-in-the-list-in-scala), that's what scala does.

 Currently we're forced to do things like:

```java
final Option<RideInfo> rideOne = recentRidesFromNewerToOlder.headOption();
final Option<RideInfo> rideTwo = recentRidesFromNewerToOlder.length() >= 2 ? Option.of(recentRidesFromNewerToOlder.get(1)) : Option.none();
```

emmanueltouzery:
i believe the PR is correct, however I am annoyed by the code duplication between get, getOrElse and getOption, especially for List, Stream and Queue.

However I think it's unavoidable if we don't want to create "unnecessary" intermediary objects at runtime, which would enable code reuse (or worse, catching exceptions as part of normal operation). Since you suggested getOrElse especially for performance, I'm guessing you'd rather not to that... Unless if we did that only for List, Stream and Queue, which have less trivial get() implementations.. and List and Stream are not very performance oriented.

### DIFF
--- a/src/main/java/io/vavr/collection/Vector.java
+++ b/src/main/java/io/vavr/collection/Vector.java
@@ -742,24 +742,6 @@ public final class Vector<T> implements IndexedSeq<T>, Serializable {
     private boolean isValid(int index) { return (index >= 0) && (index < length()); }
 
     @Override
-    public Option<T> getOption(int index) {
-        if (isValid(index)) {
-            return Option.some(trie.get(index));
-        } else {
-            return Option.none();
-        }
-    }
-
-    @Override
-    public T getOrElse(int index, T defaultValue) {
-        if (isValid(index)) {
-            return trie.get(index);
-        } else {
-            return defaultValue;
-        }
-    }
-
-    @Override
     public T head() {
         if (nonEmpty()) {
             return get(0);
--- a/src/main/java/io/vavr/collection/Vector.java
+++ b/src/main/java/io/vavr/collection/Vector.java
@@ -742,6 +742,24 @@ public final class Vector<T> implements IndexedSeq<T>, Serializable {
     private boolean isValid(int index) { return (index >= 0) && (index < length()); }
 
     @Override
+    public Option<T> getOption(int index) {
+        if (isValid(index)) {
+            return Option.some(trie.get(index));
+        } else {
+            return Option.none();
+        }
+    }
+
+    @Override
+    public T getOrElse(int index, T defaultValue) {
+        if (isValid(index)) {
+            return trie.get(index);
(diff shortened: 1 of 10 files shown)


emmanueltouzery:
That does mean double bounds check on getOption & getOrElse but then if you want to do that in a loop, there are better ways than looping on an index.. I'll have the change soon.

emmanueltouzery:
I've applied the changes. I agree that for Stream we can reuse getOption for getOrElse, and maybe we could even share the implementation for get(). It's not really a high-performance collection, a constant cost shouldn't be a big problem.

emmanueltouzery:
What if for instance in vector, we'd call directly true.getIndex, and similar in other collections, to avoid double bounds checking?

emmanueltouzery:
but you want to have a public `apply` method without bounds check? I wouldn't expect that to be public, as it has no guarantee as to the type of exceptions it'll throw, right?
But appart from the 'public' part, it sounds good, on it.

danieldietrich:
@emmanueltouzery yeah, that's right. In fact, we already have a public method `T apply(Integer)` on `Seq` because `Seq` implements `PartialFunction`.

A `PartialFunction f` works exactly that way:

* `f.isDefinedAt(t)` states whether the function is defined at a specific value `t`
* `f.apply(t)` is defined when, and only when `f.isDefinedAt(t) == true`, otherwise the behavior isn't predictable (a call may throw or may return an arbitrary value).

Please go ahead!

emmanueltouzery:
done. I didn't manage to do a javadoc link to isDefinedAt :blush:, otherwise it's starting to look nice!

src/main/java/io/vavr/collection/Vector.java
```diff
@@ -732,14 +732,9 @@
...
-    public T get(int index) {
-        if (isValid(index)) {
-            return trie.get(index);
-        } else {
-            throw new IndexOutOfBoundsException("get(" + index + ")");
-        }
+    public T apply(Integer index) {
+        return trie.get(index);
     }
-    private boolean isValid(int index) { return (index >= 0) && (index < length()); }
```
danieldietrich, line 742: Good!

PR #2598, opened 2020-05-20
Add a version of toSortedMap that accepts a mergeFunction

ventuno:
Following up on https://github.com/vavr-io/vavr/pull/2595#issuecomment-630939379 adding a merge function to `toSortedMap`.

### DIFF
--- a/src/main/java/io/vavr/collection/Traversable.java
+++ b/src/main/java/io/vavr/collection/Traversable.java
@@ -1707,7 +1707,7 @@ public interface Traversable<T> extends Iterable<T>, Foldable<T>, io.vavr.Value<
     /**
      * Converts this collection to a {@link SortedMap}.
      *
-     * @param comparator  A comparator that induces an order of the Map keys.
+     * @param keyComparator  A comparator that induces an order of the Map keys.
      * @param keyMapper   A function that maps an element to a key
      * @param valueMapper A function that maps an element to a value
      * @param merge A function that merges values that are associated with the same key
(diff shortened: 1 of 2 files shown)


src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -1704,6 +1704,43 @@ default Number sum() {
...
+     * @param <V>         The value type
+     * @return A new {@link TreeMap}.
+     */
+    default <K, V> SortedMap<K, V> toSortedMap(
+            Comparator<? super K> comparator,
+            Function<? super T, ? extends K> keyMapper,
+            Function<? super T, ? extends V> valueMapper,
+            BiFunction<? super V, ? super V, ? extends V> merge) {
+        Objects.requireNonNull(comparator, "comparator is null");
+        return TreeMap.ofEntries(comparator, toMap(keyMapper, valueMapper, merge));
```
danieldietrich, line 1724: Please add a shortcut for the empty case:

```java
if (isEmpty()) {
    return TreeMap.empty(comparator);
} else {
    ...
}
```

We could do the same for the other PRs, `toMap` and `toLinkedMap`.

danieldietrich: Please rename the `comparator` to `keyComparator` to clarify it a bit.

src/test/java/io/vavr/collection/CollectionsTest.java
```diff
@@ -112,6 +113,27 @@ public void shouldUseMergeFunctionToHandleKeysCollisions() {
...
+        assertThat(m.toVector()).isEqualTo(Vector.of(Tuple.of("bb", 2), Tuple.of("a", 63)));
+
+        Map<String, Integer> m2 = input.toSortedMap(
+                Tuple2::_1,
+                Tuple2::_2,
+                Integer::sum);
+
+        assertThat(m2.toVector()).isEqualTo(Vector.of(Tuple.of("a", 63), Tuple.of("bb", 2)));
+    }
+
```
danieldietrich, line 136: Please split the unit test. We strive for one assertion per test method.
I know that we often have more than one assert but the _testee_  should be called only one time per test.

In our case, we could create 4 test methods, two tests for each of the two `toSortedMap` conversion methods:

* converting an empty collection
* converting a non-empty collection

(I'm just thinking loud here: It is an implementation detail that the once conversion method delegates to the other. We don't expose that knowledge as public API.)

ventuno: That makes total sense: one test per use case. I added all tests for the `isEmpty` case as well.

I couldn't find any usages of a "spy" helper to check that `keyMapper`/`valueMapper`/`comparator`/`mergeFunction` are never called which I think makes the test a little less useful, but I'm happy to update the code if there's something that I can use to this effect.

src/main/java/io/vavr/collection/Traversable.java
ventuno: I just did:
```Java
if (...) {
 ...
}
return ...;
```
as there's no real need for the `else` branch, let me know if you prefer the explicit style.
I also added the same case for `toMap`.

PR #2601, opened 2020-05-22
Add a version of toLinkedMap that accepts a mergeFunction

ventuno:
Following up on https://github.com/vavr-io/vavr/pull/2595#issuecomment-630939379 adding a merge function to `toLinkedMap`.

@danieldietrich I'm not sure if I managed to capture all challenges you mentioned in [your comment](https://github.com/vavr-io/vavr/pull/2595#issuecomment-630939379). So please let me know if I missed something.

### DIFF
--- a/src/main/java/io/vavr/collection/Traversable.java
+++ b/src/main/java/io/vavr/collection/Traversable.java
@@ -1752,6 +1752,23 @@ public interface Traversable<T> extends Iterable<T>, Foldable<T>, io.vavr.Value<
      * @param merge A function that merges values that are associated with the same key
      * @param <K>         The key type
      * @param <V>         The value type
+     * @return A new {@link LinkedHashMap}.
+     */
+    default <K extends Comparable<? super K>, V> Map<K, V> toLinkedMap(
+            Function<? super T, ? extends K> keyMapper,
+            Function<? super T, ? extends V> valueMapper,
+            BiFunction<? super V, ? super V, ? extends V> merge) {
+        return LinkedHashMap.ofEntries(toMap(keyMapper, valueMapper, merge));
+    }
+
+    /**
+     * Converts this collection to a {@link Map}.
+     *
+     * @param keyMapper   A function that maps an element to a key
+     * @param valueMapper A function that maps an element to a value
+     * @param merge A function that merges values that are associated with the same key
+     * @param <K>         The key type
+     * @param <V>         The value type
      * @return A new {@link HashMap}.
      */
     default <K, V> Map<K, V> toMap(
--- a/src/main/java/io/vavr/collection/Traversable.java
+++ b/src/main/java/io/vavr/collection/Traversable.java
@@ -1758,7 +1758,15 @@ public interface Traversable<T> extends Iterable<T>, Foldable<T>, io.vavr.Value<
             Function<? super T, ? extends K> keyMapper,
             Function<? super T, ? extends V> valueMapper,
             BiFunction<? super V, ? super V, ? extends V> merge) {
-        return LinkedHashMap.ofEntries(toMap(keyMapper, valueMapper, merge));
+        final Map<K, V> map = toMap(keyMapper, valueMapper, merge);
+        final Iterator<Tuple2<K, V>> entries = iterator()
+                .map(keyMapper)
+                .distinct()
+                .map(key -> {
+                    V value = map.getOrElse(key, null); // the default value will not be used
+                    return Tuple.of(key, value);
+                });
+        return LinkedHashMap.ofEntries(entries);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -1744,6 +1744,23 @@ default Number sum() {
...
+     * @param <K>         The key type
+     * @param <V>         The value type
+     * @return A new {@link LinkedHashMap}.
+     */
+    default <K extends Comparable<? super K>, V> Map<K, V> toLinkedMap(
+            Function<? super T, ? extends K> keyMapper,
+            Function<? super T, ? extends V> valueMapper,
+            BiFunction<? super V, ? super V, ? extends V> merge) {
+        return LinkedHashMap.ofEntries(toMap(keyMapper, valueMapper, merge));
+    }
```
danieldietrich, line 1762: I think we do not maintain key order here, which is iteration order of the elements of this Traversable.

The result of `toMap` does not guarantee any specific key order. So we need to do some extra work to order the keys of the LinkedHashMap:

```java
default <K extends Comparable<? super K>, V> Map<K, V> toLinkedMap(
        Function<? super T, ? extends K> keyMapper,
        Function<? super T, ? extends V> valueMapper,
        BiFunction<? super V, ? super V, ? extends V> merge) {
    final Map<K, V> map = toMap(keyMapper, valueMapper, merge);
    final Iterator<K, V> entries = iterator().map(t -> {
        K key = keyMapper.apply(t);
        V value = map.getOrElse(key, null); // the default value will not be used
        return Tuple.of(key, value);
    });
    return LinkedHashMap.ofEntries(entries);
}
```

ventuno: Thanks @danieldietrich. I was thinking that as the implementation of [`groupBy`](https://github.com/vavr-io/vavr/blob/9a40af5cec2622a8ce068d5833a2bf07671f5eed/src/main/java/io/vavr/collection/Collections.java#L212) is based on a `LinkedHashMap` this would work.

The issue I'm seeing with the suggested code is that (with reference to the code in `shouldUseMergeFunctionToHandleKeysCollisionsLinkedMap`):
1) given `Vector.of(Tuple.of("b",20), Tuple.of("a",21), Tuple.of("c",21), Tuple.of("b",1), Tuple.of("a",21), Tuple.of("a",21))` as the `input` `Vector`;
1) `entries` will be an `Iterator<Tuple2<K,V>>` whose items are: `(Tuple.of("b", 21), Tuple.of("a", 63), Tuple.of("c", 21), Tuple.of("b", 21), Tuple.of("a", 63), Tuple.of("a", 63))` (notice repeated entries);
1) when the above is converted to a `LinkedHashMap` then the key order is `c -> a -> b`. My understanding is that we should expect the same ordering we see in the `input` `Vector` (`b -> a -> c`).

If my assumptions are correct, then an alternative (working) solution could be:
```Java
final Map<K, V> map = toMap(keyMapper, valueMapper, merge);
final Iterator<Tuple2<K, V>> entries = iterator()
            .map(keyMapper)
            .distinct()
            .map(key -> {
                V value = map.getOrElse(key, null); // the default value will not be used
                return Tuple.of(key, value);
            });
return LinkedHashMap.ofEntries(entries);
```

Thoughts?

danieldietrich: @ventuno your solution is great (and mine is buggy 😅)! Please do it like that.

danieldietrich: That push was fast :-) I will way for the CI and merge it!

ventuno: Thanks @danieldietrich I pushed one extra commit to check that the output map is an `instanceof` the expected type. Let me know if that make sense and I can do it for `SortedMap` as well.

danieldietrich: For `toLinkedMap` this extra check makes sense because the return type is of the more general type `Map`. We have to notion for the _linked_ property on the type level.

We don't need an additional `instanceof` check for `toSortedMap` because the return type is a `SortedMap`. That is sufficient.

I'm happy with your PR! If you are happy, too, give me a sign and I will merge it ;)

ventuno: I agree @danieldietrich we should let the compiler do the "assertion" for us. I just changed the return type to an explicit `SortedMap` in tests. Does that make sense?

danieldietrich: Yes, it makes it more explicit that we expect a `SortedMap`. Especially the test would fail if we accidentally changed the return type of the conversion method.

ventuno: Perfect. Let's take a minute to internally celebrate type safety in programming languages!

PR #2608, opened 2020-07-27
Issue #2552 Skip Executing Actions if the Future is cancelled.

### LINKED ISSUE #2552: Chained futures keep executing although future was cancelled
Further investigation of #2551 shows that CompletableFutures behave differently in a Spring application.

The following code using CompletableFutures stops before the `thenRun` method is executed:
```java
CompletableFuture<Void> f = CompletableFuture.runAsync(() -> {
        try {
            System.out.println(i);
            Thread.sleep(200);
        } catch (InterruptedException e) {
        }
    }
}).thenRun(() -> {
        try {
            System.out.println(i);
            Thread.sleep(200);
        } catch (InterruptedException e) {
        }
    }
});
Thread.sleep(500);

f.cancel(true);
try{
    f.get();
}
catch(CancellationException | ExecutionException e){
    System.out.println("future cancelled");
}
System.out.println("end");
```

Output:
```
1
2
future cancelled
end
3
4
5
6
7
8
9
```

However, when using the vavr future implementation, the second part is executed as well:
```java
Future<Void> f = Future.run(() -> {
    int i = 1;
    while(i < 10){
        try {
            Thread.sleep(200);
        } catch (InterruptedException e) {
        }
        System.out.println(i++);
    }
}).andThen(v -> {
    int i = 1;
    while(i < 10){
        try {
            Thread.sleep(200);
        } catch (InterruptedException e) {
        }
        System.out.println(i++);
    }
});
Thread.sleep(500);
f.cancel();
try{
    f.get();
}
catch(CancellationException e){
    System.out.println("future cancelled");
}

System.out.println("end");
```

Output:
```
1
2
future cancelled
end
3
4
5
6
7
8
9
1
2
3
4
5
6
7
8
9
```

charvakcpatel007:
Fixes #2552

### DIFF
--- a/src/main/java/io/vavr/concurrent/FutureImpl.java
+++ b/src/main/java/io/vavr/concurrent/FutureImpl.java
@@ -376,6 +376,9 @@ final class FutureImpl<T> implements Future<T> {
             if (waiters != null) {
                 waiters.forEach(this::unlock);
             }
+            if(isCancelled()) {
+                return true;
+            }
             if (actions != null) {
                 actions.forEach(this::perform);
                 return true;


src/main/java/io/vavr/concurrent/FutureImpl.java
```diff
@@ -376,6 +376,9 @@ boolean tryComplete(Try<? extends T> value) {
             if (waiters != null) {
                 waiters.forEach(this::unlock);
             }
+            if(isCancelled()) {
```
danieldietrich, line 379: (I don't want to be a nitpicker but you missed one space `if(is...` --> `if (is...`)

theqp:
@charvakcpatel007 add the missed space to this branch so it can be merged.

pivovarit:
@charvakcpatel007 are you willing to finish it? we just need tests

pivovarit:
closing due to lack of activity - feel free to reopen

PR #2612, opened 2020-08-21
 Removed Function type from Set

### LINKED ISSUE #2249: Why is Set a Function1?
I found out after spending several minutes on a compilation error in my code, that `Set` extends `Function1`. I see it has been done in #963 but I don't understand the rationale behind.

However I do see the problems it brings to the party for pattern matching:
```java
Set<String> abc = Set("abc");
Set<String> result = Match(abc).of( // Does not compile: the Java inference engine sees abc as a Function1<String, Boolean> before a Set<String> thus expects result to be of type Boolean
    Case($(), abc)
);
```

Even without considering this pattern matching issue, I think this doesn't make sense. A Set<T> is not a function taking a T and returning whether this Set contains it.
It is also not intuitive: I don't think many (any?) developers expect this.
Last: this looks like "favoritism" towards `io.vavr.collection.Set#contains` method. Why isn't `Set<T>` a `Function1<T, T>` in the name of `io.vavr.collection.Set#add`?

There's also no hint/explanation in the Javadoc.

theqp:
As suggested here: https://github.com/vavr-io/vavr/issues/2249#issuecomment-670188131

resolves #2249

### DIFF
--- a/src/main/java/io/vavr/collection/Set.java
+++ b/src/main/java/io/vavr/collection/Set.java
@@ -78,7 +78,7 @@ import java.util.function.*;
  * @param <T> Component type
  */
 @SuppressWarnings("deprecation")
-public interface Set<T> extends Traversable<T>, Function1<T, Boolean>, Serializable {
+public interface Set<T> extends Traversable<T>, Serializable {
 
     long serialVersionUID = 1L;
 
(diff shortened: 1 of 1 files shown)


danieldietrich (approved):
Awesome, thank you (and sorry for the little delay)! 👋

PR #2616, opened 2020-09-04
added Option.zip

chisui:
When working with `Option` extensively it is quit useful to have a `zip` function to combine multiple `Options` into one.

Maybe a `Option<TupleN<...>> zip(Option<A> optionA, ....)` could also be useful.


chisui:
This is essentially the same as `API.For`. My bad.

PR #2618, opened 2020-10-25
Wrong function name in Either.filterOrElse JavaDoc

gualtierotesta:
Small fix on Either.filterOrElse JavaDoc: it was referring to function filterVal while its name is zero


danieldietrich:
Thank you!

PR #2624, opened 2020-12-20
feat(io.vavr.control): add a model that expresses the return value of functions

kevinten10:
use `LogicalResult` model to expresses return value of logical
function,
use `ProcessResult` model to expresses return value of
instruction function.

### DIFF
--- a/src/main/java/io/vavr/control/LogicalResult.java
+++ b/src/main/java/io/vavr/control/LogicalResult.java
@@ -1,0 +1,212 @@
+package io.vavr.control;
+
+import io.vavr.Tuple1;
+
+import java.util.Objects;
+
+/**
+ * The Results of Logical function returns.
+ */
+@SuppressWarnings("deprecation")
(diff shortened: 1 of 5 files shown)


kevinten10:
This PR is to provide a prototype implementation for the discussion of the issue #2625

kevinten10:
I see that in the v2.0.0 branch, the inheritance of `Value` has been removed.

If I want to continue to improve, should I develop based on _v2.0.0_ or _master_?

Looking forward to your reply.

danieldietrich:
Thank you for your engagement, I really appreciate it!

Before a PR, there needs to be a discussion on the _why_ and _how_ a requested feature strategically fits into Vavr.

I took a look at your source code. You follow the general Vavr style (e.g. factory methods etc), that is great. But I don't see a place for this change in the Vavr core library, maybe because I don't fully understand the domain of application, yet.

Please let's move the further discussion to #2625, I will close this PR for now.

\- Daniel

PR #2627, opened 2021-01-10
Add `prepend` for TupleN

EntryPointKR:
Added the `prepend` in TupleN template.

After we can use:
```
<T> Tuple0.prepend(T t0) = Tuple1<T>
<T> Tuple2<A, B>.prepend(T t0) = Tuple3<T, A, B>
```

### DIFF
--- a/generator/Generator.sc
+++ b/generator/Generator.sc
@@ -2169,6 +2169,19 @@ def generateMainClasses(): Unit = {
 
             ${(i < N).gen(xs"""
               /$javadoc
+               * Prepend a value to this tuple.
+               *
+               * @param <T0> type of the value to prepend
+               * @param t0 the value to prepend
+               * @return a new Tuple with the value prepended
+               */
+              public <T0> Tuple${i+1}<${(0 to i).gen(j => s"T$j")(", ")}> prepend(T0 t0) {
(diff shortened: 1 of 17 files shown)


danieldietrich (changes requested):
Thank you @EntryPointKR,

your PR looks good to me but I have one request for change:

I recognized that the index '0' of the parameters `T0 t0` is unnecessary because tuples always start with index 1. Therefore I would choose the simpler `T t` instead for all prepend methods. Could you change that please?

And you are right, the parameters of the append method should also not have an index to be in line with the prepend method. Therefore I would change `append(Tn tn)` (for n = 0...8) to `append(T t)`. Could you please also do that in another PR?

Many thanks for your support!

\- Daniel

danieldietrich (approved):

danieldietrich:
Hi @EntryPointKR,
I approved the PR now and will perform cross-cutting rename operations of generics and args in a separate PR.
Thx!

EntryPointKR:
@danieldietrich Sorry for not response to the request, I couldn't manage my time. Can I open the PR on weekend?

danieldietrich:
@EntryPointKR sure! Please don't feel bad in any way, I really appreciate your help!

PR #2628, opened 2021-01-24
Rename type parameter T0 for `prepend` in TupleN

EntryPointKR:
Following the [previous discussion](https://github.com/vavr-io/vavr/pull/2627#pullrequestreview-565344945), we thought that the type parameter `T0` should be renamed to `T`.

```
<T> Tuple0.prepend(T t): Tuple1<T>
<T> Tuple2<T1, T2>.prepend(T t): Tuple3<T, T1, T2>
```


danieldietrich (approved):
Awesome, many thanks for the change!

PR #2629, opened 2021-01-24
Rename type parameter Tn for `append` in TupleN

EntryPointKR:
Following the [previous discussion](https://github.com/vavr-io/vavr/pull/2627#pullrequestreview-565344945), we thought that the type parameter `Tn` should be renamed to `T`.

```
<T> Tuple0.append(T t): Tuple1<T>
<T> Tuple2<T1, T2>.append(T t): Tuple3<T1, T2, T>
```

### DIFF
--- a/generator/Generator.sc
+++ b/generator/Generator.sc
@@ -2184,12 +2184,12 @@ def generateMainClasses(): Unit = {
               /$javadoc
                * Append a value to this tuple.
                *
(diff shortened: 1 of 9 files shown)


danieldietrich (approved):
That looks great!
Bonus points for creating a second PR for append :)
Thanks

PR #2642, opened 2021-02-14
more robust slideby classifier function handling

jwbargsten:
When the classifier function is pure everything works fine, but with
state, `slideBy` might end up in an infinite loop. The reason: `slideBy`
calls the classifier two times on the same item. If the second call
returns a different value, `slideBy` will not advance anymore and
iterate indefinitely on the same item.


danieldietrich (approved):
Nice finding! Sorry for the delay and many thanks!

PR #2644, opened 2021-03-12
Modernize the Gradle build

melix:
Hi!

This pull request improves the Gradle build in different manners:

1. it makes it cacheable, so that the code generation step is typically cached and doesn't need to be re-executed always
2. it migrates to the `maven-publish` plugin in preparation for the Gradle 7 upgrade
3. it configures as much as possible lazily, in order to make the builds faster (but the `release` plugin unfortunately configures many things eagerly)

Here are a few build scans showing the result:

- Initial clean build (`test`) executed in 1m 49s : https://scans.gradle.com/s/zreseb3bobn5m
- Subsequent fully up-to-date build executed in <1s: https://scans.gradle.com/s/gk7zeabuhnjzm
- Clean cached build (`test`) executed in <1s: https://scans.gradle.com/s/ffpfbf673cova

Please take a look at each commit message for details about the migration path.


danieldietrich (changes requested):
Hi Cédric,

I was very happy to see your contribution - from a Javaslang/Vavr user of the first hour :) It really shows that you work for Gradle! Your changes are well thought-through. You followed all the best practices and made the build really fast.

I had to think a bit about your changes because they are impactful. Some I want to keep, regarding others I want to handle it a bit differently.

Generally, my thoughts can be classified like this. It is (more) important to me to

- 👉 keep the **build simple** _than_ making it fast
- 👉 show generated **source code** _than_ hiding it
- 👉 replace deprecated Gradle APIs with modern ones

We should do the following changes:

- ✅ the additional release docs
- ✅ use Gradle lazy APIs
- ✅ use Java 8 for builds instead of the version used to start gradle
- ✅ migrate to maven-publish plugin
- ✅ use Nexus Staging repository plugin

We need to keep the current things:

- ⛔️ **Stick to a central ./build.gradle instead of ./build-logic/**
    I also recognized that a single build.gradle mixes u different concerns. However, Vavr's is straightforward: gen / build / test / publish / release. The single gradle reads like a book. If we don't write our own plugins (like you propose), it stays simple.
- ⛔️ **Use ammonite like before, we could use a plugin if s.o. else publishes one to the Gradle plugin registry.**
    The cost-income ratio isn't worth the effort. I do not want to maintain such an amount of additional code just for the purpose of building the project. Mid-term, I plan to replace the Scala code generator. There is something new in the make ;) Build speed isn't prio 1 for me. A PR always starts clean. And a dev locally only starts tests if there is something to test.
- ⛔️ **We need the generated sources for PR reviews and to browse the code on Github.**
    Including the sources enables our users to read the sources on GitHub. Especially we make transparent which code portions changed on PRs. The only bad thing that could happen is that we accidentally commit manually changes of generated sources. Our CI build checks that and the PR will turn red in such a case.

---

Instead of applying the changes to this PR I would love if we could keep it as-is, close it and open a new PR. This PR is the best reference I can imagine on how to change an existing Gradle build by applying best practices a Gradle engineer recommends :)

What do you think?

Thanks
Daniel

melix:
Thanks for the review, @danieldietrich ! I am however afraid I won't make the (important) changes you are requesting on this, because I don't have the bandwidth to do it. The PR as is it is what I, as a Gradle representative, think what represents a state of the art, idiomatic Gradle build. It's best practices aimed at reducing the mental overhead for future readers and maintainers. As such it's the result of several years of experience, both in developing the tool obviously, but also reading comments from people who are pro and anti-Gradle.

In particular:

1. I think it is _very_ important to split the `build-logic` from the main build script. This is the most important thing to do, as it helps keeping the main build file declarative instead of having a mix of build logic and build configuration. This is also important because it means that readers of the build can immediately see what is relevant, rather than having to understand what the custom logic does, like I had to do when migrating. You can [read my blog post](https://melix.github.io/blog/2019/11/gradle-scripting.html) about this topic or watch my 10 minutes video about [Authoring Gradle plugins for readability, maintainability and user experience](https://www.youtube.com/watch?v=MaansFoPHKg) 
2. I'm fine if the ammonite plugin is externalized, but it's not worth the effort as you mentioned. The performance benefit was extremely important for me, especially because I have a slow internet connection. Benefiting from the build cache was important. You can keep it in `build-logic` and it's perfectly fine, that's precisely the role of custom build logic and why it _must not_ live in the main `build.gradle` file.
3. putting generating files in checked in sources is considered a bad practice. It's confusing because, in most (all?) projects you only check in generators, not generated sources, so there's a mental overhead in figuring out why it's there. It also unnecessarily increases reviews, IMO, since you see code you shouldn't care about. If the intent is to see what is generated just by reading a review, I can understand the benefit but it's IMO not worth the cost. Also you should probably have tests for generated code or generate it locally before merging because you can _never_ know if what is submitted by a user is actually the result of the modified code.

danieldietrich:
I understand that you do not want to create builds that look dirty in your eyes, that is perfectly ok. I would do the same.

Vavr's build script is just 140 LOC without whitespace and comments and even only 103 LOC without lines that only contain closing braces '}'. That's nothing. To me, it is overkill to explode this simple script to 7 files and multiple directories:

build-logic

```
# before
8.0K	build.gradle

# after
4.0K	build.gradle
 28K	build-logic
```

Personally, I use (and love) Gradle because I am able to write my build in one (small) script. I want to keep it that way in such a small project. The build is blazing fast. Even code generation takes nearly no time on my machine. The most time consuming part are the tests (which can't be made faster by the build).

> ... because you can never know if what is submitted by a user is actually the result of the modified code.

Sure, we can know that! Our CI build generates the sources and overwrites the generated files that where checked in. After that, the CI build checks the `git status`. It returns 0, if no files were changed and 1, if generated files changed. The latter is a clear sign that the user manually changed generated files. In 7 years of Vavr, no user ever changed the output directories (which are listed in a central place).

---

If you want to get some feedback by me, I think you are a bit too dogmatic in what you do. Sometimes it is good to be pragmatic and choose a compromise. I tried to do that by [outlining](https://github.com/vavr-io/vavr/pull/2644#pullrequestreview-666249615) a build strategy for Vavr. I really appreciate your help & your work. I think I am able to improve the build based on your changes.

Thank you Cédric

melix:
I don't think that comparing the number of lines of code is fair here, for multiple reasons:

1. the new build supports caching, including of downloaded toolchains, making it much faster. I get that you don't think it's important and definitely disagree here: it's both faster to execute and configure, making it a pleasure to work with the codebase
2. it adds an extension to make the declaration of ammonite configuration declarative, which wasn't the case before. You could get rid of this easily, even if I think it's not a good idea :)
3. goal shouldn't be to have the minimal number of lines of code, but to be maintainable and readable
4. changing _configuration_ wouldn't invalidate the build with those changes, while any change to any configuration would invalidate the whole build in your setup
5. I'd be glad to show the `vavr` build as an example of what an idiomatic build _should look like_ if i was implemented as in this PR, but not if everything is put in a `build.gradle` file

So I don't think I'm dogmatic here: the changes have been made for a very good reason. Not caring about performance isn't, IMO, and I see dogmatism in another direction ;)

danieldietrich:
> the new build supports caching, including of downloaded toolchains, making it much faster. 

That's _effectively_ not the case for the following reasons:

On the CI server, the whole toolchain needs to be built from a clean state. Every time. Because of the additional Gradle plugins, it got slower.

> goal shouldn't be to have the minimal number of lines of code, but to be maintainable and readable

That right and that's exactly the reason why I stick to one build.gradle file in this case. It is so easy at the moment that the mental load of viewing one small file is less effort than switching between different files. For bigger projects, you are absolutely right but Vavr is super simple.

> changing configuration wouldn't invalidate the build with those changes, while any change to any configuration would invalidate the whole build in your setup

Changing configuration isn't a typical use case. 99% of the time, we change code.

---

I will apply some of your goodies to the build now. I really appreciate your suggestions.
I think this PR can be closed now, I want to take the build in a different direction - best of both worlds: apply improvements where possible, without exploding the structure.

Thx!

melix:
> On the CI server, the whole toolchain needs to be built from a clean state. Every time. Because of the additional Gradle plugins, it got slower.

Right. However it's much faster locally. Note that if you use GH actions, you can use https://github.com/eskatos/gradle-command-action to restore the cache and avoid rebuilding.

danieldietrich:
Just for the protocol: let me thank you, I have learned much about best practices from you PR! I'm sure I will apply them in other projects.

melix:
My pleasure :)

build-logic/src/main/groovy/io.vavr.library.gradle
```diff
@@ -0,0 +1,43 @@
...
+ * additional artifacts like test sources jar, automatic module name...
+ */
+plugins {
+    id 'java-library'
+    id 'jacoco'
+}
+
+java {
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(8))
```
danieldietrich, line 12: @melix One question: it is common practice to compile & test a project with different JDKs (-> travis build matrix) in order to see if the code behaves as expected with different Java versions.

Currently, each build of our Travis CI matrix falls back to the alternate Java 8 JDK:

```bash
Compiling with toolchain '/usr/lib/jvm/java-8-openjdk-amd64'
```

Do you have any suggestion what's the best way to circumvent this?

I have in mind: using an optional env variable that defaults to `8`.

danieldietrich: I have this in mind:

```
def val = findProperty('java-language-version') ?: '8'
```

danieldietrich: Ah, I've found the Gradle toolchain docs --> https://docs.gradle.org/current/userguide/toolchains.html

melix: There's no built-in support in Gradle for test matrix but it's a common practice. It's usually a waste of time and resources, though, to rebuild the project with a different JDK just to test on a different one. So the simplest solution is what you've described, which can be implemented as this:

```
    toolchain {
        languageVersion.set(JavaLanguageVersion.of(
            providers.environmentVariable("JAVA_VERSION")
                .forUseAtConfigurationTime()
                .map { it.toInt() } // Kotlin!!
                .getOrElse(8)
        ))
    }
```

Alternatively, you can configure different test tasks to use different JDKs, which is what I would personally prefer, e.g (sorry it's Kotlin DSL because I just copied it from a project of mine):

```
tasks.register<Test>("java11Test") {
    group = "verification"
    val test = tasks.test.get()
    classpath = test.classpath
    testClassesDirs = test.testClassesDirs
    javaLauncher.set(javaToolchains.launcherFor {
        languageVersion.set(JavaLanguageVersion.of(11))
    })
}
```

danieldietrich: Nice one, thanks!

PR #2648, opened 2021-03-28
Fix Array#update(int, T) complexity

anton0xf:
io.vavr.collection.Array#update(int, T) [contains](https://github.com/vavr-io/vavr/blob/2fe989583fee00842e3fc4a5f806b7f2d99a6eba/src/main/java/io/vavr/collection/Array.java#L1507) java.util.Arrays#copyOf(T[], int) call. So it can't be of constant complexity.


danieldietrich (approved):
Many thanks, this looks good.

PR #2650, opened 2021-04-07
Add convenience method for Value#toMap(keyMapper, v -> v)

MrcJkb:


src/main/java/io/vavr/Value.java
```diff
@@ -1053,6 +1053,18 @@ default CharSeq toCharSeq() {
...
+    /**
+     * Converts this to a {@link Map}, with the elements as values.
+     *
+     * @param keyMapper   A function that maps an element to a key
+     * @param <K>         The key type
+     * @return A new {@link HashMap}.
+     */
+    default <K> Map<K, T> toMapBy(Function<? super T, ? extends K> keyMapper) {
+        Objects.requireNonNull(keyMapper, "keyMapper is null");
+        return toMap(keyMapper, v -> v);
```
frecco75, line 1065: identity() instead of v -> v

frecco75 (approved):

danieldietrich:
Hi & thank you for your suggestion. We would also need additional methods for `toSortedMap`, `toLinkedMap` and maybe for `toJavaMap`. I think the cost/income ratio isn't worth it to add such an amount of additional methods.

danieldietrich (commented):
Hi & thank you for your suggestion. We would also need additional methods for toSortedMap, toLinkedMap and maybe for toJavaMap. I think the cost/income ratio isn't worth it to add such an amount of additional methods.

danieldietrich:
I will close this as discussed.

PR #2651, opened 2021-04-10
Option.none(Class) with type

MrcJkb:
The reason for this addition is because there is no convenient way of generating Lists (or infinite Streams) of Option.none() with a given type.

The current workaround:

```
String nullString = null;
// Calling Option.none() without arguments won"t compile
Stream<Option<String>>  optionStringStream = Stream.iterate(() -> Option.none(nullString));
```

Can now be simplified to:

```
Stream<Option<String>>  optionStringStream = Stream.iterate(() -> Option.none(String.class));
```

### DIFF
--- a/src/main/java/io/vavr/control/Option.java
+++ b/src/main/java/io/vavr/control/Option.java
@@ -176,6 +176,25 @@ public abstract class Option<T> implements Iterable<T>, io.vavr.Value<T>, Serial
     }
 
     /**
+     * Returns the single instance of {@code None}
+     *
+     * <pre>{@code
+     * // = None
+     * Option<String> none = Option.none();
+     * }</pre>
+     *
+     * @param <T> component type
+     * @param type the class of the component type
+     * @return the single instance of {@code None}
+     */
+    @SuppressWarnings("unused")
+    public static <T> Option<T> none(Class<T> type) {
+        @SuppressWarnings("unchecked")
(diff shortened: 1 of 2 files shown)


mincong-h:
What about:

```java
Stream<Option<String>> optionStringStream = Stream.iterate(() -> Option.<Option<String>>none());
// jshell: optionStringStream ==> Stream()
```

It seems working fine for me.

MrcJkb:
> Hi @MrcJkb , what about:
> 
> ```java
> Stream<Option<String>> optionStringStream = Stream.iterate(() -> Option.<Option<String>>none());
> // jshell: optionStringStream ==> Stream()
> ```
> 
> It seems working fine for me.

Oh, niche, I didn't know you could do that. Thanks!
P.S., that's one option too many in the type parameter.

mincong-h:
Yes, I agree that it's very verbose... If you are using Java 10+, you can use `var` (introduced by [JEP 286](https://openjdk.java.net/jeps/286)):

```java
var optionStringStream = Stream.iterate(() -> Option.<Option<String>>none());
```

Otherwise, you may consider:

```java
Stream<Option<String>> optStream = Stream.of(Option.<String>none());
```

... but this is very different from what you wanted at the beginning, because this stream is not generated lazily, but eagerly. And it does not contain infinite number of elements :(

PR #2659, opened 2021-05-16
Vector.of(T element) performance improvement

sleepytomcat:
As discussed here https://github.com/vavr-io/vavr/issues/2656#issuecomment-841811218


danieldietrich (approved):
Thank you, that was fast! It looks good to me.

PR #2661, opened 2021-05-18
Minor wording change in error message for Try().peek(Consumer, Consumer)

sleepytomcat:


danieldietrich (approved):
Good catch, thank you!

PR #2662, opened 2021-05-21
Optimized Traversable<>.unzip(Function, Function)

sleepytomcat:
New unzip() method accepts two functions to avoid extensive Tuple creation under the hood.

### DIFF
--- a/src/main/java/io/vavr/collection/Iterator.java
+++ b/src/main/java/io/vavr/collection/Iterator.java
@@ -1201,18 +1201,6 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
 
     @Override
     default <T1, T2> Tuple2<Iterator<T1>, Iterator<T2>> unzip(
-            Function<? super T, Tuple2<? extends T1, ? extends T2>> unzipper) {
-        Objects.requireNonNull(unzipper, "unzipper is null");
-        if (!hasNext()) {
-            return Tuple.of(empty(), empty());
-        } else {
-            final Stream<Tuple2<? extends T1, ? extends T2>> source = Stream.ofAll(() -> map(unzipper));
-            return Tuple.of(source.map(t -> (T1) t._1).iterator(), source.map(t -> (T2) t._2).iterator());
-        }
-    }
-
-    @Override
-    default <T1, T2> Tuple2<Iterator<T1>, Iterator<T2>> unzip(
             Function<? super T, ? extends T1> unzipper1, Function<? super T, ? extends T2> unzipper2) {
         Objects.requireNonNull(unzipper1, "unzipper1 is null");
         Objects.requireNonNull(unzipper2, "unzipper2 is null");
@@ -1223,18 +1211,6 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
             final Iterator<T1> iter1 = stream.iterator().map(unzipper1);
             final Iterator<T2> iter2 = stream.iterator().map(unzipper2);
             return Tuple.of(iter1, iter2);
-        }
-    }
-
-    @Override
-    default <T1, T2, T3> Tuple3<Iterator<T1>, Iterator<T2>, Iterator<T3>> unzip3(
-            Function<? super T, Tuple3<? extends T1, ? extends T2, ? extends T3>> unzipper) {
-        Objects.requireNonNull(unzipper, "unzipper is null");
-        if (!hasNext()) {
-            return Tuple.of(empty(), empty(), empty());
-        } else {
-            final Stream<Tuple3<? extends T1, ? extends T2, ? extends T3>> source = Stream.ofAll(map(unzipper));
-            return Tuple.of(source.map(t -> (T1) t._1).iterator(), source.map(t -> (T2) t._2).iterator(), source.map(t -> (T3) t._3).iterator());
         }
     }
 
--- a/src/main/java/io/vavr/collection/Iterator.java
+++ b/src/main/java/io/vavr/collection/Iterator.java
@@ -1219,9 +1219,11 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
         if (!hasNext()) {
             return Tuple.of(empty(), empty());
         } else {
-            final Iterator<T1> iter1 = map(unzipper1);
-            final Iterator<T2> iter2 = map(unzipper2);
-            return Tuple.of(iter1, iter2);        }
+            final Stream<T> stream = Stream.ofAll(() -> this);
+            final Iterator<T1> iter1 = stream.iterator().map(unzipper1);
+            final Iterator<T2> iter2 = stream.iterator().map(unzipper2);
+            return Tuple.of(iter1, iter2);
+        }
     }
 
     @Override
--- a/src/main/java/io/vavr/collection/Iterator.java
+++ b/src/main/java/io/vavr/collection/Iterator.java
@@ -1238,6 +1238,25 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
         }
     }
 
+    @Override
+    default <T1, T2, T3> Tuple3<Iterator<T1>, Iterator<T2>, Iterator<T3>> unzip3(
+            Function<? super T, ? extends T1> unzipper1,
+            Function<? super T, ? extends T2> unzipper2,
+            Function<? super T, ? extends T3> unzipper3) {
+        Objects.requireNonNull(unzipper1, "unzipper1 is null");
+        Objects.requireNonNull(unzipper2, "unzipper2 is null");
+        Objects.requireNonNull(unzipper3, "unzipper3 is null");
+        if (!hasNext()) {
+            return Tuple.of(empty(), empty(), empty());
+        } else {
+            final Stream<T> stream = Stream.ofAll(() -> this);
+            final Iterator<T1> iter1 = stream.iterator().map(unzipper1);
+            final Iterator<T2> iter2 = stream.iterator().map(unzipper2);
+            final Iterator<T3> iter3 = stream.iterator().map(unzipper3);
+            return Tuple.of(iter1, iter2, iter3);
+        }
+    }
+
     /**
      * Creates an iterator from a seed value and a function.
      * The function takes the seed at first.
--- a/src/main/java/io/vavr/collection/Iterator.java
+++ b/src/main/java/io/vavr/collection/Iterator.java
@@ -1212,6 +1212,20 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
     }
 
     @Override
+    default <T1, T2> Tuple2<Iterator<T1>, Iterator<T2>> unzip(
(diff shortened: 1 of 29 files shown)


sleepytomcat:
This is a PR to address https://github.com/vavr-io/vavr/issues/2503

src/main/java/io/vavr/collection/Iterator.java
```diff
@@ -1211,6 +1211,20 @@ public U next() {
...
+    @Override
+    default <T1, T2> Tuple2<Iterator<T1>, Iterator<T2>> unzip(
+            Function<? super T, ? extends T1> unzipper1, Function<? super T, ? extends T2> unzipper2) {
+        // return unzip(i -> Tuple.of(unzipper1.apply(i), unzipper2.apply(i)));
+        Objects.requireNonNull(unzipper1, "unzipper1 is null");
+        Objects.requireNonNull(unzipper2, "unzipper2 is null");
+        if (!hasNext()) {
+            return Tuple.of(empty(), empty());
+        } else {
+            final Stream<Tuple2<? extends T1, ? extends T2>> source = Stream.ofAll(() -> map(i -> Tuple.of(unzipper1.apply(i), unzipper2.apply(i))));
```
sleepytomcat, line 1222: This one looks particularly questionable, as we do create a Stream of Tuples here, and then split them into Tuple of Iterators. However I could not find a better solution yet, as Iterator has mutable state, and generally traversable only once.

sleepytomcat: Are there better ways to do this?

danieldietrich: Right, I would do it as follows:

```java
// ...
final Iterator<T1> iter1 = map(unzipper1);
final Iterator<T2> iter2 = map(unzipper2);          
return Tuple.of(iter1, iter2);
```

danieldietrich (changes requested):
Hi Sergei,

many thanks for your PR!
I've thought a bit about the unzip method. I think we should provide only one implementation in `Traversable`:

```java
default <T1, T2> Tuple2<Iterator<T1>, Iterator<T2>> unzip(...) {
    Objects.requireNonNull(unzipper1, "unzipper1 is null");
    Objects.requireNonNull(unzipper2, "unzipper2 is null");
    final Iterator<T1> iter1 = iterator().map(unzipper1);
    final Iterator<T2> iter2 = iterator().map(unzipper2);
    return Tuple.of(iter1, iter2);
}
```

And we will have the additional implementation in `Iterator`, like described below.

The reasons for the central implementation are:

1. There emerge problems in sorted collections because we do not know, if the target types T1, T2 are comparable. That might lead to ClassCastExceptions at runtime.
2. I do not want to be too restrictive. The users should decide on their own which collection they target. The Tuple2 of Iterators allows them to choose their collection later. They even could proceed with the fluent API of our rich Iterator.

I know that it was hard work to provide all the implementations. This is really a drawback for maintainers of the collection hierarchy. If you don't have any objections regarding my suggestion, it would be great if you could change your PR accordingly.

Thanks
Daniel

sleepytomcat:
Thank you @danieldietrich for detailed explanation! Let me follow-up with the code change.

sleepytomcat:
I've changed the return type of `Traversable<>.unzip(Function, Function)` to `Tuple2<Iterator<T1>, Iterator<T2>`, and removed all the unnecessary container-specific code. Once the code looks good, I'll take another step to remove previous implementation of `unzip(Function<T, Tuple2<T1,T2>>)`.

sleepytomcat:
@danieldietrich ,
```
// ...
final Iterator<T1> iter1 = map(unzipper1);
final Iterator<T2> iter2 = map(unzipper2);          
return Tuple.of(iter1, iter2);
```

—  I do not think `Iterator.unzip(Function, Function)` implementation would work this way. My reasoning: iterator is mutable, so traversing over, say, first Iterator `_1` in the resulting `Tuple2` will affect traversing of the iterator in `_2`.

Can you please share your perspective on this?

sleepytomcat:
There's what I observe under debugger with said `Iterator.unzip(Function, Function)` implementation:

image

Please note empty `a_2` - my explanation this is because `a_1` iterator was already traversed, and there's no more data yielded by `a_2`.

Do you think this explanation may seem reasonable?

danieldietrich:
> I do not think Iterator.unzip(Function, Function) implementation would work this way

You are absolutely right - it was a fast shot without really thinking about it. If we can't re-use an existing iterator (one of those internals in the Iterator interface), we have to write a new one that is able to feed two iterators. (Something like `tee` in bash).

Let be think about it, I will come back with a suggestion.

danieldietrich:
Sadly, there is no chance to do it without allocating memory. This is the easiest way to do it (I did not test the code in an editor):

```java
    @Override
    default <T1, T2> Tuple2<Iterator<T1>, Iterator<T2>> unzip(
            Function<? super T, ? extends T1> unzipper1, Function<? super T, ? extends T2> unzipper2) {
        Objects.requireNonNull(unzipper1, "unzipper1 is null");
        Objects.requireNonNull(unzipper2, "unzipper2 is null");
        if (!hasNext()) {
            return Tuple.of(empty(), empty());
        } else {
            final Stream<T> stream = Stream.ofAll(() -> this);
            final Iterator<? extends T1> iter1 = stream.iterator().map(unzipper1);
            final Iterator<? extends T2> iter2 = stream.iterator().map(unzipper2);
            return Tuple.of(iter1, iter2);
        }
    }
```

Could you do this for Iterator.unzip und Iterator.unzip3, please?

sleepytomcat:
Will do, thank you @danieldietrich

sleepytomcat:
@danieldietrich , side question: from the `Iterator` class description, and from code implementation, it's not clear to me — generally speaking, is there a legitimate way to create a copy of an iterator instance? Or, they are inherently non-copyable? (I think copying an iterator instance could help us avoid memory allocation while traversing Tuple's _1, _2).

sleepytomcat:
@danieldietrich , I've updated the code of `Tuple2<Iterator<T1>, Iterator<T2>> unzip(Function, Function)` according to your recommendations. `gradlew check` is successful now. Please kindly review and let know if it looks Ok. Once we're there, I will..
1) ...update `unzip3(Function<T, Tuple3>`) to `unzip3(Function<T, T1>, Function<T, T2>, Function<T, T2>)` — as you suggested;
2) ...add necessary tests coverage;
3) ...remove old `unzip` method using `Function<T, Tuple>` arguments, along with their tests.

Please let know if it looks Ok to you.

danieldietrich:
@sleepytomcat

> side question (...) is there a legitimate way to create a copy of an iterator instance?

Currently there exists not way because Iterator is stateful (other than Vavr collections, which are persistent). However, the idea is tempting. I could imagine that an Iterator could have a copy() aka clone() aka duplicate() method that creates a new Iterator instance by copying the state. This might involve recursively copying wrapped Iterator instances. I will think a bit about it, but it is out of scope for this PR.

danieldietrich:
@sleepytomcat Thank you, please go on with 1)-3) like you suggested!
Great that you updated the links to the methods in the javadoc at the top of the file! 👍🏼

src/main/java/io/vavr/collection/Traversable.java
```diff
@@ -1602,31 +1602,48 @@ default Number sum() {
...
+     * @param unzipper1 a function which converts elements of this to first set
+     * @param unzipper2 a function which converts elements of this to second set
+     * @param unzipper3 a function which converts elements of this to third set
+     * @param <T1>      element type returned by unzipper1
+     * @param <T2>      element type returned by unzipper2
+     * @param <T3>      element type returned by unzipper3
+     * @return A triplet of {@code Iterator} to iterate over elements of sets split by {@code unzipper1}, {@code unzipper2}, {@code unzipper3}
+     * @throws NullPointerException if any of {@code unzipper1}, {@code unzipper2}, {@code unzipper3} is null
+     */
+    default <T1, T2, T3> Tuple3<? extends Iterator<T1>, ? extends Iterator<T2>, ? extends Iterator<T3>> unzip3(
```
sleepytomcat, line 1635: @danieldietrich , should this be 
`... Tuple3<? extends Iterator<T1>, ? extends Iterator<T2>, ? extends Iterator<T3>> unzip3(...) `
or
`... Tuple3<Iterator<T1>, Iterator<T2>, Iterator<T3>> unzip3(...)`?

Same question goes to `unzip(...)` above, line 1614.

sleepytomcat:
@danieldietrich , the items (1)-(3) completed, please review the code and share your feedback. Thank you!

sleepytomcat:
I think I messed up with pulling upstream changes. Fixing.

src/main/java/io/vavr/collection/Traversable.java
danieldietrich: Please do it like this (for return types in general):

```java
Tuple3<Iterator<T1>, Iterator<T2>, Iterator<T3>>
```

Thank you!

danieldietrich (changes requested):
Awesome, thank you!
I like it to remove all the code while adding more flexibility at the same time.
Please adjust the return type(s) like commented, then we are able to merge this beauty!

sleepytomcat:
Return type changed to `Tuple3<Iterator<T1>, Iterator<T2>, Iterator<T3>>` as suggested. Thank you!

src/test/java/io/vavr/collection/AbstractTraversableTest.java
```diff
@@ -2411,26 +2411,29 @@ public void shouldReturnSomeTailWhenCallingTailOptionOnNonNil() {
...
     @Test
     public void shouldUnzip3NonNil() {
-        final Tuple actual = of(0, 1).unzip3(i -> Tuple.of(i, (char) ((short) 'a' + i), (char) ((short) 'a' + i + 1)));
-        final Tuple expected = Tuple.of(of(0, 1), of('a', 'b'), of('b', 'c'));
-        assertThat(actual).isEqualTo(expected);
+        final Tuple3<Iterator<Integer>, Iterator<Character>, Iterator<Character>> actual
+                = of(0, 1).unzip3(i -> i, i -> (char) ((short) 'a' + i), i -> (char) ((short) 'a' + i + 1));
+        final Tuple3<Iterator<Integer>, Iterator<Character>, Iterator<Character>> expected
+                = Tuple.of(of(0, 1).iterator(), of('a', 'b').iterator(), of('b', 'c').iterator());
+        assertThat(actual.map(Value<Integer>::toArray, Value<Character>::toArray, Value<Character>::toArray)).isEqualTo(expected.map(Value<Integer>::toArray, Value<Character>::toArray, Value<Character>::toArray));
```
sleepytomcat, line 2436: @danieldietrich , I keep using `.toArray()` to 'materialize' data and compare `Iterators` in unit tests. Is there a better way to compare two iterators?

danieldietrich: I think this is ok!

danieldietrich (approved):
Great, many thanks for this!

PR #2664, opened 2021-05-23
remove sneakyThrow

googl3r:
It's my first ever contributions in an open source project, I use vavr a lot and a appreciate the author view in java.
if their is some errors I can update.

### DIFF
--- a/src/main/java/io/vavr/control/Try.java
+++ b/src/main/java/io/vavr/control/Try.java
@@ -1303,7 +1303,7 @@ public abstract class Try<T> implements Iterable<T>, io.vavr.Value<T>, Serializa
 
         @Override
         public T get() {
-            return sneakyThrow(cause);
+            throw new RuntimeException(cause);
         }
 
         @Override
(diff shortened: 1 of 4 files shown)


danieldietrich:
Hi @googl3r, thanks a lot, I really appreciate it!

I don't think it is that simple, and I won't recommend to take this issue as a "first time committer" issue.
Sneaky throw was introduced with Vavr 0.9.0. [Here is the PR that added it](https://github.com/vavr-io/vavr/pull/1976).
Especially, we removed the Fatal and NonFatal exception wrappers in [Try](https://github.com/vavr-io/vavr/pull/1976/files#diff-95fadb44425ac06924a37066a537a4be05819dace1207e155cbe26ec47144813).

There have been additional additions regarding sneaky throw, like [this commit](https://github.com/vavr-io/vavr/pull/1991).

Please let me perform this change. It is more cross-cutting throughout the lib as it might seem. Also some methods need to explicitly declare which runtime exceptions are thrown, e.g. NoSuchElementException for Option.get() and NonFatalException for Try.get() etc.

Maybe we will find an issue for you that is better suited. 😅 From time to time I label issues with 'help wanted'. Look out for these.

Thank you!
Daniel

googl3r:
Hi @danieldietrich thank you for replying.

I will do as you said, search for simpler issue to start with. thanks a lot.

danieldietrich:
Thank you!

PR #2685, opened 2021-09-26
solution for problem 44

smthelusive:
solution for euler problem 44: 
https://projecteuler.net/problem=44


src/test/java/io/vavr/collection/euler/Utils.java
```diff
@@ -115,4 +115,15 @@ static boolean isPalindrome(String val) {
     static boolean isPalindrome(int val) {
         return isPalindrome(Long.toString(val));
     }
+
+    static Stream<Long> pentagonal() {
```
cal101, line 119: The t._1 is only based on t._2.
_2 is 1 2 3 4
So they can be split into Stream.from and map.

src/test/java/io/vavr/collection/euler/Euler44Test.java
```diff
@@ -0,0 +1,65 @@
...
+     * problem 44</a>.
+     */
+    @Test
+    public void shouldSolveProblem44() {
+        final Future<Long> computation = API.Future(this::minimalPentagonalDiff)
+                .await(2, TimeUnit.SECONDS);
+        assertThat(computation.getOrElse(-1L)).isEqualTo(5482660);
+    }
+
+    private Long minimalPentagonalDiff() {
```
cal101, line 58: I like the idea of looping over the sum.
Just a few remarks.
You are generating two pentagonal streams. From what i understand vavr streams  are just lazy but can be
Iterated over multiple times.
So we should be able to define a variable pentagonals in method and use that twice. It should be possible to do it in a single expression, not sure how.
Replacing the compareTo with a simple "x < y" would make it more readable IMO.

src/test/java/io/vavr/collection/euler/Euler44Test.java
```diff
@@ -0,0 +1,65 @@
...
+        return Utils.pentagonal().flatMap(sumPentagonal ->
+                Utils.pentagonal().takeWhile(pentagonal -> sumPentagonal.compareTo(pentagonal) > 0)
```
danieldietrich, line 60: Yes, @cal101 is right, `sumPentagonal.compareTo(pentagonal) > 0` can be simplified by `sumPentagonal > pentagonal`

src/test/java/io/vavr/collection/euler/Utils.java
```diff
@@ -115,4 +115,15 @@ static boolean isPalindrome(String val) {
...
+        return Stream.of(Tuple.of(1L, 1)).appendSelf(self ->
+                self.map(t ->
+                        Tuple.of((t._2 + 1) * (3L * (t._2 + 1) - 1) / 2, t._2 + 1)))
+                .map(t -> t._1);
```
danieldietrich, line 123: @cal101 could you provide the implementation you suggested? That way @smthelusive can verify if the test is still green...

danieldietrich (changes requested):
Hi @smthelusive 👋,
thank you, great solution!
Could you please take a look at the review comments?
Thank you :)

danieldietrich:
Thank you Nataliia! I will merge it now to go forward.
@cal101 we can do any improvements in a subsequent PR...

PR #2689, opened 2021-11-02
Performance improvement for List::unfold, List::unfoldLeft

sleepytomcat:
Performance improvement proposed by @tvelka here https://github.com/vavr-io/vavr/issues/2688


danieldietrich (approved):
Nice one, thank you ☺️

Issue #2690, opened 2021-11-12
Release a new alpha build.


NicklasWallgren:
The release 1.0.0-alpha-4 is currently blocked by a change made in #2665.

When I run `./gradlew release --info`, I get the following error message:

```
> Task :vavr:vavr:initializeSonatypeStagingRepository FAILED
Caching disabled for task ':vavr:vavr:initializeSonatypeStagingRepository' because:
  Build cache is disabled
Task ':vavr:vavr:initializeSonatypeStagingRepository' is not up-to-date because:
  Task has not declared any outputs despite executing actions.
No stagingProfileId set, querying for packageGroup 'io.vavr'
:vavr:vavr:initializeSonatypeStagingRepository (Thread[Execution worker for ':vavr:vavr' Thread 11,5,main]) completed. Took 0.661 secs.

> Task :vavr:runBuildTasks FAILED
:vavr:runBuildTasks (Thread[Execution worker for ':vavr',5,main]) completed. Took 4.099 secs.

> Task :release FAILED
Release process failed, reverting back any changes made by Release Plugin.
Running [git, checkout, gradle.properties] in [/Users/cafebab3/git/vavr-io/vavr]
Running [git, checkout, gradle.properties] produced output: []
:release (Thread[Execution worker for ':',5,main]) completed. Took 40.157 secs.

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':vavr:vavr:initializeSonatypeStagingRepository'.
> Failed to load staging profiles, server at https://oss.sonatype.org/service/local/ responded with status code 401

* Try:
Run with --stacktrace option to get the stack trace. Run with --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 41s
1 actionable task: 1 executed
```

danieldietrich:
Ok, we are one step further, the oss sonatype credentials are now found by the build after we merged #2691.

Now there are problems signing the release...

```
> Task :vavr:vavr:signMavenPublication FAILED
Caching disabled for task ':vavr:vavr:signMavenPublication' because:
  Build cache is disabled
Task ':vavr:vavr:signMavenPublication' is not up-to-date because:
  Task has failed previously.
:vavr:vavr:signMavenPublication (Thread[Execution worker for ':vavr:vavr',5,main]) completed. Took 0.004 secs.

> Task :vavr:runBuildTasks FAILED
[...]

PR #2691, opened 2021-11-17
#2690 sontatype 401 fix

jarekratajski:
very trivial fix - actually in gradle 7.0 username and password from USER gradle.properties were not correctly read.
I have no idea why.

This small change seems to work - at least solves the error visible in #2690 

Edited:
Just small note - if you have 
`systemProp.org.gradle.project.ossrhUsername=xxx `
in your `HOME/.gradle/gradle.properties`
it will interfere with line
`ossrhUsername=xyz` 

That confused me - I had the first line (used in other project).
Probably the second form (as documented in `vavr` build file ) is a proper one - I am not a gradle specialist.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -182,8 +182,8 @@ signing {
 nexusPublishing {
     repositories {
         sonatype {
-            username.set(providers.systemProperty("ossrhUsername").forUseAtConfigurationTime())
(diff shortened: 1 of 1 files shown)


danieldietrich (approved):
Many thanks Jarek, I am not sure how much time I would have spent to find that error. 🙏
The next alpha release is near :)

PR #2695, opened 2021-12-03
Add Validation.getOrElseThrow(Function<E, X>)

sir4ur0n:
Closes https://github.com/vavr-io/vavr/issues/2693

### DIFF
--- a/src/test/java/io/vavr/control/ValidationTest.java
+++ b/src/test/java/io/vavr/control/ValidationTest.java
@@ -300,6 +300,30 @@ public class ValidationTest extends AbstractValueTest {
         assertThat(result).isEqualTo(3);
     }
 
+
+    // getOrElseThrow(Function)
+
+    @Test
+    public void shouldReturnValidWhenGetOrElseThrowWithFunctionOnValid() {
+      final Integer actual = Validation.<String, Integer> valid(1).getOrElseThrow(s -> new RuntimeException(s));
+      assertThat(actual).isEqualTo(1);
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void shouldThrowWhenGetOrElseThrowWithFunctionOnInvalid() {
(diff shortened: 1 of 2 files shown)


src/test/java/io/vavr/control/ValidationTest.java
```diff
@@ -300,6 +300,20 @@ public void shouldConvertFailureToU() {
...
+    public void shouldReturnValidWhenGetOrElseThrowWithFunctionOnValid() {
+      final Integer actual = Validation.<String, Integer> valid(1).getOrElseThrow(s -> new RuntimeException(s));
+      assertThat(actual).isEqualTo(1);
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void shouldThrowWhenGetOrElseThrowWithFunctionOnInvalid() {
+      Validation.<String, Integer> invalid("some error").getOrElseThrow(s -> new RuntimeException(s));
+    }
+
```
sleepytomcat, line 316: We may also consider to add a test for an edge case, where `null` provided instead of `exceptionFunction`.

sleepytomcat: Something like 
```
@Test(expected = NullPointerException.class)
public void shouldThrowNullPointerExceptionWhenExceptionFunctionIsNullOnValid() {
    Validation.<String, Integer> valid(1).getOrElseThrow(null);
}

@Test(expected = NullPointerException.class)
public void shouldThrowNullPointerExceptionWhenExceptionFunctionIsNullOnInvalid() {
    Validation.<String, Integer> invalid("some error").getOrElseThrow(null);
}
```

sir4ur0n: Done (I adapted the test names a bit, and added types to make it compile)

sleepytomcat: Thank you @sir4ur0n !
I believe we need @danieldietrich review and apporoval.

danieldietrich (approved):
Many thanks, well done!

PR #2701, opened 2021-12-10
Fixing Methodref contant errors by changing Option back to an interface

nexx512:
I encountered many `Method 'io.vavr.control.Option io.vavr.control.Option.none()' must be Methodref constant` errors when trying to use the current alpha version of vavr with a current version of spring boot.
I'm not quite sure what this error means but changing it the abstract class back to an interface seemed to solve this problem.
I guess there is a reason why you changed it to an abstract class. If there is a better way to solve this I'm interested in how this can be done.

### DIFF
--- a/src/main/java/io/vavr/control/Option.java
+++ b/src/main/java/io/vavr/control/Option.java
@@ -46,12 +46,9 @@ import java.util.function.Supplier;
  * @param <T> The type of the optional value.
  */
 @SuppressWarnings("deprecation")
-public abstract class Option<T> implements Iterable<T>, io.vavr.Value<T>, Serializable {
+public interface Option<T> extends Iterable<T>, io.vavr.Value<T>, Serializable {
 
-    private static final long serialVersionUID = 1L;
-
-    // sealed
-    private Option() {}
+    long serialVersionUID = 1L;
 
     /**
      * Creates a new {@code Option} of a given value.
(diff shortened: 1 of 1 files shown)


danieldietrich (commented):
Hi, thanks for the hint!

This is because Spring internally uses a different version of Vavr. When running two incompatible versions in the same runtime, we might receive a `IncompatibleClassChangeError` error.

I want to baseline the final Vavr 1.0 version to Java 16, [like Spring does for v6](https://spring.io/blog/2021/09/02/a-java-17-and-jakarta-ee-9-baseline-for-spring-framework-6). The solution is to update the Vavr dependency in Spring. We will have several backward-incompatible changes in Vavr.

This is why it does not make sense to merge this PR (beside that, several other classes are abstract classes now...).

You need to stay on Vavr 0.10.x for now ... that's the only solution until Vavr 1.0 is final and the Spring dependency was updated.

Thx

danieldietrich:
Many thanks for trying to solve this issue. I will close this PR because we need another solution like described above.

nexx512:
Oh, I see. That makes sense. Thanks for the explanation.

yanye666:
> Hi, thanks for the hint!
> 
> This is because Spring internally uses a different version of Vavr. When running two incompatible versions in the same runtime, we might receive a `IncompatibleClassChangeError` error.
> 
> I want to baseline the final Vavr 1.0 version to Java 16, [like Spring does for v6](https://spring.io/blog/2021/09/02/a-java-17-and-jakarta-ee-9-baseline-for-spring-framework-6). The solution is to update the Vavr dependency in Spring. We will have several backward-incompatible changes in Vavr.
> 
> This is why it does not make sense to merge this PR (beside that, several other classes are abstract classes now...).
> 
> You need to stay on Vavr 0.10.x for now ... that's the only solution until Vavr 1.0 is final and the Spring dependency was updated.
> 
> Thx

I would like to ask, why can not start in jdk11（zull，11.0.11）, jdk8（1.8.0_345) can start?

PR #2703, opened 2022-01-19
Add method in validation on success

adityamandrekar:
Added onSuccess method in validation.

Purpose is to continue to next validation if current validation is successful. If a validation fails then skip the subsequent validations in the chain

ex:

 return rangeValidation(range)
                .onSuccess(() -> dateValidation(date))
                .onSuccess(() -> emailValidation(email));

### DIFF
--- a/src/main/java/io/vavr/control/Validation.java
+++ b/src/main/java/io/vavr/control/Validation.java
@@ -558,6 +558,24 @@ public abstract class Validation<E, T> implements Iterable<T>, Value<T>, Seriali
         return isValid() ? this : (Validation<E, T>) supplier.get();
     }
 
+     /**
+     * Returns result of evaluating supplier if it is valid, otherwise return this.
+     *
+     * <pre>{@code
+     * // following code return an alternative Supplier validation("vavr") when there is an error in first validation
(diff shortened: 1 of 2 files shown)


danieldietrich:
The `onSuccess` and `onFailure` methods of other Vavr types, like those of `Try`, have a different meaning. We use them to perform side-effects and finally return the same instance.

You suggest to do something different. You search a way to transform the instance, if a specific condition holds, like the _valid_ resp. _invalid_ case.

For that purpose, there already exist already a couple of methods:

* [`orElse(Supplier)`](https://javadoc.io/static/io.vavr/vavr/0.10.4/io/vavr/control/Validation.html#orElse-java.util.function.Supplier-) covers the _invalid_ case
* [`flatMap(Function)`](https://javadoc.io/static/io.vavr/vavr/0.10.4/io/vavr/control/Validation.html#flatMap-java.util.function.Function-) covers the _valid_ case. It is a bit more general than your `onSuccess` method, so it covers more uses cases

I will close this PR because my strategy is to reduce the API surface area. We shall not introducing another method for the same purpose.

Nevertheless, thank you for your suggestion and your PR! 🙏
And also thank you for kindly waiting so long for my reply.

PR #2719, opened 2022-04-26
Introduce leftJoin on Seq

tnagelkerke:
Very similar to the concept of left joining in a relational database, this method allows joining 2 Seq's based on keys that are extracted from each element.


pivovarit:
Thanks! Looking good! I will merge it to 1.x as-is and then potentially readjust when refining API for actual 1.x release

pivovarit (approved):

PR #2724, opened 2022-10-28
Propose Try.toEither(Throwable => L)

eslep:
### Thank You
Thank you for all of the work on Vavr! It has quickly become my favorite Java library, and usually the first thing I add to every new project ❤️  Aside from the benefit to my own code, it has also been a great tool to help teach others about FP and get them curious about it.

### Proposal / What is this?
I come across cases where I want to convert a `Try` to an `Either` while mapping the `Failure` value onto a `Left` value.
One example would be safely handling code which isn't mine, then mapping the `Throwable` onto an error type.

I do this normally either with `fold(throwableMapper, identity())` or `.toEither().mapLeft(throwableMapper)`, but find that `.toEither(throwableMapper)` would be a nice shortcut to replace either case.
I have also a few colleagues who would appreciate the addition from a readability standpoint I think, so I thought I'd suggest it.

### Implementation
By my own style, I'd normally just write this as a wrapper around a `fold` call. However, as a first contribution, I am trying to stick as close as possible to the style I already see in the code here.

By that reasoning, the code style, namings, and Javadoc are all created based on examples I find existing within the file.

Since the `toEither(Supplier)` implementation assumes the `Supplier` is pure/will not throw an exception and does not handle such a case, I have followed the same assumption here as well.

### DIFF
--- a/src/main/java/io/vavr/control/Try.java
+++ b/src/main/java/io/vavr/control/Try.java
@@ -1133,6 +1133,21 @@ public abstract class Try<T> implements Iterable<T>, io.vavr.Value<T>, Serializa
     }
 
     /**
+     * Converts this {@code Try} to an {@link Either}, converting the Throwable (if present)
+     * to a left value using the passed {@link Function}
+     *
+     * @param throwableMapper A transformation from throwable to the left type of the new {@code Either}
+     * @return A new {@code Either}
+     */
+    public final <L> Either<L, T> toEither(Function<? super Throwable, ? extends L> throwableMapper) {
+        if (isFailure()) {
+            return Either.left(throwableMapper.apply(getCause()));
+        } else {
+            return Either.right(get());
+        }
+    }
+
+    /**
      * Converts this {@code Try} to a {@link Validation}.
      *
      * @return A new {@code Validation}
--- a/src/test/java/io/vavr/control/TryTest.java
+++ b/src/test/java/io/vavr/control/TryTest.java
@@ -969,6 +969,15 @@ public class TryTest extends AbstractValueTest {
         assertThat(failure().toEither(() -> "test").isLeft()).isTrue();
     }
 
+    @Test
+    public void shouldConvertFailureToEitherLeftMapper() {
+        final Try<Object> failure = Try.failure(new RuntimeException("a certain value"));
+        final Either<Boolean, Object> either = failure.toEither(t ->
+                t.getMessage().equals("a certain value")
+        );
+        assertThat(either).isEqualTo(Either.left(true));
+    }
+
 
     // -- toValidation
 


src/test/java/io/vavr/control/TryTest.java
```diff
@@ -969,6 +969,15 @@ public void shouldConvertFailureToEitherLeftSupplier() {
         assertThat(failure().toEither(() -> "test").isLeft()).isTrue();
     }
 
+    @Test
+    public void shouldConvertFailureToEitherLeftMapper() {
+        final Try<Object> failure = Try.failure(new RuntimeException("a certain value"));
+        final Either<Boolean, Object> either = failure.toEither(t ->
+                t.getMessage().equals("a certain value")
+        );
+        assertThat(either).isEqualTo(Either.left(true));
```
nfekete, line 978: I'd probably replace the test code in a way that limits the number of comparisons to one, which would be the one in the assertion itself.
You could use a mapper function that extracts the message from the exception and use that as the left value to compare to in the assertion. This would make the code both shorter, and — in my opinion — easier to understand, since it would have a slightly less degree of indirection.
```suggestion
        final Either<Boolean, Object> either = failure.toEither(Throwable::getMessage)
        assertThat(either).isEqualTo(Either.left("a certain value"));
```

src/main/java/io/vavr/control/Try.java
```diff
@@ -1132,6 +1132,21 @@ public final Either<Throwable, T> toEither() {
         }
     }
 
+    /**
+     * Converts this {@code Try} to an {@link Either}, converting the Throwable (if present)
```
nfekete, line 1136: I'm proposing this change for the documentation so that it is more precise in meaning, because one could legitimately use a `Try` value to wrap a `Throwable` instance as a success value.
```suggestion
     * Converts this {@code Try} to an {@link Either}, converting the Throwable (in the failure case)
```

nfekete (approved):
Looks good to me, with a few minor suggestions.

src/test/java/io/vavr/control/TryTest.java
eslep: Sure, sounds much better that way 👍 I also felt a bit weird using `Throwable => Boolean` as my function where many people would maybe expect that to then be `Predicate<Throwable>`. I can see also how the `equals` looks like an assertion.

Main point was just to have a clear expected value returned by the function, so the suggestion sounds great

src/main/java/io/vavr/control/Try.java
eslep: I had the same impression that the wording was strange, but I kept this to stay consistent with `toValidation(Throwable => U)`.

I would make this doc change to both of these methods then, since your point applies to both cases.

pivovarit:
Hi! It's been a while, I know :) I'd love to include this in v0.11.0 - would you be willing to finish or should I take over? keep in mind that the current `main` branch differers a lot, and you'd need to probably reapply those changes from scratch.

adamkopec:
I ported this to new main branch here: https://github.com/vavr-io/vavr/pull/3088

PR #2729, opened 2023-01-01
Upgrade  Gradle wrapper to version 7.6

bmscomp:
Minimal pull request for build tools upgrade 
- upgrade Gradle wrapper to version 7.6 
- add entry for jdk 17 to travis script


danieldietrich (approved):
Thanks Said, LGTM!

.travis.yml
```diff
@@ -20,6 +20,8 @@ matrix:
     env: JAVA_VERSION=11 DEPLOY=true
   - jdk: openjdk16
     env: JAVA_VERSION=16
+  - jdk: openjdk17
+    env: JAVA_VERSION=17
```
mincong-h, line 24: Unfortunately this does not work because Travis CI didn't run since June 6, 2021, see https://travis-ci.org/github/vavr-io/vavr Perhaps it's time to move to GitHub actions 😢 Here is what I did few days ago for vavr-jackson https://github.com/vavr-io/vavr-jackson/pull/195

bmscomp: I will try to move to GitHub actions :)

danieldietrich: Good point! Yep, GH Actions is the way to go!

PR #2732, opened 2023-01-23
add copyright information to jar

tobi5775:
### Add copyright information to Jar.

For a customer project an OpenSource report has to be generated. Therefore a scanner was created to extract the copyright and licence information from our dependencies. This scanner can extract all informaton from the Jar File. Therefore it would be great to have the copyright information included in the Jar File as COPYRIGHT File.


danieldietrich (approved):
Thank you so much! Great addition :)

PR #2735, opened 2023-02-09
Propose Validation.all() to combine multiple validations on the same value.

evantill:
Hello all,

I'm submitting this pull request to add a method to combine multiple validation of the same type into a single validation.

I've added the tests and it's working as expected. I've also updated the relevant documentation to reflect the changes.
Please let me know if there's anything that needs to be changed or improved upon. I'm happy to make any necessary updates before merging.
Thank you for your time!

### Example:
We have a Path and we want to validate that it’s an absolute path on a writable directory.
We already have simple validations like :
```java
Validation<String,Path> isAbsolute(Path)
Validation<String,Path> isDirectory(Path)
Validation<String,Path> isWritable(Path)
```
We can then combine those like this :
```java
var pathValidation=Validation.all(
	isAbsolute(path),
	isDirectory(path),
	isWritable(path)
);
```

### DIFF
--- a/src/main/java/io/vavr/control/Validation.java
+++ b/src/main/java/io/vavr/control/Validation.java
@@ -19,9 +19,11 @@
 package io.vavr.control;
 
 import io.vavr.*;
+import io.vavr.collection.Array;
 import io.vavr.collection.Iterator;
 import io.vavr.collection.List;
 import io.vavr.collection.Seq;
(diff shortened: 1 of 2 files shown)


evantill:
Hi, any update on this PR please ?

danieldietrich (approved):
This is a helpful addition, many thanks.
LGTM

Issue #2711, opened 2022-03-01
Checking for null _creates_ a null reference problem


jbrains:
My apologies if you've already discussed this to death; just let me know.

I ran into this today.

```
var fallback = ...;
myOption.fold(fallback::g, null);
```

I was in a test where I knew `myOption` is `none()`, so I didn't bother supplying a function for the "some" case---instead I left it `null`. I reasoned that since only `fallback` was being called in this path, the `fold()` would silently work. Instead, it blew up with a failed `null` check.

Admittedly, I haven't thought this through for more than a few moments, but I infer that `fold()` is eagerly checking for `null`. I think I would prefer if it didn't. Is `fold()` checking for `null` as a defensive habit (for the sake of consistency, I imagine) or to avoid a genuine problem?

danieldietrich:
Hi, I would stick to eager evaluation here because generally it is a good idea to fail early. Lazy evaluation might lead to unexpected errors. Your specific problem could be solved by just using () -> null instead of null.

jbrains:
Thank you. I believe I understand your position, but I use the ability to pass `null` for a collaborator as a design tool to alert me to when code might better be split into two smaller parts. The idea is to document examples (tests) that can safely pass `null`, which this behavior now artificially prevents.

As you say, I could merely pass a collaborator that intentionally does nothing or I could use a Crash Test Dummy, which would blow up spectacularly if accidentally used. That would provide a similar benefit, although perhaps not be quite as easy to spot when skimming the code.

I think this comes down to personal preference. I don't think I care enough about the difference to cling to my preference. As you point out, lazy evaluation might cause more headache for others than it causes for me in this one case.

Thank you.

danieldietrich:
I thought about it a bit..

I agree that it is artificial. In the early days of Vavr (Javaslang) having explicit null-checks all over the place was questioned. Especially when called internally (where I am sure that function refs are not null) it is an overhead of performing null-checks n times.

Others asked for adding (non-)nullability annotations to method parameters in order to be able to check it at runtime (which would not help you because you want to set a non-nullable param to null).

To be honest, I would be okay with it to completely remove all the null-checks in favor of performance. Personally, the concrete message does not matter to me if a NPE occurs at runtime. Removing all the checks would change the semantics a bit but hopefully no one relies on such exceptions for the control flow of their program (fingers crossed 🤞).

jbrains:
[...]

PR #2744, opened 2023-05-06
Remove explicit null checks

patrickguenther:
https://github.com/vavr-io/vavr/issues/2711

### DIFF
--- a/src/main/java/io/vavr/collection/Iterator.java
+++ b/src/main/java/io/vavr/collection/Iterator.java
@@ -1033,7 +1033,11 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
     // -- Additional methods of Iterator
 
     @Override
+    @SuppressWarnings("unchecked")
     default <R> Iterator<R> collect(PartialFunction<? super T, ? extends R> partialFunction) {
+        if (!hasNext()) {
+            return (Iterator<R>)this;
+        }
         return filter(partialFunction::isDefinedAt).map(partialFunction);
     }
 
--- a/src/main/java/io/vavr/collection/Iterator.java
+++ b/src/main/java/io/vavr/collection/Iterator.java
@@ -74,7 +74,6 @@ public interface Iterator<T> extends java.util.Iterator<T>, Traversable<T> {
     @SuppressWarnings("varargs")
     @SafeVarargs
     static <T> Iterator<T> concat(Iterable<? extends T>... iterables) {
-        Objects.requireNonNull(iterables, "iterables is null");
         if (iterables.length == 0) {
             return empty();
         } else {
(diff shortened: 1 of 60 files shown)


src/main/java/io/vavr/collection/Iterator.java
```diff
@@ -1033,7 +1033,11 @@ public T next() {
     // -- Additional methods of Iterator
 
     @Override
+    @SuppressWarnings("unchecked")
     default <R> Iterator<R> collect(PartialFunction<? super T, ? extends R> partialFunction) {
+        if (!hasNext()) {
```
patrickguenther, line 1038: I'm not sure if this is adheres to coding guidelines.
If `partialFunction == null` should not throw when empty, this additional check is required, because the expression partialFunction::isDefinedAt will otherwise throw immediately.

szarnekow:
What is the reasoning behind this change? At first glance it seems to cloak a lot of programming mistakes by shifting away from fail-early-at-root-cause to fail-at-random-point-in-time.

patrickguenther:
@szarnekow see discussion in #2711 
But also: the whole shift we see in the last couple of years for programming languages becoming more functional is because we, the software developer community, have figured out that the best solution to avoid NPEs is to completely eradicate the use of the nullpointer itself.
If that goal is reached, however, all code for checking nulls just becomes obsolete boilerplate.

The other argument, you brought up: "Throwing close to the source" is maybe only valid when the potential throwing of the stack trace and the injection point of the null value are in completely different execution stacks. For that, I would support placing some null checks in strategic places like Futures, where an uncaught NPE might even lead it to never terminate.
Otherwise IDEs usually highlight your own classes in the stack trace or when looking at the stacktrace in a terminal or textfile you anyway quickly scan the trace for your own namespace.

ezoerner:
> What is the reasoning behind this change? At first glance it seems to cloak a lot of programming mistakes by shifting away from fail-early-at-root-cause to fail-at-random-point-in-time.

See the discussion of this in #2711. Evidently the reasoning is that it is for performance reasons that the null checks are being removed. Ideally null checks should not be needed in a functional library as nulls have no business being in functional code—but unfortunately in Java that minefield always lurks under the surface. Personally I think it’s ok to remove the null checks because 1) there shouldn’t be any nulls when your code properly protects against them coming from 3rd party libraries, and 2) a resulting NPE almost always provides enough locality info to quickly determine where the null came from.

For me the compelling reason is to keep the code cleaner by removing the null-check noise; I am skeptical about performance being a compelling reason to remove them.

pivovarit:
Thank you for all the effort that went into this, but we'll not be moving forward: https://github.com/vavr-io/vavr/issues/2711#issuecomment-2708167247

> For me the compelling reason is to keep the code cleaner by removing the null-check noise; I am skeptical about performance being a compelling reason to remove them.

I agree with the statement that those null-checks are noise, but I believe the convenience of library creators is less important than user experience.

Issue #2417, opened 2019-06-13
Explore reimplementing HashMap as CHAMP (Compressed Hash-Array Mapped Prefix-tree) 


pivovarit:
Related Epic: https://github.com/vavr-io/vavr/issues/2308

Scala 2.13 brought us new and improved collections. One of them is a fully reimplemented `HashMap` as CompressedHAMP.

https://github.com/scala/collection-strawman/issues/192
https://github.com/scala/scala/commit/d5ae93e1b35fac4002cfbffff9ec741356549e08

https://michael.steindorfer.name/publications/oopsla15.pdf

We should follow up and investigate if it makes sense to do the same move as Scala.

Would love to work on this myself.

pivovarit:
Great, I was aware of the library - wasn't aware of that thread. Definitely helpful!

danieldietrich:
@pivovarit I plan to include it in 2.0.0, which is already in the make but in a very early alpha state. Collections aren't part of it, currently.

I also plan to get my hands on the essential data structures BMT, Red/Black Tree and CHAMP. I'm not sure if it makes sense that you invest time.

wrandelshofer:
I made an experimental (and very dirty) port of CHAMP collections in a fork of vavr:
https://github.com/wrandelshofer/vavr#readme

I decided to implement linked collections with sequence numbers. This is much simpler to do, than keeping a linked list or a vector in sync with the CHAMP trie. However performance is a mixed bag. Most operations are quite fast, except accesses to the first/last entry of the collection - which are linear in time. Also, occasionally, we need to renumber the sequence numbers, which may cause unexpected slow-downs. Memory efficiency should be quite good though.

So, maybe my fork is just a demonstration on how not to do it. 😂

wrandelshofer:
I made some progress now. 

I have now implemented four different CHAMP-based collections:

- vavr.ChampMap
- vavr.SequencedChampMap
- vavr.ChampSet
- vavr.SequencedChampSet

Here is the current README file. It shows a performance comparison with vavr.HashMap, vavr.LinkedHashMap and scala.HashMap, scala.VectorMap, scala.TreeSeqMap.

https://github.com/wrandelshofer/vavr/blob/6569c67e3d5c3b4dae681fa762c714c8d074d7ab/README.md

The performance of the sequenced collections (the ones that preserve the insertion order) is now O(1) for all operations. 
There are no performance cliffs anymore.

The code is in a very rough draft state. Let me know, what you think about it.

[...]

PR #2745, opened 2023-05-10
#2417 Explore reimplementing HashMap as CHAMP (Compressed Hash-Array Mapped Prefix-tree) 

wrandelshofer:
This is a draft intended for discussion.

This merge requests explores how a migration of the HAMT trie to a CHAMP trie could look like:

- HashSet, HashMap: Replaces the internal HAMT data structure by a CHAMP data structure. 
- LinkedHashSet, LinkedHashMap: Replaces the internal HAMT+Seq data strucures by CHAMP+Vector data structures.
- The changes are intended to affect the performance of vavr, but are not supposed to change any API.

The changes are somewhat sub-optimal and somewhat over-complicated, because I strived to not break any of the unit tests. (I believe that I should change a number of unit tests, because some rely on the iteration sequence of the HAMT implementation. However, the vavr API specifies that the iteration sequence is unspecified.)

The code is a port from the JHotDraw 8 collections module. https://github.com/wrandelshofer/jhotdraw8/tree/main/org.jhotdraw8.collection
The JHotDraw 8 collections are in turn a port of the 'capsule' collection library https://github.com/usethesource/capsule, and of vavr.

I do not expect that this draft can be merged in the current form. In JHotDraw 8 we have a mutable 'partner' collection for every immutable collection. This allows to test all immutable collections with the 'guava-testlib' library. Which provides very through tests. I did not want to change the API of vavr - so I have downgraded the mutable collections to transient collections. But I have not written new tests yet. So code coverage by unit tests is currently lower than I want it to be.


wrandelshofer:
I made now a release that is binary compatible with vavr 0.10.5.

You can get it here: https://github.com/wrandelshofer/vavr/releases/tag/v0.10.5

pivovarit:
What do you think about including those new implementations in a dedicated package `io.vavr.collection.champ`, shipping it as part of v0.11.0, and then potentially replacing existing implementations for v1.0.0?

Issue #2811, opened 2024-09-11
JDK 21-23 Compatibility for 0.x branch


pivovarit:
Branch `version/0.x` represents a release train for "legacy" vavr and it still needs to be able to be built and verified under the newest JDKs. 

At the moment, it collapses under JDK21 with:

```
(...)
  last tree to typer: EmptyTree
       tree position: <unknown>
            tree tpe: <notype>
              symbol: null
           call site: <none> in <none>

== Source file context for tree position ==

error: scala.reflect.internal.FatalError:
  bad constant pool index: 0 at pos: 48445
     while compiling: <no file>
        during phase: globalPhase=<no phase>, enteringPhase=<some phase>
     library version: version 2.12.6
    compiler version: version 2.12.6
```

micrommer:
Hi @pivovarit ,

First of all, I’m really excited to see a new, passionate leader taking charge of the wonderful VAVR!

I compiled using the OpenJDK 64-Bit Server VM Zulu21.32+17-CA (build 21.0.2+13-LTS, mixed mode, sharing) and encountered the same exception. From what I understand, this appears to be an issue with the bytecode.

If you could provide some guidance on how to resolve this issue, I’d be happy to assist.

pivovarit:
Thanks! Actually, I have no clue what is the core issue here - my bet is that it's caused by outdated dependencies responsible for code generation, or Scala itself (according to [scala docs](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html), _2.12.18_ is the first JDK21-friendly version)

micrommer:
Hi dear @pivovarit 
I hope you're doing well!

Please take a look at the PR. It is probably not a comprehensive solution for the mentioned issue, but it could possibly be a path to it.
Based on my observation, the warnings are mostly related to serialization. I’ll attach the log file that shows only the warnings.
I think removing -Werror cannot be a permanent solution, and the serialization problem should be fixed.
[warnings.log](https://github.com/user-attachments/files/17118208/warnings.log)

 Best regards

KrnSaurabh:
Hi @pivovarit , I also see -Werror warnings related to serialization problem with JDK21. I am able to build `version/0.x` branch on JDK8/JDK17/JDK21 simultaneously. I can work on fixing those serialization issues. Please let me know if we want to put effort towards the same.

pivovarit:
Yes, we should handle those - I guess that would be the last thing we could fix before releasing 0.10.5 :)

[...]

PR #2841, opened 2024-09-24
Ignore the warningAsError (-Werror) maven build flag

micrommer:
Update Scala to the latest version and ignore the maven strict warning checking flag!
With these changes, the project has been built with no problem using the following command
`mvn clean install`

It is a showcase pull request and more investigation is required. 
A draft for #2811.


pivovarit:
Thanks for the PR - it's clear the only thing left is to deal with those serialization issues :) someone is already on it

PR #2842, opened 2024-09-24
Replace synchronized method/block with reentrant lock 

KrnSaurabh:
Virtual threads suffer when performing a blocking operation inside a `synchronized` method/block, and it is recommended that the `synchronized` method/block be replaced with a `ReentrantLock`.

More details: https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html#GUID-04C03FFC-066D-4857-85B9-E5A27A875AF9 

----

related: https://github.com/vavr-io/vavr/issues/2760

### DIFF
--- a/src/main/java/io/vavr/Lazy.java
+++ b/src/main/java/io/vavr/Lazy.java
@@ -37,6 +37,7 @@ import java.io.Serializable;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Proxy;
 import java.util.Objects;
+import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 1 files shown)


pivovarit:
Would you be willing to backport it to [version/0.x](https://github.com/vavr-io/vavr/tree/version/0.x)? If not, I will handle that for you

KrnSaurabh:
> Would you be willing to backport it to [version/0.x](https://github.com/vavr-io/vavr/tree/version/0.x)? If not, I will handle that for you

Yes sure, I can backport it to version/0.x branch.

PR #2845, opened 2024-09-25
Replace synchronized method/block with reentrant lock  (#2842)

KrnSaurabh:
Virtual threads suffer when performing a blocking operation inside a `synchronized` method/block, and it is recommended that the `synchronized` method/block be replaced with a `ReentrantLock`.

More details: https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html#GUID-04C03FFC-066D-4857-85B9-E5A27A875AF9

----

related: https://github.com/vavr-io/vavr/issues/2760


pivovarit:
Thanks :)

PR #2857, opened 2024-09-29
Update Scala to 3.5

achinaou:
Made the minimal changes to update Scala to `3.5` and fix the resulting issues.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,8 +40,9 @@ plugins {
     id 'io.github.gradle-nexus.publish-plugin' version '2.0.0'
 }
 
-ext.ammoniteScalaVersion = '2.13'
-ext.ammoniteVersion = '2.5.11'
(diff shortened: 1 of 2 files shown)


achinaou:
I started by migrating from Ammonite to using Scala directly, as the [3.5+](https://scala-lang.org/blog/2024/08/22/scala-3.5.0-released.html) versions use the [Scala CLI](https://scala-cli.virtuslab.org) as the default runner, which provides similar functionality. But, this led to a larger download size and more complex build script.

pivovarit:
Thanks! Amazing work! It was on my list as well, but I postponed it because it seemed like way more work 🙈

achinaou:
I also was expecting it to be more complicated...

It's nice for a change to overestimate a task! :sweat_smile:

PR #2858, opened 2024-09-29
Update Scala to 3.5

achinaou:
Backport of: https://github.com/vavr-io/vavr/pull/2857


achinaou:
I saw that the build fails when upgrading the `scala-maven-plugin` or the `scala-version` individually (#2851, #2833).
But combining those changes did the trick.

PR #2869, opened 2024-09-30
Backport: # 2870 Fix Java Serialization warnings for Tuple

KrnSaurabh:
backport https://github.com/vavr-io/vavr/pull/2870

### DIFF
--- a/vavr/generator/Generator.scala
+++ b/vavr/generator/Generator.scala
@@ -1953,7 +1953,7 @@ def generateMainClasses(): Unit = {
               /$javadoc
                * The ${j.ordinal} element of this tuple.
                */
-              public final T$j _$j;
+              transient public T$j _$j;
             """)("\n\n")}
 
             ${if (i == 0) xs"""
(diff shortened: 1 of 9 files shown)


KrnSaurabh:
@pivovarit There are total 73 warnings related to Java Serialization when built on JDK21 and many of them requires to mark the instance variable as `transient` which additionally requires us to implement readObject()/writeObject() methods in order to serialize/deserialize objects.

In this PR I have fixed issues related to tuples only and there are many more. Hence fixing all 73 issues in one single PR will cause too many lines of code changes and will be hard to review as well.

Hence I am planning to publish fixes for all 73 issues in multiple PRs. Let me know if you have better suggestion on this.

vavr/generator/Generator.scala
```diff
@@ -1953,7 +1953,7 @@ def generateMainClasses(): Unit = {
               /$javadoc
                * The ${j.ordinal} element of this tuple.
                */
-              public final T$j _$j;
+              transient public T$j _$j;
```
pivovarit, line 1956: `transient` means that you want to ignore field during serialization - tuple elements can't be ignored

szarnekow: Indeed. The transient modifier will likely break other serialization libraries, too.

KrnSaurabh: There is an implementation added in this class for readObject() and writeObject() to explicitly serialize and deserialize these fields so that they are not ignored.

pivovarit:
I've taken care of Tuples here: https://github.com/vavr-io/vavr/pull/2872, feel free to have a look at the rest

PR #2870, opened 2024-09-30
Fix Java Serialization warnings for Tuple

KrnSaurabh:
Fix Java Serialization warnings for Tuple.
Maven build throws warnings related to Java Serialization when built with JDK21
>  non-transient instance field of a serializable class declared with a non-serializable type

https://github.com/vavr-io/vavr/issues/2811

### DIFF
--- a/generator/Generator.sc
+++ b/generator/Generator.sc
@@ -1963,7 +1963,7 @@ def generateMainClasses(): Unit = {
               /$javadoc
                * The ${j.ordinal} element of this tuple.
                */
-              public final T$j _$j;
+              transient public T$j _$j;
             """)("\n\n")}
 
             ${if (i == 0) xs"""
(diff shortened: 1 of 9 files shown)


KrnSaurabh:
@pivovarit There are total 73 warnings related to Java Serialization when built on JDK21 and many of them requires to mark the instance variable as `transient` which additionally requires us to implement readObject()/writeObject() methods in order to serialize/deserialize objects.

In this PR I have fixed issues related to tuples only and there are many more. Hence fixing all 73 issues in one single PR will cause too many lines of code changes and will be hard to review as well.

Hence I am planning to publish fixes for all 73 issues in multiple PRs. Let me know if you have better suggestion on this.

pivovarit:
@KrnSaurabh thanks for the hard work - however, I think this needs a different approach. The approach with changing final fields to `transient` and writing custom serialization/deserialization logic solves the problem by moving it somewhere else - the code won't work if some of those fields are not serializable.

In this case, it's probably enough to just ignore those serialization issues. This is also what JDK does in `AbstractMap.SimpleEntry`:

```java
public static class SimpleEntry<K,V> implements Entry<K,V>, java.io.Serializable {
    // ...
    @SuppressWarnings("serial") // Conditionally serializable
    private final K key;
    @SuppressWarnings("serial") // Conditionally serializable
    private V value;
    // ...
```

pivovarit:
Also, I think that's it's all that needs to be done here - other cases depended on fixing Tuple :)

KrnSaurabh:
Sure, make sense. Let me try to change the code and suppress the warning.

PR #2880, opened 2024-10-04
[version/0.x branch] Fix remaining Java Serialization issues with JDK21

KrnSaurabh:
Fix remaining Java Serialization issues with JDK21


pivovarit:
Looks good to me - we still need to figure out what's wrong with JDK23

KrnSaurabh:
> Looks good to me - we still need to figure out what's wrong with JDK23

Sorry, I didn't worry about JDK23 since it was not mandatory check. I can look into it separately.
Thanks for reviewing the PRs.

pivovarit:
Those were not mandatory, because they never worked :) once they work, I will upgrade them to mandatory (just like I'm about to do now with JDK 21/22)

PR #2881, opened 2024-10-04
Fix remaining Java Serialization issues with JDK21

KrnSaurabh:
Fix remaining Java Serialization issues with JDK21

### DIFF
--- a/generator/Generator.sc
+++ b/generator/Generator.sc
@@ -1105,7 +1105,7 @@ def generateMainClasses(): Unit = {
                   private static final long serialVersionUID = 1L;
 
                   private final Pattern0<T> pattern;
-                  private final $FunctionType<? super T, ? extends R> f;
+                  private transient final $FunctionType<? super T, ? extends R> f;
 
                   private Case0(Pattern0<T> pattern, $FunctionType<? super T, ? extends R> f) {
                       this.pattern = pattern;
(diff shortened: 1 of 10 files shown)


generator/Generator.sc
```diff
@@ -1131,13 +1131,18 @@ def generateMainClasses(): Unit = {
                   case 2 => BiFunctionType
                   case _ => s"Function$i"
                 }
+                val accessModifier = i match {
+                  case 1 => "transient final"
+                  case 2 => "transient final"
+                  case _ => "final"
```
pivovarit, line 1137: Can't it be `transient final` everywhere?

KrnSaurabh: Yes we can keep `transient final` for all `Case`, Although `Case3` onwards uses `io.vavr.Function` which is already serializable, hence there is no serializable warning for `Case3` onwards.

Let me know if I should keep `transient final` for all `Case`.

pivovarit: Ah, thanks for the clarification - this is ok

pivovarit (approved):

KrnSaurabh:
@pivovarit PR to backport in version/0.x branch --> https://github.com/vavr-io/vavr/pull/2880

pivovarit:
@KrnSaurabh looking now, trying to understand what fails when building with JDK 23

PR #2882, opened 2024-10-05
Fix maven build failure with JDK23

KrnSaurabh:
Fix maven build failure with JDK23.
Attached failure logs: 
[build_logs.txt](https://github.com/user-attachments/files/17265414/build_logs.txt)

### DIFF
--- a/vavr/pom.xml
+++ b/vavr/pom.xml
@@ -64,6 +64,13 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessors>
+                        <annotationProcessor>
+                            io.vavr.match.PatternsProcessor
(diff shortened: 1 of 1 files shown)


KrnSaurabh:
@pivovarit Please review.

pivovarit:
Looks good to me, but any clue why it suddenly stopped working on JDK23?

KrnSaurabh:
Not sure exactly why we need to specify `AnnotationProcessor` class specifically in pom file. Ideally `javax.annotation.processing.Processor` in Resource/META-INF should be enough for Maven to invoke Annotation Processor.

I found it from the logs that `AnnotationProcessor` is not being invoked and some classes were not being generated so the build was throwing compilation errors. It work when `AnnotationProcessor` specified in `maven-compiler-plugin` config.

pivovarit:
@KrnSaurabh I think this is the reason: https://inside.java/2024/06/18/quality-heads-up/

PR #2909, opened 2024-10-15
Replace synchronized by lock in functions generator

Koziolek:
Similar to https://github.com/vavr-io/vavr/pull/2845 

I replaced synchronized blocks by `Lock` because of virtual threads. More details: https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html#GUID-04C03FFC-066D-4857-85B9-E5A27A875AF9


pivovarit:
Thanks, looks good to me :)

PR #2912, opened 2024-10-18
Replace synchronization by Lock in FutureImpl

Koziolek:
I replaced _synchronized_ blocks with Lock for the sake of Virtual Threads support. More details: https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html#GUID-04C03FFC-066D-4857-85B9-E5A27A875AF9

----
Similar to https://github.com/vavr-io/vavr/pull/2845


pivovarit:
It's ok for now, but I'd definitely revisit the design of Vavr's Future. That locking in constructor feels excessive - I believe that _immutable_ objects can be actually mutable inside as long as this this doesn't leak. It also feels that the current implementation tries to be overly smart by inducing _happens-before_ using _synchronized_ instead of plain _volatile_

PR #2950, opened 2024-12-29
Add mapTry to Option

bvkatwijk:
* Add `mapTry` method to Option, which is a shortcut for `.toTry().mapTry`
  * I propose this because when mapping an optional using a checked method, I find it intuitive that the type changes to a Try, and `mapTry` methods exist on other types already
  * I initially added it onto `Value` but this ran into an issue where a method with same name but return value already exists on subtype`Future`

Unrelated but maybe helpful (I can remove it if desired)
* Add toolchain resolver
   * After forking the project I could not load it in IntelliJ, Gradle complained that it could not resolve the toolchain. Added a resolver as per the [Gradle Docs](https://docs.gradle.org/current/userguide/toolchains.html#sub:download_repositories)
* Add Nested test class for `mapTry` tests instead of a comment
   * This achieves organizational separation just like the comment, but now `JUnit` also understands it and can e.g. run only these tests in isolation.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,1 +1,5 @@
+plugins {
+    id 'org.gradle.toolchains.foojay-resolver-convention' version '0.9.0'
+}
+
 rootProject.name = 'vavr'
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,5 +1,1 @@
-plugins {
-    id 'org.gradle.toolchains.foojay-resolver-convention' version '0.9.0'
-}
-
 rootProject.name = 'vavr'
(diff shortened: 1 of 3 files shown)


pivovarit:
Thanks for your contribution :) looking good!

> Add Nested test class for mapTry tests instead of a comment

makes sense - we've migrated to JUnit5 recently, but we're not leveraging it fully

> After forking the project I could not load it in IntelliJ, Gradle complained that it could not resolve the toolchain.

sounds good - could you add it a a separate change, though? so that we don't mix two things together

> I initially added it onto Value but this ran into an issue where a method with same name but return value already exists on subtypeFuture

that's something we might want to straighten out before releasing 1.0.0

bvkatwijk:
Proposing to migrate to nested classes in #2951

Issue #2924, opened 2024-10-28
Enforce CONTRIBUTING rules with ArchUnit


pivovarit:
At the moment, contributing rules are paper-based with protein-based verification and enforcement - https://github.com/vavr-io/vavr/blob/master/CONTRIBUTING.md

We could enforce those using https://www.archunit.org

bvkatwijk:
I think that ArchUnit does not support checking Javadoc.
An option for checking those rules could be [CheckStyle](https://checkstyle.sourceforge.io/checks/javadoc/index.html)

PR #2952, opened 2024-12-30
Enforce package depth with ArchUnit

bvkatwijk:
- Enforce package depth using ArchUnit test

CONTRIBUTING.md mentions:
> The maximum package depth is two

Violation results in a unit test failure:
> Architecture Violation [Priority: MEDIUM] - Rule 'classes should have a package depth of 3 or less' was violated (1 times):
Class io.vavr.too.deep.SomeClass has a package depth of 4, which exceeds the allowed maximum of 3

Small step in fixing issue #2924

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,6 +43,7 @@ plugins {
 ext.ammoniteScalaVersion = '3.5'
 ext.ammoniteVersion = '3.0.0'
 ext.ammoniteVersionQualifier = '2-6342755f'
+ext.archUnitVersion = '1.3.0'
 ext.assertjVersion = '3.27.0'
 ext.junitVersion = '5.11.4'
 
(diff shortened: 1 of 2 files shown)


bvkatwijk:
I'm not sure if i understand "The maximum package depth is two" correctly - I would assume that means no subpackages under `io.vavr`. It could also mean "max two subpackages under io.vavr" (i.e. `io.vavr.a.b.` is compliant).
I currently implemented it based on what I saw in the project itself (`io.vavr` source files have at most 1 subpackage level).
Could you clarify what is the desired rule?

pivovarit:
I wasn't around when the `The maximum package depth is two` rule was established, but my guess is that `io.vavr` is, kind of, considered a single namespace and `collection` or `control` are 2nd level depth packages - this is done to leverage package-private visibility and I think that's what we should stick to

pivovarit (approved):
Anyway, this is looking good :) thanks

Do you want to backport it to 0.x?

PR #2954, opened 2024-12-31
Fix CheckedRunnableTest#shouldApplyAnUncheckedFunctionThatThrows

bvkatwijk:
* `shouldApplyAnUncheckedFunctionThatThrows` always passed,  when removing the `throw new Error()` because the failed asserting was immediately caught.

```java
        final Runnable runnable = CheckedRunnable.of(() -> { /*throw new Error();*/ }).unchecked(); // Test would still pass
```

Fixed by using `assertThrows` from JUnit


pivovarit:
Looks good, thanks!

pivovarit:
fyi: https://github.com/vavr-io/vavr/pull/2956

PR #2955, opened 2025-01-01
Enforce method order with ArchUnit

bvkatwijk:
* Enforce alphabetical method order within method kinds:
  * static methods
  * non-static methods
  * adjusted return types

Notes:
* To interpret "adjusted return types" I currently used "All overridden methods", perhaps this is not fully accurate.
* ArchUnit does not yet have a built-in override predicate so I copied code from https://github.com/TNG/ArchUnit/issues/359#issuecomment-975456116
* Build will fail due to 264 detected violations
* Not yet done: Validating that entire sections are ordered (static methods > non-static methods > adjusted return types)

Example test output:
>Architecture Violation [Priority: MEDIUM] - Rule 'classes should have methods alphabetically ordered in sections' was violated (264 times):
Methods in section 'adjusted return types' of class 'io.vavr.Function1$1' are not alphabetically ordered: [isDefinedAt, apply]

Part of implementing #2924

### DIFF
--- a/src/test/java/linter/ClassSectionTest.java
+++ b/src/test/java/linter/ClassSectionTest.java
@@ -1,0 +1,70 @@
+package linter;
+
+import com.tngtech.archunit.core.domain.JavaClass;
+import com.tngtech.archunit.core.domain.JavaMethod;
+import com.tngtech.archunit.core.domain.JavaModifier;
+import com.tngtech.archunit.core.importer.ClassFileImporter;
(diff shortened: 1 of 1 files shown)


bvkatwijk:
Personally I don't think method ordering is very important - and if it is, i'm not sure why alphabetical would be the chosen ordering.

If an ordering is desired, I'll check if I can use freezing, that sounds like a good solution.

pivovarit:
https://github.com/vavr-io/vavr/pull/2984 :)

PR #2960, opened 2025-01-03
Use JUnit @Nested classes for tests instead of comment sections

### LINKED ISSUE #2951: Use @Nested classes instead of comments for grouping tests
JUnit Jupiter, tests can be grouped using [Nested classes](https://junit.org/junit5/docs/current/user-guide/#writing-tests-nested)

> @Nested tests give the test writer more capabilities to express the relationship among several groups of tests. Such nested tests make use of Java’s nested classes and facilitate hierarchical thinking about the test structure

This seems to fit the exact purpose of the `-- comment` divisors in test., and enables a number of useful features like being able to run the set of nested tests individually.

bvkatwijk:
Fix #2951

- [x] Update `Generator.sc` so that generated tests use `@Nested` classes
- [x] Use `@Nested` classes in all standalone tests
- [ ] Use `@Nested` classes in inherited tests
    - Not sure if this is possible at the moment - Inheritance and nested classes don't mix well, especially in java 8 which is the current language level
- [ ] Decide on style for `@Nested` classes
    -  Class name sometimes clashes with other class names, so I appended `Test` and used `@DisplayName`

Notes:
- Used a script to do some bulk refactoring (sed may behave differently across OSes): 
```shell-script
find ./src -name "*Test.java" -type f -exec sed -E -i '' \
    's|// -- ([a-zA-Z0-9]+)$|}\n\n    @Nested\n    @DisplayName("\1")\n    class \CAPS\1Test {|' {} +
find ./src -type f -name '*Test.java' -exec perl -i -pe 's/\bCAPS([a-zA-Z])([a-zA-Z]*)/\u\1\2/g' {} +
```


generator/Generator.sc
```diff
@@ -2925,157 +2930,168 @@ def generateTestClasses(): Unit = {
...
-            })("\n\n")}
+
+            }
+
+            @Nested
+            class MatchPatterns {
+
+                class ClzMatch {}
+                class ClzMatch1 extends ClzMatch {}
+                class ClzMatch2 extends ClzMatch {}
```
bvkatwijk, line 3058: I changed these to non-static since static inner classes are not supported in JDK 8

pivovarit:
note: there's now `main` branch, which is ex-`version/0.x` - for now, there's no reason to target `version/1.x`

pivovarit:
Perhaps we could start merging it bit by bit? starting with generator?

bvkatwijk:
> Perhaps we could start merging it bit by bit? starting with generator?

Good approach, done in #2999

pivovarit:
I tried to finish this one up, but it looks like all of it is already on main. Thanks for help, but there's nothing more to do here :)

PR #2976, opened 2025-01-06
Tool to simplify "if" code blocks.

qin-liang-dev:
``` java 
String planet = getPlanet();  
if (planet == null) {   
   planet = "Earth";  
}
```
To simplify this code, you can try using a ternary expression
``` java
 String planet = null==getPlanet()?"Earth":"getPlanet()";
 ```
Obviously, calling the 'getPlanet' method twice is inappropriate.
But you can simplify this code by using ifNew(Object, Predicate, Object) as follows: 
``` java  
 String planet =  If.ifNew(getPlanet(),Objects::isNull,"Earth");
```

### DIFF
--- a/vavr/src/main/java/io/vavr/NonePredicate.java
+++ b/vavr/src/main/java/io/vavr/NonePredicate.java
@@ -1,0 +1,7 @@
+package io.vavr;
+
+
+@FunctionalInterface
+public interface NonePredicate {
+    boolean test();
+}
--- a/vavr/src/main/java/io/vavr/control/If.java
+++ b/vavr/src/main/java/io/vavr/control/If.java
@@ -1,0 +1,213 @@
+package io.vavr.control;
+
+import io.vavr.NonePredicate;
+
+import java.io.Serializable;
+import java.util.Objects;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+/**
+ * <pre>{@code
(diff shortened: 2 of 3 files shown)


valery1707:
Original problem:
```java
String planet = getPlanet();
if (planet == null) {
   planet = "Earth";
}
```
can be solved within pure JDK classes:
```java
String planet = Objects.requireNonNullElse(getPlanet(), "Earth");
```
More generic problem with custom condition it can be solved too:
```java
String planet = Optional.ofNullable(getPlanet()).filter(it -> isActual(it)).orElse("Earth");
```

And both of this solutions can calculate "else" value in lazy manner:
```java
String planet1 = Objects.requireNonNullElseGet(getPlanet(), () -> "Earth");
String planet2 = Optional.ofNullable(getPlanet()).filter(it -> isActual(it)).orElseGet(() -> "Earth");
```

talios:
Under "new java" you can use pattern matching and do something like:

```java
var result = switch(getPlanet() {
    case Planet planet -> planet;
    default -> "Earth";
}
```

As @pivovarit says, I don't think we need anything special in Vavr for this, but if anything - that could be something for maybe an OpenRewrite extension (https://docs.openrewrite.org/) - it already has _a lot_ of good, automated code improvement modules/recipes without anything Vavr specific existing.

qin-liang-dev:
Thank you for your correction. I didn't know enough about Vavr before, but now I've learned more ways to use it.

PR #2977, opened 2025-01-10
Add Option#mapTry()

bvkatwijk:
Add `mapTry` method to Option, which is a shortcut for `.toTry().mapTry`:
  * when mapping an optional using a checked method, I find it intuitive that the type changes to a Try, and `mapTry` methods exist on other types already
  * I initially added it onto `Value` but this ran into an issue where a method with same name but return value already exists on subtype`Future`
  
----

Rebase of #2950


pivovarit (approved):

pivovarit:
Looking good, thanks!

PR #2999, opened 2025-03-12
Use @Nested in APITest (generated via Generator.scala)

bvkatwijk:
Partial Fix for #2951

- [x] Update `Generator.sc` so that generated tests use `@Nested` classes


pivovarit:
Thanks!

PR #3000, opened 2025-03-12
Replace deterministic random values with actual randomness in tests

bvkatwijk:
- [x] Use random numbers instead of fixed sequence in loop

Two tests used identical fixed random instance inside a loop, causing the extact same sequence of numbers being generated:

```java
            final Random random = getRandom(123456789);
                // <rest of code>
        }
```

Unless I am missing the purpose, this means that each loop is identical and thus redundant.
Generating the same sequence of numbers via a Random does not seem to be the intented design here.

If my assumptions are correct, this fixes both issues by using an unseeded random instance, creating a test run with actual randomness.

If each loop iteration should be different but deterministic values are desired, a different fix could be to move the random instance out of the loop:

```java
        final Random random = getRandom(123456789);
                // <rest of code>
        }
```

### DIFF
--- a/vavr/src/test/java/io/vavr/AbstractValueTest.java
+++ b/vavr/src/test/java/io/vavr/AbstractValueTest.java
@@ -64,18 +64,6 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 @ExtendWith(AbstractMultimapTest.TestTemplateProvider.class)
 public abstract class AbstractValueTest {
 
-    protected Random getRandom(int seed) {
-        if (seed >= 0) {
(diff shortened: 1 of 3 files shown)


pivovarit:
I agree that there's a problem, but I don't agree with the solution - this is an attempt to introduce property-based-like tests, but it's missing the point: those tests should be random, but with an ability to make them deterministic during debugging by passing a custom seed.

So the solution would be to:
- pass negative seed to let it randomize runs
- introduce parameterless `getRandom()` alongside the existing method and use it in tests

Your call :)

PR #3003, opened 2025-03-15
create MatchV2

tltwuyu:
My English is poor so I use translation software. Therefore, my expressions may not be accurate. In my project, I need to handle some situations where if statements are used to obtain a final value. I used the API.Match, but I don't like its syntax. So, I created my own Match class in my project. It is inspired by Optional. I prefer this syntax, and I think adding flatMap makes it more flexible. You can observe the following code:
```
String matchString = "1";
String value = MatchV2.of(matchString, "def")
    .caseNull("is null")
    .caseValue("null", "null")
    .flatMap((m, r) -> MatchV2.of(Integer.valueOf(m), r))
    .caseValue(0, "zero")
    .orElse("none");
```
is equal to
```
String matchString = "1";
String value="def";
if(Objects.isNull(matchString)){
    value="is null";
}else if(Objects.equals(matchString,"null")){
    value="null";
}else if(Objects.equals(Integer.valueOf(matchString),"0")){
    value="zero";
} else{
    value = "none"
}
```
I am not sure if Vavr considers this MatchV2 as a replacement for the API's Match, or as a class similar to Either. Or perhaps you are not considering adding this MatchV2 at all. I have limited understanding of Vavr's code standards and structure. If you are considering adding this class, could you please advise on how to optimize it? Alternatively, you could create a similar pattern matching implementation.
Please let me know your thoughts. Thank you!

### DIFF
--- a/vavr/src/main/java/io/vavr/MatchV2.java
+++ b/vavr/src/main/java/io/vavr/MatchV2.java
@@ -1,0 +1,289 @@
+package io.vavr;
+
+import io.vavr.control.Option;
+
+import java.util.Objects;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+/**
+ * Optional-like structural pattern matching for Java。
+ * <pre>
+ * <code>
+ *      MatchV2.of(matchString, "def")
+ *          .caseNull("is null")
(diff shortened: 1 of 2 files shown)


pivovarit:
Thanks for your contribution, but Match is effectively in maintenance mode, and we're not going to invest more into it - _switch-case_ is too good of an alternative nowadays

PR #3006, opened 2025-03-19
Added distinctByKeepLast for Sequences

sam-chakrabarty:
### About
This PR addresses a feature request mentioned [here](https://github.com/vavr-io/vavr/issues/2712).

The intent behind this change is to introduce a method(`distinctByKeepLast`) to sequences that can distinguish and retain the last seen value based on a specified key.

```
List.of("1a", "2a", "3a", "2b", "4a").distinctByKeepLast(s -> s.substring(1)); 
//Should return ["2b", "4a"]
```

--- 
closes: https://github.com/vavr-io/vavr/issues/2712

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Array.java
+++ b/vavr/src/main/java/io/vavr/collection/Array.java
@@ -747,6 +747,18 @@ public final class Array<T> implements IndexedSeq<T>, Serializable {
     }
 
     @Override
+    public Array<T> distinctByKeepLast(Comparator<? super T> comparator) {
+        Objects.requireNonNull(comparator, "comparator is null");
+        return ofAll(iterator().distinctByKeepLast(comparator));
+    }
+
+    @Override
+    public <U> Array<T> distinctByKeepLast(Function<? super T, ? extends U> keyExtractor) {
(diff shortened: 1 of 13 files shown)


pivovarit:
It's my mistake - I forgot that unordered data structures are also `Traversables` and this operation makes sense only for ordered data structures.  This should be indeed implemented only for `Seq`

The implementation looks good to me

sam-chakrabarty:
I have updated the code to address that comment, the latest PR should be reflective of the most recent changes

> It's my mistake - I forgot that unordered data structures are also `Traversables` and this operation makes sense only for ordered data structures. This should be indeed implemented only for `Seq`
> 
> The implementation looks good to me

No worries. I have just addressed that bottleneck and put it up for consumption.

sam-chakrabarty:
I believe all tests pass and comments are addressed, we should be good to merge after an approval 🚀

sam-chakrabarty:
@pivovarit just checking in if you had any other feedback prior to merging this PR in, I have refactored the code to limit exposure to sequences.

pivovarit (approved):
That looks good :) thanks for the great work! I might adjust JavaDocs slightly, but that can be easily done later on

PR #3016, opened 2025-05-09
Add `compose1`, ..., `composeN` methods to `FunctionN`

bvkatwijk:
I was in need of function composition for 2+ arity functions, but these were not present on the function types.

- [x] Add (generated) `composeN` method(s) to `Function`, adding a composition function to the nth parameter
  - `composeN` method name is numbered (e.g. `compose1`) to prevent ambiguity
- [x] Add (generated) tests

Open considerations:
- `Function1#compose` now has a (virtually) duplicate `compose1`, is this an issue?
- Should the input argument be `Function` or `Function1`?
- This pattern and naming scheme could also be used to add `curryN`, `applyN` etc.,

### DIFF
--- a/vavr/src-gen/main/java/io/vavr/CheckedFunction1.java
+++ b/vavr/src-gen/main/java/io/vavr/CheckedFunction1.java
@@ -160,6 +160,20 @@ public interface CheckedFunction1<T1, R> extends Serializable {
     R apply(T1 t1) throws Throwable;
 
     /**
+     * Returns a composed function that first applies the {@linkplain Function} {@code before1} to the
+     * 1st argument and then applies this CheckedFunction1 to the result.
+     *
+     * @param <S> argument type of before1
+     * @param before1 the function applied before this
+     * @return a function composed of before1 and this
+     * @throws NullPointerException if before1 is null
+     */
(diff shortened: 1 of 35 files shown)


vavr/src-gen/main/java/io/vavr/CheckedFunction1.java
```diff
@@ -298,6 +298,20 @@ default <V> CheckedFunction1<V, R> compose(CheckedFunction1<? super V, ? extends
...
+    /**
+     * Returns a composed function that first applies the {@linkplain Function} {@code before1} to the
+     * 1st argument and then applies this CheckedFunction1 to the result.
+     *
+     * @param <S> argument type of before1
+     * @param before1 the function applied before this
+     * @return a function composed of before1 and this
+     * @throws NullPointerException if before1 is null
+     */
+    default <S> CheckedFunction1<S, R> compose1(Function1<? super S, ? extends T1> before1) {
```
bvkatwijk, line 311: This is now (almost) a duplicate of `compose`, but including it does make the API more consistent.

pivovarit:
Thanks, looks great :) I will think again what to do with `compose1` before the release, but it will probably stay

pivovarit (approved):

pivovarit (approved):

PR #3017, opened 2025-05-10
Replace string literal for comment initiation to help IDE

bvkatwijk:
- [x] Replace "//" in heavily nested string interpolations to help IDE understand that it is a string, not a comment

Before, scala code is green (IDE thinks it is a string) and literal text is highlighted (IDE thinks it is code):
Screenshot 2025-05-10 at 19 27 12

After:
Screenshot 2025-05-10 at 19 27 22


pivovarit (approved):
Nice :)

PR #3018, opened 2025-05-15
Add Value.to() general conversion methods

joytools:
There are lot of useful type conversion to turn a `Value` into another well known data type, such as `java.util.List`, `java.util.Map`, `io.vavr.collection.Map`, etc.

The goal of this request is adding to `Value` a new generic `to(Function)` method in order to enable the conversion of a Value into a custom type by using a fluent syntax, as per following example:

```
record Book(int id, String title, String author) { }
    
class Library {
    Library(Iterable<Book> books) {
        this.books = Vector.ofAll(books);
    }
    public void display() {
        books.forEach(System.out::println);
    }
    private final Seq<Book> books;
}
    
void test() {
    Stream.of(1, 2, 3)
        .map(id -> new Book(id, "Book " + id, "John Doe"))
        .to(Library::new)
        .display();
}
```

This is also useful to enhance interoperability with other library data types, such as:
- Guava -> `Table`, `ImmutableList`, etc
- Apache Commons -> `Bag`, `MultiMap`, etc

There are other 3 variants to deal with more advanced convertions when the conversion `Function` takes in input an `Iterable` of `Tuple2`, `Tuple3 `or `Tuple4`, as per following example:

```  
class SortedLibrary {
    SortedLibrary(Iterable<Tuple2<Integer, Book>> books) {
        this.books = TreeMap.ofEntries(books);
    }
    public void display() {
        books.forEach(System.out::println);
    }
    private final SortedMap<Integer, Book> books;
}
    
void test() {
    Stream.of(1, 2, 3)
            .to(id -> id * 2,                                 // Key mapper
                id -> new Book(id, "Book " + id, "John Doe"), // Value mapper
                SortedLibrary::new)
            .display();
}
```

### DIFF
--- a/vavr/src/main/java/io/vavr/Value.java
+++ b/vavr/src/main/java/io/vavr/Value.java
@@ -1472,6 +1472,149 @@ public interface Value<T> extends Iterable<T> {
         return ValueModule.toTraversable(this, Vector.empty(), Vector::of, Vector::ofAll);
     }
 
+    /**
+     * Generic conversion method to transform this {@code Iterable<T>} into 
+     * an instance of type {@code R} by applying the given function.
+     *
+     * <p>This method is a convenience for applying a custom transformation
+     * to the current object with a fluent syntax.</p>
+     *
+     * @param <R> the type of the result produced by the transformation function
+     * @param function the function to apply to this {@code Iterable<T>}
+     * @return the result of applying the transformation function
+     * @throws NullPointerException if the provided function is {@code null}
+     * @author Andrea Rombaldi
+     */    
+    default <R> R to(final Function<? super Iterable<T>, R> function) {
+        Objects.requireNonNull(function, "function is null");
+        return function.apply(this);
+    }
+    
+    /**
+     * Generic conversion method to transform this {@code Iterable<T>} into 
+     * a result of type {@code R} by first mapping each element into a tuple 
+     * of values using the provided mapping functions, and then applying 
+     * a transformation function to the resulting collection of tuples.
+     *
+     * <p>Each element {@code x} of the iterable is transformed into a {@code Tuple2<T1, T2>}
+     * where {@code T1 = mapper1.apply(x)} and {@code T2 = mapper2.apply(x)}.
+     * The resulting {@code Iterable<Tuple2<T1, T2>>} is then passed to the final
+     * transformation function.</p>
+     *
+     * <p>This method is a convenience for applying a custom transformation
+     * to the current object with a fluent syntax.</p>
+     * 
+     * @param <T1> the result type of the first mapping function
+     * @param <T2> the result type of the second mapping function
+     * @param <R> the type of the result produced by the final transformation function
+     * @param mapper1 the first mapping function to apply to each element
+     * @param mapper2 the second mapping function to apply to each element
+     * @param function the function to apply to the resulting {@code Iterable} of tuples
+     * @return the result of applying the transformation function to the collection of tuples
+     * @throws NullPointerException if any of the provided functions is {@code null}
+     * @author Andrea Rombaldi
+     */
+    default <T1, T2, R> R to(
+            final Function<? super T, T1> mapper1,
+            final Function<? super T, T2> mapper2,
+            final Function<? super Iterable<Tuple2<T1, T2>>, R> function) {
+        Objects.requireNonNull(mapper1, "mapper1 is null");
+        Objects.requireNonNull(mapper2, "mapper2 is null");
+        Objects.requireNonNull(function, "function is null");
+        return function.apply(map(x -> Tuple(
+                mapper1.apply(x), 
+                mapper2.apply(x))));
+    }
+
+    /**
+     * Generic conversion method to transform this {@code Iterable<T>} into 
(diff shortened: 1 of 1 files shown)


pivovarit:
Thanks for your submission! I got it, and I need some time to think it through. I need to weigh pros and cons - there's some overlap with `collect()` and `toMap(km, vm).toStream()`, and I needto  play with it for a bit - we're getting close to releasing a new major version and new APIs are likely to stay for a very long time

joytools:
> Thanks for your submission! I got it, and I need some time to think it through. I need to weigh pros and cons - there's some overlap with `collect()` and `toMap(km, vm).toStream()`, and I needto play with it for a bit - we're getting close to releasing a new major version and new APIs are likely to stay for a very long time

IMHO `collect()` is more a compatibility layer between Vavr and JDK's stream API, so that you can use existing `java.util.stream.Collector` with `io.vavr.Value` based type.

The new `to(Function)` method proposed here aims to be a generalized version of the various Vavr's `toXXXX()` convertion methods.

To illustrate the idea, let's see how to rewrite some of the existing `toXXXX()` methods using `to(Function)`

```
Value.toArray() --> Value.to(Array::ofAll)
Value.toList() --> Value.to(List::ofAll)
Value.toLinkedSet() --> Value.to(LinkedSet::ofAll)
Value.toLinkedMap() --> Value.to(k -> k, v -> v, LinkedHashMap::ofEntries)
```

I am using a lot this approach combined with a library of singleton convertion functions in my [JoyTools Commons Library](https://github.com/joytools/joytools-commons) the you may see here:

[Element](https://github.com/joytools/joytools-commons/blob/main/src/main/java/org/joytools/commons/interop/Element.java) : get an element from any `Iterable`.
- only()
- first() / last()
- firstOrNull() / lastOrNull()
- firstOrFail() / lastOrFail()
- firstOrElse(defValue) / lastOrElse(defValue)

[Java](https://github.com/joytools/joytools-commons/blob/main/src/main/java/org/joytools/commons/interop/Java.java) : convertion to JDK data structures.
- array(Class)
- arrayList()
- optional()
- hashMap() / linkedHaspMap() / treeMap() / concurrentHashMap()
- hashSet() / linkedHashSet()
- etc

[Guava](https://github.com/joytools/joytools-commons/blob/main/src/main/java/org/joytools/commons/interop/Guava.java) : convertion to Guava's data structures.
- hashBasedTable() / treeBasedTable() / treeBaseTableIgnoreCase() / immutableTable()
- hashBiMap() / immutableBiMap()
- etc

[Vavr](https://github.com/joytools/joytools-commons/blob/main/src/main/java/org/joytools/commons/interop/Vavr.java) : convertion to Vavr'sdata structures.
- array()
- hashMap() / linkedHashMap() / treeMap() / treeMapIgnoreCase()
- hashSet() / linkedHashSet() / treeSet() / treeSetIgnoreCase()
- stream()
- option() / some() / none()
- etc

Let's combine all the above with some easy examples:

Create a new case insensitive Vavr SortedMap
```
Vector.of("a", "b", "c")
    .to(Function.identity(), 
        v -> v + v + v,
        Vavr.treeMapIgnoreCase());
```

Create a new case insensitive Java SortedMap
```
Vector.of("a", "b", "c")
    .to(Function.identity(), 
        v -> v + v + v,
        Java.treeMapIgnoreCase());
```

Create a new case insensitive Java concurrentHashMap()
```
Vector.of("a", "b", "c")
    .to(Function.identity(), 
        v -> v + v + v,
        Java.concurrentHashMap());
```

Create a new Guava Table
```
Vector.of("a", "b", "c")
    .to(Function.identity(), 
        v -> v + v + v,
        v -> v.length()
        Guava.hashBasedTable());
```

Get the only existing element
```
Stream.range(0, 1000)
    .map(...)
    .filter(...)
    .to(Element.only());
```

The advantages of this approach:
- the syntax is elegant and fluent
- if a new convertion method is needed, it is just matter of creating a new singleton convertion function instead of adding new toXXXX() methods in Vavr.
- the convertion functions may be implemented to work on any `Iterable`, not just on Vavr data types.

Hope this may help to clarify why I consider this proposal so interesting ;-)

pivovarit:
So, I considered both options, and I think:
- a lot of those use-cases could be implemented with a dedicated `to()` Collector: ```integers.collect(to(...))```, but this involves iteration which is not always needed, for example, when we just want to wrap the result
- `to()` as a new method allows skipping iteration when not needed and requires way less hassle than `integers.collect(to()))`
- `to()` with additional Tuple mappings is redundant - the same can be achieved by using `map().to()`, so I'd drop those companion methods for Tuple mappings and consider either expanding `map()` to accept multiple mapping functions

The new `to()` method should make better use of generics: at the moment, users are forced to provide a function of `Iterable`, which can be problematic in cases like:

image

```java
import io.vavr.collection.List;
import java.util.Collections;

class Main {
        List<Integer> integers = List.of(1);
        // reason: no instance(s) of type variable(s) T exist so that Iterable<Integer> conforms to List<? extends T>
        List<Integer> r = integers.to(l -> Collections.unmodifiableList(l));
    }
}
```

We also need tests :)

PR #3085, opened 2025-11-03
Lazy For() Comprehension Generator

pivovarit:
Introduces a generator for lazy for-comprehension variants that produce intermediate `ForLazyN<M>` classes for supported monadic types (e.g., _Option_, _Try_, _Either_, etc.).

This addition complements the existing eager for-comprehension generator by introducing a lazy variant.

- Added generation of `ForLazyN<M`> classes (arity 2 to N) that:
  - Use one monadic source (ts1) followed by a chain of dependent functions (ts2, ts3, …).
  - Allow lazy composition of monadic operations through nested lambdas.
- Added corresponding `For(...)` factory methods for convenient construction.
- Implemented a recursively generated `yield(...)` method with proper indentation and block structure for readability.

The generator handles mixed monadic/function parameter patterns such as:

```java
Option<T1> ts1;
Function1<T1, Option<T2>> ts2;
Function2<T1, T2, Option<T3>> ts3;
```

expands for-comprehension with lazily-evaluated variants, so that the existing is possible:

```java
Option<Integer> result = API.For(
  calculate1(),
  r1 -> calculate2(r1),
  (r1, r2) -> calculate3(r1, r2)).yield((r1, r2, r3) -> r1 + r2 + r3);
```

The generator enforces that the first argument is always a monadic value, not a function. This is necessary because:
- The first argument acts as the entry point of the comprehension; it provides the initial context for all later `flatMap` calls.
- If _ts1_ were also a function (e.g., `Function0<M<T1>>`), **Java’s generic type inference fails to resolve chained type variables** properly due to the presence of methods with similar erasure.
- Subsequent arguments can safely be functions (`FunctionN<T1, ..., TN, M<T(N+1)>>`), since their input types are already determined by previous monadic values.

----
related: https://github.com/vavr-io/vavr/issues/3038

### DIFF
--- a/vavr/src/main/java/io/vavr/Lazy.java
+++ b/vavr/src/main/java/io/vavr/Lazy.java
@@ -224,6 +224,16 @@ public final class Lazy<T> implements Value<T>, Supplier<T>, Serializable {
     }
 
     @Override
+    public <U> Lazy<U> mapTo(U value) {
+        return map(ignored -> value);
+    }
+
+    @Override
+    public Lazy<Void> mapToVoid() {
+        return map(ignored -> null);
+    }
+
+    @Override
     public Lazy<T> peek(Consumer<? super T> action) {
         action.accept(get());
         return this;
--- a/vavr/src/main/java/io/vavr/Value.java
+++ b/vavr/src/main/java/io/vavr/Value.java
@@ -70,6 +70,8 @@ import static io.vavr.API.*;
  * <li>{@link #getOrElseTry(CheckedFunction0)}</li>
  * <li>{@link #getOrNull()}</li>
  * <li>{@link #map(Function)}</li>
+ * <li>{@link #mapTo(Object)}</li>
+ * <li>{@link #mapToVoid()}</li>
  * <li>{@link #stringPrefix()}</li>
  * </ul>
  *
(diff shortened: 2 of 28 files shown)


vavr/src-gen/main/java/io/vavr/API.java
```diff
@@ -5740,6 +5740,3660 @@ public <R> Validation<L, R> yield(Function8<? super T1, ? super T2, ? super T3,
...
+     * @param ts4 the 4th Option
+
+     * @param <T1> the component type of the 1st Option
+     * @param <T2> the component type of the 2nd Option
+     * @param <T3> the component type of the 3rd Option
+     * @param <T4> the component type of the 4th Option
+     * @return a new {@code ForLazy4Option} builder of arity 4
+     * @throws NullPointerException if any argument is {@code null}
+     */
+    public static <T1, T2, T3, T4> ForLazy4Option<T1, T2, T3, T4> For(Option<T1> ts1, Function1<T1, Option<T2>> ts2, Function2<T1, T2, Option<T3>> ts3, Function3<T1, T2, T3, Option<T4>> ts4) {
```
adamkopec: just wondering about type variance here, wouldn't it be a good moment to introduce it now, than eg. in a later release, because of compatibility?

PR #3089, opened 2025-11-07
Fix: Missing javadocs for types in For comprehensions

adamkopec:
Partly (but just partly) tackles:

https://github.com/vavr-io/vavr/issues/2981

If the format is OK, I can fix other generated elements, like Patterns and Cases and maybe refactor the generator a bit.


pivovarit (approved):
Looking good! Thanks!

PR #3096, opened 2025-11-14
Add Maven wrapper

kressi:
Since not everyone has/wants Maven locally installed, I'm proposing to add Maven wrapper to the project. Additionally, this ensures the project is built with the desired Maven version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
           cache: 'maven'
 
       - name: Build with Maven
-        run: mvn test
+        run: ./mvnw test
 
   deploy-snapshot:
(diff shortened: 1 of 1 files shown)


kressi:
Yes, only [.m2/repository](https://github.com/actions/setup-java/blob/main/src/cache.ts#L26) is being cached. It does not look like cache can easily be extended with additional paths. I think setting up cache for `.m2/wrapper/dist` will complicate things unnecessarily. Is it ok if I add wrapper as `type=bin`, so it's part of the repository?

pivovarit:
please also sync with the main branch - this should be easy (just accept what's on main and replace commands with wrapper)

pivovarit (approved):
thanks!

PR #3256, opened 2026-04-04
Modularised the generator

adamkopec:
Hey,

I got frustrated by the generator - it was such a huge file, difficult to read and modify. I changed the structure only, the output stayed the same - byte to byte.

Would love to know what you think!

Cheers,
Adam


adamkopec:
@pivovarit I still ask for the review. I think it's better than the original super turbo mega file. and that merging of this change encourages further work (like making the generated file follow the conventions more strictly or splitting the generator further). please take a look 🙏

PR #3260, opened 2026-04-19
Improve fold, foldLeft, and foldRight Javadoc

toolforger:
I have split this up into multiple small commits, and gone to some lengths to explain why I think each one is an improvement.
Apologies for the verbosity; I felt this required a bit more explanation than usual as some changes may go against implicit ideas about what to document where, while others try to clarify concepts but I may simply be wrong.

I am able and willing to modify the commit history in any way you wish.

### DIFF
--- a/vavr/src/main/java/io/vavr/collection/Foldable.java
+++ b/vavr/src/main/java/io/vavr/collection/Foldable.java
@@ -82,6 +82,12 @@ public interface Foldable<T> {
      * Folding from the left means that elements are combined in the order they are encountered,
      * associating each step with the accumulated result so far.
      * <p>
+     * Note that {@code zero} is not necessarily a zero in the mathematical sense,
+     * which would be the neutral element of {@code combine};
+     * it may even be of a different type than the elements of this {@code Foldable}.<br>
+     * Think of it as the initial value for the accumulator inside the {@code foldLeft} function,
+     * which is updated with each call to {@code combine}.
+     * <p>
      * <strong>Example:</strong>
      * <pre>{@code
      * // Result: "cba!"
@@ -89,7 +95,10 @@ public interface Foldable<T> {
      * }</pre>
      *
      * @param <U>     the type of the accumulated result
-     * @param zero    the initial value to start folding with
+     * @param zero    the initial value to start folding with.<br>
+     *                <i>This is not a zero in the mathematical sense!</i><br>
+     *                Think of this as the starting value for the accumulator,
+     *                replaced with a new accumulator value with every call to {@code combine}.
      * @param combine a function that combines the accumulated value and the next element
      * @return the folded result
      * @throws NullPointerException if {@code combine} is null
--- a/vavr/src/main/java/io/vavr/collection/Foldable.java
+++ b/vavr/src/main/java/io/vavr/collection/Foldable.java
@@ -45,7 +45,7 @@ public interface Foldable<T> {
 
     /**
      * Folds the elements of this structure using the given associative binary operator,
-     * starting with the provided {@code zero} element and successively applying {@code combine}.
+     * starting with the provided {@code zero} value and successively applying {@code combine}.
      * <p>
      * The order in which elements are combined is non-deterministic. Therefore, {@code combine}
      * must be associative to guarantee a consistent result regardless of traversal order.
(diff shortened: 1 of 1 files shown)


pivovarit (changes requested):
looks good to me, but I think we should get more precise in some places

> combine.apply(zero, x) must give the same result as combine.apply(x, zero)

the actual requirement is stronger than this - zero must be _an identity element_ (_combine(zero, x) must equal x_)

vavr/src/main/java/io/vavr/collection/Foldable.java
```diff
@@ -87,14 +82,23 @@ default T fold(T zero, @NonNull BiFunction<? super T, ? super T, ? extends T> co
...
+     * it may even be of a different type than the elements of this {@code Foldable}.<br>
+     * Think of it as the initial value for the accumulator inside the {@code foldLeft} function,
+     * which is updated with each call to {@code combine}.
+     * <p>
      * <strong>Example:</strong>
      * <pre>{@code
-     * // Result: "cba!"
-     * List.of("a", "b", "c").foldLeft("!", (acc, x) -> x + acc);
+     * // Result: 42
+     * List.of('4', '2').foldLeft(0, (acc, x) -> acc * 10 + x - '0');
```
pivovarit:
```suggestion
     * List.of('4', '2').foldLeft(0, (acc, x) -> acc * 10 + (x - '0'));
```

it should make it a bit more clear

toolforger:
> looks good to me, but I think we should get more precise in some places
> 
> > combine.apply(zero, x) must give the same result as combine.apply(x, zero)
> 
> the actual requirement is stronger than this - zero must be _an identity element_ (_combine(zero, x) must equal x_)



> looks good to me, but I think we should get more precise in some places
> 
> > combine.apply(zero, x) must give the same result as combine.apply(x, zero)
> 
> the actual requirement is stronger than this - zero must be _an identity element_ (_combine(zero, x) must equal x_)

I'm assuming `combine` is a `java.lang.BiFunction`, so _combine(zero, x) must equal x_ is a shorthand for `combine(foo,bar)`.
Note that I'm striving for making this accessible to Java programmers.

Nitpick: For non-commutative folds such as when they involve string concatenation, identity needs to be defined on both sides, i.e. _combine(zero, x) = combine(x, zero) = x_.
Or, translated from math to Java: `combine(zero, x) == zero` and `combine(x, zero) == x` - still sloppy, it should really be a deep equals instead of `==`.

I don't think dwelling on deep equality is really worth it in this Javadoc (but it's surprisingly tricky so Vavr should probably define it somewhere, if only as a recommendation how to define `isEqual` on Pojos.
Here, it's probably best to drop from Java to math, i.e. say: _The results of `combine(zero, x)` and `combine(x, zero)` must be equal to `x`._

(Trickiness of equality: The definition is recursive so you need to define a fixed point; fixed points are somewhat arbitrary so you have to choose, the absurd extremes are "nothing is equal" and "everything is equal", the middle grounds being that you can include or exclude Pojo properties at will; equality for infinite data structures is undecidable, so you comparing two equal infinite lists will run into an infinite loop, insisting that lambdas/`Function`s/`BiFunction`s are to be compared using reference equality will possibly turn equal values into inequal ones, which the definition of equality is just fine with but it might not be intended, as it misses both `Objects.isEqual(() -> List.of(1, 2, 3), () -> List.of(1, 2, 3)` and `Objects.isEqual(List.of(1, 2, 3), () -> List.of(1, 2, 3))`. The former can be circumvented by being careful to define each lambda exactly once, the latter requires depth-controlled partial evaluation and an exception when the limit is exceeded. Oh, and you probably want to de-recursify `isEqual()`, too, at least for the Vavr-aware Pojos, as the Java stack can be surprisingly small.
It's really a lot of complications for such a conceptually simple thing... but then it turns out this mess is related to mathematical messes where equality of uncountable sets is just as problematic, just that the limitations pop up for computing even in the countably infinite case as we insist on finite calculations.
It all makes sense, but the mess is still there...)

pivovarit:
Good catch, your suggestion makes sense:

> The results of combine(zero, x) and combine(x, zero) must be equal to x.

toolforger:
> looks good to me, but I think we should get more precise in some places

Are there more places that can be made more precise?
I have tunnel vision and can't identify any.

pivovarit:
No, that was it - there are probably dozens of places that would benefit from extra clarify, but this change is self-sufficient, big thanks for shipping this!

pivovarit (approved):

PR #3266, opened 2026-05-08
Annotate Option's generic parameter as @NonNull

jkstrauss:
This is a trivial change, but a game-changer when trying to avoid nullability warnings.


pivovarit:
oh :) a great catch! thanks!

pivovarit (approved):

pivovarit:
I will merge this once GitHub recovers.

image

jkstrauss:
> oh :) a great catch! thanks!

Actually, IntelliJ caught it. I was just annoyed with making unnecessary null checks to make the warning go away that I figured I might as well do something about it. :-)

PR #3280, opened 2026-05-21
Fix future cancellation callbacks

### LINKED ISSUE #2552: Chained futures keep executing although future was cancelled

puneetdixit200:
## Summary
- stop cancellation from dispatching queued or newly registered completion callbacks
- make `andThen` use a cancellable result future so cancelling the chained future prevents the delayed side-effect action
- add regression coverage for direct callbacks and cancelled `andThen` chains

Fixes #2552

## Tests
- `JAVA_HOME=/opt/homebrew/Cellar/openjdk/25.0.2/libexec/openjdk.jdk/Contents/Home PATH=/opt/homebrew/Cellar/openjdk/25.0.2/libexec/openjdk.jdk/Contents/Home/bin:$PATH ./mvnw -pl vavr -DskipCheckstyle -DskipCoverage -Dmaven.repo.local=$PWD/.m2/repository -Dtest=io.vavr.concurrent.FutureTest test`
- `git diff --check`

### DIFF
--- a/vavr/src/main/java/io/vavr/concurrent/Future.java
+++ b/vavr/src/main/java/io/vavr/concurrent/Future.java
@@ -864,7 +864,7 @@ public interface Future<T> extends Value<T> {
     /**
      * Returns the value of this {@code Future}.
      *
-     * @return {@code None} if the {@code Future} is not yet completed or was cancelled; otherwise, {@code Some(Try)} containing the result or failure
+     * @return {@code None} if the {@code Future} is not yet completed; otherwise, {@code Some(Try)} containing the result, failure, or cancellation failure
      */
     Option<Try<T>> getValue();
 
--- a/vavr/src/main/java/io/vavr/concurrent/Future.java
+++ b/vavr/src/main/java/io/vavr/concurrent/Future.java
@@ -664,12 +664,14 @@ public interface Future<T> extends Value<T> {
      */
     default Future<T> andThen(@NonNull Consumer<? super Try<T>> action) {
         Objects.requireNonNull(action, "action is null");
-        return run(executor(), complete ->
-                onComplete(t -> {
-                    Try.run(() -> action.accept(t));
-                    complete.with(t);
-                })
-        );
+        final FutureImpl<T> result = FutureImpl.of(executor());
+        onComplete(t -> {
+            if (!result.isCancelled()) {
+                Try.run(() -> action.accept(t));
+                result.tryComplete(t);
+            }
+        });
+        return result;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


pivovarit:
Hi! Thanks for this PR!

I added extra Future tests that cover important parts of the Future contract and those are failing now

pivovarit:
Looks good to me. However, huge chunk of your logic duplicates `tryComplete`. You could probably avoid it if you set cancelled to true and then called `tryComplete`

puneetdixit200:
Fixed in d5b4f545. `cancel(...)` now marks the future as cancelled and delegates completion, waiter release, and callback delivery to `tryComplete(Try.failure(new CancellationException()))` instead of duplicating that path.

Fresh local check on this head:

```console
JAVA_HOME=/opt/homebrew/opt/openjdk@21/libexec/openjdk.jdk/Contents/Home PATH="/opt/homebrew/opt/openjdk@21/bin:$PATH" ./mvnw -q -pl vavr -Dtest=FutureTest test -Dmaven.repo.local=/tmp/vavr-3280-m2
# TEST-io.vavr.concurrent.FutureTest.xml: tests=227, failures=0, errors=0
```

pivovarit:
Thanks! Looks good to me! I know this is AI, but I don't care - the change is good :)

pivovarit (approved):

PR #3343, opened 2026-08-30
Fix NullAway 0.14 override-bound errors

skrcode:
## Summary

Companion fix for #3338, targeting its exact Dependabot branch.

NullAway 0.14 checks generic upper bounds on overridden methods more strictly. This change:

- makes regular `Function1.andThen` and `Function1.compose` retain the non-null implicit bound required by `java.util.function.Function`, while preserving nullable bounds for checked and higher-arity Vavr functions;
- repeats the nullable parent bounds on the affected map and multimap overrides; and
- updates the function generator as well as its tracked generated output.

There is no runtime behavior change. The annotations make the existing override contracts explicit.

## Verification

- `JAVA_HOME=<JDK 21> ./mvnw -B -Pnullaway compile` — passed, including Spotless.
- `JAVA_HOME=<JDK 21> ./mvnw -B clean test` — passed: 23,325 tests, 0 failures, 0 errors, 43 skipped.
- `JAVA_HOME=<JDK 21> ./mvnw -B -pl vavr generate-sources` — passed; `Function1.java` was byte-for-byte stable on regeneration.
- `git diff --check` and complete diff review — passed.
- JAIPilot Remote (JDK 21): `./mvnw -B -Pnullaway compile` — passed in 44.7s, including Spotless.

The documented `javadoc:javadoc` goal currently fails, with and without `-pl vavr`, because Maven Javadoc refuses to aggregate the repository's named and unnamed modules. The remote medium-profile full test attempt also exceeded its 4 GiB heap and tripped one wall-clock performance assertion; the complete suite passed locally as recorded above.

Built and verified with [JAIPilot](https://github.com/JAIPilot/jaipilot):
- Skills: `jaipilot-maintainer-intent`, `jaipilot-fast-execution`, `jaipilot-review-diff`, and `jaipilot-remote-java`
- Execution profile: `gpt-5.6-sol · xhigh · fast`

### DIFF
--- a/vavr/generator/Generator.scala
+++ b/vavr/generator/Generator.scala
@@ -1800,6 +1800,7 @@ def generateMainClasses(): Unit = {
         val genericsTryReturnType = s"<${genericsFunction}${im.getType("io.vavr.control.Try")}<R>>"
         val curried = if (i == 0) "v" else (1 to i).gen(j => s"t$j")(using " -> ")
         val compositionType = if(checked) "CheckedFunction1" else im.getType("java.util.function.Function")
+        val compositionTypeVariableBound = if (!checked && i == 1) "" else s" $nullableBound"
 
         // imports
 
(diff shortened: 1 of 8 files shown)


pivovarit:
Looks good to me! Thanks!

PR #3352, opened 2026-09-19
Fix sequence rotations by Integer.MIN_VALUE

jakezwang:
Rotating a nonempty sequence by `Integer.MIN_VALUE` throws `StackOverflowError`: the negative-distance branches call each other because negating that value leaves it unchanged.

Normalize the distance with `Math.floorMod` in both directions. Existing empty, zero-distance, and whole-cycle rotations still return the original instance. Add shared sequence tests for the minimum integer at lengths five and four, covering rotation and no-op identity respectively.

Validation on JDK 21: `./mvnw -B package` (23,345 tests, 43 skipped, no failures), including formatting and Javadoc checks; `./mvnw -B -Pnullaway compile`. The four new List tests fail with `StackOverflowError` before the fix. An additional public-API comparison against JDK `Collections.rotate` passed for all six sequence implementations with boundary and random distances.


pivovarit:
Nice catch, thanks!

pivovarit (approved):

